PR #2, opened 2025-08-29
Use new sedona_internal_err to avoid misleading datafusion internal err

petern48:
I've found the use of Datafusion's `internal_err!` to be very misleading during development because it suggests the error is DataFusion's and not Sedona's (see the example texts). This PR adds a new macro to use instead.

Example (code was modified to intentionally trigger this internal error):

Before:
```sql
> select st_m(st_geomfromtext('point (1 1)'));
Internal error: unexpected dimension.
This issue was likely caused by a bug in DataFusion's code. Please help us to resolve this by filing a bug report in our issue tracker: https://github.com/apache/datafusion/issues
```

Now:
```sql
> select st_m(st_geomfromtext('point (1 1)'));
External error: SedonaDB internal error: unexpected dimension.
This issue was likely caused by a bug in SedonaDB's code. Please help us to resolve this by filing a bug report in our issue tracker: https://github.com/apache/sedona-db/issues
```

It would be nice to avoid the "External error:" prepending the message, but that seems to be the most reasonable of the available options without making an upstream change.

### DIFF
--- a/rust/sedona-common/src/error.rs
+++ b/rust/sedona-common/src/error.rs
@@ -8,10 +8,10 @@ macro_rules! sedona_internal_err {
             Please help us to resolve this by filing a bug report in our issue tracker: \
             https://github.com/apache/sedona-db/issues",
             std::format!($($args),*),
-            DataFusionError::get_back_trace(),
+            datafusion_common::DataFusionError::get_back_trace(),
         );
         // We avoid using Internal to avoid the message suggesting it's internal to DataFusion
-        Err(DataFusionError::External(msg.into()))
+        Err(datafusion_common::DataFusionError::External(msg.into()))
     }};
 }
 
--- a/rust/sedona-common/src/error.rs
+++ b/rust/sedona-common/src/error.rs
@@ -1,0 +1,34 @@
+/// Macro to create Sedona Internal Error that avoids the misleading error message from
+/// DataFusionError::Internal.
+#[macro_export]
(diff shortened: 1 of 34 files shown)


rust/sedona-common/src/error.rs
```diff
@@ -0,0 +1,34 @@
...
+/// DataFusionError::Internal.
+#[macro_export]
+macro_rules! sedona_internal_err {
+    ($($args:expr),*) => {{
+        let msg = std::format!(
+            "SedonaDB internal error: {}{}.\nThis issue was likely caused by a bug in SedonaDB's code. \
+            Please help us to resolve this by filing a bug report in our issue tracker: \
+            https://github.com/apache/sedona-db/issues",
+            std::format!($($args),*),
+            datafusion_common::DataFusionError::get_back_trace(),
```
paleolimbot, line 11: If we get the same linking errors in CI that we got before, you could try removing the back trace feature (we can always add it back in later if it turns out that we're having trouble debugging user errors)

paleolimbot (commented):
Thank you!

We'll need to get CI running to make sure, but this is great in principle!

rust/sedona-common/src/error.rs
petern48: Ah, that's what you meant

petern48: it passed !

PR #8, opened 2025-08-29
Add more python integration tests + fix pre-commit ruff-format

petern48:
- Moved st_dwithin to test_predicates
- Add tests for st_buffer, st_asbinary, st_geo(m/g)fromwkb
- Removed a duplicate ruff-format entry (the older version) in pre-commit that was conflicting with the other, preventing successful passing the check when they wanted different things


.pre-commit-config.yaml
```diff
@@ -44,13 +44,6 @@ repos:
       name: rustfmt
       args: ["--all", "--"]
 
-  - repo: https://github.com/astral-sh/ruff-pre-commit
-    rev: v0.1.5
-    hooks:
-      - id: ruff
-        args: [ --fix ]
-      - id: ruff-format
-
```
paleolimbot, line 53: Thanks!

python/sedonadb/python/sedonadb/testing.py
```diff
@@ -260,11 +269,15 @@ def assert_result(self, result, expected, **kwargs) -> "DBEngine":
             self.assert_result(result, [(expected,)], **kwargs)
         elif isinstance(expected, bool):
             self.assert_result(result, [(str(expected).lower(),)], **kwargs)
+        elif isinstance(expected, bytes):
+            self.assert_result(result, [(expected,)], datatype="binary", **kwargs)
```
paleolimbot: It may make more sense to implement this in the same way that you implemented the numeric comparison (i.e., take the first cell of the Pandas output). I am not sure that passing the datatype makes much sense for the tuples output (which should always be strings).

python/sedonadb/python/sedonadb/testing.py
```diff
@@ -183,6 +188,10 @@ def result_to_tuples(self, result, *, wkt_precision=None) -> List[Tuple[str]]:
             # isinstance() does not always work with pyarrow in pytest
             if _type_is_geoarrow(col.type):
                 columns.append(ga.format_wkt(col, precision=wkt_precision).to_pylist())
+            elif datatype == "binary":
+                binary_lst = col.cast(pa.binary()).to_pylist()
+                # Convert to hex format for comparison e.g b'0101000000000000000000F03F000000000000F03F',
+                columns.append([b.hex().upper().encode() for b in binary_lst])
```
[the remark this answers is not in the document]
paleolimbot: I thought that `col.cast(pa.string())` does this already. Does it not generate an uppercase hex string?

python/sedonadb/python/sedonadb/testing.py
```diff
@@ -260,11 +269,15 @@ def assert_result(self, result, expected, **kwargs) -> "DBEngine":
...
+        elif isinstance(expected, bytes):
+            self.assert_result(result, [(expected,)], datatype="binary", **kwargs)
         elif isinstance(expected, (int, float)):
             result_df = self.result_to_pandas(result)
             assert result_df.shape == (1, 1)
             result_value = result_df.iloc[0, 0]
-            assert result_value == expected, f"Expected {expected}, got {result_value}"
+            assert math.isclose(result_value, expected), (
+                f"Expected {expected}, got {result_value}"
+            )
```
paleolimbot: This is something we should handle explicitly as an argument (i.e., `numeric_epsilon`, or perhaps align it with the name that `pandas.testing.assert_dataframe_equal()` uses to specify an epsilon). Most of the time we want to assert identicalness and only in specific cases do we want to relax that (and we want that relaxing to be explicit in the test code).

python/sedonadb/tests/functions/test_functions.py
```diff
@@ -50,6 +50,106 @@ def test_st_area(eng, geom, expected):
     eng.assert_query_result(f"SELECT ST_Area({geom_or_null(geom)})", expected)
 
 
+@pytest.mark.parametrize("eng", [SedonaDB, PostGIS])
+@pytest.mark.parametrize(
+    ("geom", "expected"),
+    [
+        ("POINT (1 1)", b"0101000000000000000000F03F000000000000F03F"),
```
```suggestion
        ("POINT (1 1)", "0101000000000000000000F03F000000000000F03F"),
```

If we're asserting bytes output as uppercase hex strings, the `b"` is confusing. If we're asserting bytes output as bytes, use a bytestring with `\x` escapes (e.g., `b"\x01\x01\x00..."`). I think hex strings will work out of the box here without any modifications to the tester (because that is how Arrow C++ casts them to string).

python/sedonadb/tests/functions/test_functions.py
```diff
@@ -50,6 +50,106 @@ def test_st_area(eng, geom, expected):
...
+    eng = eng.create_or_skip()
+    eng.assert_query_result(f"SELECT ST_AsBinary({geom_or_null(geom)})", expected)
+
+
+@pytest.mark.parametrize("eng", [SedonaDB, PostGIS])
+@pytest.mark.parametrize(
+    ("geom", "expected"),
+    [
+        (None, None),
+        # Arrow-rs returns POINT (nan nan) instead of POINT EMPTY
```
```suggestion
        # geoarrow-c returns POINT (nan nan) instead of POINT EMPTY
```

python/sedonadb/tests/functions/test_functions.py
```diff
@@ -261,6 +327,60 @@ def test_st_geomfromtext(eng, wkt, expected):
...
+        (
+            "GEOMETRYCOLLECTION (POINT (0 0), LINESTRING (0 0, 1 1), POLYGON ((0 0, 0 1, 1 1, 1 0, 0 0)))",
+            "GEOMETRYCOLLECTION (POINT (0 0), LINESTRING (0 0, 1 1), POLYGON ((0 0, 0 1, 1 1, 1 0, 0 0)))",
+        ),
+    ],
+)
+def test_st_geogfromwkb(eng, geom, expected):
+    eng = eng.create_or_skip()
+    eng.assert_query_result(
+        f"SELECT ST_GeogFromWKB(ST_AsBinary({geom_or_null(geom)}))", expected
```
paleolimbot: This would be a less circular test if it used `shapely` to convert to WKB. That would also let you check `flavor="iso"` and `flavor="extended"`, which both engines should accept.

It's also worth checking what happens on invalid input (we error...I'm not sure what PostGIS does).

python/sedonadb/tests/functions/test_functions.py
```diff
@@ -50,6 +50,106 @@ def test_st_area(eng, geom, expected):
...
+            "GEOMETRYCOLLECTION (POINT (0 0), POLYGON ((0 0, 1 0, 1 1, 0 1, 0 0)), LINESTRING (0 0, 1 1), GEOMETRYCOLLECTION (POLYGON ((0 0, -1 0, -1 -1, 0 -1, 0 0))))",
+            "GEOMETRYCOLLECTION (POINT (0 0), POLYGON ((0 0, 1 0, 1 1, 0 1, 0 0)), LINESTRING (0 0, 1 1), GEOMETRYCOLLECTION (POLYGON ((0 0, -1 0, -1 -1, 0 -1, 0 0))))",
+        ),
+        ("POINT Z (0 0 0)", "POINT Z (0 0 0)"),
+        ("POINT ZM (0 0 0 0)", "POINT ZM (0 0 0 0)"),
+    ],
+)
+def test_st_astext(eng, geom, expected):
+    eng = eng.create_or_skip()
+    # Engines vary on where they place spaces in the text, so we convert back to geometry to compare
```
paleolimbot: I think this would be a better test if checked the output directly (which should eliminate the need to repeat the parameter twice). Because there's differing behaviour, parameterizing by engine doesn't make sense but you can do something like:

```python
sedona = SedonaDB()
sedona_wkt = geom.replace(" (", "").replace(", ", ",")
sedona.assert_query_equal(..., sedona_wkt)

postgis = PostGIS.create_or_skip()
postgis.assert_query_equal(..., geom)
```

python/sedonadb/python/sedonadb/testing.py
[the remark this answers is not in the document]
petern48: `col.cast(pa.string())` leads to this error, so it seems like I needed to use `pa.binary()` instead
```
python/sedonadb/testing.py:196: in result_to_tuples
    columns.append(col.cast(pa.string()).to_pylist())
pyarrow/table.pxi:597: in pyarrow.lib.ChunkedArray.cast
    ???
/Users/peter/.pyenv/versions/3.10.18/lib/python3.10/site-packages/pyarrow/compute.py:412: in cast
    return call_function("cast", [arr], options, memory_pool)
pyarrow/_compute.pyx:604: in pyarrow._compute.call_function
    ???
pyarrow/_compute.pyx:399: in pyarrow._compute.Function.call
    ???
pyarrow/error.pxi:155: in pyarrow.lib.pyarrow_internal_check_status
    ???
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

>   ???
E   pyarrow.lib.ArrowInvalid: Invalid UTF8 payload
```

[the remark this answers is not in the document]
paleolimbot: I see! Maybe:

```python
elif pa.types.is_binary(col.type) or pa.types.is_binary_view(col.type) or pa.types.is_large_binary(col.type):
   columns.append([None if b is None else b.hex().upper() for b in binary_lst])
```

(i.e., use the column type and not an extra argument, handle nulls, and return a string instead of utf-8 bytes).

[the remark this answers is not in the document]
petern48: I found a cleaner way using the escaped binary strings

.pre-commit-config.yaml
petern48: Yep ... i was left very confused why pre-commit was complaining and not making any changes until I realized there were two! 🙃

python/sedonadb/tests/functions/test_functions.py
petern48: changed to specifying the '\x' cleaned the `testing.py` code up

python/sedonadb/tests/functions/test_functions.py
```diff
@@ -147,6 +149,7 @@ def test_st_buffer(eng, geom, dist, expected_area):
     eng.assert_query_result(
         f"SELECT ST_Area(ST_Buffer({geom_or_null(geom)}, {val_or_null(dist)}))",
         expected_area,
+        numeric_epsilon=1e-1,
```
paleolimbot: This seems like a pretty high epsilon for things we think should be almost equal. Can this be lower?

petern48: Oh that's right ... I forgot how numbers work for a sec. This is a higher value, not a lower one. I changed it to the same as the default value.

paleolimbot (approved):
Thank you!

PR #11, opened 2025-09-02
feat(rust/sedona-functions,c/sedona-geos): Add remaining required sedona bench functions

paleolimbot:
Apologies for stuffing these all in the same PR (I didn't think they would be as complicated as they were!). This PR adds ST_Collect(), ST_MakeLine(), and ST_ConvexHull().

A few other things were required to make these cleanly implemented:

- Benchmark configuration for MultiPoint input (none of the existing random geometries were realistic input)
- Iterating over geometry arrays where the array is of NULL type (happens when a SQL literal NULL makes its way to a scalar function and the scalar function calls `.to_array(num_iterations)`.

### DIFF
--- a/rust/sedona-functions/src/st_makeline.rs
+++ b/rust/sedona-functions/src/st_makeline.rs
@@ -58,8 +58,8 @@ fn doc() -> Documentation {
         "Construct a line".to_string(),
         "ST_MakeLine (g1: Geometry or Geography, g2: Geometry or Geography)".to_string(),
     )
-    .with_argument("x", "double: X value")
-    .with_argument("y", "double: Y value")
+    .with_argument("g1", "Geometry or Geography: The first point or geometry")
+    .with_argument("g2", "Geometry or Geography: The second point or geometry")
     .with_sql_example("SELECT ST_MakeLine(ST_Point(0, 1), ST_Point(2, 3)) as geom")
     .build()
 }


rust/sedona-functions/src/st_makeline.rs
```diff
@@ -0,0 +1,204 @@
...
+    use super::*;
+    use datafusion_expr::ScalarUDF;
+
+    #[test]
+    fn udf_metadata() {
+        let geom_from_point: ScalarUDF = st_makeline_udf().into();
+        assert_eq!(geom_from_point.name(), "st_makeline");
+        assert!(geom_from_point.documentation().is_some());
+    }
+}
```
paleolimbot: Mental note to add tests here

rust/sedona-geometry/src/types.rs
```diff
@@ -147,6 +147,27 @@ impl GeometryTypeAndDimensions {
         }
     }
 
+    /// Create from [GeometryTrait]
+    pub fn try_from_geom(geom: &impl GeometryTrait) -> Result<Self, SedonaGeometryError> {
```
paleolimbot, line 151: Mental note that this needs a test

rust/sedona-testing/src/benchmark_util.rs
```diff
@@ -272,6 +272,8 @@ pub enum BenchmarkArgSpec {
     LineString(usize),
     /// Randomly generated polygon input with a specified number of vertices
     Polygon(usize),
+    /// Randomly generated linestring input with a specified number of vertices
+    MultiPoint(usize),
```
paleolimbot, line 276: Mental note to add a test here

PR #13, opened 2025-09-02
feat: Implement native st_pointz, st_pointm, st_pointzm

petern48:
- Implements native functions st_pointz, st_pointm, st_pointzm
- Cleans up and makes that file more readable (previously a lot of magic numbers)
- Add support for ArrayArrayArray and ArrayArrayArrayArray benchmark args


rust/sedona-testing/src/testers.rs
```diff
@@ -368,7 +368,7 @@ impl ScalarUdfTester {
         }
     }
 
-    fn invoke_arrays(&self, arrays: Vec<ArrayRef>) -> Result<ArrayRef> {
+    pub fn invoke_arrays(&self, arrays: Vec<ArrayRef>) -> Result<ArrayRef> {
```
petern48, line 372: I didn't create separate new public test functions like `invoke_arrays_arrays_arrays_arrays` this time because
1. I didn't need them for testing for this particular case
2. Functions with this number of args are so rare

But if you feel strongly about creating new functions and keeping this private, just let me know

paleolimbot: I like your approach! (Although this needs a tiny docstring now that it's public)

rust/sedona-functions/src/st_point.rs
```diff
@@ -38,53 +46,166 @@ pub fn st_point_udf() -> SedonaScalarUDF {
...
-        format!("{name} (x: Double, y: Double)"),
-    )
-    .with_argument("x", "double: X value")
-    .with_argument("y", "double: Y value")
-    .with_sql_example(format!("{name}(-64.36, 45.09)"))
-    .build()
+        3 if params[2] == "z" => format!(
+            "Construct a Point {} from X, Y and Z",
+            out_type_name.to_lowercase()
+        ),
```
paleolimbot: Even though it's slightly repetitive, I think this would be better off just split into four `doc()` sections.

rust/sedona-functions/src/st_point.rs
```diff
@@ -93,60 +214,108 @@ impl SedonaScalarKernel for STGeoFromPoint {
...
+                    coords.push(array.value(i));
                 }
             }
+
+            if has_null {
+                builder.append_null();
+            } else {
+                populate_wkb_item(&mut item, &coords);
+                builder.append_value(&item);
+            }
```
paleolimbot: The inside of this loop is a very performance-sensitive section and I think there are a few things we can do to help:

```rust
for ... {
  let values = (0..num_dimensions).iter().map(|j| coord_f64_arrays[j].value(i).collect::<Vec<_>>();
  let values_non_null = values.iter().take(num_dimensions).flatten().collect::<Vec<_>>();
  if values_non_null.len() == num_dimensions {
    write_wkb_point_header(&mut builder, dimensions)?;
    // I forget exactly how to write the f64s as little endian doubles to the builder but it's
    // the same as in write_wkb_xxxx()
   builder.append_value([]);
  } else {
    builder.append_null();
 }
}
```

This was the first ever function I wrote in sedonadb and we now have `write_wkb_point_xx()`. The `builder` implements `Write`, so you can write directly into it (pass it directly to `write_wkb_xxx()` functions).

Make sure you run `cd rust/sedona-functions` + `cargo bench -- st_point` to make sure your implementation doesn't slow down the XY case (the most common). It is probably worth separating the implementation completely for Z, M, and ZM (where performance is not as critical).

rust/sedona-functions/src/st_point.rs
```diff
@@ -38,53 +46,166 @@ pub fn st_point_udf() -> SedonaScalarUDF {
...
+            .with_argument("m", "double: M coordinate");
+    }
+
+    builder.with_sql_example(example).build()
 }
 
 #[derive(Debug)]
 struct STGeoFromPoint {
     out_type: SedonaType,
+    wkb_type: u32,
```
paleolimbot: Probably the right thing to use as a parameter here is `Dimensions`. You can use `GeometryTypeAndDimensions` to get the WKB value (or use `write_wkb_xxx()` from sedona-geometry, which I believe takes a Dimensions). You can use `dimensions.size()` to get the count.

paleolimbot (commented):
We have some new helpers that I think will make the implementation a bit more concise that are worth a try...the test code looks great!

rust/sedona-functions/src/st_point.rs
petern48: Yeah, you were right. It was a notable perf regression (at least 100%), so I reverted `st_point.rs` back to the original version and moved everything else over into a new `st_pointzm.rs

jiayuasu (approved):

PR #18, opened 2025-09-04
perf: Add more python benchmarks

petern48:
More benchmarks, I've been investigating lately.

### DIFF
--- a/benchmarks/README.md
+++ b/benchmarks/README.md
@@ -28,6 +28,12 @@ Install pytest-benchmark:
 pip install pytest-benchmark
 ```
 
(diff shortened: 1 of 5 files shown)


petern48:
Just pushing this somewhere for now

petern48:
I randomly looked back at this PR to do some cleanup. It originally mostly had a few benchmarks back when I was trying to add pytest-benches for every function. It no longer makes sense to do so (instead we can create them when we need them), so I removed those changes. Instead, I've left the one change I do think is pretty valuable in here, which is the doc change telling users to use release mode and not debug mode 🙂

PR #19, opened 2025-09-04
 bug: Add support for GeometryCollection to sd_random_geometry

### LINKED ISSUE #15: bug: Setting geom_type to GeometryCollection sd_random_geometry just always returns points
```sql
> select count(*), st_geometrytype(geometry) as geom_type FROM sd_random_geometry('{"geom_type": "GeometryCollection"}') group by geom_type;
┌──────────┬───────────┐
│ count(*) ┆ geom_type │
│   int64  ┆    utf8   │
╞══════════╪═══════════╡
│     1024 ┆ ST_Point  │
└──────────┴───────────┘
```

petern48:
Closes #15

I also personally prefer to avoid defaulting to points (since it can be misleading), so I implemented the last possible case so there's no default.


petern48:
Not urgent or too important. Just like having it for using for benchmarking functions

paleolimbot (approved):
Thank you!

Issue #20, opened 2025-09-04
bug: integer arguments can't be casted to floats


petern48:
Despite the error saying it's DataFusion's code, I suspect this is really our bug and we're just using `DataFusionError::Internal` somewhere or `internal_err!`, which returns this misleading bug, as mentioned in this previous PR: https://github.com/apache/sedona-db/pull/2

```sql
> select st_buffer(st_geomfromtext('point (0 0)'), 1);
Internal error: Cannot convert Int64(1) to f64.
This issue was likely caused by a bug in DataFusion's code. Please help us to resolve this by filing a bug report in our issue tracker: https://github.com/apache/datafusion/issues

-- Now, add a ".0" to make it a float, and it now works
> select st_buffer(st_geomfromtext('point (0 0)'), 1.0);
┌──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┐
│                                   st_buffer(st_geomfromwkt(Utf8("point (0 0)")),Float64(1))                                  │
│                                                              wkb                                                             │
╞══════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════╡
│ POLYGON((1 0,0.9807852804032304 -0.19509032201612825,0.9238795325112867 -0.3826834323650898,0.8314696123025452 -0.555570233… │
└──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┘
```

The same happens with `select st_dwithin(st_geomfromtext('point (0 0)'), st_geomfromtext('point (0 0)'), 1);`

Datafusion functions like max() still work with ints properly
```sql
select max(1);
┌───────────────┐
│ max(Int64(1)) │
│     int64     │
╞═══════════════╡
│             1 │
└───────────────┘
```

paleolimbot:
This is because we forgot to cast it!

https://github.com/apache/sedona-db/blob/4b74b560a8a2572864c77c335e638378163153e7/c/sedona-geos/src/st_buffer.rs#L61-L62

(`args[1].cast_to(&DataType::Float64)` should do it)

petern48:
Hmm, you're right, looking at some of datafusion's functions, they types inside of these functions. I originally thought it would be a higher level cast that applies to all functions, but I guess this is what datafusion does as well. Submitted a fix.

PR #32, opened 2025-09-05
bug: Casts ints to floats in st_buffer and st_dwithin args

petern48:
Closes https://github.com/apache/sedona-db/issues/20


paleolimbot (approved):
Thank you!

Issue #29, opened 2025-09-05
enhancement:  Split out ST_SetSRID string implementation into ST_SetCRS and add ST_CRS


jesspav:
Currently we have an implementation for `ST_SetSRID` that allows for a CRS string. Breaking it out into `ST_SetCRS` (string) and `ST_SetSRID` (int) to align `ST_SetSRID` with external implementations and so that we can pass the results of `ST_CRS` into transforms, etc.  Also need an implementation for `ST_SetCrs`.

## Requirements:
* `ST_SetSRID` split into new `ST_SetSRID` and `ST_SetCRS`
* `ST_CRS` is impemented
* integration tests for both (although only `ST_SetSRID` will run against PostGIS)
* bench marks for all three

paleolimbot:
Closed in #37 and #38!

PR #38, opened 2025-09-08
Add ST_CRS

jesspav:
## Summary

Adding `ST_CRS` so that we can pull the full CRS object from geometry where SRIDs do not apply (for example with non-EPSG codes).

Issue: #29 

## Examples

```
> select ST_crs(ST_SetCrs(ST_GeomFromText('POINT (1 1)'), 'EPSG:4329')) as crs;
┌───────────┐
│    crs    │
│    utf8   │
╞═══════════╡
│ EPSG:4329 │
└───────────┘

> select ST_crs(ST_SetCrs(ST_GeomFromText('POINT (1 1)'), 'EPSG:4326')) as crs;
┌───────────┐
│    crs    │
│    utf8   │
╞═══════════╡
│ OGC:CRS84 │
└───────────┘


> select ST_crs(ST_SetSrid(ST_GeomFromText('POINT (1 1)'), null)) as crs;
┌──────┐
│  crs │
│ utf8 │
╞══════╡
│      │
└──────┘

```


python/sedonadb/tests/functions/test_transforms.py
```diff
@@ -64,3 +80,20 @@ def test_st_setcrs_sedonadb(eng, geom, crs, expected_srid):
...
     assert df.crs.to_epsg() == expected_srid
+
+
+@pytest.mark.parametrize("eng", [SedonaDB])
+@pytest.mark.parametrize(
+    ("geom", "crs", "expected_crs"),
+    [
+        ("POINT (1 1)", "'EPSG:26920'", "EPSG:26920"),
+        ("POINT (1 1)", f"'{pyproj.CRS('EPSG:26920').to_json()}'", "EPSG:26920"),
+        ("POINT (1 1)", None, None),
```
paleolimbot: optional nit: These parameters become part of the test name when we run `pytest`, which in this case is a very long string for the second parameter! It may make sense to have multiple `eng.assert_query_result()` instead.

rust/sedona-functions/src/st_srid.rs
```diff
@@ -94,6 +118,52 @@ impl SedonaScalarKernel for StSrid {
...
+
+        executor.execute_wkb_void(|maybe_wkb| {
+            match maybe_wkb {
+                Some(_wkb) => builder.append_option(crs_opt.clone()),
+
+                _ => builder.append_null(),
+            }
+
+            Ok(())
+        })?;
```
paleolimbot: If we're going to return an array here, I think `ScalarValue::Utf8(crs_opt).to_array(executor.num_iterations())` would work too (and avoid parsing all the WKB in the array).

I'm not sure the internals support it very well, but returning a `ScalarValue` even for arrays (when the crs is attached to the type) would probably help with SQL like `ST_SetCRS(geom, ST_CRS(other_geom))`. This might mean that you can't use the `ScalarUdfTester` (which might error for such cases).

rust/sedona-functions/src/st_srid.rs
```diff
@@ -94,6 +118,52 @@ impl SedonaScalarKernel for StSrid {
...
+        let executor = WkbExecutor::new(arg_types, args);
+        let preallocate_bytes = "EPSG:4326".len() * executor.num_iterations();
+        let mut builder =
+            StringBuilder::with_capacity(executor.num_iterations(), preallocate_bytes);
+        let crs_opt: Option<String> = match &arg_types[0] {
+            SedonaType::Wkb(_, Some(crs)) | SedonaType::WkbView(_, Some(crs)) => {
+                match crs.to_authority_code()? {
+                    Some(auth_code) => Some(auth_code),
+                    None => Some(crs.to_json()),
+                }
```
paleolimbot: While the authority:code output is definitely a lot nicer to look at, I think we want to return `crs.to_json()` here because it does not loose information. I think that `ST_SRID()` will support most uses where the user wants a code. In some future we could also have a "crs" type that prints out nicely but keeps the JSON under the hood.

rust/sedona-functions/src/st_srid.rs
```diff
@@ -94,6 +118,52 @@ impl SedonaScalarKernel for StSrid {
...
 }
 
+#[derive(Debug)]
+struct StCrs {}
+
+impl SedonaScalarKernel for StCrs {
+    fn return_type(&self, args: &[SedonaType]) -> Result<Option<SedonaType>> {
+        let matcher = ArgMatcher::new(
+            vec![ArgMatcher::is_geometry_or_geography()],
+            SedonaType::Arrow(DataType::Utf8),
```
paleolimbot: A `Utf8View` is probably what we want to return here (it more effectively handles repeated values). We can change that in the future if it causes an error in some internal somewhere (string views are a bit newer).

jesspav: Thanks, fixed.

rust/sedona-functions/src/st_srid.rs
jesspav: Interesting. Thanks. Fixed.

python/sedonadb/tests/functions/test_transforms.py
jesspav: :) yeah, I see what you mean.  I ended up removing the long one anyways since we switched to returning json instead of auth code and the the json returned is slightly different.

rust/sedona-functions/src/st_srid.rs
jesspav: Discussed offline with @paleolimbot .  We agreed that there is a deeper issue that blocks us from being able to do  `ST_SetCRS(geom, ST_CRS(other_geom))`, since the CRS is currently embedded in the type so we need to know the CRS at the planning stage in order to create the correct return type.   Fixing that is out of scope for this review.

jesspav: Created #46 to track fixing that.

jesspav: Outside of that, I 100% agree that parsing the binary is a waste, but I thought we needed to return null if the geometry is null, so `ScalarValue::Utf8(crs_opt).to_array(executor.num_iterations())` wouldn't quite work, right?

paleolimbot: That's a great point...it would only work if the source null_count was zero. We can circle back and optimize later!

python/sedonadb/tests/functions/test_transforms.py
```diff
@@ -64,3 +80,15 @@ def test_st_setcrs_sedonadb(eng, geom, crs, expected_srid):
...
+def test_st_crs_sedonadb(eng):
+    eng = eng.create_or_skip()
+    eng.assert_query_result(
+        f"SELECT ST_CRS(ST_SetCrs(ST_GeomFromText('POINT (1 1)'), 'EPSG:26920'))",
+        "\"EPSG:26920\""
+    )
+    eng.assert_query_result(
+        f"SELECT ST_CRS(ST_SetCrs(ST_GeomFromText('POINT (1 1)'), NULL))",
+        None,
+    )
```
paleolimbot: `pre-commit run --all-files` may complain about the end of this file (because it lacks a newline)

rust/sedona-functions/src/st_srid.rs
```diff
@@ -79,16 +103,58 @@ impl SedonaScalarKernel for StSrid {
...
-                    builder.append_option(srid_opt);
-                }
-                _ => builder.append_null(),
+        if let ColumnarValue::Array(array) = &args[0] {
+            (0..array.len()).for_each(|i| {
+                builder.append_option(if array.is_null(i) { None } else { srid_opt });
+            });
+        } else if let ColumnarValue::Scalar(scalar) = &args[0] {
+            builder.append_option(if scalar.is_null() { None } else { srid_opt });
+        }
```
paleolimbot: This works too!

I usually see this written as

```rust
match &args[0] {
  ColumnarValue::Array(array) => {...}
  ColumnarValue::Scalar(scalar) => {...}
}
```

...but either is ok!

rust/sedona-functions/src/st_srid.rs
```diff
@@ -152,4 +222,39 @@ mod test {
...
+            .invoke_scalar("POLYGON ((0 0, 1 0, 0 1, 0 0))")
+            .unwrap();
+        tester.assert_scalar_result_equals(
+            result,
+            ScalarValue::Utf8(Some("\"EPSG:4837\"".to_string())),
+        );
+
+        // Test with a CRS but null geom
+        let result = tester.invoke_scalar(ScalarValue::Null).unwrap();
+        tester.assert_scalar_result_equals(result, ScalarValue::Null);
```
paleolimbot, line 290: This (and the SRID test) would benefit from a test that uses `tester.invoke_array(create_array(&[Some(...), None]))` since that's now special-cased in the implementation.

paleolimbot (approved):
A few things to consider now or in a follow-up!

(Also the merge conflict so that CI can run!)

rust/sedona-functions/src/st_srid.rs
jesspav: yeah, that does look more rusty.  Forgive my noob formatting. :)

python/sedonadb/tests/functions/test_transforms.py
jesspav: fixed in next revision.

rust/sedona-functions/src/st_srid.rs
jesspav: switched in next revision.

rust/sedona-functions/src/st_srid.rs
jesspav: Good point. added them in next iteration.

PR #41, opened 2025-09-09
Add geopandas interop

MrPowers:


docs/geopandas-interop.ipynb
```diff
@@ -0,0 +1,211 @@
...
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 9,
+   "id": "2691bd24-9b2d-4cf9-958d-4ef01d967cb3",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "gdf = gpd.read_file(\"some_data.json\")"
```
paleolimbot, line 43: This would be a more compelling example with a real file.

Maybe:

- Read a .fgb file from an https:// url from https://geoarrow.org/data with GeoPandas
- Create the SedonaDB DataFrame

There are probably also other compelling GeoJSON examples that can be loaded from https:// urls (or you could use GeoPandas to write one to demonstrate interop in the other direction).

paleolimbot (approved):
Minor suggestion if you feel inclined...thank you!

docs/geopandas-interop.ipynb
MrPowers: Good idea, I added a FlatGeobuf example as well.  I kept the GeoJSON example if that's alright cause it's just nice and compact and easily understandable, especially for folks that aren't familiar with geo data.  It's really nice that SedonaDB supports all these file types via GeoPandas!

README.md
```diff
@@ -39,7 +37,7 @@ Start by establishing a connection:
 
 ```python
 import sedonadb
```
jiayuasu, line 39: Can we use sedona.db everywhere? This will help us get all downloads go to apache-sedona[db]?

jiayuasu:
@MrPowers can you address the git conflict?

PR #44, opened 2025-09-09
feat: Add options parameter to read_parquet() for S3 configuration

zhangfengcdt:
Adds an optional `options` parameter to the `read_parquet()` function to support S3 anonymous access and other DataFusion table configuration options.

  - Add `options: Optional[Dict[str, Any]]` parameter to Python `read_parquet()` function

  - Extend `GeoParquetReadOptions` to handle custom table options via `from_table_options()`
  - Update Python-Rust bindings to convert Python dict to Rust `HashMap<String, String>`
  - Maintain full backward compatibility - all existing code continues to work unchanged
  - Support for S3 anonymous access via `{"aws.nosign": True}` and similar options
  - Comprehensive unit test coverage (23 tests) for all usage scenarios

  ## Usage Examples

  ```python
  import sedonadb

  sd = sedonadb.connect()

  # Anonymous S3 access to public buckets
  df = sd.read_parquet(
      "s3://public-bucket/data.parquet",
      options={"aws.nosign": True, "aws.region": "us-west-2"}
  )

  # Backward compatibility - existing code unchanged
  df = sd.read_parquet("local/file.parquet")  # Still works

  # Empty options (equivalent to no options)
  df = sd.read_parquet("file.parquet", options={})

## Test Coverage: 

  - /tests/io/test_parquet_options.py (14 tests)
  - /tests/test_context.py (5 integration tests)

### DIFF
--- a/python/sedonadb/python/sedonadb/context.py
+++ b/python/sedonadb/python/sedonadb/context.py
@@ -17,7 +17,7 @@
 import os
 import sys
 from pathlib import Path
-from typing import Iterable, Literal, Union
+from typing import Any, Dict, Iterable, Literal, Optional, Union
 
 from sedonadb._lib import InternalContext, configure_proj_shared
 from sedonadb.dataframe import DataFrame, _create_data_frame
@@ -78,12 +78,18 @@ class SedonaContext:
         """
         self._impl.drop_view(name)
 
-    def read_parquet(self, table_paths: Union[str, Path, Iterable[str]]) -> DataFrame:
+    def read_parquet(
+        self,
+        table_paths: Union[str, Path, Iterable[str]],
+        options: Optional[Dict[str, Any]] = None,
+    ) -> DataFrame:
         """Create a [DataFrame][sedonadb.dataframe.DataFrame] from one or more Parquet files
 
         Args:
             table_paths: A str, Path, or iterable of paths containing URLs to Parquet
                 files.
+            options: Optional dictionary of options to pass to the Parquet reader.
+                For S3 access, use {"aws.nosign": True} for anonymous access to public buckets.
 
         Examples:
 
@@ -92,12 +98,23 @@ class SedonaContext:
             >>> sedonadb.connect().read_parquet(url)
             <sedonadb.dataframe.DataFrame object at ...>
 
+            >>> # Access public S3 bucket anonymously
+            >>> sedonadb.connect().read_parquet(
+            ...     "s3://public-bucket/data.parquet",
+            ...     options={"aws.nosign": True}
+            ... )
+            <sedonadb.dataframe.DataFrame object at ...>
+
         """
         if isinstance(table_paths, (str, Path)):
             table_paths = [table_paths]
 
+        if options is None:
+            options = {}
+
         return DataFrame(
-            self._impl, self._impl.read_parquet([str(path) for path in table_paths])
+            self._impl,
+            self._impl.read_parquet([str(path) for path in table_paths], options),
         )
 
     def sql(self, sql: str) -> DataFrame:
--- a/python/sedonadb/src/context.rs
+++ b/python/sedonadb/src/context.rs
@@ -14,7 +14,7 @@
 // KIND, either express or implied.  See the License for the
 // specific language governing permissions and limitations
 // under the License.
-use std::sync::Arc;
+use std::{collections::HashMap, sync::Arc};
 
 use pyo3::prelude::*;
 use sedona::context::SedonaContext;
@@ -74,11 +74,29 @@ impl InternalContext {
         &self,
         py: Python<'py>,
         table_paths: Vec<String>,
+        options: HashMap<String, PyObject>,
     ) -> Result<InternalDataFrame, PySedonaError> {
+        // Convert Python options dict to Rust HashMap<String, String>
+        let rust_options: HashMap<String, String> = options
+            .into_iter()
+            .map(|(k, v)| {
+                let v_str = if v.is_none(py) {
+                    String::new()
+                } else {
+                    // Convert PyObject to string
+                    match v.call_method0(py, "__str__") {
+                        Ok(str_obj) => str_obj.extract::<String>(py)?,
+                        Err(_) => v.extract::<String>(py)?,
+                    }
+                };
+                Ok((k, v_str))
+            })
+            .collect::<Result<HashMap<String, String>, PySedonaError>>()?;
+
         let df = wait_for_future(
             py,
             &self.runtime,
-            self.inner.read_parquet(table_paths, Default::default()),
+            self.inner.read_parquet(table_paths, rust_options),
         )??;
         Ok(InternalDataFrame::new(df, self.runtime.clone()))
     }
--- a/python/sedonadb/tests/test_context.py
+++ b/python/sedonadb/tests/test_context.py
@@ -49,7 +49,104 @@ def test_dynamic_object_stores():
     schema = pa.schema(con.read_parquet(url))
     assert schema.field("geometry").type.extension_name == "geoarrow.wkb"
 
-    # Fresh context
-    con = sedonadb.connect()
-    schema = pa.schema(con.sql(f"SELECT * FROM '{url}'"))
-    assert schema.field("geometry").type.extension_name == "geoarrow.wkb"
+
+def test_read_parquet_options_parameter(con, geoarrow_data):
+    """Test the options parameter functionality for read_parquet()"""
+    test_file = geoarrow_data / "quadrangles/files/quadrangles_100k_geo.parquet"
+
+    # Test 1: Backward compatibility - no options parameter
+    tab1 = con.read_parquet(test_file).to_arrow_table()
+    assert tab1["geometry"].type.extension_name == "geoarrow.wkb"
+
+    # Test 2: options=None (explicit None)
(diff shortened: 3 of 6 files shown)


python/sedonadb/python/sedonadb/context.py
```diff
@@ -120,12 +126,28 @@ def read_parquet(self, table_paths: Union[str, Path, Iterable[str]]) -> DataFram
...
+            >>> sedonadb.connect().read_parquet(url, options={})
+            <sedonadb.dataframe.DataFrame object at ...>
+
+            >>> # Test with None options (should work the same as no options)
+            >>> sedonadb.connect().read_parquet(url, options=None)
+            <sedonadb.dataframe.DataFrame object at ...>
+
+            >>> # Test S3 options (using HTTP URL for reliable testing)
+            >>> sedonadb.connect().read_parquet(url, options={"aws.nosign": True})
+            <sedonadb.dataframe.DataFrame object at ...>
```
paleolimbot: These examples are a bit confusing...probably just the third one makes sense but only if `url` is `s3://` (I don't have a plausible but tiny s3 example off the top of my head). We can also omit this until we find a better example (we'll use this example in the Python user guides/README, so there will be example usage somewhere).

python/sedonadb/src/context.rs
```diff
@@ -74,11 +74,29 @@ impl InternalContext {
...
+            .into_iter()
+            .map(|(k, v)| {
+                let v_str = if v.is_none(py) {
+                    String::new()
+                } else {
+                    // Convert PyObject to string
+                    match v.call_method0(py, "__str__") {
+                        Ok(str_obj) => str_obj.extract::<String>(py)?,
+                        Err(_) => v.extract::<String>(py)?,
+                    }
```
paleolimbot: I think that if you make `options: HashMap<String, Option<String>>` you can remove this

rust/sedona/src/context.rs
```diff
@@ -229,19 +229,20 @@ impl SedonaContext {
     pub async fn read_parquet<P: DataFilePaths>(
         &self,
         table_paths: P,
-        options: GeoParquetReadOptions<'_>,
+        options: HashMap<String, String>,
```
paleolimbot: It would probably make more sense to keep this as `GeoParquetReadOptions` and push the `GeoParquetReadOptions::from_table_options()` one level up

python/sedonadb/tests/test_context.py
```diff
@@ -63,7 +63,104 @@ def test_dynamic_object_stores():
...
+    assert len(tab3) == len(tab1)  # Should be identical
+
+    # Test 4: Options with string values
+    tab4 = con.read_parquet(
+        test_file, options={"test.option": "value"}
+    ).to_arrow_table()
+    assert tab4["geometry"].type.extension_name == "geoarrow.wkb"
+    assert len(tab4) == len(
+        tab1
+    )  # Should be identical (option ignored but not errored)
```
paleolimbot, line 92: I appreciate the comprehensiveness of the other tests, but I think this is probably the only test we need to maintain for this particular feature. I think we should use an actual s3 URL to make sure our parameters made it through, though. (There are some non-geometry ones in the sedona-cli tests, or we could use something in a wherobots public bucket if you know if a small one).

rust/sedona-geoparquet/src/provider.rs
```diff
@@ -78,22 +78,38 @@ pub async fn geoparquet_listing_table(
...
         config: &SessionConfig,
-        table_options: TableOptions,
+        mut table_options: TableOptions,
     ) -> ListingOptions {
+        // Merge custom table options if provided
+        if let Some(ref custom_options) = self.table_options {
+            for (key, value) in custom_options {
+                let _ = table_options.set(key, value);
+            }
+        }
```
paleolimbot: Is there a way to reject unknown options? (Probably better for the Python user experience if they misspelled something)

paleolimbot (commented):
I know you're still working on this...just some notes while I took a look before I'm done the work day!

python/sedonadb/python/sedonadb/context.py
zhangfengcdt: Sure, i just removed them from the pydoc to avoid confusion.

python/sedonadb/tests/test_context.py
zhangfengcdt: Sure, I have removed the other tests and add a new one specific to S3 anaonymous access: test_read_parquet_s3_anonymous_access

rust/sedona-geoparquet/src/provider.rs
zhangfengcdt: I added some logic to validate the aws options from the end users.

python/sedonadb/src/context.rs
zhangfengcdt: This should be much simplified using built-in .str() method.

python/sedonadb/src/context.rs
```diff
@@ -74,11 +74,33 @@ impl InternalContext {
...
+                        .map(|s: String| (k, s))
+                        .ok()
+                }
+            })
+            .collect();
+
+        let geo_options =
+            sedona_geoparquet::provider::GeoParquetReadOptions::from_table_options(rust_options)
+                .map_err(|e| {
+                    PySedonaError::SedonaPython(format!("Invalid table options: {}", e))
```
paleolimbot:
```suggestion
                    PySedonaError::SedonaPython(format!("Invalid table options: {e}"))
```

(Causes a lint error for newer rust versions)

python/sedonadb/tests/test_context.py
```diff
@@ -63,7 +63,81 @@ def test_dynamic_object_stores():
...
+
+
+def test_read_parquet_s3_options_parameter():
+    """Test that S3 options are accepted without errors (using HTTP URL for reliability)"""
     con = sedonadb.connect()
-    schema = pa.schema(con.sql(f"SELECT * FROM '{url}'"))
-    assert schema.field("geometry").type.extension_name == "geoarrow.wkb"
+    url = "https://raw.githubusercontent.com/geoarrow/geoarrow-data/v0.2.0/example/files/example_geometry_geo.parquet"
```
paleolimbot: Should this be `"s3://wherobots-examples/data/onboarding_1/nyc_buildings.parquet"`?

python/sedonadb/tests/test_context.py
```diff
@@ -63,7 +63,81 @@ def test_dynamic_object_stores():
...
+        s3_url, options={"aws.skip_signature": True, "aws.region": "us-west-2"}
+    ).to_arrow_table()
+    assert len(tab) > 0
+    assert "geom" in tab.schema.names  # This dataset uses 'geom' instead of 'geometry'
+
+
+def test_read_parquet_invalid_aws_option():
+    """Test that invalid AWS options are caught and provide helpful error messages"""
+    con = sedonadb.connect()
+    url = "https://raw.githubusercontent.com/geoarrow/geoarrow-data/v0.2.0/example/files/example_geometry_geo.parquet"
```
paleolimbot: Should this be an S3 URL?

rust/sedona/src/object_storage.rs
```diff
@@ -97,6 +122,11 @@ pub async fn ensure_object_store_registered(state: &mut SessionState, name: &str
     Ok(())
 }
 
+// Backward compatibility wrapper
+pub async fn ensure_object_store_registered(state: &mut SessionState, name: &str) -> Result<()> {
+    ensure_object_store_registered_with_options(state, name, None).await
+}
```
paleolimbot, line 128: Is this still needed?

rust/sedona/src/context.rs
```diff
@@ -232,18 +231,21 @@ impl SedonaContext {
...
+        // Pre-register object store with our custom options before creating GeoParquetReadOptions
+        if !urls.is_empty() {
+            use crate::object_storage::ensure_object_store_registered_with_options;
+            // Extract the table options from GeoParquetReadOptions for object store registration
+            let table_options_map = options.table_options().cloned().unwrap_or_default();
+            ensure_object_store_registered_with_options(
+                &mut self.ctx.state(),
+                urls[0].as_str(),
+                Some(&table_options_map),
+            )
```
paleolimbot: I'm curious if this would prevent future object store accesses from accessing credentials. For example, a user loads Overature Data with `aws.nosign` and then wants to load something from a private bucket (e.g., managed cloud storage). (I don't think we have to solve that now, just wondering 🙂 )

paleolimbot (approved):
Thank you!

A few optional things to consider now or in the future!

rust/sedona/src/context.rs
zhangfengcdt: Yes, I think currenly due to the way we register the storage store by scheme, once the store is initialized with nosig, it cannot be used to access the private bucket. However, if user initializes it with private credentials and then use it to access the public bucket, it will work.

For future fix, we can change to register the storage store by bucket (e.g., s3://bucket_name) instead of s3://, this will enable use to use different access patterns (public vs private) for different buckets.

I will add a todo comment to implement this.

rust/sedona/src/object_storage.rs
zhangfengcdt: This one is actually used in catalog.rs, and it was public.

PR #50, opened 2025-09-10
chore(ci): Move pyo3 extension module feature to Python args

paleolimbot:
The pyo3 feature extension-module enables the actual Python extension, which causes problems for some cargo commands of the parent repo (e.g., `cargo build --all`). I had added this so that we could add the s2geography feature optionally for MacOS and Linux builds and they failed without it; however, that can be specified elsewhere.

### DIFF
--- a/python/sedonadb/Cargo.toml
+++ b/python/sedonadb/Cargo.toml
@@ -38,7 +38,7 @@ datafusion = { workspace = true }
 datafusion-common = { workspace = true }
 datafusion-ffi = { workspace = true }
 futures = { workspace = true }
-pyo3 = { version = "0.25.1", features = ["extension-module"] }
+pyo3 = { version = "0.25.1" }
 sedona = { path = "../../rust/sedona" }
 sedona-adbc = { path = "../../rust/sedona-adbc" }
 sedona-geoparquet = { path = "../../rust/sedona-geoparquet" }
--- a/ci/scripts/wheels-build-linux.sh
+++ b/ci/scripts/wheels-build-linux.sh
@@ -48,7 +48,7 @@ BEFORE_ALL_MANYLINUX="yum install -y curl zip unzip tar clang perl"
 # add quite a bit of complexity but could save time if we build wheels for linux frequently.
 # The native and Rust builds are cached on each image such that compile work is effectively
 # cached between Python versions (just not between invocations of this script).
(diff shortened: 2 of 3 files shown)


python/sedonadb/Cargo.toml
```diff
@@ -38,7 +38,7 @@ datafusion = { workspace = true }
 datafusion-common = { workspace = true }
 datafusion-ffi = { workspace = true }
 futures = { workspace = true }
-pyo3 = { version = "0.25.1", features = ["extension-module"] }
+pyo3 = { version = "0.25.1" }
```
kylebarron, line 41: https://pyo3.rs/v0.26.0/faq#i-cant-run-cargo-test-or-i-cant-build-in-a-cargo-workspace-im-having-linker-issues-like-symbol-not-found-or-undefined-reference-to-_pyexc_systemerror

paleolimbot: Thanks! That was definitely the issue (and I think that the pyprojects.toml value was overridden when I passed `--features` in the wheel build 😬 )

Kontinuation:
Can we add a plain `cargo build` to our GitHub Workflow to catch this issue?

```diff
diff --git a/.github/workflows/rust.yml b/.github/workflows/rust.yml
index 1c57218..1c6bd19 100644
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -49,7 +49,7 @@ jobs:
   rust:
     strategy:
       matrix:
-        name: ["clippy", "docs", "test"]
+        name: ["clippy", "build", "docs", "test"]
 
     name: "${{ matrix.name }}"
     runs-on: ubuntu-latest
@@ -117,6 +117,11 @@ jobs:
         run: |
           cargo clippy --workspace --all-targets --all-features -- -Dwarnings
 
+      - name: Build
+        if: matrix.name == 'build'
+        run: |
+          cargo build --workspace --all-targets --all-features
+
       - name: Test
         if: matrix.name == 'test'
         run: |
```

Kontinuation:
> Sure, in parallel maybe? I don't use `cargo build` for anything...is it useful? (`cargo clippy` and `cargo test` have always worked for me).

I use `cargo build` to build everything including sedona-cli. I think keeping the project building cleanly using the most commonly used command for building most Rust project is a good idea. Unless using `cargo build` is generally a bad practice for building any Rust projects.

PR #51, opened 2025-09-10
bug: Fix `SpatialFilter` for `st_covers` and `st_contains` for correct pruning

### LINKED ISSUE #49: Reading GeoParquet with `ST_Within()` or `ST_Contains()` predicate produces incorrect results
I think that the bounding box containment for `ST_Within()` and `ST_Contains()` might be flipped?

```python
import sedonadb

sd = sedonadb.connect()

url = "https://github.com/geoarrow/geoarrow-data/releases/download/v0.2.0/microsoft-buildings_point_geo.parquet"
sd.read_parquet(url).to_view("buildings")

filter = "POLYGON ((-73.4341 44.0087, -73.4341 43.7981, -73.2531 43.7981, -73.2531 43.8889, -73.1531 43.8889, -73.1531 44.0087, -73.4341 44.0087))"

sd.sql(f"""
SELECT * FROM buildings
  WHERE ST_Intersects(ST_SetSRID(ST_GeomFromText('{filter}'), 4326), geometry)
""").count()
#> 6710 (takes ~30s)

sd.sql(f"""
SELECT * FROM buildings
  WHERE ST_Contains(ST_SetSRID(ST_GeomFromText('{filter}'), 4326), geometry)
""").count()
#> 0 (takes ~2s)

sd.sql(f"""
SELECT * FROM buildings
  WHERE ST_Within(ST_SetSRID(ST_GeomFromText('{filter}'), 4326), geometry)
""").count()
#> 0 (takes ~30s)
```

Also I think we may not be respecting the value of the pruning configuration option:

```python
sd.sql("SET datafusion.execution.parquet.pruning = false").show()
```

petern48:
This PR fixes an issue where the incorrect predicates were being used in pruning for the st_contains and st_covers functions. This now follows the same logic as used in the original Apache Sedona [here](https://github.com/apache/sedona/blob/3a1e9b7515b4dea994969f7f3c9ff43cf700e177/spark/common/src/main/scala/org/apache/spark/sql/sedona_sql/optimization/SpatialFilterPushDownForGeoParquet.scala#L116-L119)

closes #49

### DIFF
--- a/rust/sedona-expr/src/spatial_filter.rs
+++ b/rust/sedona-expr/src/spatial_filter.rs
@@ -231,21 +231,21 @@ impl SpatialFilter {
 
                 match (&args[0], &args[1]) {
                     (ArgRef::Col(column), ArgRef::Lit(literal)) => {
-                        // column contains/covers literal -> Intersects filter
-                        // (column must potentially intersect literal to contain it)
+                        // column contains/covers literal -> CoveredBy filter
+                        // (column's bbox must fully contain literal's bbox)
                         match literal_bounds(literal) {
                             Ok(literal_bounds) => {
-                                Ok(Some(Self::Intersects(column.clone(), literal_bounds)))
+                                Ok(Some(Self::CoveredBy(column.clone(), literal_bounds)))
                             }
                             Err(e) => Err(DataFusionError::External(Box::new(e))),
                         }
                     }
                     (ArgRef::Lit(literal), ArgRef::Col(column)) => {
-                        // literal contains/covers column -> CoveredBy filter
-                        // (equivalent to st_within(column, literal))
+                        // literal contains/covers column -> Intersects filter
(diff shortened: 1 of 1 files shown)


petern48:
Hi, it's me 🤠. I originally picked this up thinking this would be a simple one-liner fix. However, when I ran all of the predicates against geopandas output (the integration test), I saw there were more buggy predicate functions (e.g st_covers, st_contains, st_coveredby, st_within) than the original ones I expected.

Using new integration tests, I specifically found that the `(column, literal)` logic for `coveredby` was buggy (e.g [CI failure](https://github.com/apache/sedona-db/actions/runs/17628690672/job/50091513064)). I honestly didn't think too hard about why that logic is buggy, but I figured it's easier and apparently more correct to switch to the `Covers` filter logic the original Apache Sedona code uses.

You're welcome to propose a different direction, if you can think of one.

PR #52, opened 2025-09-10
feat(rust/sedona-geoparquet): Add GeoParquet writer for non-spatial output

paleolimbot:
This PR adds a writer for Parquet that writes GeoParquet metadata for output with spatial columns. Currently this doesn't write a bbox because that requires something a bit more sophisticated than just popping some metadata into the Parquet write options.

```python
import sedonadb

sd = sedonadb.connect()

sd.sql("SELECT ST_SetSRID(ST_GeomFromText('POINT (0 1)'), 4326) as geometry").to_parquet("foofy.parquet")
sd.read_parquet("foofy.parquet").show()
#> ┌────────────┐
#> │  geometry  │
#> │  geometry  │
#> ╞════════════╡
#> │ POINT(0 1) │
#> └────────────┘
```

### DIFF
--- a/rust/sedona-geoparquet/src/options.rs
+++ b/rust/sedona-geoparquet/src/options.rs
@@ -1,0 +1,32 @@
+use datafusion::config::TableParquetOptions;
+
+/// [TableParquetOptions] wrapper with GeoParquet-specific options
+#[derive(Debug, Default, Clone)]
+pub struct TableGeoParquetOptions {
+    pub inner: TableParquetOptions,
+    pub geoparquet_version: GeoParquetVersion,
+}
+
+impl From<TableParquetOptions> for TableGeoParquetOptions {
+    fn from(value: TableParquetOptions) -> Self {
+        Self {
+            inner: value,
+            geoparquet_version: GeoParquetVersion::default(),
+        }
+    }
+}
+
+#[derive(Debug, Clone, Copy)]
+pub enum GeoParquetVersion {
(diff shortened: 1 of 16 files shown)


rust/sedona-geoparquet/src/options.rs
```diff
@@ -0,0 +1,58 @@
...
+    /// of spatial columns is Parquet native geometry, where the Parquet writer will include
+    /// native statistics according to the underlying Parquet options. Some readers
+    /// (e.g., SedonaDB) can use these statistics to prune row groups on read.
+    V2_0,
+
+    /// Do not write GeoParquet metadata
+    ///
+    /// This option suppresses GeoParquet metadata; however, spatial types will be written as
+    /// Parquet native Geometry/Geography when this is supported by the underlying writer.
+    Omitted,
```
paleolimbot, line 51: @jiayuasu Only `V1_0` is implemented in this PR, but is this is what I had in mind for how a user will specify output.

jiayuasu: I think this is good enough. Since this PR follows `V1_0`, does it have bbox value in its file metadata?

paleolimbot: Not in this PR, because that requires a true override/copy of the `ParquetSink`. I will probably try `V1_1` before attempting that (also doesn't involve an override of the sink because it is a simple projection).

PR #53, opened 2025-09-10
Troubleshoot performance issue in spatial bench issue in knn join - q9

zhangfengcdt:


jiayuasu:
Which side got cached?  The trip table will grow linearly together with Scale Factor. If this just caches the buildings table, we should be good.

rust/sedona-spatial-join/src/index.rs
```diff
@@ -331,6 +358,14 @@ pub(crate) struct SpatialIndex {
...
     #[expect(dead_code)]
     reservation: MemoryReservation,
+
+    /// Cached vector of geometries for KNN queries to avoid repeated WKB-to-geometry conversions
+    /// This is computed once during index building for performance optimization
+    cached_geometries: Vec<Geometry<f64>>,
+
+    /// Cached position mapping for geometries to avoid recomputing during KNN queries
+    /// Maps geometry index to (batch_idx, row_idx) pairs
+    cached_geometry_positions: Vec<(i32, i32)>,
```
Kontinuation: There's no need to add a new field. We can use `data_id_to_batch_pos` for mapping the index retrieved from rtree to (batch_idx, row_idx).

rust/sedona-spatial-join/src/index.rs
```diff
@@ -271,6 +292,10 @@ impl SpatialIndexBuilder {
             ConcurrentReservation::try_new(REFINER_RESERVATION_PREALLOC_SIZE, refiner_reservation)
                 .unwrap();
 
+        // Pre-compute geometries for KNN queries to avoid repeated WKB-to-geometry conversions
+        let (cached_geometries, cached_geometry_positions) =
+            Self::build_cached_geometries(&self.indexed_batches);
```
Kontinuation: Additionally memory needs to be reserved for cached geometries. It is hard to know the exact amount of memory taken by `geo::Geometry` objects, but at least we can have a rough estimation.

zhangfengcdt:
> Which side got cached? The trip table will grow linearly together with Scale Factor. If this just caches the buildings table, we should be good.

Yes, this is only cache geometries on the buildings dataset for the Q9.

zhangfengcdt:
> 1. I regret not spotting this issue when I first reviewed the code.
> 2. [suggestion] The `neighbors_geometry` method defined in geo-index does not have to take an array of `geo::Geometry` as a parameter. We can play all kinds of tricks in the `distance_metric` function to compute a distance value given 2 indices of indexed items. This makes the interface more flexible, and we are free to play all kinds of tricks inside our custom distance metric function. For instance, decoding the WKB on demand and caching decoded geometries for reuse.

It's a great suggestion! And, yes, using a custom distance metric with index-based lookups would be more flexible and could potentially offer even better performance optimizations. However, I'd like to keep this PR focused on the immediate performance fix we're addressing. This keeps the current fix simple and focused while leaving room for future optimization. 

The suggested approach would require:
  - Changes to the geo-index crate's neighbors_geometry API
  - A custom distance metric implementation
  - More complex testing across two repositories
  - Additional coordination between geo-index and sedona-spatial-join

I think this optimization deserves its own focused PR where we can change both repos and test them accordingly.

What do you think?

zhangfengcdt:
CC: @prantogg This is a fix we need to fix the performance issues in spatial bench Q9.

rust/sedona-spatial-join/src/index.rs
```diff
@@ -235,25 +235,76 @@ impl SpatialIndexBuilder {
...
+                                    .map(|ring| ring.coords().count() * COORDINATE_SIZE)
+                                    .sum();
+                                exterior + interiors
+                            })
+                            .sum::<usize>(),
+                        _ => 256, // Conservative estimate for other geometry types
+                    }
+            })
+            .sum()
     }
```
Kontinuation: This is overly complicated but even less accurate than directly using the length of WKB buffer. I'd rather just take the length of WKB buffer plus or multiply some overhead.

benchmarks/test_knn.py
```diff
@@ -0,0 +1,225 @@
...
+from sedonadb.testing import SedonaDB
+
+
+class TestBenchKNN(TestBenchBase):
+    def setup_class(self):
+        """Setup test data for KNN benchmarks"""
+        self.sedonadb = SedonaDB.create_or_skip()
+
+        # Create building-like polygons (index side - fewer, larger geometries)
+        # Note: Dataset sizes are limited to avoid DataFusion CoalescePartitionsExec issues
```
Kontinuation: What is the "DataFusion CoalescePartitionsExec issue"? Subsequent comments also mention "partition issues". Can you explain in the comment what the so-called "partition issue" is?

zhangfengcdt: This is from issues I see when i was debugging into the Datafusion code and find when calling .collect() from python api, the CoalescePartitionsExec was triggered and caused OOM issues for very large test data sizes.  It may not be a real "partition issue", but I assume CoalescePartitionsExec might have issues when combining so many small partitions in preparing the join data. It is an assumption though cause I cannot find real-world data to verify it.  But yes, I've rephrased the comments to be more accurate.

PR #57, opened 2025-09-11
Disable optimized spatial join and GeoParquet data pruning for geography types

Kontinuation:
As mentioned in https://github.com/apache/sedona-db/pull/39, our GeoParquet data pruning optimization and spatial join optimization does not handle geography types correctly, so we'd better disable them to avoid generating incorrect query results.

### DIFF
--- a/python/sedonadb/tests/test_sjoin.py
+++ b/python/sedonadb/tests/test_sjoin.py
@@ -78,29 +78,63 @@ def test_spatial_join(join_type, on):
     eng_postgis.assert_query_result(sql, sedonadb_results)
 
 
-@pytest.mark.xfail(reason="https://github.com/apache/sedona-db/issues/63")
-def test_spatial_join_geography(geoarrow_data):
+@pytest.mark.parametrize(
+    "join_type", ["INNER JOIN", "LEFT OUTER JOIN", "RIGHT OUTER JOIN"]
+)
+@pytest.mark.parametrize(
+    "on",
+    [
+        "ST_Intersects(sjoin_geog1.geog, sjoin_geog2.geog)",
+        "ST_Distance(sjoin_geog1.geog, sjoin_geog2.geog) < 100000",
+    ],
+)
+def test_spatial_join_geography(join_type, on):
     eng_sedonadb = SedonaDB.create_or_skip()
     eng_postgis = PostGIS.create_or_skip()
 
-    eng_sedonadb.create_table_parquet(
-        "sjoin_geog_all",
-        geoarrow_data
-        / "natural-earth/files/natural-earth_countries-geography_geo.parquet",
+    # Select two sets of bounding boxes that cross the antimeridian,
+    # which would be disjoint on a Euclidean plane. A geography join will produce non-empty results,
+    # whereas a geometry join would not.
+    west_most_bound = [-190, -10, -170, 10]
+    east_most_bound = [170, -10, 190, 10]
+    options = json.dumps(
+        {
+            "geom_type": "Point",
+            "num_parts_range": [2, 10],
+            "vertices_per_linestring_range": [2, 10],
+            "bounds": west_most_bound,
+            "size_range": [0.1, 5],
+            "seed": 42,
+        }
     )
-    test_data = eng_sedonadb.execute_and_collect(
-        "SELECT * FROM sjoin_geog_all LIMIT 100"
+    df_point = eng_sedonadb.execute_and_collect(
+        f"SELECT id, ST_SetSRID(ST_GeogFromWKB(ST_AsBinary(geometry)), 4326) geog, dist FROM sd_random_geometry('{options}') LIMIT 100"
     )
-    eng_sedonadb.create_table_arrow("sjoin_geog", test_data)
-    eng_postgis.create_table_arrow("sjoin_geog", test_data)
+    options = json.dumps(
+        {
+            "geom_type": "Polygon",
+            "polygon_hole_rate": 0.5,
+            "num_parts_range": [2, 10],
+            "vertices_per_linestring_range": [2, 10],
+            "bounds": east_most_bound,
+            "size_range": [0.1, 5],
+            "seed": 43,
+        }
+    )
+    df_polygon = eng_sedonadb.execute_and_collect(
+        f"SELECT id, ST_SetSRID(ST_GeogFromWKB(ST_AsBinary(geometry)), 4326) geog, dist FROM sd_random_geometry('{options}') LIMIT 100"
+    )
+    eng_sedonadb.create_table_arrow("sjoin_geog1", df_point)
+    eng_sedonadb.create_table_arrow("sjoin_geog2", df_polygon)
+    eng_postgis.create_table_arrow("sjoin_geog1", df_point)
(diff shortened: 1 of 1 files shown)


rust/sedona-geoparquet/src/file_opener.rs
```diff
@@ -299,6 +314,9 @@ fn parse_column_coverings(
             if !metadata.columns.contains_key(field.name()) {
                 return Ok(None);
             }
+            if !is_prunable_geospatial_field(field) {
+                return Ok(None);
+            }
```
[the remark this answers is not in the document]
Kontinuation: It is probably a good idea. Fixed.

rust/sedona-geoparquet/src/file_opener.rs
```diff
@@ -200,7 +200,13 @@ fn geoparquet_file_geo_stats(
...
             // (otherwise, construct unspecified statistics)
             if let Some(column_metadata) = metadata.columns.get(field.name()) {
-                Ok(column_metadata.to_geo_statistics())
+                if is_prunable_geospatial_field(field) {
+                    Ok(column_metadata.to_geo_statistics())
+                } else {
+                    // Bounding box based pruning does not work for geography fields, so we remove
+                    // the bbox from statistics to ensure that they are not used for pruning.
+                    Ok(column_metadata.to_geo_statistics().with_bbox(None))
+                }
```
paleolimbot: Should we handle this in `literal_bounds()` by returning a Full bounding box?

https://github.com/apache/sedona-db/blob/3e9b7a97f9b5b925ec2f01e0a235190c96e3fc89/rust/sedona-expr/src/spatial_filter.rs#L317-L319

(Strictly speaking there's nothing wrong with the information in the file, it's just that we can't bound the literal)

rust/sedona-spatial-join/src/optimizer.rs
```diff
@@ -2418,4 +2466,39 @@ mod tests {
         let result = transform_join_filter(&join_filter);
         assert!(result.is_none()); // Should fail - k must be a literal value
     }
+
+    #[test]
+    fn test_is_spatial_predicate_supported() {
+        // Planar geometry field
+        let geom_field = WKB_GEOMETRY.to_storage_field("geom", false).unwrap();
+        let schema = Arc::new(Schema::new(vec![geom_field.clone()]));
```
paleolimbot, line 2474: This would benefit from a high-level integration test in Python (i.e., make sure that a join with geography returns correct results). A join between `"submodules/geoarrow-data/natural-earth/files/natural-earth_countries-geography_geo.parquet"` and `"submodules/geoarrow-data/natural-earth/files/natural-earth_countries-geography_geo.parquet"` might be a good candidate.

rust/sedona-geoparquet/src/file_opener.rs
Kontinuation: Good idea.

Kontinuation: I found that we cannot return a full bounding box for geography. We are pushing down ST_Contains and ST_Covers predicates as `SpatialFilter::Covers`, which requires the bbox of the geospatial column to contain the query window. Returning a full bounding box for geography will skip everything in such cases.

I'll try a different approach. The modification will still go into spatial_filter.rs and keep the bboxes in geo-statistics retrieved from GeoParquet files intact.

Kontinuation:
I tried to use `submodules/geoarrow-data/natural-earth/files/natural-earth_countries-geography_geo.parquet` for running GeoParquet data pruning test, but found that the bbox of this file cannot be parsed because it does not conform to the standard.

Here is the file level geo metadata of that Parquet file:

```json
{"version": "1.0.0", "primary_column": "geometry", "columns": {"geometry": {"encoding": "WKB", "geometry_types": [], "crs": {"$schema": "https://proj.org/schemas/v0.7/projjson.schema.json", "type": "GeographicCRS", "name": "WGS 84 (CRS84)", "datum_ensemble": {"name": "World Geodetic System 1984 ensemble", "members": [{"name": "World Geodetic System 1984 (Transit)", "id": {"authority": "EPSG", "code": 1166}}, {"name": "World Geodetic System 1984 (G730)", "id": {"authority": "EPSG", "code": 1152}}, {"name": "World Geodetic System 1984 (G873)", "id": {"authority": "EPSG", "code": 1153}}, {"name": "World Geodetic System 1984 (G1150)", "id": {"authority": "EPSG", "code": 1154}}, {"name": "World Geodetic System 1984 (G1674)", "id": {"authority": "EPSG", "code": 1155}}, {"name": "World Geodetic System 1984 (G1762)", "id": {"authority": "EPSG", "code": 1156}}, {"name": "World Geodetic System 1984 (G2139)", "id": {"authority": "EPSG", "code": 1309}}], "ellipsoid": {"name": "WGS 84", "semi_major_axis": 6378137, "inverse_flattening": 298.257223563}, "accuracy": "2.0", "id": {"authority": "EPSG", "code": 6326}}, "coordinate_system": {"subtype": "ellipsoidal", "axis": [{"name": "Geodetic longitude", "abbreviation": "Lon", "direction": "east", "unit": "degree"}, {"name": "Geodetic latitude", "abbreviation": "Lat", "direction": "north", "unit": "degree"}]}, "scope": "Not known.", "area": "World.", "bbox": {"south_latitude": -90, "west_longitude": -180, "north_latitude": 90, "east_longitude": 180}, "id": {"authority": "OGC", "code": "CRS84"}}, "edges": "spherical"}}}
```

The bbox field is:

```json
"bbox": {
  "south_latitude": -90,
  "west_longitude": -180,
  "north_latitude": 90,
  "east_longitude": 180
}
```

rust/sedona-spatial-join/src/optimizer.rs
Kontinuation: I tried to add this test case, but found that SedonaDB gives wildly different results from PostGIS. S2 is less likely to consider polygons that touch at boundaries as intersections. I can still add this Python test, but mark it as xfail.

python/sedonadb/tests/test_sjoin.py
```diff
@@ -76,3 +76,31 @@ def test_spatial_join(join_type, on):
...
+    eng_sedonadb.create_table_parquet(
+        "sjoin_geog_all",
+        geoarrow_data
+        / "natural-earth/files/natural-earth_countries-geography_geo.parquet",
+    )
+    test_data = eng_sedonadb.execute_and_collect(
+        "SELECT * FROM sjoin_geog_all LIMIT 100"
+    )
+    eng_sedonadb.create_table_arrow("sjoin_geog", test_data)
+    eng_postgis.create_table_arrow("sjoin_geog", test_data)
```
paleolimbot: I pasted this into the issue so we can track it there...in the meantime, we can either remove this or pick a less hard join (e.g., pick a small number of hard-coded or randomly generated points). This test is mostly just making sure that it works at all.

Kontinuation: I switched to using randomly generated data for this geography join test. The bounds of generated data are disjoint on planar surface but intersects the antimeridian on spherical surface. This distribution of data ensures that we won't pass the test when handling geography objects as planar geometry objects.

Issue #61, opened 2025-09-11
Accept SRIDs as args to ST_Transform


jesspav:
`ST_Transform` is not parsing SRIDs as arguments.

For example:
```
> SELECT ST_Transform(ST_GeomFromText('POINT (1 1)'), '4326', '3857') as geom;
proj_create: unrecognized format / unknown name
proj_create: unrecognized format / unknown name
proj_create: unrecognized format / unknown name
proj_create: unrecognized format / unknown name
proj_create: unrecognized format / unknown name
proj_create: unrecognized format / unknown name
proj_create: unrecognized format / unknown name
proj_create: unrecognized format / unknown name
proj_create: unrecognized format / unknown name
proj_create: unrecognized format / unknown name
proj_create: unrecognized format / unknown name
proj_create: unrecognized format / unknown name
proj_create: unrecognized format / unknown name
Execution error: Transform error: Failed to create CRS from source '4326': Invalid PROJ string syntax
```

jesspav:
Fixed!
```
> SELECT ST_Transform(ST_GeomFromText('POINT (1 1)'), '4326', '3857') as geom;
┌─────────────────────────────────────────────┐
│                     geom                    │
│                   geometry                  │
╞═════════════════════════════════════════════╡
│ POINT(111319.49079327357 111325.1428663851) │
└─────────────────────────────────────────────┘

1 row(s)/1 column(s) fetched. 
Elapsed 0.043 seconds.

> SELECT ST_Transform(ST_GeomFromText('POINT (1 1)'), 4326, 3857) as geom;
┌─────────────────────────────────────────────┐
│                     geom                    │
│                   geometry                  │
╞═════════════════════════════════════════════╡
│ POINT(111319.49079327357 111325.1428663851) │
└─────────────────────────────────────────────┘

1 row(s)/1 column(s) fetched. 
Elapsed 0.008 seconds.

> 
```

PR #62, opened 2025-09-11
[#61] Accept SRIDs as args to ST_Transform 

jesspav:
Currently, `ST_Transform` is only accepting utf8 args. 
This change allows for SRIDs to be passed in as ints.  It also adds some better argument checks.

Issue: #61

I tried going through a `CRS` and passing the `to_json()` output as the CRS args to the transform, but proj was failing to parse the json.  I'll need to figure that out separately, but accepting the SRID seemed like a P0 issue, so fixing that first.

## Testing

Unit tests and manual tests.

## Examples

```
> SELECT ST_Transform(ST_SetSrid(ST_GeomFromText('POINT (1 1)'), 4326), 3857) as geom;
┌─────────────────────────────────────────────┐
│                     geom                    │
│                   geometry                  │
╞═════════════════════════════════════════════╡
│ POINT(111319.49079327357 111325.1428663851) │
└─────────────────────────────────────────────┘

> SELECT ST_Transform(ST_SetSrid(ST_GeomFromText('POINT (1 1)'), 4326), 'EPSG:3857') as geom;
┌─────────────────────────────────────────────┐
│                     geom                    │
│                   geometry                  │
╞═════════════════════════════════════════════╡
│ POINT(111319.49079327357 111325.1428663851) │
└─────────────────────────────────────────────┘

```

### DIFF
--- a/c/sedona-proj/src/st_transform.rs
+++ b/c/sedona-proj/src/st_transform.rs
@@ -158,21 +158,19 @@ impl SedonaScalarKernel for STTransform {
         scalar_args: &[Option<&ScalarValue>],
     ) -> Result<Option<SedonaType>> {
         let matcher = ArgMatcher::new(
-           vec![
+            vec![
                 ArgMatcher::is_geometry_or_geography(),
-                ArgMatcher::or(vec![
-                    ArgMatcher::is_numeric(),
-                    ArgMatcher::is_string()
-                ]),
+                ArgMatcher::or(vec![ArgMatcher::is_numeric(), ArgMatcher::is_string()]),
                 ArgMatcher::optional(ArgMatcher::or(vec![
                     ArgMatcher::is_numeric(),
-                    ArgMatcher::is_string()
+                    ArgMatcher::is_string(),
                 ])),
                 ArgMatcher::optional(ArgMatcher::is_boolean()),
-            ], SedonaType::Wkb(Edges::Planar, None)
+            ],
+            SedonaType::Wkb(Edges::Planar, None),
         );
 
-        if !matcher.match_args(arg_types) {
+        if !matcher.matches(arg_types) {
             return Ok(None);
         }
 
@@ -216,7 +214,7 @@ impl SedonaScalarKernel for STTransform {
         let mut indexes = TransformArgIndexes::new();
         define_arg_indexes(arg_types, &mut indexes);
 
-        let first_crs = get_crs_str(args, indexes.first_crs).ok_or_else(|| {;
+        let first_crs = get_crs_str(args, indexes.first_crs).ok_or_else(|| {
             DataFusionError::Execution(
                 "First CRS argument must be a string or numeric scalar".to_string(),
             )


c/sedona-proj/src/st_transform.rs
```diff
@@ -135,7 +136,9 @@ fn define_arg_indexes(arg_types: &[SedonaType], indexes: &mut TransformArgIndexe
     indexes.first_crs = 1;
 
     for (i, arg_type) in arg_types.iter().enumerate().skip(2) {
-        if *arg_type == SedonaType::Arrow(DataType::Utf8) {
+        if ArgMatcher::is_numeric().match_type(arg_type)
```
paleolimbot, line 139: Is it worth adding `ArgMatcher::is_integer_or_string()`?

c/sedona-proj/src/st_transform.rs
```diff
@@ -157,19 +160,40 @@ impl SedonaScalarKernel for STTransform {
         let mut indexes = TransformArgIndexes::new();
         define_arg_indexes(arg_types, &mut indexes);
 
-        let to_crs_opt = if let Some(second_crs_index) = indexes.second_crs {
+        if (arg_types.len() < 2) || (arg_types.len() > 4) {
+            return Ok(None);
+        }
```
paleolimbot: I wonder if this would be any cleaner by registering multiple kernels:

```rust
struct STTransformTo {} (matches `vec![ArgMatcher::is_geometry_or_geography(), ArgMatcher::is_integer_or_string())`)
struct STTransformFromTo {} (matches `vec![ArgMatcher::is_geometry_or_geography(), ArgMatcher::is_integer_or_string()ArgMatcher::is_integer_or_string())`)
```

I am still lukewarm on including `lenient` here (we can solve compatibility with Sedona in other ways and the default assumption of a lon/lat CRS isn't something we do elsewhere in SedonaDB), although it could be included in this way as `ArgMatcher::is_optional()`.

(Or not!)

jesspav: I don't love lenient either, I just want folks to have a nice experience bringing their queries from Sedona to Sedona-db, and was told at some point that was important.  I wish we had better data on how many people use it.

jiayuasu: I prefer we enforce the lon lat order in SedonaDB otherwise this will create other issues in other functions such as H3 functions, GeoHash, S2 cell ids, and Kepler as well

jiayuasu: Since these functions are also sensitive to Lon lat order

jesspav: I could add an `or()` similar to the way I implement `is_optional()` (which I probably should have called `optional()`

jesspav: Something like: https://github.com/apache/sedona-db/compare/main...jesspav:sedona-db:or

jesspav: On the long/lat issue, it seems like we are missing: `ST_FlipCoordinates`.  Having that could help with those functions.  I can quickly implement it.  Thoughts?

paleolimbot: - I'm OK with any version of `lenient` here....the situation of an "unset" CRS is actually pretty rare in SedonaDB (because we always propagate CRSes from our sources) and so the implicit assumption that it is lon/lat *sometimes* is something I don't mind fixing later on if it comes up. Here I was more observing that it created some complexity with the arguments that might have been related to the failed CI, so it might be worth removing it. Also fine to leave it alone!
- Jess is right that the lon/lat lat/lon ordering is a separate issue. Feel free to open an issue but I'm also not super worried about that one as a priority for the initial release.

paleolimbot: > I could add an or() similar to the way I implement is_optional() (which I probably should have called optional()

All good ideas! I think you've tested this appropriately and as long as they pass I'm happy with whatever you come up with!

jesspav: PR for `ArgMatcher::or`: https://github.com/apache/sedona-db/pull/69

jesspav: PR for `ST_FlipCoordinates`: #67

c/sedona-proj/src/st_transform.rs
jesspav: Added an `or` arg matcher (https://github.com/apache/sedona-db/pull/69) and used it here.

PR #69, opened 2025-09-12
Add argument matcher or

jesspav:
Adding an or argument matcher for when multiple argument types are ok, for example accepting both SRIDs (numeric) and CRS authority codes (utf8) to a transform.

### DIFF
--- a/rust/sedona-schema/src/matchers.rs
+++ b/rust/sedona-schema/src/matchers.rs
@@ -271,7 +271,7 @@ impl TypeMatcher for OrMatcher {
     }
 
     fn type_if_null(&self) -> Option<SedonaType> {
-        self.matchers.get(0).and_then(|m| m.type_if_null())
+        None
     }
 }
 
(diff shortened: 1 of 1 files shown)


rust/sedona-schema/src/matchers.rs
```diff
@@ -253,6 +260,21 @@ impl TypeMatcher for OptionalMatcher {
...
+    matchers: Vec<Arc<dyn TypeMatcher + Send + Sync>>,
+}
+
+impl TypeMatcher for OrMatcher {
+    fn match_type(&self, arg: &SedonaType) -> bool {
+        self.matchers.iter().any(|m| m.match_type(arg))
+    }
+
+    fn type_if_null(&self) -> Option<SedonaType> {
+        self.matchers.get(0).and_then(|m| m.type_if_null())
```
paleolimbot: I think we might have to return `None` here (i.e., we're not sure what the creator of this matcher wants a null to be cast to).

jesspav: sounds good.

PR #71, opened 2025-09-12
Fix pytest hanging when test fails by closing PostgreSQL connection using `with` statement

Kontinuation:
This patch depends on https://github.com/apache/sedona-db/pull/70.

The pytest may hang when some of the test fails. This is a workflow run that exhibited this problem: https://github.com/apache/sedona-db/actions/runs/17668086933/job/50213528921

Pytest does not free up resources held by local variables by calling `__del__` in time when a test fails. When pytest captured the AssertionError, it effectively adds a reference to the local c object, preventing it from being __del__-ed when the function ends. This leaves the transaction started by the failed test active when the next test starts running.

We generate test data by dropping a temporary table, inserting data into the temporary table, and finally renaming the temporary table. Dropping the temporary table will be blocked by a lock held by the still-active transaction started by the previously failed test:

```
postgres=# select pid, query, state, wait_event, wait_event_type, backend_type from pg_stat_activity;
  pid  |                                           query                                            |        state        |     wait_event      | wait_event_type |         backend_type         
-------+--------------------------------------------------------------------------------------------+---------------------+---------------------+-----------------+------------------------------
 99895 |                                                                                            |                     | AutoVacuumMain      | Activity        | autovacuum launcher
 99896 |                                                                                            |                     | LogicalLauncherMain | Activity        | logical replication launcher
 83813 | select pid, query, state, wait_event, wait_event_type, backend_type from pg_stat_activity; | active              |                     |                 | client backend
 47073 | DROP TABLE IF EXISTS "public" . "sjoin_polygon_xtemp"                                      | idle in transaction | ClientRead          | Client          | client backend
 51565 | DROP TABLE IF EXISTS "public" . "sjoin_point_xtemp"                                        | active              | relation            | Lock            | client backend
 66330 | DROP TABLE IF EXISTS "public" . "sjoin_point_xtemp"                                        | active              | relation            | Lock            | client backend
 71494 | DROP TABLE IF EXISTS "public" . "sjoin_point_xtemp"                                        | active              | relation            | Lock            | client backend
 99892 |                                                                                            |                     | BgWriterHibernate   | Activity        | background writer
 99891 |                                                                                            |                     | CheckpointerMain    | Activity        | checkpointer
 99894 |                                                                                            |                     | WalWriterMain       | Activity        | walwriter

```

This patch fixes this problem by managing connections using `with` statement. The transaction will be committed or rolled back when the connection is closed. This prevents us from leaking transactions and blocking ourselves on table locks.

### DIFF
--- a/python/sedonadb/python/sedonadb/testing.py
+++ b/python/sedonadb/python/sedonadb/testing.py
@@ -63,6 +63,17 @@ class DBEngine:
     of that name.
     """
 
+    def close(self):
+        """Close the connection - base implementation does nothing"""
+        pass
+
+    def __enter__(self):
+        return self
+
+    def __exit__(self, exc_type, exc_val, exc_tb):
+        self.close()
+        return False
+
     @classmethod
     def name(cls) -> str:
         """This engine's name
@@ -398,6 +409,11 @@ class PostGIS(DBEngine):
         if uri is None:
             uri = "postgresql://localhost:5432/postgres?user=postgres&password=password"
         self.con = adbc_driver_postgresql.dbapi.connect(uri)
+
+    def close(self):
+        """Close the connection"""
+        if self.con:
+            self.con.close()
 
     @classmethod
     def name(cls):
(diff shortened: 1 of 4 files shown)


jiayuasu:
Please address the git conflict. Thanks! @Kontinuation

Issue #68, opened 2025-09-12
The barrier() operator is missing from the KNN join doc


jiayuasu:
@zhangfengcdt 

We need to update the md file here: https://github.com/apache/sedona-db/blob/main/docs/reference/sql-joins.md#k-nearest-neighbor-knn-join

PR #72, opened 2025-09-12
docs: Add barrier function documentation and use it in KNN join example

zhangfengcdt:
Address https://github.com/apache/sedona-db/issues/68


jiayuasu:
LGTM overall. Can you elaborate a bit more in the doc about why applying the filter before and after the KNN actually have different semantic meanings?

zhangfengcdt:
> LGTM overall. Can you elaborate a bit more in the doc about why applying the filter before and after the KNN actually have different semantic meanings?

Sure, explanations have been added with examples to the doc.

PR #73, opened 2025-09-12
ST_Distance and ST_DWithin based on georust/geo

Kontinuation:
This patch implements ST_Distance and ST_DWithin using georust/geo's Euclidean distance function.

Performance comparison with GEOS-based implementation:

GEO:

```
geo-st_distance-ArrayScalar(Point, Polygon(10))
                        time:   [19.120 ms 19.171 ms 19.227 ms]
geo-st_distance-ArrayScalar(Point, Polygon(500))
                        time:   [446.14 ms 448.10 ms 450.64 ms]
```

GEOS:

```
geos-st_distance-ArrayScalar(Polygon(10), Polygon(10))
                        time:   [187.51 ms 188.38 ms 189.72 ms]
geos-st_distance-ArrayScalar(Polygon(10), Polygon(500))
                        time:   [2.8651 s 2.8735 s 2.8822 s]
```

Benchmarking ternary functions such as ST_DWithin takes forever, it could be a problem with our benchmarking framework.

### DIFF
--- a/rust/sedona-geo/src/st_dwithin.rs
+++ b/rust/sedona-geo/src/st_dwithin.rs
@@ -18,17 +18,15 @@ use std::sync::Arc;
 
 use arrow_array::builder::BooleanBuilder;
 use arrow_schema::DataType;
-use datafusion_common::{error::Result, DataFusionError};
+use datafusion_common::{cast::as_float64_array, error::Result};
 use datafusion_expr::ColumnarValue;
-use geo::{Distance, Euclidean};
+use geo_generic_alg::line_measures::DistanceExt;
 use sedona_expr::scalar_udf::{ScalarKernelRef, SedonaScalarKernel};
 use sedona_functions::executor::WkbExecutor;
 use sedona_schema::{datatypes::SedonaType, matchers::ArgMatcher};
 use wkb::reader::Wkb;
 
-use crate::to_geo::item_to_geometry;
-
-/// ST_DWithin() implementation using [Euclidean] [Distance]
+/// ST_DWithin() implementation using [DistanceExt]
 pub fn st_dwithin_impl() -> ScalarKernelRef {
     Arc::new(STDWithin {})
 }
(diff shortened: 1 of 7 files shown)


jesspav:
Looks great!
Do these functions not apply to geographies?

Kontinuation:
> Looks great! Do these functions not apply to geographies?

Geographies has its own implementation of ST_Distance: https://github.com/apache/sedona-db/blob/3d0122730166c6b9eb5e9fc303d0262829673191/c/sedona-s2geography/src/scalar_kernel.rs#L85-L91

rust/sedona-geo/src/st_dwithin.rs
```diff
@@ -0,0 +1,166 @@
...
+                distance = None;
+            } else {
+                distance = Some(f64::try_from(scalar_arg.clone())?);
+            }
+        } else {
+            return Err(DataFusionError::Execution(format!(
+                "Invalid distance: {:?}",
+                args[2]
+            )));
+        }
```
paleolimbot: This will error when `distance` isn't a constant? I think you could probably handle this case with:

```rust
let arg2_array = arg2.to_array(executor.num_iterations())?;
let arg2_f64_array = as_float64_array(&arg2_array)?;
let arg2_iter = arg2_f64_array.iter();

// In the loop
let distance = arg2_iter.next().unwrap()
```

That is possibly slower for the scalar case and could be kept separate if it ends up mattering.

rust/sedona-geo/src/st_dwithin.rs
```diff
@@ -0,0 +1,166 @@
...
+            Ok(())
+        })?;
+
+        executor.finish(Arc::new(builder.finish()))
+    }
+}
+
+fn invoke_scalar(wkb_a: &Wkb, wkb_b: &Wkb, distance_bound: f64) -> Result<bool> {
+    let geom_a = item_to_geometry(wkb_a)?;
+    let geom_b = item_to_geometry(wkb_b)?;
```
paleolimbot: Just double-checking: do we need these conversions, or is the distance metric implemented for the generic case?

Also, it should be more efficient to use the `GeoTypesExecutor`, which should ensure that scalar inputs on either the right or left are only converted once.

rust/sedona-geo/src/st_dwithin.rs
Kontinuation: I am simply replicating what st_dwithin for GEOS does: https://github.com/apache/sedona-db/blob/aa4f80a5732e0799e8637930304e86ad2d7c46bb/c/sedona-geos/src/st_dwithin.rs#L56-L70

I'll address the same problem in GEOS UDF code in this patch.

rust/sedona-geo/src/st_dwithin.rs
Kontinuation: The generic distance PR was merged 1 hour ago. I'll remove the conversion and call distance_ext directly on WKB values.

Issue #30, opened 2025-09-05
bug: Reduce error output for ST_Transform


jesspav:
Invalid transformations using proj results in a mountain of text to stdout. 

Need to investigate whether we can set the log level.

## Repro:

```
SELECT ST_Transform(ST_Point(-200, 0), 4326, 3857) 
```

jesspav:
Based on these queries I believe that it is the `proj_create: unrecognized format / unknown name` that we are being asked to suppress.

```
> SELECT ST_Transform(ST_GeomFromText('POINT (1 1)'), 4326, 3857) as geom;
Execution error: First argument must be a scalar string

> SELECT ST_Transform(ST_GeomFromText('POINT (1 1)'), '4326', '3857') as geom;
proj_create: unrecognized format / unknown name
proj_create: unrecognized format / unknown name
proj_create: unrecognized format / unknown name
proj_create: unrecognized format / unknown name
proj_create: unrecognized format / unknown name
proj_create: unrecognized format / unknown name
proj_create: unrecognized format / unknown name
proj_create: unrecognized format / unknown name
proj_create: unrecognized format / unknown name
proj_create: unrecognized format / unknown name
proj_create: unrecognized format / unknown name
proj_create: unrecognized format / unknown name
proj_create: unrecognized format / unknown name
Execution error: Transform error: Failed to create CRS from source '4326': Invalid PROJ string syntax

> SELECT ST_Transform(ST_GeomFromText('POINT (1 1)'), 'EPSG:4326', 'EPSG:3857') as geom;
┌─────────────────────────────────────────────┐
│                     geom                    │
│                   geometry                  │
╞═════════════════════════════════════════════╡
│ POINT(111319.49079327357 111325.1428663851) │
└─────────────────────────────────────────────┘

1 row(s)/1 column(s) fetched. 
Elapsed 0.007 seconds.
```

jesspav:
I think that we should also be changing `4326` to `EPSG:4326` before passing it to proj.

paleolimbot:
> I think that we should also be changing 4326 to EPSG:4326 before passing it to proj.

Agreed! I'd thought that it was handled here:

https://github.com/apache/sedona-db/blob/5633898f591c88b9d1c4db8f292d40cb330bfb0e/rust/sedona-schema/src/crs.rs#L136-L168

...but it's maybe not making its way all the way through the system (or maybe we're not deserializing a CRS when it's used as input to transform and/or setsrid/crs).

[...]

PR #77, opened 2025-09-13
Default proj log level to none

jesspav:
Currently, invalid inputs to proj via `ST_Transform` are resulting in large log dumps to std out.

For example, for a single row:
```
> SELECT ST_Transform(ST_GeomFromText('POINT (1000 1000)'), 'EPSG:4326', 'EPSG:3857') as geom;
webmerc: Invalid latitude
webmerc: Invalid latitude
webmerc: Invalid latitude
webmerc: Invalid latitude
webmerc: Invalid latitude
webmerc: Invalid latitude
webmerc: Invalid latitude
webmerc: Invalid latitude
webmerc: Invalid latitude
webmerc: Invalid latitude
webmerc: Invalid latitude
webmerc: Invalid latitude
webmerc: Invalid latitude
Execution error: Transform error: PROJ coordinate transformation failed with error: Invalid coordinate
```

Since we are already capturing and sharing the error message, this change lowers the log level to none/0, but allows users to override it.

After this change, the excess error messages no longer appear and the user can jump straight to the useful error message. For example:
```
> SELECT ST_Transform(ST_GeomFromText('POINT (1000 1000)'), 'EPSG:4326', 'EPSG:3857') as geom;
Execution error: Transform error: PROJ coordinate transformation failed with error: Invalid coordinate
```

Tracking issue: #30

### DIFF
--- a/c/sedona-proj/src/transform.rs
+++ b/c/sedona-proj/src/transform.rs
@@ -14,6 +14,9 @@
 // KIND, either express or implied.  See the License for the
 // specific language governing permissions and limitations
 // under the License.
+use crate::error::SedonaProjError;
+use crate::proj::{Proj, ProjContext};
+use proj_sys::PJ_LOG_LEVEL_PJ_LOG_NONE;
 use sedona_geometry::bounding_box::BoundingBox;
 use sedona_geometry::error::SedonaGeometryError;
 use sedona_geometry::interval::IntervalTrait;
@@ -21,9 +24,6 @@ use sedona_geometry::transform::{CrsEngine, CrsTransform};
 use std::cell::RefCell;
 use std::path::PathBuf;
 use std::rc::Rc;
-
-use crate::error::SedonaProjError;
-use crate::proj::{Proj, ProjContext};
 
 /// Builder for a [ProjCrsEngine]
 ///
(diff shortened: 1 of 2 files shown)


c/sedona-proj/src/transform.rs
```diff
@@ -14,6 +14,9 @@
 // KIND, either express or implied.  See the License for the
 // specific language governing permissions and limitations
 // under the License.
+use crate::error::SedonaProjError;
+use crate::proj::{Proj, ProjContext};
+use proj_sys::PJ_LOG_LEVEL_PJ_LOG_NONE;
```
paleolimbot:
```suggestion
```

c/sedona-proj/src/transform.rs
```diff
@@ -102,6 +122,13 @@ impl ProjCrsEngineBuilder {
             ctx.set_search_paths(&string_vec)?;
         }
 
+        if let Some(log_level) = &self.log_level {
+            ctx.set_log_level(*log_level)?;
+        } else {
+            // Default log level to none
+            ctx.set_log_level(PJ_LOG_LEVEL_PJ_LOG_NONE)?;
```
paleolimbot:
```suggestion
            ctx.set_log_level(0)?;
```

c/sedona-proj/src/transform.rs
jesspav: yeah, sorry about that. removed it.

c/sedona-proj/src/transform.rs
jesspav: oops... made the switch.

jesspav:
Yeah, you are right, I was not thinking about whether the constant I referenced would be available without proj being installed. Good thing the build here caught that.

paleolimbot (approved):
Thank you!

PR #83, opened 2025-09-15
Fix KNN joins and bypasses the geography type checking for KNN predicates

zhangfengcdt:
**Root Cause**

The ST_KNN regression was introduced in commit df1e0b7 (September 12, 2025) which added geography type checking to prevent optimized spatial joins from being used with unsupported geography types.

**The Problem**

The new is_spatial_predicate_supported() function was checking ALL spatial predicates (including KNN) for geography types, but it was overly restrictive and blocked valid planar geometry KNN operations. When KNN predicates were blocked, the spatial join optimizer would skip optimization and fall back to NestedLoopJoinExec, which doesn't support ST_KNN function calls, resulting in the error:

Implementation for st_knn([WkbView(Planar, None), WkbView(Planar, None), Arrow(Int64), Arrow(Boolean)]) was not registered

### DIFF
--- a/rust/sedona-spatial-join/src/optimizer.rs
+++ b/rust/sedona-spatial-join/src/optimizer.rs
@@ -892,9 +892,10 @@ fn is_spatial_predicate_supported(
     }
 
     match spatial_predicate {
+        // Always allow KNN predicates to preserve existing functionality
+        SpatialPredicate::KNearestNeighbors(_) => true,
         SpatialPredicate::Relation(RelationPredicate { left, right, .. })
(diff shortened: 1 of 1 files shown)


zhangfengcdt:
Looks like we need to register to binary_predicate to have the correct fix:
https://github.com/apache/sedona-db/blob/main/c/sedona-geos/src/register.rs

Kontinuation:
I have submitted https://github.com/apache/sedona-db/pull/87 as a finer-grained fix to keep disabling ST_KNN joins for geographies but fixing the failed queries involving planar geometries.

PR #86, opened 2025-09-15
[WIP] Rename collect to to_memtable

jesspav:
The current implementation of `collect` does not maintain ordering, breaking with user's expectation for that API.  We are renaming it `to_memtable()`.  We can create an ordered `collect` at a later date.


paleolimbot (approved):
Thank you!

Kontinuation:
Do we need to change the [overture example notebook](https://github.com/apache/sedona-db/blob/0e6fe515a26e6af01a25382efd8a1e16d6ae972b/docs/overture-examples.ipynb) as well?

jesspav:
Nice catch @Kontinuation .  Putting out a PR now to fix.

PR #87, opened 2025-09-16
Fix distinguishing the joined geospatial types for ST_KNN

Kontinuation:
This PR fixes a regression introduced by https://github.com/apache/sedona-db/pull/57, which added geography type checking to prevent optimized spatial joins from being used with unsupported geography types.

The problematic patch incorrectly matches `left` and `right` expressions in `KNNPredicate` to query plans. Actually, how `left` and `right` maps to the original query plans depends on the value of `probe_side` field. This is quite misleading and may be addressed in a future PR by renaming `left` and `right` to `build` and `probe`.


paleolimbot (approved):
Thank you!

PR #92, opened 2025-09-16
feat: Support explaining DataFrame when using Python API

Kontinuation:
This patch fixes the following error when `show`ing a DataFrame backed by an `EXPLAIN ...` query:

```python
>>> import sedonadb
>>> con = sedonadb.connect()
>>> con.sql("EXPLAIN SELECT 1 as one").show()
Traceback (most recent call last):
  File "<python-input-4>", line 1, in <module>
    con.sql("EXPLAIN SELECT 1 as one").show()
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^
  File "/Users/bopeng/workspace/wherobots/sedona-db/python/sedonadb/python/sedonadb/dataframe.py", line 362, in show
    print(self._impl.show(self._ctx, limit, width, ascii), end="")
          ~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
sedonadb._lib.SedonaError: Unsupported logical plan: Explain must be root of the plan.
This issue was likely caused by a bug in DataFusion's code. Please help us to resolve this by filing a bug report in our issue tracker: https://github.com/apache/datafusion/issues
```

The above code will run successfully after applying this patch:

```
>>> import sedonadb
>>> con = sedonadb.connect()
>>> con.sql("EXPLAIN SELECT 1 as one").show()
┌───────────────┬─────────────────────────────────┐
│   plan_type   ┆               plan              │
│      utf8     ┆               utf8              │
╞═══════════════╪═════════════════════════════════╡
│ logical_plan  ┆ Projection: Int64(1) AS one     │
│               ┆   EmptyRelation                 │
├╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌┤
│ physical_plan ┆ ProjectionExec: expr=[1 as one] │
│               ┆   PlaceholderRowExec            │
│               ┆                                 │
└───────────────┴─────────────────────────────────┘
```

This patch also added an `explain` method to `DataFrame`.

### DIFF
--- a/python/sedonadb/python/sedonadb/dataframe.py
+++ b/python/sedonadb/python/sedonadb/dataframe.py
@@ -363,37 +363,33 @@ class DataFrame:
 
     def explain(
         self,
-        verbose: bool = False,
-        analyze: bool = False,
+        type: str = "standard",
         format: str = "indent",
-        width: Optional[int] = None,
-        ascii: bool = False,
-    ) -> str:
-        """Print the execution plan for this DataFrame
+    ) -> "DataFrame":
+        """Return the execution plan for this DataFrame as a DataFrame
 
-        Shows the logical and physical execution plans that will be used to
+        Retrieves the logical and physical execution plans that will be used to
         compute this DataFrame. This is useful for understanding query
         performance and optimization.
 
         Args:
-            verbose: Use True to show additional details in the plan output.
-            analyze: Use True to actually execute the plan and report metrics.
+            type: The type of explain plan to generate. Supported values are:
+                "standard" (default) - shows logical and physical plans,
+                "extended" - includes additional query optimization details,
+                "analyze" - executes the plan and reports actual metrics.
             format: The format to use for displaying the plan. Supported formats are
-                "indent", "tree", "pgjson" and "graphviz"
-            width: The number of characters to use to display the output.
-                If None, uses `Options.width` or detects the value from the
-                current terminal if available.
-            ascii: Use True to disable UTF-8 characters in the output.
+                "indent" (default), "tree", "pgjson" and "graphviz".
 
         Returns:
-            A string representation of the execution plan.
+            A DataFrame containing the execution plan information with columns
+            'plan_type' and 'plan'.
 
         Examples:
 
             >>> import sedonadb
             >>> con = sedonadb.connect()
             >>> df = con.sql("SELECT 1 as one")
-            >>> print(df.explain())
+            >>> df.explain().show()
             ┌───────────────┬─────────────────────────────────┐
             │   plan_type   ┆               plan              │
             │      utf8     ┆               utf8              │
@@ -405,10 +401,8 @@ class DataFrame:
             │               ┆   PlaceholderRowExec            │
             │               ┆                                 │
             └───────────────┴─────────────────────────────────┘
-            <BLANKLINE>
         """
-        width = _out_width(width)
-        return self._impl.explain(self._ctx, verbose, analyze, format, width, ascii)
+        return DataFrame(self._ctx, self._impl.explain(type, format))
 
     def __repr__(self) -> str:
         if global_options().interactive:
--- a/python/sedonadb/src/dataframe.rs
+++ b/python/sedonadb/src/dataframe.rs
@@ -187,39 +187,24 @@ impl InternalDataFrame {
         Ok(content)
     }
 
-    #[allow(clippy::too_many_arguments)]
-    fn explain<'py>(
-        &self,
-        py: Python<'py>,
-        ctx: &InternalContext,
-        verbose: bool,
-        analyze: bool,
-        format: &str,
-        width_chars: usize,
-        ascii: bool,
-    ) -> Result<String, PySedonaError> {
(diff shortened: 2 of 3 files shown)


python/sedonadb/src/dataframe.rs
```diff
@@ -186,6 +187,41 @@ impl InternalDataFrame {
...
+    ) -> Result<String, PySedonaError> {
+        let format = ExplainFormat::from_str(format)?;
+        let explain_option = ExplainOption::default()
+            .with_analyze(analyze)
+            .with_verbose(verbose)
+            .with_format(format);
+        let explain_df = self.inner.clone().explain_with_options(explain_option)?;
+
+        let mut options = DisplayTableOptions::new();
+        options.table_width = width_chars.try_into().unwrap_or(u16::MAX);
```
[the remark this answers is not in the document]
Kontinuation: I prefer following the convention of existing code.

rust/sedona/src/context.rs
```diff
@@ -292,9 +292,23 @@ impl SedonaDataFrame for DataFrame {
...
+        ) {
+            // Show multi-line output without truncation for plans like `EXPLAIN`
+            options.max_row_height = usize::MAX;
+
+            // We don't want to apply an additional .limit() to plans like `Explain`
+            // as that will trigger an internal error: Unsupported logical plan: Explain must be root of the plan
+            self
+        } else {
+            // Apply limit if specified
+            self.limit(0, limit)?
```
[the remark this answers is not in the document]
Kontinuation: This approach introduces additional copies, and probably make the code look worse.

python/sedonadb/python/sedonadb/dataframe.py
```diff
@@ -361,6 +361,55 @@ def show(
...
+
+        Returns:
+            A string representation of the execution plan.
+
+        Examples:
+
+            >>> import sedonadb
+            >>> con = sedonadb.connect()
+            >>> df = con.sql("SELECT 1 as one")
+            >>> print(df.explain())
```
paleolimbot: I would have expected this function to do the `print()`ing (DuckDB's does) so that the user doesn't have to. In the future we might want to return a custom object like `class Explain` whose `__repr__` is the text output to more easily provide access to the explanation (e.g., to render the JSON into a jupyter widget).

python/sedonadb/src/dataframe.rs
```diff
@@ -186,6 +187,41 @@ impl InternalDataFrame {
...
+    #[allow(clippy::too_many_arguments)]
+    fn explain<'py>(
+        &self,
+        py: Python<'py>,
+        ctx: &InternalContext,
+        verbose: bool,
+        analyze: bool,
+        format: &str,
+        width_chars: usize,
+        ascii: bool,
```
paleolimbot: If this function returned a `DataFrame` we could probably avoid having to duplicate the show logic

python/sedonadb/python/sedonadb/dataframe.py
```diff
@@ -361,6 +361,55 @@ def show(
         width = _out_width(width)
         print(self._impl.show(self._ctx, limit, width, ascii), end="")
 
+    def explain(
+        self,
+        verbose: bool = False,
+        analyze: bool = False,
```
paleolimbot: I'm not sure we need to maintain DuckDB compatibility here, but DuckDB parameterizes this as `explain(type="standard")/explain(type="analyze")`.

Kontinuation: Shall we replace these 2 parameters with a `type` parameter with the following possible values?

* "standard": verbose = False, analyze = False
* "analyze": verbose = False, analyze = True
* "extended": verbose = True, analyze = False (this shows how optimization rules have changed the query plan over optimization passes)

python/sedonadb/python/sedonadb/dataframe.py
Kontinuation: This interface actually mimics [DuckDB's Python API](https://duckdb.org/docs/stable/clients/python/relational_api#explain): it returns a str silently without printing anything to stdout/stderr.

Returning another DataFrame containing execution plans might be a better idea. This avoids duplicating the show parameters and logic.

python/sedonadb/python/sedonadb/dataframe.py
paleolimbot: I think so!

python/sedonadb/python/sedonadb/dataframe.py
paleolimbot: Ah, I was confused since that doesn't happen in notebooks:

https://github.com/duckdb/duckdb-python/blob/14245ff33f591863296261bf07cabd48a7e803fa/src/duckdb_py/pyrelation.cpp#L1765-L1766

In any case anything is Ok here (it's all a vast improvement over the opaque limit error we got before)

PR #94, opened 2025-09-16
Update overture notebook with new API names

jesspav:
Updating API names that have been changed in the overture example Jupyter notebook.

## Testing

Opened up the Jupyter notebook in a sedonadb python environment and ran all of the cells to ensure everything worked.

### DIFF
--- a/docs/overture-examples.ipynb
+++ b/docs/overture-examples.ipynb
@@ -17,13 +17,13 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "import sedonadb\n",
+    "import sedona.db\n",
     "import os\n",
     "\n",
     "os.environ[\"AWS_SKIP_SIGNATURE\"] = \"true\"\n",
     "os.environ[\"AWS_DEFAULT_REGION\"] = \"us-west-2\"\n",
     "\n",
-    "sd = sedonadb.connect()"
+    "sd = sedona.db.connect()"
    ]
   },
   {
--- a/docs/overture-examples.ipynb
+++ b/docs/overture-examples.ipynb
@@ -424,7 +424,7 @@
    ],
    "source": [
     "%%time\n",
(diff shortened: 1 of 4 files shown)


jesspav:
This is my first docs change. Is there anything I need to do to get it to the website?

docs/overture-examples.ipynb
```diff
@@ -17,13 +17,13 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "import sedona.db\n",
+    "import sedonadb\n",
```
paleolimbot:
```suggestion
    "import sedona.db\n",
```

docs/overture-examples.ipynb
```diff
@@ -17,13 +17,13 @@
...
     "import os\n",
     "\n",
     "os.environ[\"AWS_SKIP_SIGNATURE\"] = \"true\"\n",
     "os.environ[\"AWS_DEFAULT_REGION\"] = \"us-west-2\"\n",
     "\n",
-    "sd = sedona.db.connect()"
+    "sd = sedonadb.connect()"
```
paleolimbot:
```suggestion
    "sd = sedona.db.connect()"
```

paleolimbot (commented):
Thank you!

We need the `sedona DOT db`...the version of `sedona` that contains the `db` part was just released (i.e., if you `pip install --upgrade apache-sedona` locally just one time this should work locally for you).

docs/overture-examples.ipynb
jesspav: Yeah `import sedona.db` didn't work for me because I was running it from within the environment and the other 3 notebooks had it as `sedona.db`.  After chatting with Powers  I've updated the other notebooks now to use the new convention `sedona.db`, but it is a little annoying that the notebooks won't just work from within the sedonadb python environment anymore.

PR #97, opened 2025-09-16
feat(python/sedonadb): Expose random_geometry as a Python function

paleolimbot:
We've been using the SQL table function `sd_random_geometry()` quite a bit for testing and benchmarking and I think this may make those types of simulations much nicer to write.

This still needs a bit of testing/moving of the parameter validation to Rust but it's not essential for release. I'll circle back after working on some more releasy-type tasks.

### DIFF
--- a/python/sedonadb/python/sedonadb/testing.py
+++ b/python/sedonadb/python/sedonadb/testing.py
@@ -14,11 +14,11 @@
 # KIND, either express or implied.  See the License for the
 # specific language governing permissions and limitations
 # under the License.
-import os
 import math
+import os
 import warnings
 from pathlib import Path
-from typing import TYPE_CHECKING, List, Tuple
+from typing import TYPE_CHECKING, Iterable, List, Literal, Optional, Tuple, Union
 
 import geoarrow.pyarrow as ga
 import pyarrow as pa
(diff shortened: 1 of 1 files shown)


petern48:
This reminds me that when I was writing up the pytest-benchmarking code, I realized it would be nice to have a random geometry rust function (instead of a table provider) that returns a single column, so that we could create multiple unique random geometry columns in a single query to use with functions that take two geometry inputs (e.g predicates). I can't think of many other use cases tho, so it doesn't seem very high priority.

python/sedonadb/python/sedonadb/testing.py
```diff
@@ -29,6 +29,85 @@
...
+    bounds = [float(b) for b in bounds]
+    if len(bounds) != 4:
+        raise ValueError(
+            f"Expected bounds as [xmin, ymin, xmax, ymax] but got {bounds}"
+        )
+
+    width = bounds[2] - bounds[0]
+    height = bounds[3] - bounds[1]
+    if size_min > width or size_min > height:
+        raise ValueError("size > height / 2 or width / 2 of bounds")
```
martin-g: The error message does not match the check. The error message talks about halfs.

python/sedonadb/python/sedonadb/testing.py
```diff
@@ -29,6 +29,85 @@
...
+    *,
+    num_vertices: Union[int, Tuple[int, int]] = 4,
+    num_parts: Union[int, Tuple[int, int]] = (1, 3),
+    size: Union[float, Tuple[float, float]] = (5.0, 20.0),
+    bounds: Iterable[float] = (-170, -80, 170, 80),
+    hole_rate: float = 0.0,
+    empty_rate: float = 0.0,
+    null_rate: float = 0.0,
+    seed: Optional[int] = None,
+) -> "sedonadb.dataframe.DataFrame":
```
martin-g: nit: Missing docstring

python/sedonadb/python/sedonadb/testing.py
```diff
@@ -29,6 +29,85 @@
...
+    if size_min > size_max:
+        raise ValueError("size_min > size_max")
+
+    bounds = [float(b) for b in bounds]
+    if len(bounds) != 4:
+        raise ValueError(
+            f"Expected bounds as [xmin, ymin, xmax, ymax] but got {bounds}"
+        )
+
+    width = bounds[2] - bounds[0]
```
martin-g: Should there be checks that xmin is smaller than xmax ?
Same for the `y` bounds below.

PR #108, opened 2025-09-18
Use distance_ext for evaluating ST_DWithin/ST_Distance in spatial join and upgrade dependencies

Kontinuation:
The generic `DistanceExt` function should work well for now, switching the distance evaluator of spatial join to DistanceExt passes all tests.

This patch also upgraded the dependencies, especially the commit hash of `geo-generic-alg`, to ensure that we'll use the updated `DistanceExt` function.

### DIFF
--- a/.github/workflows/dev.yml
+++ b/.github/workflows/dev.yml
@@ -38,6 +38,19 @@ jobs:
       - uses: actions/setup-python@v5
         with:
           python-version: '3.x'
+      - name: Install pinned Rust toolchain
+        id: rust
+        run: |
+          set -euxo pipefail
+          # Extract pinned channel from rust-toolchain.toml (keeps single source of truth)
+          TOOLCHAIN=$(awk -F '"' '/^channel/ {print $2}' rust-toolchain.toml)
+          echo "Pinned toolchain: ${TOOLCHAIN}"
+          rustup toolchain install "${TOOLCHAIN}" --profile minimal --no-self-update
+          rustup component add --toolchain "${TOOLCHAIN}" rustfmt clippy
+          rustup default "${TOOLCHAIN}"
+          rustup show active-toolchain
+          rustc -V
+          cargo -V
       - name: pre-commit (cache)
         uses: actions/cache@v4
         with:
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1534,9 +1534,9 @@ dependencies = [
 
 [[package]]
 name = "dary_heap"
-version = "0.3.7"
+version = "0.3.8"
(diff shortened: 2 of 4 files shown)


Kontinuation:
More of such kind of regression happens in the latest 1.90.0 Rust compiler. I believe that the GAT recursion problem will never be addressed by Rust compiler developers. The geo-traits needs to be redesigned to get rid of recursion.

Kontinuation:
This compilation failure is tough to resolve. I tried to tweak the design of geo-trait-ext but got overwhelmed by more and more compile errors. Shall we specify the Rust toolchain using `rust-toolchain.toml` to unblock the upcoming release of the Python packages? I know this is far from ideal, but it could buy us some time to resolve this on the latest stable Rust.

jiayuasu:
@paleolimbot Is it normal for the Python unit tests take almost an hour to finish?

Kontinuation:
The latest DistanceExt generic implementation hangs indefinitely on particular inputs. I can reproduce it locally. The hanging test is:

```
python/sedonadb/tests/functions/test_predicates.py::test_st_dwithin[GEOMETRYCOLLECTION (POINT (0 0), LINESTRING (0 0, 1 1))-GEOMETRYCOLLECTION (POINT (0 0), LINESTRING (0 0, 1 1))-1-True-SedonaDB]
```

I'll look into this tomorrow. Also CC @zhangfengcdt

zhangfengcdt:
> The latest DistanceExt generic implementation hangs indefinitely on particular inputs. I can reproduce it locally. The hanging test is:
> 
> ```
> python/sedonadb/tests/functions/test_predicates.py::test_st_dwithin[GEOMETRYCOLLECTION (POINT (0 0), LINESTRING (0 0, 1 1))-GEOMETRYCOLLECTION (POINT (0 0), LINESTRING (0 0, 1 1))-1-True-SedonaDB]
> ```
> 
> I'll look into this tomorrow. Also CC @zhangfengcdt

I think i found the case that causes the infinite recursive loop and trying to get a workaround for it:

https://github.com/wherobots/geo/pull/10/files

The macro expansion causes issues in this specific case, and looks like the rust trait bouds checking is the issue and why we will need this fix.

Kontinuation:
> I think i found the case that causes the infinite recursive loop and trying to get a workaround for it:
> 
> https://github.com/wherobots/geo/pull/10/files
> 
> The macro expansion causes issues in this specific case, and looks like the rust trait bouds checking is the issue and why we will need this fix.

Thank you. I have merged the fix for wherobots/geo and updated the Cargo.lock. Now the tests are passing.

PR #112, opened 2025-09-19
[DOCS] aligning nav

kadolor:


kadolor:
@jiayuasu 

- Removing the string, "Python", will make the "Sedona Homepage" link visible again. We've just hit a character limit. 
image

PR #114, opened 2025-09-20
feat(r/sedonadb): Add tweaks on sedona-proj and sedona-tg for Windows build with Rtools

yutannihilation:
This pull request adds a mechanism to avoid running bindgen in sedona-proj and sedona-tg. This fixes two of the three remaining problems we found in https://github.com/apache/sedona-db/pull/78#issuecomment-3289180614 related to cross-compilation on Windows with Rtools.

The mechanism is

1. If there's already `bindings/{target_triple}.rs`, use it.
2. Otherwise, run bindgen and use the newly generated bindings.
    - If `SEDONA_*_BINDINGS_OUTPUT_PATH` envvar is set, output the bindings to the path.
    - If not, use `OUT_DIR`, which Rust uses for intermediate artifacts.

So, if we want to update the binding, we can set `SEDONA_*_BINDINGS_OUTPUT_PATH` and build the crate. For example, run the following to generate bindings to `./tmp.rs` using Rtools45 (I tried both `stable-x86_64-pc-windows-msvc` and `stable-x86_64-pc-windows-gnu` toolchain, but it seems the result bindings is the same).

```ps1
$env:PATH = "C:\rtools45\x86_64-w64-mingw32.static.posix\bin;$env:PATH"
$env:PKG_CONFIG_SYSROOT_DIR = "C:\rtools45\x86_64-w64-mingw32.static.posix\"
$env:SEDONA_PROJ_BINDINGS_OUTPUT_PATH = "./tmp.rs"

cargo build -p sedona-proj
```

Note that, the output bindings file doesn't contain the license header (I couldn't find a bindgen's option to inject this...), so this needs manual edit at least at the moment.

If it's better not to add the prebuilt bindings into the crates, I can tweak the mechanism and generate bindings in R's `Makevars.win` (i.e. run `cargo build -p sedona-proj` without cross-compilation first, and then use the generated bindings via some envvar).


c/sedona-proj/build.rs
```diff
@@ -16,19 +16,64 @@
...
+
 fn main() {
     println!("cargo:rerun-if-changed=src/proj_dyn.c");
     cc::Build::new().file("src/proj_dyn.c").compile("proj_dyn");
 
+    println!("cargo:rerun-if-env-changed=SEDONA_PROJ_BINDINGS_OUTPUT_PATH");
+
+    let os = std::env::var("CARGO_CFG_TARGET_OS").unwrap();
+    let arch = env::var("CARGO_CFG_TARGET_ARCH").unwrap();
+    let prebuilt_bindings_path = format!("src/bindings/{os}-{arch}.rs");
```
paleolimbot: Should this also include the `gnu` part of the target (i.e., I imagine the bindings *might* be different under the gnu vs. msvc targets in the future?)

paleolimbot (commented):
Thank you! The runtime generated bindings have been a pain in a few places (conda environments were another I recently ran into) and moving towards providing them for a few common targets is something I have seen other projects (e.g., geos-sys) do as well.

CI may have a few things to say...my only high level comment is that I would love to have a script used to generate the prebuilt bindings somewhere so that when we update the libraries/headers we can just run the script. This could also be in a README for the crate since it seems like it's just:

```
SEDONA_PROJ_BINDINGS_OUTPUT_PATH=src/bindings cargo build
```

?

yutannihilation:
> a few common targets is something I have seen other projects (e.g., geos-sys) do as well.

Oh, thanks. I didn't notice geos-sys does. It seems it doesn't use different bindings for different OS/arch, but for different library versions. If this works on PROJ and TG, probably it's better than mine.

https://github.com/georust/geos/tree/master/sys/prebuilt-bindings

yutannihilation:
Thanks, but what I was thinking was, probably it matters a bit on this part:

> a script used to generate the prebuilt bindings

If it's independent of the OS, the arch, and the target, it would be easy to add a simple script. But, if it's not, probably it needs to be done on the CI to generate multiple versions of bindings. So, I need to think a bit about to what extent I complicate the mechanism in this pull request...

yutannihilation:
Ah, thanks, I forgot about the license header...

yutannihilation:
Oh, didn't know README also needs the license header...

c/sedona-proj/build.rs
yutannihilation: Now I changed this to use the target triple.

yutannihilation:
> I think it's OK to just add instructions for how to do it for one OS/architecture to each crate's README for now...

I was hoping I can add some nicer mechanism in this pull request, but I gave up and decided just to add READMEs. Should be ready for review now!

c/sedona-proj/README.md
```diff
@@ -0,0 +1,48 @@
...
+SEDONA_PROJ_BINDINGS_OUTPUT_PATH=src/bindings/x86_64-unknown-linux-gnu.rs cargo build -p sedona-proj
+```
+
+Note that, the generated file doesn't contain the license header, so you need
+to add it manually (hopefully, this will be automated).
+
+### Windows
+
+On Windows, you might need to set `PATH` and `PKG_CONFIG_SYSROOT_DIR` for
+`pkg-config`. For example, when using Rtools45, you progbably need these
```
paleolimbot, line 38:
```suggestion
`pkg-config`. For example, when using Rtools45, you probably need these
```

c/sedona-tg/README.md
```diff
@@ -0,0 +1,48 @@
...
+SEDONA_TG_BINDINGS_OUTPUT_PATH=src/bindings/x86_64-unknown-linux-gnu.rs cargo build -p sedona-tg
+```
+
+Note that, the generated file doesn't contain the license header, so you need
+to add it manually (hopefully, this will be automated).
+
+### Windows
+
+On Windows, you might need to set `PATH` and `PKG_CONFIG_SYSROOT_DIR` for
+`pkg-config`. For example, when using Rtools45, you progbably need these
```
paleolimbot, line 38:
```suggestion
`pkg-config`. For example, when using Rtools45, you probably need these
```

paleolimbot (approved):
Thank you!

yutannihilation:
Aw, thanks for catching the typos... Probably I'll fix in another pull request.

PR #115, opened 2025-09-20
Turn on the s2geography feature explicitly in verify release script to pass pytests

petern48:
When running the raw `dev/release/verify-release-candidate.sh` script, I constantly ran into pytest errors for `test_spatial_join_geography` inside test_sjoin.py. The error message wasn't straightforward to debug, but I eventually found that I needed to turn on the `s2geography` feature explicitly to get the tests to pass. This PR sets it inside the verify release script explicitly to smoothen out the process.

### DIFF
--- a/.github/workflows/python.yml
+++ b/.github/workflows/python.yml
@@ -116,6 +116,7 @@ jobs:
 
       - name: Install
         run: |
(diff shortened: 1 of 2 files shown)


jiayuasu:
@paleolimbot should we merge it?

PR #116, opened 2025-09-20
fix(r/sedonadb): Add missing settings to configure.win and Makevars.win.in

yutannihilation:
I succeeded to compile R sedonadb package on my local with https://github.com/apache/sedona-db/pull/114 and https://github.com/apache/datafusion/issues/17687. But, besides it, I found `configure.win` and `Makevars.win.in` needs a few tweaks to match with `configure` and `Makevars.in`.


paleolimbot (approved):
Thank you!

PR #117, opened 2025-09-21
chore(r/sedonadb): Add necessary tweaks for Windows CI on the R package

yutannihilation:
I confirmed my pull request to datafusion (https://github.com/apache/datafusion/pull/17688) will enable the Windows CI for R package. Note that the latest version of datafusion is already v50 while sedona-db uses v49. So, I think this won't happen very soon.

my experiment: https://github.com/yutannihilation/sedona-db/pull/1
diff: https://github.com/apache/datafusion/compare/49.0.2...yutannihilation:datafusion:49.0.2-pr-17688

Until the day, this pull request adds some necessary tweaks for Windows. Also, this includes a typo fix found in https://github.com/apache/sedona-db/pull/114#discussion_r2365915732

### DIFF
--- a/.github/workflows/r.yml
+++ b/.github/workflows/r.yml
@@ -86,14 +86,6 @@ jobs:
           # Update this key to force a new cache
           prefix-key: "r-v1"
 
-      - name: Install build dependencies
-        run: |
-            R -e 'install.packages(c("nanoarrow", "geoarrow"))'
-
-      - name: Install R Package
-        run: |
-          R CMD INSTALL r/sedonadb --preclean
-
       - uses: r-lib/actions/setup-r-dependencies@v2
         with:
           needs: check
--- a/.github/workflows/r.yml
+++ b/.github/workflows/r.yml
@@ -67,7 +67,7 @@ jobs:
           rustup toolchain install stable --no-self-update
           rustup default stable
 
-      # GHA runnder installs GNU target by default, but --no-self-update might
+      # GHA runner installs GNU target by default, but --no-self-update might
       # remove it, so run `target add` to ensure the GNU target installed.
       - name: Setup Rust (Windows)
         if: matrix.config.os == 'windows-latest'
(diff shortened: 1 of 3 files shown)


.github/workflows/r.yml
```diff
@@ -36,13 +36,17 @@ jobs:
   test:
     runs-on: ${{ matrix.config.os }}
     name: ${{ matrix.config.os }} (${{ matrix.config.r }})
+    defaults:
+      run:
+        shell: bash
```
yutannihilation, line 41: This is necessary because otherwise `R` is an alias to `Invoke-History` on PowerShell...

.github/workflows/r.yml
```diff
@@ -76,19 +86,15 @@ jobs:
           # Update this key to force a new cache
           prefix-key: "r-v1"
 
-      - name: Install build dependencies
-        run: |
-            R -e 'install.packages(c("nanoarrow", "geoarrow"))'
```
yutannihilation, line 81: It seems `install.packages()` doesn't work on Windows before setting CRAN mirror. So, just run `setup-r-dependencies` action instead.

.github/workflows/r.yml
```diff
@@ -63,6 +67,12 @@ jobs:
           rustup toolchain install stable --no-self-update
           rustup default stable
 
+      # GHA runnder installs GNU target by default, but --no-self-update might
```
paleolimbot:
```suggestion
      # GHA runner installs GNU target by default, but --no-self-update might
```

.github/workflows/r.yml
```diff
@@ -36,13 +36,17 @@ jobs:
...
+      run:
+        shell: bash
 
     strategy:
       fail-fast: false
       matrix:
         config:
           - {os: ubuntu-latest, r: 'release'}
           - {os: macos-latest, r: 'release'}
+          # - {os: windows-latest, r: 'release'}
```
paleolimbot, line 49: Can you open an issue to track adding this and link it here? (We do this for most instances of commented-out code elsewhere in the project)

.github/workflows/r.yml
yutannihilation: Thanks for catching!

.github/workflows/r.yml
yutannihilation: Sure. I created an issue for this.

https://github.com/apache/sedona-db/issues/120

PR #146, opened 2025-09-24
feat(r/sedonadb): Use the "bootstrap.R" mechanism to copy necessary Rust source codes

yutannihilation:
I found @paleolimbot invented a cool mechanism (https://github.com/r-lib/pkgbuild/pull/157) that can be used for vendoring the necessary source code when installed via `pkgbuild` (i.e. pak). Currently, the README requires users to install the sedonadb R package by:

```sh
git clone https://github.com/apache/sedona-db.git
cd sedona-db/r/sedonadb
R CMD INSTALL .
```

This pull request will allows us to simplify the installation to this:

```r
pak::pkg_install("apache/sedona-db/r/sedonadb")
```

(and probably enables it to publish the R package to R-universe?)

### DIFF
--- a/.github/workflows/r.yml
+++ b/.github/workflows/r.yml
@@ -88,14 +88,13 @@ jobs:
 
       - uses: r-lib/actions/setup-r-dependencies@v2
         with:
-          extra-packages: any::rcmdcheck
           needs: check
           working-directory: r/sedonadb
 
-      - uses: r-lib/actions/check-r-package@v2
-        with:
-          args: 'c("--no-manual")' # no --as-cran
-          upload-snapshots: true
-          working-directory: r/sedonadb
-          error-on: '"error"'      # ignore CRAN warnings
-          upload-results: 'never'  # results are too heavy to upload
+      - name: Install R Package
+        run: |
+          R CMD INSTALL r/sedonadb --preclean
+
+      - name: Test R Package
+        run: |
+          R -e 'library(testthat); test_local("r/sedonadb", MultiReporter$new(list(CheckReporter$new(), LocationReporter$new())))'


r/sedonadb/bootstrap.R
eitsupi, line 1: I suppose this file should be added to `.Rbuildignore`.

yutannihilation: Ah, thanks for catching!

.github/workflows/r.yml
```diff
@@ -88,13 +88,12 @@ jobs:
...
           needs: check
           working-directory: r/sedonadb
 
-      - name: Install R Package
-        run: |
-          R CMD INSTALL r/sedonadb --preclean
-
-      - name: Test R Package
-        run: |
-          R -e 'library(testthat); test_local("r/sedonadb", MultiReporter$new(list(CheckReporter$new(), LocationReporter$new())))'
```
paleolimbot, line 100: It might make sense to keep this and add the CMD check version in a separate workflow that doesn't run whenever somebody touches `rust/**`. I *think* the R CMD INSTALL version more effectively uses the cache (but if the CMD check version is still in the ~15 minute range we should totally do that).

yutannihilation: Thanks, makes sense! I reverted `r.yml`.

yutannihilation: > add the CMD check version in a separate workflow

The reason I used the `check-r-package` action was simply that it looked easier to me. So, if not, I don't have strong opinion to use R CMD check.

eitsupi: FYI, r-polars has two jobs: a fast job that builds in debug mode and runs only test, and a job workflow that builds in release mode and runs R CMD check.
If you're interested, check them out.
https://github.com/pola-rs/r-polars/blob/66d7b73b6496dc4494fae72f8540a6958cec0c9a/.github/workflows/test-r.yml#L47-L197

paleolimbot: Let's get CMD check passing in a future PR! As I think you discovered here, there's some C file deep in the dependency stack that CMD check objects to (we either need to patch the file, remove the file after build, or ignore just that specific error).

PR #149, opened 2025-09-25
refactor(c/sedona-proj,c/sedona-tg): Remove build-time generated headers

### LINKED ISSUE #147: c/sedona-tg,c/sedona-proj: Use prebuilt bindings instead of build-time generated bindings
The build-time generated bindings by bindgen are causing issues when building on various platforms, and we don't have much surface area that we need to cover for either of those projects, anyway.

We can either use the mechanism that @yutannihilation kindly added to these crates (or just type out definitions for the functions we need, which may be easier).

Example of where a prebuilt binding is causing a failure (R-universe build): https://github.com/r-universe/paleolimbot/actions/runs/17987153646/job/51168200021#step:4:1211

paleolimbot:
Closes #147.

It turns out that bindgen-generated bindings are tricky to integrate into a lot of build systems...Windows for R was the first one but conda environments are another and the dependency on libclang is not trivial. Most recently, this caused a failure in R-universe (see the issue).

This PR just hard-codes the generated bindings...in tg's case, reducing them to only the functions we use. This would be a problem if we expected the headers to change frequently, but I don't think there's much chance of our particular set of functions changing very rapidly.

Apologies to @yutannihilation since this redoes the work in your earlier PR to prebuild these for Windows. I did play with that a bit but ran into the same problem you did, which was that it's tricky to generate bindings for platforms other than the one you're developing on.


yutannihilation:
Never mind! Mine was just a temporary solution. Actually, in savvy, I chose to write the bindings by hand after some struggle, so I was wondering if sedona-db also stops relying on bindgen at some point. This decision looks good to me.

PR #152, opened 2025-09-25
fix(r/sedonadb): Increase CMAKE path length allowed on Windows and fix CMD check WARNINGs

paleolimbot:
We're building in R Universe now! https://paleolimbot.r-universe.dev/sedonadb

We have some WARNINGs from missing declared dependencies (my fault) and a build failure for Windows (path length related to the CMake builder for the AWS dependency) and WebR (which I don't attempt to solve here).

### DIFF
--- a/.github/workflows/r-extended.yml
+++ b/.github/workflows/r-extended.yml
@@ -15,7 +15,7 @@
 # specific language governing permissions and limitations
 # under the License.
 
-name: r
+name: r-extended
 
 on:
   push:
@@ -25,11 +25,6 @@ on:
     branches:
       - main
     paths:
-      - 'Cargo.toml'
-      - 'Cargo.lock'
-      - '.github/workflows/r.yml'
-      - 'rust/**'
-      - 'c/**'
       - 'r/**'
 jobs:
 
@@ -45,8 +40,6 @@ jobs:
       matrix:
         config:
           - {os: ubuntu-latest, r: 'release'}
-          - {os: macos-latest, r: 'release'}
-          # - {os: windows-latest, r: 'release'}
 
     env:
       GITHUB_PAT: ${{ secrets.GITHUB_TOKEN }}
@@ -54,6 +47,7 @@ jobs:
 
     steps:
       - uses: actions/checkout@v4
+      - uses: r-lib/actions/setup-pandoc@v2
       - uses: r-lib/actions/setup-r@v2
         with:
           r-version: ${{ matrix.config.r }}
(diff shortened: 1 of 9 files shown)


r/sedonadb/R/zzz.R
```diff
@@ -36,8 +36,10 @@ check_interrupts <- function() {
   }, interrupt = function(...) TRUE, error = function(...) TRUE)
 }
 
-# Permissively licensed (unlicense) from
-# https://github.com/r-lib/vctrs/blob/b63a233bd8b8d6511cf3f7d9182c359a3b8c3cab/R/register-s3.R
+# From the `vctrs` package (this function is intended to be copied
+# without attribution or license requirements to avoid a hard dependency on
+# vctrs:
+# https://github.com/r-lib/vctrs/blob/c2a7710fe55e3a2249c4fdfe75bbccbafcf38804/R/register-s3.R#L25-L31
```
paleolimbot, line 42: This reverts what I had with a self-contained version that doesn't depend on rlang

r/sedonadb/bootstrap.R
```diff
@@ -29,7 +29,6 @@ writeLines(
   cargo_toml
 )
 
-dir.create("src/dep_crates/")
```
yutannihilation, line 32: Aw, thanks for catching!

yutannihilation:
Probably you need to patch datafusion-doc as well?

.github/workflows/r-extended.yml
```diff
@@ -0,0 +1,76 @@
...
+      - uses: r-lib/actions/setup-r-dependencies@v2
+        with:
+          extra-packages: any::rcmdcheck
+          needs: check
+          working-directory: r/sedonadb
+
+      - uses: r-lib/actions/check-r-package@v2
+        with:
+          working-directory: r/sedonadb
+          error-on: '"error"'
```
paleolimbot, line 76: This is the primary reason for this workflow: `R CMD check` runs some extra tests beyond just "the tests" to make sure examples run, documentation is correct, and some others. Unfortunately we generate some WARNINGs still for things that CRAN cares about but we don't (perhaps just speaking for myself)...it would be nice to ignore specific warnings so that other warnings don't go unnoticed (but perhaps a battle for a different PR).

Cargo.toml
```diff
@@ -116,6 +116,17 @@ tokio = { version = "1.44" }
...
+datafusion-catalog = { git = "https://github.com/paleolimbot/datafusion.git", branch = "local-49-with-patch", package = "datafusion-catalog" }
+datafusion-common = { git = "https://github.com/paleolimbot/datafusion.git", branch = "local-49-with-patch", package = "datafusion-common" }
+datafusion-common-runtime = { git = "https://github.com/paleolimbot/datafusion.git", branch = "local-49-with-patch", package = "datafusion-common-runtime" }
+datafusion-execution = { git = "https://github.com/paleolimbot/datafusion.git", branch = "local-49-with-patch", package = "datafusion-execution" }
+datafusion-expr = { git = "https://github.com/paleolimbot/datafusion.git", branch = "local-49-with-patch", package = "datafusion-expr" }
+datafusion-ffi = { git = "https://github.com/paleolimbot/datafusion.git", branch = "local-49-with-patch", package = "datafusion-ffi" }
+datafusion-physical-expr = { git = "https://github.com/paleolimbot/datafusion.git", branch = "local-49-with-patch", package = "datafusion-physical-expr" }
+datafusion-physical-plan = { git = "https://github.com/paleolimbot/datafusion.git", branch = "local-49-with-patch", package = "datafusion-physical-plan" }
```
paleolimbot, line 128: I'm not actually sure this is a good idea given that we're in the process of removing the other patches. If we leave them in we get R binaries for Windows via R Universe...we can remove them when we update to DataFusion 51, which I think we'd like to do soon. I'm also not offended if this is too much for right now (I can just comment out the Windows CI again).

Kontinuation: Looks like https://github.com/apache/datafusion/pull/17688 is the fix required for the R Windows build. I'm fine with this since the other forked dependencies also take time to be cleared. We can target removing all forked dependencies once DataFusion 51 is released.

yutannihilation:
(Just for notice)
It seems this didn't fix the build on R-universe. Probably there's some difference that we need to figure out.

https://github.com/r-universe/paleolimbot/actions/runs/18026582294/job/51295664262#step:5:762
```
  -- Check if the system is big endian
  -- Searching 16 bit integer
  -- Looking for sys/types.h
  -- Looking for sys/types.h - not found
  -- Looking for stdint.h
  -- Looking for stdint.h - not found
  -- Looking for stddef.h
  -- Looking for stddef.h - not found
  -- Check size of unsigned short
  -- Check size of unsigned short - failed
  -- Check size of unsigned int
  -- Check size of unsigned int - failed
  -- Check size of unsigned long
  -- Check size of unsigned long - failed
  -- Configuring incomplete, errors occurred!
```

PR #160, opened 2025-09-28
chore(r/sedonadb): Attempt to disable the CMake builder on Windows for aws-lc-sys

paleolimbot:
Apparently the R-universe build is still trying to use a CMake build (which is failing due to path length constraints). This is an attempt to force that crate to use the cc builder with perhaps success.


fithisux:
> aws-lc-sys

Here though it is mentioned that : https://crates.io/crates/aws-lc-sys

x86_64_pc_windows_gnu  (aka mingw64) is supported.

PR #165, opened 2025-09-29
[WIP] Eliminate forked dependencies: geo, wkb, wkt and adbc_core

Kontinuation:
This PR migrates generic geo algorithms in wherobots/geo's geo-traits-ext and geo-generic-alg crates into sedona-db, and switches depended versions of geo, wkb and wkt to official release versions.


rust/sedona-testing/src/compare.rs
```diff
@@ -205,9 +210,32 @@ fn assert_wkb_value_equal(
             )
         }
         (Some(actual_wkb), Some(expected_wkb)) => {
+            // Quick test: if the binary of the WKB is the same, they are equal
             if actual_wkb != expected_wkb {
-                let (actual_wkt, expected_wkt) = (format_wkb(actual_wkb), format_wkb(expected_wkb));
-                panic!("{actual_label} != {expected_label}\n{actual_label}:\n  {actual_wkt}\n{expected_label}:\n  {expected_wkt}")
+                // Binary of the WKB are not the same, but geometries could be topologically the same.
+                let expected = wkb::reader::read_wkb(expected_wkb);
+                let actual = wkb::reader::read_wkb(actual_wkb);
```
paleolimbot: I would prefer to have this be a different function (in other words, I would like to know when I am asserting topological equality vs. coordiante-for-coordinate equality (within some tolerance) or byte-for-byte equality).

rust/sedona-testing/Cargo.toml
```diff
@@ -52,3 +52,4 @@ sedona-expr = { path = "../sedona-expr" }
 sedona-schema = { path = "../sedona-schema" }
 wkb = { workspace = true }
 wkt = { workspace = true }
+geo = { workspace = true }
```
paleolimbot: I would prefer to keep `geo` out of sedona-testing, which is otherwise algorithm library-neutral (individual tests can always use geo if they want!)

Cargo.toml
```diff
@@ -22,6 +22,9 @@ members = [
     "c/sedona-s2geography",
     "c/sedona-tg",
     "r/sedonadb/src/rust",
+    "rust/geo-traits-ext",
+    "rust/geo-generic-alg",
+    "rust/geo-test-fixtures",
```
paleolimbot: Should we rename these to `sedona-*`?

c/sedona-geos/Cargo.toml
```diff
@@ -31,6 +31,7 @@ criterion = { workspace = true }
 sedona = { path = "../../rust/sedona" }
 sedona-testing = { path = "../../rust/sedona-testing", features = ["criterion"] }
 rstest = { workspace = true }
+geo = { workspace = true }
```
paleolimbot: Does this need `geo` or would `geo-types` suffice? (It seems odd to include geo here?)

paleolimbot (commented):
I know this is the tip of the iceberg on a huge, huge amount of work and that the distance work in particular was essential to ensuring we had reasonable performance for our first release!

This is a huge amount of code, and it's unclear to me which pieces are copied and which pieces are original. It's difficult to review this at once, although I understand that we needed this PR in order to ensure all the pieces were there.

I do think there are some orthogonal sets of changes here that would enable the community to more effectively understand and/or review these changes:

- ADBC changes (I'm happy to copy these over since that one was my fault)
- WKB/GEOS changes. These could use an isolated/duplicate `wkb` dependency in a subcrate if the rest of the crates can't handle the new wkb version. I wonder if the wkb dependency to support the WkbExecutor will be a problem here (isolating that is probably a good idea anyway).
- Test fixtures (https://github.com/apache/sedona-testing is perhaps a better place for these, or perhaps there is a way to expose them from the original crate from whence they came?)
- sedona-geo-traits-ext (without updating the top-level Cargo.toml). There are a few pieces of this that are missing docstrings and this might be a good opportunity to provide those.
- sedona-geo-generic-alg (without updating the top-level Cargo.toml). I think we shouldn't include anything we're not specifically using yet (we can copy over the implementation from the fork when we use it, at which point we can run the benchmarks to check the level of complexity. There are some things that duplicate internals we already have (bounds, for example) that could be folded in.
- Update the top-level Cargo.toml and use the new crates everywhere

I'm happy to review it all at once as well but I think the project will be able to better engage the community if we split it up a bit!

rust/sedona-testing/Cargo.toml
Kontinuation: How should we implement topological equality without depending on geo? Do we need to implement a simplified version of the topological equality check in sedona-testing?

Kontinuation:
The WKB/GEOS change cannot be done without the geo-traits-ext and geo-generic-alg change in place. Upgrading the wkb crate in `sedona-geos` will introduce conflicts of `geo-traits` versions. I can do this in the final step where we upgrade the top-level Cargo.toml.

The other steps looks fine. I'll submit several smaller PRs to gradually replace/upgrade the dependencies.

c/sedona-geos/Cargo.toml
Kontinuation: We only need `geo-types` here so I'll switch to use `geo-types` instead of `geo`. This is a test dependency so I thought that it should be fine.

Cargo.toml
Kontinuation: Yes. It will help avoid package name conflicts on crates.io.

rust/sedona-testing/Cargo.toml
paleolimbot: Maybe just feature flag it? It can even be in the default features (but then add `default-feautres = false` to the `sedona` dependency on `sedona-testing`, which is a little strange but we have it because of the random geometry generator).

jiayuasu:
@Kontinuation are we ready to merge this?

Kontinuation:
> @Kontinuation are we ready to merge this?

No. I'll break it into 5 smaller PRs. I have converted it to a draft to prevent it from being accidentally merged.

Kontinuation:
> * Test fixtures (https://github.com/apache/sedona-testing is perhaps a better place for these, or perhaps there is a way to expose them from the original crate from whence they came?)

The [`geo-test-fixtures` crate](https://github.com/georust/geo/tree/geo-0.31.0/geo-test-fixtures) is not published to crates.io, and it is only used as a dev-dependency in geo. I have tried depending on `geo-test-fixtures` using `geo-test-fixtures = { git = "https://github.com/georust/geo.git", tag = "geo-0.31.0"}` because we only use it as a dev-dependency as well, so having a git dependency does not block us from publishing the crates. However, I hit a strange error caused by duplicating the geo-types dependency (git rev vs. released), and there's no easy workaround.

I don't think asking georust/geo to publish geo-test-fixtures is a valid request, so I'll copy the [wkt files](https://github.com/georust/geo/tree/geo-0.31.0/geo-test-fixtures/fixtures) to apache/sedona-testing, and migrate [lib.rs](https://github.com/georust/geo/blob/geo-0.31.0/geo-test-fixtures/src/lib.rs) to rust/sedona-testing.

Kontinuation:
We've split this into several smaller PRs and merged them into main:

* https://github.com/apache/sedona-db/pull/190
* https://github.com/apache/sedona-db/pull/192
* https://github.com/apache/sedona-db/pull/193
* https://github.com/apache/sedona-db/pull/194
* https://github.com/apache/sedona-db/pull/195
* https://github.com/apache/sedona-db/pull/203

This PR can be closed.

PR #33, opened 2025-09-05
feat: Implement native Rust ST_Centroid and ST_Length functions

zhangfengcdt:
Implements native Rust versions of ST_Centroid and ST_Length functions using the geo-generic-alg library, providing substantial performance improvements over the existing GEOS-based implementations.

  - **Added ST_Centroid implementation** (`rust/sedona-geo/src/st_centroid.rs`)
    - Native Rust implementation using geo-generic-alg
    - Support for Point, LineString, Polygon, and GeometryCollection types
    - Registered as alternative to GEOS implementation

  - **Added ST_Length implementation** (`rust/sedona-geo/src/st_length.rs`)
    - Native Rust implementation using geo-generic-alg
    - Support for LineString, Polygon, and GeometryCollection types
    - Comprehensive length calculation including polygon perimeters

  - **Updated benchmark tests** (`benchmarks/test_functions.py`)
    - Modified ST_Length tests to use segments_large, collections_simple, and collections_complex tables
    - Enhanced test coverage for performance validation

## ST_Centroid

Before the fix:
```
----------------------------------------------- benchmark 'table=polygons_complex': 2 tests ------------------------------------------------
Name (time in ms)                                Median               Mean            StdDev                Min                Max
--------------------------------------------------------------------------------------------------------------------------------------------
test_st_centroid[polygons_complex-SedonaDB]      7.3047 (1.0)       7.3327 (1.0)      0.1581 (1.0)       7.0635 (1.0)       8.0731 (1.0)
test_st_centroid[polygons_complex-DuckDB]       22.6901 (3.11)     22.9047 (3.12)     0.5560 (3.52)     22.4189 (3.17)     24.6498 (3.05)
--------------------------------------------------------------------------------------------------------------------------------------------

---------------------------------------------- benchmark 'table=polygons_simple': 2 tests ---------------------------------------------
Name (time in ms)                              Median              Mean            StdDev               Min               Max
---------------------------------------------------------------------------------------------------------------------------------------
test_st_centroid[polygons_simple-DuckDB]       1.4665 (1.0)      1.4766 (1.0)      0.0496 (1.0)      1.4512 (1.0)      2.1513 (1.0)
test_st_centroid[polygons_simple-SedonaDB]     2.3950 (1.63)     2.3920 (1.62)     0.0915 (1.84)     1.7250 (1.19)     2.7119 (1.26)
---------------------------------------------------------------------------------------------------------------------------------------


```

After the fix:
```
[...]

PR #169, opened 2025-09-30
[WIP] feat: optimize KNN join with lock-free shared geometry cache

zhangfengcdt:
Optimizes KNN join performance by implementing a lock-free shared geometry cache that eliminates redundant WKB-to-geometry conversions across multiple queries.

Changes:
  - Use the new indexed distance metric and adapter pattern
  - Replace per-query geometry cache with shared `Vec<OnceCell<Geometry<f64>>>`
  - Use atomic operations for thread-safe lazy geometry initialization
  - Simplify memory tracking using WKB buffer size as proxy

NOTE:
This is PR is pending for the refactor changes be merged:
https://github.com/wherobots/geo-index/pull/6

### DIFF
--- a/rust/sedona-spatial-join/src/index.rs
+++ b/rust/sedona-spatial-join/src/index.rs
@@ -14,8 +14,7 @@
 // KIND, either express or implied.  See the License for the
 // specific language governing permissions and limitations
 // under the License.
-use std::cell::RefCell;
-use std::collections::HashMap;
+use once_cell::sync::OnceCell;
 use std::sync::{
     atomic::{AtomicUsize, Ordering},
     Arc,
@@ -248,6 +247,7 @@ impl SpatialIndexBuilder {
                 self.options,
                 AtomicUsize::new(self.probe_threads_count),
                 self.reservation,
+                self.memory_pool.clone(),
             ));
         }
 
@@ -275,6 +275,10 @@ impl SpatialIndexBuilder {
             ConcurrentReservation::try_new(REFINER_RESERVATION_PREALLOC_SIZE, refiner_reservation)
                 .unwrap();
 
+        let cache_size = batch_pos_vec.len();
+        let knn_components =
+            KnnComponents::new(cache_size, &self.indexed_batches, self.memory_pool.clone())?;
+
         Ok(SpatialIndex {
             schema,
             evaluator,
@@ -286,6 +290,7 @@ impl SpatialIndexBuilder {
             geom_idx_vec,
             visited_left_side,
             probe_threads_counter: AtomicUsize::new(self.probe_threads_count),
+            knn_components,
             reservation: self.reservation,
         })
     }
@@ -330,6 +335,9 @@ pub(crate) struct SpatialIndex {
     /// The last finished probe thread will produce the extra output batches for unmatched
     /// build side when running left-outer joins. See also [`report_probe_completed`].
     probe_threads_counter: AtomicUsize,
+
+    /// Shared KNN components (distance metrics and geometry cache) for efficient KNN queries
+    knn_components: KnnComponents,
 
     /// Memory reservation for tracking the memory usage of the spatial index
     /// Cleared on `SpatialIndex` drop
@@ -378,6 +386,7 @@ impl SpatialIndex {
         options: SpatialJoinOptions,
         probe_threads_counter: AtomicUsize,
         mut reservation: MemoryReservation,
+        memory_pool: Arc<dyn MemoryPool>,
     ) -> Self {
         let evaluator = create_operand_evaluator(&spatial_predicate, options.clone());
         let refiner = create_refiner(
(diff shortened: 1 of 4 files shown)


rust/sedona-spatial-join/src/index.rs
```diff
@@ -557,32 +537,15 @@ impl SpatialIndex {
...
-        // Choose distance metric based on use_spheroid parameter
-        let distance_metric: Box<dyn DistanceMetric<f32>> = if use_spheroid {
-            // For spheroid (geodesic) distance, we use the Haversine formula as an approximation for now.
-            // The distance metric will be used to calculate distances between geometries for ranking purposes.
-            Box::new(HaversineDistance::default())
-        } else {
-            Box::new(EuclideanDistance)
-        };
+        // Create an indexed distance metric adapter for KNN queries with cached distance metrics
+        let indexed_metric = self.create_knn_adapter(use_spheroid);
```
Kontinuation, line 541: The geometry cache in the KNN adapter is not reused across multiple `query_knn` calls. This renders the cache mostly useless.

rust/sedona-spatial-join/src/index.rs
```diff
@@ -937,6 +905,150 @@ async fn collect_build_partition(
...
+
+/// Enhanced indexed distance metric adapter for SedonaDB KNN queries.
+/// This adapter provides on-demand WKB decoding and geometry caching for efficient
+/// distance calculations with support for both Euclidean and Haversine distance metrics.
+struct SedonaKnnAdapter<'a> {
+    indexed_batches: &'a [IndexedBatch],
+    data_id_to_batch_pos: &'a [(i32, i32)],
+    distance_metric: &'a dyn DistanceMetric<f32>,
+    // Reference to KNN components for cache and memory tracking
+    knn_components: &'a KnnComponents,
```
Kontinuation, line 961: This is a shared reference owned by the spatial index, so the geometry cache was correctly reused. My previous review comment was incorrect. This should be OK.

rust/sedona-spatial-join/src/index.rs
```diff
@@ -937,6 +903,109 @@ async fn collect_build_partition(
...
+    /// Estimate the maximum memory usage for decoded geometries based on WKB sizes
+    fn estimate_max_memory_usage(indexed_batches: &[IndexedBatch]) -> usize {
+        let mut total_wkb_size = 0;
+
+        for batch in indexed_batches {
+            for wkb in batch.geom_array.wkbs().iter().flatten() {
+                total_wkb_size += wkb.buf().len();
+            }
+        }
+        total_wkb_size
```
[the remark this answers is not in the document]
zhangfengcdt: Using the wkb size is conservative and we are using the same approach as before.

.github/workflows/rust.yml
```diff
@@ -55,9 +55,9 @@ jobs:
     runs-on: ubuntu-latest
     env:
       CARGO_INCREMENTAL: 0
-      # Reduce debug info to save disk space
-      CARGO_PROFILE_DEV_DEBUG: 1
-      CARGO_PROFILE_TEST_DEBUG: 1
+      # Disable debug info completely to save disk space
+      CARGO_PROFILE_DEV_DEBUG: 0
+      CARGO_PROFILE_TEST_DEBUG: 0
```
paleolimbot: Can you see if bumping the cache id for the Rust cache fixes this instead of these changes?

.github/workflows/rust.yml
```diff
@@ -106,33 +109,50 @@ jobs:
...
+          # Also clean target/.rustc_info.json and other cache files
+          find target/debug -name "*.rlib" -delete
+          find target/debug -name "*.rmeta" -delete
 
       - name: Doctests
         if: matrix.name == 'test'
         run: |
           cargo test --workspace --doc --all-features
+          # Clean up after doctests
+          cargo clean
```
Kontinuation, line 167: Can we bring https://github.com/apache/sedona-db/pull/33/commits/facc7e7076d80f6b03d761d1196ec0795f97b345 back if the available disk space is tight for this workflow?

rust/sedona-spatial-join/src/index.rs
```diff
@@ -937,6 +903,112 @@ async fn collect_build_partition(
...
+            indexed_batches,
+            data_id_to_batch_pos,
+            knn_components,
+        }
+    }
+}
+
+impl<'a> GeometryAccessor for SedonaKnnAdapter<'a> {
+    /// Get geometry for the given item index with lock-free caching
+    fn get_geometry(&self, item_index: usize) -> Option<Geometry<f64>> {
```
Kontinuation: Can we avoid returning a clone? Given that all geo_type::Geometry values will be cached and owned by self, we can return `Option<&Geometry<f64>>` here.

zhangfengcdt: Yes, I will need to change the trait definition in geo-index first before changing this to return reference.

zhangfengcdt: @Kontinuation Here is the change to geo-index: https://github.com/wherobots/geo-index/pull/8

zhangfengcdt: The clone should have been removed by using the new geo-index refactoring.

Kontinuation (approved):

Issue #162, opened 2025-09-28
bug: Parser can't infer ZM dimensionality correctly for certain WKT strings


petern48:
Creating either one of the following geometries fails with the following error on SedonaDB, even though they succeed on PostGIS. This behavior occurs in the cli, the tests, etc.

```sql
select st_geomfromtext('GEOMETRYCOLLECTION Z (POINT (1 2 3))')
-- fails

select st_geomfromtext('GEOMETRYCOLLECTION (POINT (1 2 3))');
-- postgis infers this to be xyz based on st_hasz, st_hasm calls

select st_geomfromtext('point (1 1 1)');
-- postgis infers this to be xyz based on st_hasz, st_hasm calls
```

Error:
```
Internal error: WKT parse error: Missing closing parenthesis for type.
This issue was likely caused by a bug in DataFusion's code. Please help us to resolve this by filing a bug report in our issue tracker: https://github.com/apache/datafusion/issues
```
(Ignore that it misleadingly says "Datafusion's code")

paleolimbot:
Thank you for opening!

This would be coming from the `wkt` crate, although we could also use geoarrow-c to speed this up considerably. There's definitely a lot of WKT out there along the lines of `POINT(1 2 3)` even though I don't think it is strictly correct. I'm not sure if that is worth "fixing" (it's slower to parse the WKT if you have to infer the dimensions from the coordinates).

(We should also fix it to use an `execution_err!()`!)

petern48:
If / whenever this is supported, we should add these test cases to `st_has(z/m)`, as mentioned in https://github.com/apache/sedona-db/issues/170

paleolimbot:
For now supporting `POINT (0 1 2)` (vs the spec compliant `POINT Z (0 1 2)` or `POINT M (0 1 2)`) is out of scope...we don't have an internal WKT Parser and neither of the ones we could use are planning to support this. We can repoen if this comes up from a user perspective!

PR #171, opened 2025-10-01
perf: Optimize st_has(z/m) using WKBBytesExecutor

petern48:
See issue #162. Turns out more complicated then you'd think

### DIFF
--- a/python/sedonadb/tests/functions/test_functions.py
+++ b/python/sedonadb/tests/functions/test_functions.py
@@ -473,6 +473,8 @@ def test_st_geomfromwkb(eng, geom):
         ("POINT Z EMPTY", True),
         ("POINT M EMPTY", False),
         ("POINT ZM EMPTY", True),
+        # SedonaDB can't parse this yet: https://github.com/apache/sedona-db/issues/162
+        # ("POINT (0 0 0)", True),
         ("POINT Z (0 0 0)", True),
         ("POINT M (0 0 0)", False),
         ("POINT ZM (0 0 0 0)", True),
@@ -481,8 +483,18 @@ def test_st_geomfromwkb(eng, geom):
         ("LINESTRING Z (0 0 0, 1 1 1)", True),
         ("POLYGON EMPTY", False),
         ("MULTIPOINT ((0 0), (1 1))", False),
+        ("MULTIPOINT Z ((0 0 0))", True),
+        # SedonaDB can't parse this yet: https://github.com/apache/sedona-db/issues/162
+        # ("MULTIPOINT ((0 0 0))", True),
+        ("MULTIPOINT ZM ((0 0 0 0))", True),
         ("GEOMETRYCOLLECTION EMPTY", False),
+        # Z-dim specified only in the nested geometry
         ("GEOMETRYCOLLECTION (POINT Z (0 0 0))", True),
+        # SedonaDB can't parse this yet: https://github.com/apache/sedona-db/issues/162
+        # Z-dim specified only on the geom collection level but not the nested geometry level
+        # ("GEOMETRYCOLLECTION Z (POINT (0 0 0))", True),
+        # Z-dim specified on both levels
+        ("GEOMETRYCOLLECTION Z (POINT Z (0 0 0))", True),
         ("GEOMETRYCOLLECTION (GEOMETRYCOLLECTION (POINT Z (0 0 0)))", True),
     ],
 )
--- a/rust/sedona-functions/src/st_haszm.rs
+++ b/rust/sedona-functions/src/st_haszm.rs
@@ -158,6 +158,10 @@ fn infer_haszm(buf: &[u8], dim_index: usize) -> Result<Option<bool>> {
         // If empty geometry (num_geometries == 0), fallback to below logic to check the geom collection's dimension
         // GEOMETRY COLLECTION Z EMPTY hasz -> true
     }
+
+    // TODO: Last check: check how many dimensions the 1st coordinate has (all other coordinates must have the same)
+    // e.g handle this case: POINT (0 0 0) -> xyz dimension
+
     // If code was unspecified / xy and we couldn't infer the dimension, it must be xy
     Ok(Some(false))
 }
--- a/rust/sedona-functions/src/st_haszm.rs
+++ b/rust/sedona-functions/src/st_haszm.rs
@@ -136,7 +136,7 @@ fn infer_haszm(buf: &[u8], dim_index: usize) -> Result<Option<bool>> {
         _ => return sedona_internal_err!("Unexpected code: {code}"),
     };
 
-    // If GeometryCollection (7), we need to check the dimension of the next geometry
+    // If GeometryCollection (7), we need to check the dimension of the first geometry
     if code & 0x7 == 7 {
         // The next 4 bytes are the number of geometries in the collection
         let num_geometries = match byte_order {
@@ -160,9 +160,9 @@ fn infer_haszm(buf: &[u8], dim_index: usize) -> Result<Option<bool>> {
     }
 
     // TODO: Last check: check how many dimensions the 1st coordinate has (all other coordinates must have the same)
-    // e.g handle this case: POINT (0 0 0) -> xyz dimension
+    // e.g handle this case: POINT (0 0 0) -> xyz dimension, POINT (0 0 0 0) -> xyzm dimension
 
-    // If code was unspecified / xy and we couldn't infer the dimension, it must be xy
+    // If code was unspecified and we couldn't infer the dimension, it must be xy
     Ok(Some(false))
 }
 
--- a/rust/sedona-functions/src/st_haszm.rs
+++ b/rust/sedona-functions/src/st_haszm.rs
@@ -16,7 +16,7 @@
 // under the License.
 use std::sync::Arc;
 
-use crate::executor::WkbExecutor;
+use crate::executor::WkbBytesExecutor;
 use arrow_array::builder::BooleanBuilder;
 use arrow_schema::DataType;
 use datafusion_common::error::Result;
@@ -90,13 +90,13 @@ impl SedonaScalarKernel for STHasZm {
             _ => unreachable!(),
         };
 
-        let executor = WkbExecutor::new(arg_types, args);
+        let executor = WkbBytesExecutor::new(arg_types, args);
         let mut builder = BooleanBuilder::with_capacity(executor.num_iterations());
 
         executor.execute_wkb_void(|maybe_item| {
             match maybe_item {
                 Some(item) => {
-                    builder.append_option(invoke_scalar(&item, dim_index)?);
+                    builder.append_option(infer_haszm(&item, dim_index)?);
                 }
                 None => builder.append_null(),
             }
@@ -107,28 +107,59 @@ impl SedonaScalarKernel for STHasZm {
     }
 }
 
-fn invoke_scalar(item: &Wkb, dim_index: usize) -> Result<Option<bool>> {
-    match item.as_type() {
-        geo_traits::GeometryType::GeometryCollection(collection) => {
-            use geo_traits::GeometryCollectionTrait;
-            if collection.num_geometries() == 0 {
-                Ok(Some(false))
-            } else {
-                // PostGIS doesn't allow creating a GeometryCollection with geometries of different dimensions
-                // so we can just check the dimension of the first one
-                let first_geom = unsafe { collection.geometry_unchecked(0) };
-                invoke_scalar(first_geom, dim_index)
-            }
-        }
-        _ => {
-            let geom_dim = item.dim();
-            match dim_index {
-                2 => Ok(Some(matches!(geom_dim, Dimensions::Xyz | Dimensions::Xyzm))),
-                3 => Ok(Some(matches!(geom_dim, Dimensions::Xym | Dimensions::Xyzm))),
-                _ => sedona_internal_err!("unexpected dim_index"),
-            }
-        }
+/// Fast-path inference of geometry type name from raw WKB bytes
+/// An error will be thrown for invalid WKB bytes input
+///
+/// Spec: https://libgeos.org/specifications/wkb/
+fn infer_haszm(buf: &[u8], dim_index: usize) -> Result<Option<bool>> {
+    if buf.len() < 5 {
+        return sedona_internal_err!("Invalid WKB: buffer too small ({} bytes)", buf.len());
     }
+
+    let byte_order = buf[0];
+    let code = match byte_order {
+        0 => u32::from_be_bytes([buf[1], buf[2], buf[3], buf[4]]),
+        1 => u32::from_le_bytes([buf[1], buf[2], buf[3], buf[4]]),
+        other => return sedona_internal_err!("Unexpected byte order: {other}"),
+    };
+
+    // 0000 -> xy or unspecified
+    // 1000 -> xyz
+    // 2000 -> xym
+    // 3000 -> xyzm
+    match code / 1000 {
+        // If xy, it's possible we need to infer the dimension
+        0 => {},
+        1 => return Ok(Some(dim_index == 2)),
+        2 => return Ok(Some(dim_index == 3)),
+        3 => return Ok(Some(true)),
+        _ => return sedona_internal_err!("Unexpected code: {code}"),
+    };
+
+    // If GeometryCollection (7), we need to check the dimension of the next geometry
+    if code & 0x7 == 7 {
+        // The next 4 bytes are the number of geometries in the collection
+        let num_geometries = match byte_order {
+            0 => u32::from_be_bytes([buf[5], buf[6], buf[7], buf[8]]),
+            1 => u32::from_le_bytes([buf[5], buf[6], buf[7], buf[8]]),
+            other => return sedona_internal_err!("Unexpected byte order: {other}"),
+        };
+        // Check the dimension of the first geometry since they all have to be the same dimension
+        // Note: Attempting to create the following geometries error and are thus not possible to create:
+        // - Nested geometry dimension doesn't match the **specified** geom collection z-dimension
+        //   - GEOMETRYCOLLECTION M (POINT Z (1 1 1))
+        // - Nested geometry doesn't have the specified dimension
+        //   - GEOMETRYCOLLECTION Z (POINT (1 1))
+        // - Nested geometries have different dimensions
+        //   - GEOMETRYCOLLECTION (POINT Z (1 1 1), POINT (1 1))
+        if num_geometries >= 1 {
+            return infer_haszm(&buf[9..], dim_index);
+        }
+        // If empty geometry (num_geometries == 0), fallback to below logic to check the geom collection's dimension
+        // GEOMETRY COLLECTION Z EMPTY hasz -> true
+    }
+    // If code was unspecified / xy and we couldn't infer the dimension, it must be xy
+    Ok(Some(false))
 }
 
 #[cfg(test)]
@@ -172,6 +203,16 @@ mod tests {
         let result = m_tester.invoke_scalar("POINT (1 2)").unwrap();
         m_tester.assert_scalar_result_equals(result, false);
 
+        // SedonaDB can't parse this yet: https://github.com/apache/sedona-db/issues/162
+        // Infer z-dimension
+        // let result = z_tester.invoke_scalar("POINT (1 2 3)").unwrap();
+        // z_tester.assert_scalar_result_equals(result, true);
+
+        // SedonaDB can't parse this yet: https://github.com/apache/sedona-db/issues/162
+        // Infer z-dimension
+        // let result = m_tester.invoke_scalar("POINT (1 2 3)").unwrap();
+        // m_tester.assert_scalar_result_equals(result, false);
+
         let result = z_tester.invoke_scalar("POINT M (1 2 3)").unwrap();
         z_tester.assert_scalar_result_equals(result, false);
 
@@ -184,10 +225,26 @@ mod tests {
         let result = m_tester.invoke_wkb_scalar(None).unwrap();
         m_tester.assert_scalar_result_equals(result, ScalarValue::Null);
 
+        // Z-dimension specified only in the nested geometry, but not the geom collection level
         let result = z_tester
             .invoke_wkb_scalar(Some("GEOMETRYCOLLECTION (POINT Z (1 2 3))"))
             .unwrap();
         z_tester.assert_scalar_result_equals(result, ScalarValue::Boolean(Some(true)));
+
+        // Z-dimension specified on both the geom collection and nested geometry level
+        // Geometry collection with Z dimension both on the geom collection and nested geometry level
+        let result = z_tester
+            .invoke_wkb_scalar(Some("GEOMETRYCOLLECTION Z (POINT Z (1 2 3))"))
+            .unwrap();
+        z_tester.assert_scalar_result_equals(result, ScalarValue::Boolean(Some(true)));
+
+        // SedonaDB can't parse this yet: https://github.com/apache/sedona-db/issues/162
+        // Z-dimension specified only on the geom collection level but not the nested geometry level
+        // Geometry collection with Z dimension both on the geom collection and nested geometry level
+        // let result = z_tester
+        //     .invoke_wkb_scalar(Some("GEOMETRYCOLLECTION Z (POINT (1 2 3))"))
+        //     .unwrap();
+        // z_tester.assert_scalar_result_equals(result, ScalarValue::Boolean(Some(true)));
 
         let result = m_tester
             .invoke_wkb_scalar(Some("GEOMETRYCOLLECTION (POINT M (1 2 3))"))
@@ -203,5 +260,16 @@ mod tests {
             .invoke_wkb_scalar(Some("GEOMETRYCOLLECTION EMPTY"))
             .unwrap();
         m_tester.assert_scalar_result_equals(result, ScalarValue::Boolean(Some(false)));
+
+        // Empty geometry collections with Z or M dimensions
+        let result = z_tester
+            .invoke_wkb_scalar(Some("GEOMETRYCOLLECTION Z EMPTY"))
+            .unwrap();
+        z_tester.assert_scalar_result_equals(result, ScalarValue::Boolean(Some(true)));
+
+        let result = m_tester
+            .invoke_wkb_scalar(Some("GEOMETRYCOLLECTION M EMPTY"))
+            .unwrap();
+        m_tester.assert_scalar_result_equals(result, ScalarValue::Boolean(Some(true)));
     }
 }


rust/sedona-functions/src/st_haszm.rs
```diff
@@ -107,28 +107,63 @@ impl SedonaScalarKernel for STHasZm {
...
+        //   - GEOMETRYCOLLECTION (POINT Z (1 1 1), POINT (1 1))
+        if num_geometries >= 1 {
+            return infer_haszm(&buf[9..], dim_index);
         }
+        // If empty geometry (num_geometries == 0), fallback to below logic to check the geom collection's dimension
+        // GEOMETRY COLLECTION Z EMPTY hasz -> true
     }
+
+    // TODO: Last check: check how many dimensions the 1st coordinate has (all other coordinates must have the same)
+    // e.g handle this case: POINT (0 0 0) -> xyz dimension, POINT (0 0 0 0) -> xyzm dimension
```
petern48: This isn't fully correct yet, so it isn't quite done (nor can we test it #162). This part would require another recursive function to get the first coordinate. Definitely doable, but I'd rather wait for feedback on whether we'd want to merge a PR that complicates this function this much.

python/sedonadb/tests/functions/test_functions.py
```diff
@@ -473,6 +473,8 @@ def test_st_geomfromwkb(eng, geom):
         ("POINT Z EMPTY", True),
         ("POINT M EMPTY", False),
         ("POINT ZM EMPTY", True),
+        # SedonaDB can't parse this yet: https://github.com/apache/sedona-db/issues/162
+        # ("POINT (0 0 0)", True),
```
paleolimbot: From a WKB perspective, this is actually identical...that is a WKT parser issue, not an underlying data representation issue for us.

petern48:
I'd also want to convert that function to one that returns the dimensionality (e.g xy, xyz, etc) and then use that to implement `st_haszm`, in case that logic can be reused elsewhere.

python/sedonadb/tests/functions/test_functions.py
```diff
@@ -481,8 +483,18 @@ def test_st_geomfromwkb(eng, geom):
         ("LINESTRING Z (0 0 0, 1 1 1)", True),
         ("POLYGON EMPTY", False),
         ("MULTIPOINT ((0 0), (1 1))", False),
+        ("MULTIPOINT Z ((0 0 0))", True),
+        # SedonaDB can't parse this yet: https://github.com/apache/sedona-db/issues/162
+        # ("MULTIPOINT ((0 0 0))", True),
```
paleolimbot: This one we actually can test, but you need to generate the WKB bytes yourself. Something like `0x01, 0x04, 0x00, 0x00, 0x00, 0x01, 0x00, 0x00, 0x00, <the WKB for POINT Z (0 0 0)>` (0x01 for little endian, 4 as a little endian int 32 for multipoint, 1 as a little endian int for one point, the point).

python/sedonadb/tests/functions/test_functions.py
```diff
@@ -481,8 +483,18 @@ def test_st_geomfromwkb(eng, geom):
...
+        ("MULTIPOINT ZM ((0 0 0 0))", True),
         ("GEOMETRYCOLLECTION EMPTY", False),
+        # Z-dim specified only in the nested geometry
         ("GEOMETRYCOLLECTION (POINT Z (0 0 0))", True),
+        # SedonaDB can't parse this yet: https://github.com/apache/sedona-db/issues/162
+        # Z-dim specified only on the geom collection level but not the nested geometry level
+        # ("GEOMETRYCOLLECTION Z (POINT (0 0 0))", True),
```
paleolimbot: This is also a parser issue, not a data representation issue.

rust/sedona-functions/src/st_haszm.rs
```diff
@@ -172,6 +207,16 @@ mod tests {
         let result = m_tester.invoke_scalar("POINT (1 2)").unwrap();
         m_tester.assert_scalar_result_equals(result, false);
 
+        // SedonaDB can't parse this yet: https://github.com/apache/sedona-db/issues/162
+        // Infer z-dimension
+        // let result = z_tester.invoke_scalar("POINT (1 2 3)").unwrap();
+        // z_tester.assert_scalar_result_equals(result, true);
```
paleolimbot: See my comment above...WKT parsers that accept this (GEOS/PostGIS) would write out the WKB identically to `POINT Z (1 2 3)`.

`MULTIPOINT ((1 2 3))` is a bit different. You should be able to generate bytes for that using shapely (or use my sketch above to do it by hand). We have a place for WKB bytes of things that are hard to generate using WKT:

https://github.com/apache/sedona-db/blob/a15844b4ff9c7e9416b8f1c1c07ad81d908a89cd/rust/sedona-testing/src/fixtures.rs#L17-L24

rust/sedona-functions/src/st_haszm.rs
```diff
@@ -107,28 +107,63 @@ impl SedonaScalarKernel for STHasZm {
...
-                _ => sedona_internal_err!("unexpected dim_index"),
-            }
+/// Fast-path inference of geometry type name from raw WKB bytes
+/// An error will be thrown for invalid WKB bytes input
+///
+/// Spec: https://libgeos.org/specifications/wkb/
+fn infer_haszm(buf: &[u8], dim_index: usize) -> Result<Option<bool>> {
+    if buf.len() < 5 {
+        return sedona_internal_err!("Invalid WKB: buffer too small ({} bytes)", buf.len());
+    }
```
paleolimbot: We should consider consolidating this with the geometry type optimization like:

```rust
struct WkbHeader {
  geometry_type: u32,
  size: u32,
  first_coord_geometry_type: u32,
  first_coord_offset: u32
}

impl WkbHeader {
  pub fn geometry_type_id(&self) -> GeometryTypeId {...}
  pub fn dimensions(&self) -> Dimensions { ... }
  pub fn num_dimensions(&self) -> usize { ... }
}
```

There are a few functions that can benefit from this (npoints in a few cases, hilbert, isempty), although it might make a better PR into the `wkb` crate where there's already the logic for the parsing.

rust/sedona-functions/src/st_haszm.rs
```diff
@@ -107,28 +107,63 @@ impl SedonaScalarKernel for STHasZm {
...
+    // 2000 -> xym
+    // 3000 -> xyzm
+    match code / 1000 {
+        // If xy, it's possible we need to infer the dimension
+        0 => {}
+        1 => return Ok(Some(dim_index == 2)),
+        2 => return Ok(Some(dim_index == 3)),
+        3 => return Ok(Some(true)),
+        _ => return sedona_internal_err!("Unexpected code: {code}"),
+    };
```
paleolimbot: I am not sure whether or not we should handle EWKB here (the rest of the package does via parsing by the wkb crate). In that case you would also need to check the high bits for ZM flags too.

rust/sedona-functions/src/st_haszm.rs
petern48: Based on what you're saying, this logic be unnecessary, so I will remove it.

python/sedonadb/tests/functions/test_functions.py
petern48: Before testing it, I assumed you meant the outer dimension code would be xy instead of xyz, but based on the script below, I get the same byte representation as `MULTIPOINT Z ((0 0 0))` (type code 1004, xyz multipoint). Am I missing something? What did you have in mind?

```python
# shapely 2.1.2
wkt = "MULTIPOINT Z ((1 1 1))"
geom = shapely.from_wkt(wkt)

byte_order = geom.wkb[0]
assert byte_order == 1
iso_bytes = shapely.to_wkb(geom, flavor="iso")

print(iso_bytes)

# adds a few spaces for readability
# 1  ec 03 00 00  0100000001e9030000000000000000f03f000000000000f03f000000000000f03f

# The type code `ec 03 00 00` translates to 1004 (an xyz multipoint)
```

rust/sedona-functions/src/st_haszm.rs
petern48: Yeah, I like this idea. I was thinking just create it in another file in `sedona-geometry` for now, and then we can later decide to upstream to `wkb` if it makes sense to.

python/sedonadb/tests/functions/test_functions.py
paleolimbot: Given that it's this difficult to actually generate such well-known binary, I'm not sure we have to handle this case. What you're up against here is that you need the parser, the internal representation of the geometry (there: GEOS), and the WKB writer to all support generating what is actually very fishy output...most don't let you do that (so if you want to test this case, you're going to have to hard-code the WKB yourself).

rust/sedona-functions/src/st_haszm.rs
petern48: Interesting, I didn't think we handled EWKB anywhere in sedona-db, since the CRS info is stored at the type level (at least at the moment). Considering that, why would we use EWKB in sedona-db at all? How do we detect whether something is EWKB or WKB in other parts of the codebase? If EWKB is really possible, it would break the implementation for the GEOMETRYCOLLECTION where I need to recurse to the nested geometry since there would be an additional byte offset for the SRID after wkbType.

python/sedonadb/tests/functions/test_functions.py
petern48: I see, you're saying test it even though it may not necessarily be representable in WKT. Added the fixture and test.

rust/sedona-functions/src/st_haszm.rs
petern48: Done. Let me know what you think about the approach. I'll do a follow-up PR to make `st_geometrytype` use this, since I have another small feature to implement with it. Keep in mind that the dimension() calculation can potentially recurse a lot (e.g a bunch of nested `GEOMETRYCOLLECTION`s), so I'd like to avoid just computing all of the fields at construction and saving them as fields. I instead went the route of computing them lazily and using the fields for caching values after they are calculated.

petern48:
Added perf benchmarks to the PR description 🤠

petern48:
I can't import `sedona-testing` into `sedona-geometry` due to a circular dependency, and hence can't import the fixture into the `wkb_header.rs` to be tested. It is at least being tested in `st_has_(z/m)`, so I'm confident the logic is right. Let me know if you'd rather move something around or copy-paste the test over, otherwise, this is how I'll leave it.

The unparseable WKT strings are still left in the code as comments at the moment, though I did also mention them in https://github.com/apache/sedona-db/issues/162 as a separate reminder if / whenever that's fixed. Personally, I prefer to leave the comments in the code as an additional reminder, but if you'd rather have me delete them. Let me know.

python/sedonadb/tests/functions/test_functions.py
paleolimbot: (Not relevant to this test, which is not about parsing WKT)

```suggestion
```

python/sedonadb/tests/functions/test_functions.py
paleolimbot: (We should still remove these, though. These comments highlight WKT parsing issues and this is a WKB test)

```suggestion
```

python/sedonadb/tests/functions/test_functions.py
```suggestion
```

rust/sedona-testing/src/fixtures.rs
```diff
@@ -22,3 +22,17 @@ pub const MULTIPOINT_WITH_EMPTY_CHILD_WKB: [u8; 30] = [
...
+pub const MULTIPOINT_WITH_INFERRED_Z_DIMENSION_WKB: [u8; 38] = [
+    0x01, // byte-order
+    0x04, 0x00, 0x00, 0x00, // multipoint with xy-dimension specified
+    0x01, 0x00, 0x00, 0x00, // 1 point
+    // nested point geom
+    0x01, // byte-order
+    0xe9, 0x03, 0x00, 0x00, // point with xyz-dimension specified
+    0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, // x-coordinate of point
+    0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, // y-coordinate of point
+    0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, // z-coordinate of point
```
paleolimbot: optional nit: This would be a slightly better fixture if these were non-zero (1, 2, 3, for example).

rust/sedona-geometry/src/wkb_header.rs
```diff
@@ -0,0 +1,363 @@
...
+/// Performs operations lazily and caches them after the first computation
+pub struct WkbHeader<'a> {
+    buf: &'a [u8],
+    geometry_type_id: Option<GeometryTypeId>,
+    dimensions: Option<Dimensions>,
+}
+
+impl<'a> WkbHeader<'a> {
+    /// Creates a new [WkbHeader] from a buffer
+    pub fn new(buf: &'a [u8]) -> Result<Self> {
```
paleolimbot:
```suggestion
    pub fn try_new(buf: &'a [u8]) -> Result<Self> {
```

(I believe it's `Wkb::try_new()` as well)

rust/sedona-geometry/src/wkb_header.rs
```diff
@@ -0,0 +1,363 @@
...
+use datafusion_common::error::{DataFusionError, Result};
+use geo_traits::Dimensions;
+use sedona_common::sedona_internal_err;
+
+/// Fast-path WKB header parser
+/// Performs operations lazily and caches them after the first computation
+pub struct WkbHeader<'a> {
+    buf: &'a [u8],
+    geometry_type_id: Option<GeometryTypeId>,
+    dimensions: Option<Dimensions>,
```
paleolimbot: How about:

```suggestion
    geometry_type: u32,
    size: u32,
    srid: u32,
    first_xy: (f64, f64)
```

- I don't think you need to keep a reference to the slice here...whoever was calling this has access to that already.
- The laziness here I don't think is helping performance (these objects aren't in places where the values are getting reused and caching the very cheap calculation would help)
- Size + first_xy are enough to ensure non-empty in most cases (we need the first_xy for points in this case)
- first_xy is something I need shortly to implement spatial sorting
- srid is something I need shortly to implement EWKB parsing into our new row-level SRID type
- Other things can be implemented as methods on top of those values

rust/sedona-geometry/src/wkb_header.rs
```diff
@@ -0,0 +1,363 @@
...
+        return sedona_internal_err!("Invalid WKB: buffer too small ({} bytes)", buf.len());
+    }
+
+    let byte_order = buf[0];
+
+    let code = match byte_order {
+        0 => u32::from_be_bytes([buf[1], buf[2], buf[3], buf[4]]),
+        1 => u32::from_le_bytes([buf[1], buf[2], buf[3], buf[4]]),
+        other => return sedona_internal_err!("Unexpected byte order: {other}"),
+    };
```
paleolimbot: I would put this in `try_new()`

(Also, `sedona_internal_err!()` should be `exec_err!()`)

rust/sedona-geometry/src/wkb_header.rs
```diff
@@ -0,0 +1,363 @@
...
+    // 2000 -> xym
+    // 3000 -> xyzm
+    match code / 1000 {
+        // If xy, it's possible we need to infer the dimension
+        0 => {}
+        1 => return Ok(Dimensions::Xyz),
+        2 => return Ok(Dimensions::Xym),
+        3 => return Ok(Dimensions::Xyzm),
+        _ => return sedona_internal_err!("Unexpected code: {code}"),
+    };
```
paleolimbot: This should also handle EWKB high bit flags. Most of the time this will be ISO WKB from GeoParquet but not all tools have control over the type of WKB they generate and we're better for dealing with it (unless you can demonstrate measurable performance overhead, which I doubt is the case here). One notable data point is that WKB coming from Sedona Spark's `dataframe_to_arrow()` is EWKB.

rust/sedona-geometry/src/wkb_header.rs
```diff
@@ -0,0 +1,363 @@
...
+            other => return sedona_internal_err!("Unexpected byte order: {other}"),
+        };
+        // Check the dimension of the first geometry since they all have to be the same dimension
+        // Note: Attempting to create the following geometries error and are thus not possible to create:
+        // - Nested geometry dimension doesn't match the **specified** geom collection z-dimension
+        //   - GEOMETRYCOLLECTION M (POINT Z (1 1 1))
+        // - Nested geometry doesn't have the specified dimension
+        //   - GEOMETRYCOLLECTION Z (POINT (1 1))
+        // - Nested geometries have different dimensions
+        //   - GEOMETRYCOLLECTION (POINT Z (1 1 1), POINT (1 1))
```
paleolimbot: This part is, I believe unique to `st_hasz()` and should possibly live in the file implementing that function (or be explicit in the name of the function...I think of `dimensions` as the explicitly declared dimensions at the top-level WKB).

rust/sedona-geometry/src/wkb_header.rs
```diff
@@ -0,0 +1,363 @@
...
+        if self.geometry_type_id.is_none() {
+            self.geometry_type_id = Some(parse_geometry_type_id(self.buf)?);
+        }
+        self.geometry_type_id.ok_or_else(|| {
+            DataFusionError::External("Unexpected internal state in WkbHeader".into())
+        })
+    }
+
+    /// Returns the dimension of the WKB by only parsing what's minimally necessary instead of the entire WKB
+    pub fn dimension(mut self) -> Result<Dimensions> {
```
paleolimbot:
```suggestion
    pub fn dimensions(mut self) -> Result<Dimensions> {
```

(`dimension` is something different)

rust/sedona-functions/src/st_haszm.rs
```diff
@@ -107,28 +106,27 @@ impl SedonaScalarKernel for STHasZm {
...
-            match dim_index {
-                2 => Ok(Some(matches!(geom_dim, Dimensions::Xyz | Dimensions::Xyzm))),
-                3 => Ok(Some(matches!(geom_dim, Dimensions::Xym | Dimensions::Xyzm))),
-                _ => sedona_internal_err!("unexpected dim_index"),
-            }
-        }
+/// Fast-path inference of geometry type name from raw WKB bytes
+/// An error will be thrown for invalid WKB bytes input
+///
+/// Spec: https://libgeos.org/specifications/wkb/
```
paleolimbot:
```suggestion
```

paleolimbot (commented):
This is going to be so cool! I left some suggestions about reorganizing the `WkbHeader` to support a few of the other things I'd like to do with it 🙂

rust/sedona-geometry/src/wkb_header.rs
petern48: I think this logic should be kept here, actually. The logic of `st_hasz()` is simply to get the dimensionality of the object and see if it has a z-dimension. No other special logic. This logic here you're referring to is for handling a slight nuance in how SedonaDB converts WKT to WKB. Specifically, it translates the following geometry into WKB where the top-most dimension is specified as xy, while all of the actual coordinates in the geometry have z dimension.

e.g `GEOMETRYCOLLECTION (POINT Z (1 2 3))`
(I'd expect the same issue with `MULTIPOINT ((1 1 1))` is WKT supported parsing it)

I think of these examples as geometries that really are xyz dimension, but rely on us to infer the z-dimension.

SedonaDB parses the first example as follows:
```sql
select st_asbinary(st_geomfromtext('geometrycollection (point z (1 2 3))'));
-- 01**07000000**0100000001e9030000000000000000f03f00000000000000400000000000000840
```

Notably, the top-level dimensionality is simply `7` (xy), whereas we should really be interpretting the whole thing as an xyz.

Interestingly, the same query on PostGIS, returns the binary as the following where the top-level dimensionality is xyz.
```
01**ef030000**0100000001e9030000000000000000f03f00000000000000400000000000000840
```

I'm not sure if this is a bug in how WKT is translated into WKB, but this logic should be necessary to interpret that WKT the same way as PostGIS interprets it. We'd want to kept this logic for ST_ZMFlag, for example. Are there any concrete functions you can think of where we'd want to take the top-level dimension and ignore any potential extra dimensions in the points?

petern48: Hmm, unless you want to move this `dimensions()` method from the `WKBHeader` class entirely. I do like your idea of removing the `buf` field from the class, but considering this edge case, I see two options we could do to maintain correctness:
1. Move the `dimensions()` function outside and don't provide any method inside of `WKBHeader`
2. In `try_new()` also check the dimension of the first coordinate (e.g, it's xyz) and store that as a separate field to be retrieved in the `dimensions()` method. We could get this info during our pass to getting `first_xy`.

edit: I'm working on option 2 atm, unless you say otherwise

paleolimbot: Got it! How about two methods:

- `dimensions(&self)` (top-level dimensions as declared by WKB)
- `first_coord_dimensions(&self)`

Which one of those you want mostly depends why you're asking if a geometry has a Z component or what previous information you have (our implementation of st_z, for example, would have no need for the second version).

These are also *both* approximations...there's nothing stopping somebody from putting a Z value in in the *second* collection item (does PostGIS only check the first one?). Since neither are truly correct I don't think the `WkbHeader` should take sides...just provide information. If somebody really does need to wrangle badly written data from SQL there are other tools at their disposal (`st_dump()` maybe)...if a particular algorithm must know if if there are Z values, it should probably check the entire collection.

rust/sedona-geometry/src/wkb_header.rs
```diff
@@ -0,0 +1,728 @@
...
+    // number of rings for a polygon
+    // number of geometries for a MULTIPOINT, MULTILINESTRING, MULTIPOLYGON, or GEOMETRYCOLLECTION
+    size: u32,
+    // SRID if given buffer was EWKB. Otherwise, 0.
+    srid: u32,
+    // First x,y coordinates for a point. Otherwise (f64::NAN, f64::NAN) if empty
+    first_xy: (f64, f64),
+    // Dimensions of the first nested geometry of a collection or None if empty
+    // For POINT, LINESTRING, POLYGON, returns the dimensions of the geometry
+    first_geom_dimensions: Option<Dimensions>,
```
petern48, line 43: I made it `first_geom_dimensions` instead of `first_coord_dimensions` since that's logically what I'm actually doing, finding the first non-collection geometry (using `first_geom_idx()`) and then taking the dimensions field of that. I'm not somehow checking the values of the first coordinate and determining whether it's xy or xyz. Feel free to propose a different name if you think it's oddly named.

rust/sedona-geometry/src/wkb_header.rs
```diff
@@ -0,0 +1,728 @@
...
+    }
+
+    // #[test]
+    // fn srid() {
+    //     // This doesn't work
+    //     let wkb = make_wkb("SRID=4326;POINT (1 2)");
+    //     println!("wkb: {:?}", wkb);
+    //     let header = WkbHeader::try_new(&wkb).unwrap();
+    //     assert_eq!(header.srid(), 4326);
+    // }
```
petern48: Can you give me some advice on how to test this? Any nice helper functions for SRIDs?

rust/sedona-geometry/src/wkb_header.rs
```diff
@@ -0,0 +1,728 @@
...
+        let header = WkbHeader::try_new(&wkb).unwrap();
+        assert_eq!(header.first_xy(), (1.0, 2.0));
+
+        let wkb = make_wkb("GEOMETRYCOLLECTION (POINT (1 2), LINESTRING (1 2, 3 4), POLYGON ((0 0, 0 1, 1 0, 0 0)))");
+        let header = WkbHeader::try_new(&wkb).unwrap();
+        assert_eq!(header.first_xy(), (1.0, 2.0));
+    }
+
+    #[test]
+    fn empty_first_xy() {
```
petern48, line 703: Double-check if this is the behavior you want. I took a guess that you want (NaN, NaN) returned for any empty geoms. None seemed reasonable too.

rust/sedona-geometry/src/wkb_header.rs
```diff
@@ -0,0 +1,728 @@
...
+        assert_eq!(header.first_geom_dimensions(), Some(Dimensions::Xy));
+
+        let wkb = make_wkb("POLYGON Z EMPTY");
+        let header = WkbHeader::try_new(&wkb).unwrap();
+        assert_eq!(header.first_geom_dimensions(), Some(Dimensions::Xyz));
+
+        // Empty collections should return None
+        let wkb = make_wkb("MULTIPOINT EMPTY");
+        let header = WkbHeader::try_new(&wkb).unwrap();
+        assert_eq!(header.first_geom_dimensions(), None);
```
petern48, line 879: What `first_geom_dimensions()` returns in these cases doesn't really matter to me tbh. This is what was most convenient to implement, and it's not an illogical behavior.

What actually matters to me is that we end up returning the top-level dimension downstream. For the non-collection case, it returns that value already, for the collection case, it returns None, but I have logic elsewhere to make it return the top level dim that I want in that case.

petern48:
Nearly done, mainly just waiting for advice on how to test SRID. Might need to debug a bit. But otherwise, this is close. Variable / function renaming suggestions are welcome. Spent less time thinking about naming as things got complicated, I shifted towards just getting everything to work right.

rust/sedona-geometry/src/wkb_header.rs
petern48: Gentle ping in case this slipped under your radar @paleolimbot. Looking [here](https://github.com/georust/wkb/blob/130eb0c2b343bc9299aeafba6d34c2a6e53f3b6a/src/test/ewkb.rs#L4), it looks like WKB crate doesn't support writing EWKB, and instead relies on the `geos` crate for writing (which we can't access from `sedona-geometry`. I was hoping to write a decent number of cases originally, but it's looking like I'll need to hard-code these as fixtures. Let me know if you have any alternative ideas.

rust/sedona-geometry/src/wkb_header.rs
paleolimbot: How about `first_sequence_geometry_type: u32`? (Slightly more in tune with your existing pattern of storing raw data and calculating the value on request)

rust/sedona-geometry/src/wkb_header.rs
```diff
@@ -0,0 +1,728 @@
...
+    // Dimensions of the first nested geometry of a collection or None if empty
+    // For POINT, LINESTRING, POLYGON, returns the dimensions of the geometry
+    first_geom_dimensions: Option<Dimensions>,
+}
+
+impl WkbHeader {
+    /// Creates a new [WkbHeader] from a buffer
+    pub fn try_new(buf: &[u8]) -> Result<Self> {
+        if buf.len() < 5 {
+            return exec_err!("Invalid WKB: buffer too small -> try_new");
```
paleolimbot: We should probably use `SedonaGeometryError` here (this should avoid a datafusion-common and sedona-common dependency here)

rust/sedona-geometry/Cargo.toml
```diff
@@ -34,8 +34,10 @@ serde_json = { workspace = true }
 wkt = { workspace = true }
 
 [dependencies]
+datafusion-common = { workspace = true }
```
paleolimbot: We probably should not depend on datafusion-common or sedona-common here (this is otherwise a pretty lightweight crate).

rust/sedona-geometry/src/wkb_header.rs
```diff
@@ -0,0 +1,728 @@
...
+    /// Returns the top-level dimension of the WKB
+    pub fn dimensions(&self) -> Result<Dimensions> {
+        let dimensions = match self.geometry_type / 1000 {
+            0 => Dimensions::Xy,
+            1 => Dimensions::Xyz,
+            2 => Dimensions::Xym,
+            3 => Dimensions::Xyzm,
+            _ => exec_err!("Unexpected code: {}", self.geometry_type)?,
+        };
+        Ok(dimensions)
```
paleolimbot: This also needs to handle the EWKB Z or M mask. This match exists in a few places and would benefit from its own function.

rust/sedona-geometry/src/wkb_header.rs
```diff
@@ -0,0 +1,728 @@
...
+            .map_err(|e| DataFusionError::External(Box::new(e)))?;
+
+        let mut i = 5;
+        let mut srid = 0;
+        // if EWKB
+        if geometry_type & SRID_FLAG_BIT != 0 {
+            srid = match byte_order {
+                0 => u32::from_be_bytes([buf[5], buf[6], buf[7], buf[8]]),
+                1 => u32::from_le_bytes([buf[5], buf[6], buf[7], buf[8]]),
+                other => return sedona_internal_err!("Unexpected byte order: {other}"),
```
paleolimbot: This pattern is also repeated quite a few times and would benefit from a function

rust/sedona-geometry/src/wkb_header.rs
```diff
@@ -0,0 +1,728 @@
...
+
+    match code / 1000 {
+        0 => Ok(Dimensions::Xy),
+        1 => Ok(Dimensions::Xyz),
+        2 => Ok(Dimensions::Xym),
+        3 => Ok(Dimensions::Xyzm),
+        _ => exec_err!("Unexpected code: {code:?}"),
+    }
+}
+
```
paleolimbot, line 367: There is a lot of code here that is bookkeeping and byte swapping as you walk the buffer and a number of those elements are repeated. The part that makes this complicated is the collection part where you need to parse until the first sequence (otherwise you would just be copying the first few bytes of the buffer).

Many parsers manage abstracting that repetition with something like this:

```rust
struct WkbBuffer {
  buf: &[u8],
  offset: usize,
  remaining: usize,
  last_endian: u8
}

impl WkbBuffer {

   pub fn read_endian(&mut self) -> Result<()> {
        if self.remaining < 1 {
            return Err(...)
        }
        self.last_endian = buf[self.offset];
        self.remaining -= 1;
        self.offset += 1;
        Ok(())
   }

   pub fn read_u32(&mut self) -> Result<u32> {
        if self.remaining < 4 {
            return Err(...)
        }
        let out = match self.last_endian { ... }
        self.remaining -= 4;
        self.offset += 4;
        Ok(out)
   }
}
```

rust/sedona-geometry/src/wkb_header.rs
paleolimbot: It did...thanks for the ping!

I use R's wk package to generate these (long ago I wrote EKWT parsing and EWKB writing as a default, which was not a good idea in retrospect, but has proved very useful for generating test data). You can do this yourself or use these as fixtures (I think these are all the ones you'll need):

``` r
wk::as_wkb("SRID=4326;POINT (1 2)") |> dput()
#> structure(list(as.raw(c(0x01, 0x01, 0x00, 0x00, 0x20, 0xe6, 0x10, 
#> 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0xf0, 0x3f, 0x00, 
#> 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x40))), class = c("wk_wkb", 
#> "wk_vctr"))
wk::as_wkb("SRID=4326;POINT Z (1 2 3)") |> dput()
#> structure(list(as.raw(c(0x01, 0x01, 0x00, 0x00, 0xa0, 0xe6, 0x10, 
#> 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0xf0, 0x3f, 0x00, 
#> 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x40, 0x00, 0x00, 0x00, 0x00, 
#> 0x00, 0x00, 0x08, 0x40))), class = c("wk_wkb", "wk_vctr"))
wk::as_wkb("SRID=4326;POINT M (1 2 4)") |> dput()
#> structure(list(as.raw(c(0x01, 0x01, 0x00, 0x00, 0x60, 0xe6, 0x10, 
#> 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0xf0, 0x3f, 0x00, 
#> 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x40, 0x00, 0x00, 0x00, 0x00, 
#> 0x00, 0x00, 0x10, 0x40))), class = c("wk_wkb", "wk_vctr"))
wk::as_wkb("SRID=4326;POINT ZM (1 2 3 4)") |> dput()
#> structure(list(as.raw(c(0x01, 0x01, 0x00, 0x00, 0xe0, 0xe6, 0x10, 
#> 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0xf0, 0x3f, 0x00, 
#> 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x40, 0x00, 0x00, 0x00, 0x00, 
#> 0x00, 0x00, 0x08, 0x40, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x10, 
#> 0x40))), class = c("wk_wkb", "wk_vctr"))

wk::as_wkb("SRID=4326;GEOMETRYCOLLECTION (POINT (1 2))") |> dput()
#> structure(list(as.raw(c(0x01, 0x07, 0x00, 0x00, 0x20, 0xe6, 0x10, 
#> 0x00, 0x00, 0x01, 0x00, 0x00, 0x00, 0x01, 0x01, 0x00, 0x00, 0x00, 
#> 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0xf0, 0x3f, 0x00, 0x00, 0x00, 
#> 0x00, 0x00, 0x00, 0x00, 0x40))), class = c("wk_wkb", "wk_vctr"
#> ))
wk::as_wkb("SRID=4326;GEOMETRYCOLLECTION (POINT Z (1 2 3))") |> dput()
#> structure(list(as.raw(c(0x01, 0x07, 0x00, 0x00, 0x20, 0xe6, 0x10, 
#> 0x00, 0x00, 0x01, 0x00, 0x00, 0x00, 0x01, 0x01, 0x00, 0x00, 0x80, 
#> 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0xf0, 0x3f, 0x00, 0x00, 0x00, 
#> 0x00, 0x00, 0x00, 0x00, 0x40, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 
#> 0x08, 0x40))), class = c("wk_wkb", "wk_vctr"))
wk::as_wkb("SRID=4326;GEOMETRYCOLLECTION (POINT M (1 2 4))") |> dput()
#> structure(list(as.raw(c(0x01, 0x07, 0x00, 0x00, 0x20, 0xe6, 0x10, 
#> 0x00, 0x00, 0x01, 0x00, 0x00, 0x00, 0x01, 0x01, 0x00, 0x00, 0x40, 
#> 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0xf0, 0x3f, 0x00, 0x00, 0x00, 
#> 0x00, 0x00, 0x00, 0x00, 0x40, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 
#> 0x10, 0x40))), class = c("wk_wkb", "wk_vctr"))
wk::as_wkb("SRID=4326;GEOMETRYCOLLECTION (POINT ZM (1 2 3 4))") |> dput()
#> structure(list(as.raw(c(0x01, 0x07, 0x00, 0x00, 0x20, 0xe6, 0x10, 
#> 0x00, 0x00, 0x01, 0x00, 0x00, 0x00, 0x01, 0x01, 0x00, 0x00, 0xc0, 
#> 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0xf0, 0x3f, 0x00, 0x00, 0x00, 
#> 0x00, 0x00, 0x00, 0x00, 0x40, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 
#> 0x08, 0x40, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x10, 0x40))), class = c("wk_wkb", 
#> "wk_vctr"))
```

rust/sedona-geometry/src/wkb_header.rs
paleolimbot: I think that works!

rust/sedona-geometry/src/wkb_header.rs
paleolimbot: I think None is reasonable here (i.e., there are no sequences)

rust/sedona-geometry/src/wkb_header.rs
```diff
@@ -0,0 +1,728 @@
...
+        let wkb = make_wkb("MULTIPOLYGON M EMPTY");
+        let header = WkbHeader::try_new(&wkb).unwrap();
+        assert_eq!(header.first_geom_dimensions(), None);
+
+        let wkb = make_wkb("GEOMETRYCOLLECTION ZM EMPTY");
+        let header = WkbHeader::try_new(&wkb).unwrap();
+        assert_eq!(header.first_geom_dimensions(), None);
+    }
+}
```
paleolimbot, line 1012: There also needs to be tests here for incomplete buffers. In theory you have logic to check that if there are an insufficient number of bytes available on the buffer you don't call `buf[i]`; however, if your checks are wrong the process will crash.

This is another benefit of using something like the `WkbBuffer` I suggested above (that logic is consolidated and you don't have to test as many cases).

petern48:
I'll try to get back to this over the weekend. Takes more focus

petern48:
@paleolimbot WDYT about adding `geos` as a test dependency to avoid having to hard-code so many fixtures? Reasonable or nah? Thinking about usefulness for future PRs too. We could write EWKB similar to how this PR in WKB used it for testing https://github.com/georust/wkb/pull/46/files#diff-d5cbc1df3ceaaa6b6d928a7d04e566e05caaa85fa2eb79665fcf3d43d01b7a19

If not, then I'll proceed with hard-coding fixtures.

rust/sedona-geometry/src/wkb_header.rs
petern48: great pattern!

rust/sedona-geometry/src/wkb_header.rs
petern48: did not know what you meant here, until I had to debug it 🙃

rust/sedona-geometry/src/wkb_header.rs
petern48: I like `first_geom_dimensions()` better still, since this also applies to nested geometries. e.g `GEOMETRYCOLLECTION(POINT Z EMPTY)`

rust/sedona-geometry/src/wkb_header.rs
petern48: I still kinda like `first_geom_dimensions`. I don't see anything else we'd need the whole nested geometry_type for, so I think it's fine to just directly compute what I need here. If we do come across another case that it'd be useful for later, we can change it then. We can change what we store internally without breaking / removing the public `first_geom_dimensions()` api. It's only a different pattern internally. Externally there's already a `dimensions()` method, so I didn't find this off-pattern personally.

in reality, i got lazy (it's late haha), so if you feel strongly i don't mind doing it.

rust/sedona-geometry/src/wkb_header.rs
```diff
@@ -0,0 +1,1016 @@
...
+            GeometryTypeId::MultiPoint
+        );
+        assert_eq!(header.size(), 2);
+        assert_eq!(header.first_xy(), (1.0, 2.0));
+        assert_eq!(header.dimensions().unwrap(), Dimensions::Xy);
+    }
+
+    // #[test]
+    // fn geometrycollection_with_srid() {
+    //     use sedona_testing::fixtures::*;
```
paleolimbot: Can this be uncommented?

rust/sedona-geometry/src/wkb_header.rs
```diff
@@ -0,0 +1,1016 @@
...
+        assert_eq!(header.first_geom_dimensions(), None);
+    }
+
+    #[test]
+    fn incomplete_buffers() {
+        // Test various incomplete buffer scenarios to ensure proper error handling
+
+        // Empty buffer
+        let result = WkbHeader::try_new(&[]);
+        assert!(result.is_err());
```
paleolimbot, line 900: A possibly more compact strategy for testing these is something like:

```rust
let wkb = make_wkb("POINT EMPTY");
for i in (1..the_number_of_bytes_you_care_about) {
    assert!(WKBHeader::try_new(&wkb[i..wkb.len()]).is_err(), "{i}...end unexpectedly succeeded")
}
```

(In other words, check that you get an error for every invalid byte that is shorter).

rust/sedona-geometry/src/wkb_header.rs
```diff
@@ -0,0 +1,1016 @@
...
+        ]);
+        assert!(result.is_err());
+    }
+
+    #[test]
+    fn incomplete_ewkb_buffers() {
+        // Test incomplete EWKB buffers specifically
+
+        // EWKB with SRID flag but missing the extra 4 bytes for the SRID
+        let result = WkbHeader::try_new(&[
```
paleolimbot: I think the pattern I suggested above would help avoid hard-coding the fixtures again here

rust/sedona-geometry/src/wkb_header.rs
paleolimbot: This is opaque internal storage so we can update if there's a reason to later!

rust/sedona-geometry/src/wkb_header.rs
```diff
@@ -0,0 +1,1016 @@
...
+    /// Returns the geometry type id of the WKB by only parsing the header instead of the entire WKB
+    /// 1 -> Point
+    /// 2 -> LineString
+    /// 3 -> Polygon
+    /// 4 -> MultiPoint
+    /// 5 -> MultiLineString
+    /// 6 -> MultiPolygon
+    /// 7 -> GeometryCollection
+    ///
+    /// Spec: https://libgeos.org/specifications/wkb/
```
paleolimbot:
```suggestion
    /// Returns the [GeometryTypeId] of the WKB by only parsing the header instead of the entire WKB
```

rust/sedona-geometry/src/wkb_header.rs
```diff
@@ -0,0 +1,1016 @@
...
+        let geometry_type_id = GeometryTypeId::try_from_wkb_id(code)?;
+
+        Ok(geometry_type_id)
+    }
+
+    /// Returns the size of the geometry
+    /// Not applicable for a point
+    /// Number of points for a linestring
+    /// Number of rings for a polygon
+    /// Number of geometries for a MULTIPOINT, MULTILINESTRING, MULTIPOLYGON, or GEOMETRYCOLLECTION
```
paleolimbot:
```suggestion
    /// Returns the size of the geometry
    ///
    /// - 1 for Points
    /// - Number of points for a linestring
    /// - Number of rings for a polygon
    /// - Number of geometries for a MULTIPOINT, MULTILINESTRING, MULTIPOLYGON, or GEOMETRYCOLLECTION
```

rust/sedona-geometry/src/wkb_header.rs
```diff
@@ -0,0 +1,1016 @@
...
+            return Err(SedonaGeometryError::Invalid(format!(
+                "Invalid WKB: buffer too small. At offset: {}. Need 8 bytes.",
+                self.offset
+            )));
+        }
+
+        let buf = &self.buf;
+        let off = self.offset;
+        let coord: f64 = match self.last_endian {
+            0 => f64::from_be_bytes([
```
paleolimbot, line 299: This is great for this PR (where we only ever read two ordinates)...if we were to expand this we'd want to move this `match` outside the loop (i.e., so we only check the endian and buffer size once per coordinate sequence)

rust/sedona-geometry/src/wkb_header.rs
```diff
@@ -0,0 +1,1016 @@
...
+
+            // (NaN, NaN) for empty geometries
+            if size == 0 {
+                return Ok((f64::NAN, f64::NAN));
+            }
+
+            // For POLYGON, after the number of rings, the next 4 bytes are the
+            // number of points in the exterior ring. We must skip that count to
+            // land on the first coordinate's x value.
+            if geometry_type_id == GeometryTypeId::Polygon {
```
paleolimbot: I'm not sure if this will be cleaner, but this seems like you might want `match geometry_type_id { GeometryTypeId::LineString => ..., GeometryTypeId::Polygon => ... => {} }`

paleolimbot (commented):
Thank you for bearing with me on this! This will be useful for a lot of cheap/structural inspections of geometries. All the comments except the commented-out test are optional 🙂

> WDYT about adding geos as a test dependency to avoid having to hard-code so many fixtures?

I'd like to avoid geos as a test dependency for now (we can revisit if our fixture list gets out of control). In general being able to run tests without any system dependencies is helpful for contributors.

rust/sedona-functions/src/st_haszm.rs
paleolimbot: Looks great! Having this live inside methods is great...we can adjust the internal approach if there's a good reason to do so in the future.

rust/sedona-geometry/src/wkb_header.rs
petern48: Checking the front slices made more sense to me than checking the back. Note, due to the optimization, we only catch errors up to the first xy coord. Anything missing after that is fair game (or else there wouldn't be an optimization).

rust/sedona-geometry/src/wkb_header.rs
petern48: I actually meant to delete this, which I've now done. It's redundant. There's already GeometryCollection with SRID test cases elsewhere.

rust/sedona-geometry/src/wkb_header.rs
petern48: I see what you mean, but I think it would look / feel weird to pull it out prematurely. The loop doesn't exist yet (I assume you're talking about a loop iterating over the coords, bc I'm not seeing any loop in the existing code). If I'm understanding you right. It wouldn't make a difference now performance-wise wise since we're only reading one xy coord. I'd rather leave it like this for now, and pull it out if / when we read more coords.

rust/sedona-geometry/src/wkb_header.rs
paleolimbot: Sorry, I meant `[0..i]` (with `the_number_of_bytes_you_care_about` as the limit of the optimized reading). If we need any new cases we can consider this (your existing cases are very complete!)

rust/sedona-functions/src/st_haszm.rs
```diff
@@ -108,27 +106,35 @@ impl SedonaScalarKernel for STHasZm {
...
-        }
-        _ => {
-            let geom_dim = item.dim();
-            match dim_index {
-                2 => Ok(Some(matches!(geom_dim, Dimensions::Xyz | Dimensions::Xyzm))),
-                3 => Ok(Some(matches!(geom_dim, Dimensions::Xym | Dimensions::Xyzm))),
-                _ => sedona_internal_err!("unexpected dim_index"),
-            }
-        }
+/// Fast-path inference of geometry type name from raw WKB bytes
```
paleolimbot:
```suggestion
```

petern48:
FYI, our performance compared to duckdb is not much farther ahead as shown in the original benchmark in the PR description. Here's the updated benchmark.

image

Main differences are:
1) This updated benchmark uses the single-threaded Sedona. I'm guessing this is the major cause of the perf drop.
2) We've implemented the full WkbHeader since the original benchmark. We do more than the original implementations did, so it should be slower. Mainly, we support EWKB and SRID checks, and we also grab `first_xy`. `first_xy` shouldn't be much more effort since we already were getting to first_geom for the `first_geom_dimensions` field. I've addressed all main slowdowns I was aware of (e.g. creating nested buffers), but I wonder if there are more.

Issue #138, opened 2025-09-23
rust/sedona-geoparquet: Improve GeoParquet writer support


paleolimbot:
The current GeoParquet writer support only supports GeoParquet 1.0 and does not write a file-level bounding box in the metadata. This was because adding static metadata without peeking into the batches was comparatively easy to do:

https://github.com/apache/sedona-db/blob/aa6c8ea5d10dbfaec2216ecf7f7c35c71a3afa57/rust/sedona-geoparquet/src/writer.rs#L127-L136

The next easiest thing to support is GeoParquet 1.1 bounding box columns, because that is "just" adding a projection using operations we already support (Cartesian bounds). We'd still have no bounding box in the schema metadata but bbox-column aware readers like GDAL, GeoPandas, and SedonaDB would still be able to effectively prune row groups based on those column statistics.

Finally, we can support adding bbox information to the file metadata by reimplementing the `ParquetSink`.

PR #175, opened 2025-10-02
feat(rust/sedona-geoparquet): GeoParquet 1.1 write support

paleolimbot:
For the first half of #138, writer support for GeoParquet 1.1.

Work in progress!


Kontinuation (commented):
LGTM. The only concern is that silently overwriting existing bbox column may lead to confusion.

Issue #159, opened 2025-09-27
feat: expand the Python interface


MrPowers:
SedonaDB currently has SQL and Python interfaces.

Here's how you can get the number of rows in a SedonaDB DataFrame with Python:

```python
df.count()
```

Here's how you can get the number of rows with SQL:

```python
df.to_view("earth_cities")
sd.sql("select count(*) from earth_cities").show()
```

Maybe this isn't the best example because `df.count()` returns an integer and `sd.sql()` returns a DataFrame, but you get the point.

Other functionality can only be expressed via the SQL API.  Here is an example:

```python
nyc_bbox_wkt = (
    "POLYGON((-74.2591 40.4774, -74.2591 40.9176, -73.7004 40.9176, -73.7004 40.4774, -74.2591 40.4774))"
)

sd.sql(f"""
SELECT
    id,
    height,
    num_floors,
    roof_shape,
    ST_Centroid(geometry) as centroid
FROM
    buildings
WHERE
    is_underground = FALSE
    AND height IS NOT NULL
    AND height > 20
    AND ST_Intersects(geometry, ST_SetSRID(ST_GeomFromText('{nyc_bbox_wkt}'), 4326))
LIMIT 5;
""").show()
```

Some libraries, like Spark, have both SQL and Python interfaces.  Here's how this logic would be expressed in PySpark:

```python
result = (buildings
    .where(col("is_underground") == False)
    .where(col("height").isNotNull())
    .where(col("height") > 20)
    .where(ST_Intersects(
        col("geometry"), 
        ST_SetSRID(ST_GeomFromText(lit(nyc_bbox_wkt)), lit(4326))
    ))
    .select(
        "id",
        "height", 
        "num_floors",
        "roof_shape",
        ST_Centroid(col("geometry")).alias("centroid")
    )
    .limit(5)
    .show()
)
```

Polars syntax would be something like this (I think, feel free to chime in here Polars experts):

```python
result = (
    buildings
    .filter(
        (pl.col("is_underground") == False) &
        (pl.col("height").is_not_null()) &
        (pl.col("height") > 20) &
        pl.st_intersects(
            pl.col("geometry"),
            pl.st_set_srid(pl.st_geom_from_text(pl.lit(nyc_bbox_wkt)), pl.lit(4326))
        )
    )
    .select([
        pl.col("id"),
        pl.col("height"),
        pl.col("num_floors"),
        pl.col("roof_shape"),
[...]

PR #176, opened 2025-10-02
Implement DataFrame methods `__len__`, `columns`, and `shape`

p-vdp:
Basic/standard data frame methods like in NumPy and Pandas

### DIFF
--- a/r/sedonadb/src/Makevars.in
+++ b/r/sedonadb/src/Makevars.in
@@ -30,6 +30,11 @@ PKG_LIBS = -L$(LIBDIR) -lsedonadbr @PKG_LIBS@
 
 CARGO_BUILD_ARGS = --lib --profile $(PROFILE) $(FEATURE_FLAGS) --manifest-path=./rust/Cargo.toml --target-dir $(TARGET_DIR)
 
+# "-Wno-pedantic" is to override the "-pedantic" flag passed by pkgbuild, which
+# makes the compilation of aws-lc-sys fail. Hopefully, such C code will be fixed
+# on upstream so that we can remove this.
+CFLAGS_TWEAKED = $(CFLAGS) -Wno-pedantic
+
 all: $(SHLIB) clean_intermediate
 
 $(SHLIB): $(STATLIB)
@@ -40,7 +45,7 @@ $(STATLIB):
 	# therefore is only used if cargo is absent from the user's PATH.
 	export PATH="$(PATH):$(HOME)/.cargo/bin" && \
 	  export CC="$(CC)" && \
-	  export CFLAGS="$(CFLAGS)" && \
+	  export CFLAGS="$(CFLAGS_TWEAKED)" && \
 	  export AWS_LC_SYS_CMAKE_BUILDER=1 && \
 	  export RUSTFLAGS="$(RUSTFLAGS)" && \
 	  if [ "$(TARGET)" != "wasm32-unknown-emscripten" ]; then \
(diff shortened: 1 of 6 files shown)


python/sedonadb/python/sedonadb/dataframe.py
```diff
@@ -151,6 +151,29 @@ def count(self) -> int:
...
+        """Return the column names in the DataFrame"""
+        columns = list()
+        field_index = 0
+        while True:
+            try:
+                columns.append(self._impl.schema().field(field_index).name)
+                field_index += 1
+            except IndexError:
+                break
+
```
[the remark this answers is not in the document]
p-vdp: Yes this is an ugly loop but `self._impl.schema()` isn't iterable. This suggestion won't work without also modifying the return from `PySedonaSchema.schema()`

python/sedonadb/python/sedonadb/dataframe.py
```diff
@@ -151,6 +151,29 @@ def count(self) -> int:
...
+        return columns
```
petern48, line 170:
```suggestion
        return self._impl.columns()
```

I think we should implement this in Rust, adding the below code [here](https://github.com/apache/sedona-db/blob/a15844b4ff9c7e9416b8f1c1c07ad81d908a89cd/python/sedonadb/src/dataframe.rs#L62). To test the updated Rust code locally, you'll want to run `maturin develop` from the `python/sedonadb/` directory. You might need to `pip install maturin`.

```rust
    fn columns(&self) -> Vec<String> {
        self.inner
            .schema()
            .fields()
            .iter()
            .map(|f| f.name().to_string())
            .collect()
    }
```

petern48 (commented):
Just passing by. I'll leave it to Dewey to provide his thoughts. SedonaDB's dataframe interface hasn't been agreed upon yet (issue https://github.com/apache/sedona-db/issues/159). I think adding `__len__` and `columns` is fine since those are pretty universal, but `shape` seems to be very specific to Pandas APIs. Since it's as simple as calling `self.count()` and `len(self.columns)`, maybe we should hold back on it for now?

python/sedonadb/python/sedonadb/dataframe.py
```diff
@@ -151,6 +151,29 @@ def count(self) -> int:
         """
         return self._impl.count()
 
+    def __len__(self) -> int:
+        """Compute the number of rows in the DataFrame"""
+        return self.count()
```
paleolimbot, line 156: I wonder if we should include this or not. Our "DataFrame" hasn't been materialized yet and might well take quite a long time to do so.

As a data point, Ibis implements `__len__()` but returns an error:

```
ExpressionError: Use .count() instead
```

Another data point...duckdb implements `__len__()` and executes the query like this implementation. I'm not sure what pyspark does here.

I would personally lean towards the Ibis approach (return an error forcing a user to explicitly count).

python/sedonadb/python/sedonadb/dataframe.py
paleolimbot: Agreed...I mostly just didn't expose enough from the schema object when I wrote it 😬

python/sedonadb/python/sedonadb/dataframe.py
```diff
@@ -151,6 +151,29 @@ def count(self) -> int:
...
+                field_index += 1
+            except IndexError:
+                break
+
+        return columns
+
+    @property
+    def shape(self) -> tuple[int, int]:
+        """Return the shape of the DataFrame as a tuple of integers corresponding to (rows, columns)"""
+        return self.count(), len(self.columns)
```
paleolimbot, line 175: Neither Ibis nor DuckDB implement a `.shape` accessor. Given that this would also trigger execution, I don't think it's a good idea to include this 😬

jiayuasu:
@p-vdp great work on this PR. Thank you very much! What is your plan on this PR?

p-vdp:
@jiayuasu from the discussion it sounds like __len__() and shape() aren't great ideas at this stage, and columns() is best implemented on the Rust side. Closing this PR and I might work on the columns() implementation in Rust if no one else gets to it first. Thanks!

PR #178, opened 2025-10-03
chore: Refactor `with_aliases` to a more builder-like pattern

petern48:
I wanted to create some trivial issues for creating aliases for functions, but I figured we should fix this before we add some more. The current `new_with_aliases` function was a workaround to not having optional parameters in Rust. It wouldn't work well if wanted to add more optional parameters to the constructor. This PR refactors it to be more builder style so we can easily customize in case we want to add more optional parameters later.

### DIFF
--- a/rust/sedona-expr/src/scalar_udf.rs
+++ b/rust/sedona-expr/src/scalar_udf.rs
@@ -139,20 +139,15 @@ impl SedonaScalarUDF {
         }
     }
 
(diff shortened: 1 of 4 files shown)


paleolimbot (commented):
I see some CI errors but +1 in principle! Thank you!

paleolimbot (approved):
Thank you!

Issue #180, opened 2025-10-04
docs: Generate SQL documentation reference


paleolimbot:
Our existing SQL documentation works but is tricky to keep up-to-date with changes to sedona-functions!

https://github.com/apache/sedona-db/blob/11e299392b4d5bd17281dd4892ece764b8138795/docs/reference/sql.md#L20-L26

This file was initially generated from the documentation we've been keeping inline in the functions themselves. For example:

https://github.com/apache/sedona-db/blob/11e299392b4d5bd17281dd4892ece764b8138795/rust/sedona-functions/src/st_dimension.rs#L41-L50

We can recreate the script we used to generate the initial reference section to ensure we keep this all up to date (and potentially generate reference pages for R and Python when we have function wrappers there). Alternatively we could go in the other direction (have extremely carefully formatted markdown reference that we parse into Rust, R, and Python documentation).

It would also be great to have one page per function to allow these pages room to grow as we accumulate experience and examples with these functions!

paleolimbot:
There are a few potentially useful things we can do here. First, we could formalize the script we use to generate the one file (this would probably be a good time to also split up everything so we have one markdown file per function. I just found this out but apparently it's available from the information schema:

```python
import sedona.db

sd = sedona.db.connect()

sd.sql("""
SELECT DISTINCT routine_name 
FROM information_schema.routines 
WHERE routine_type = 'FUNCTION' AND routine_name LIKE 'st%'
ORDER BY routine_name
""").show()
#> ┌────────────────┐
#> │  routine_name  │
#> │      utf8      │
#> ╞════════════════╡
#> │ st_analyze_agg │
#> ├╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌┤
#> │ st_area        │
#> ├╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌┤
#> │ st_asbinary    │
#> ├╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌┤
```

I think after we do that generation we might want to have the individual markdown files be the canonical documentation for that function. We can make them `.qmd` files and use Quarto to execute them (among other things, this would let us include plot output in the results so we could visualize some of these functions!).

paleolimbot:
(Credit to @petern48 for the SQL to make this happen!)

yutannihilation:
[...]

PR #183, opened 2025-10-05
feat(sql): Implement `ST_Azimuth()`

yutannihilation:
This pull request is my attempt to implement `ST_Azimuth()`.

```
> sd_sql("select ST_Azimuth( ST_Point(25, 45),  ST_Point(75, 100))")
┌──────────────────────────────────────────────────────────────────────────┐
│ st_azimuth(st_point(Int64(25),Int64(45)),st_point(Int64(75),Int64(100))) │
│                                  float64                                 │
╞══════════════════════════════════════════════════════════════════════════╡
│                                                       0.7378150601204649 │
└──────────────────────────────────────────────────────────────────────────┘
Preview of up to 6 row(s)
```

This pull request is mostly for figuring out how to implement a function so that I can contribute more. So, please feel free to close if there's another plan to implement this.

I picked `ST_Azimuth()` because it looks relatively simple. I follow the implementation of Sedona, although I might not understand the code correctly as I'm poor at Java...

https://github.com/apache/sedona/blob/f9069e7dc6682d53335f0e0c6fb4bd444024d3b5/common/src/main/java/org/apache/sedona/common/Functions.java#L202-L209

I think the implementation is straightforward. One thing to note is that, if I understand the Sedona's implementation correctly, it returns `0.0` when the two points are the same. However, PostGIS returns `NULL`. [The document](https://postgis.net/docs/en/ST_Azimuth.html) says:

> NULL if the two points are coincident

### DIFF
--- a/rust/sedona-functions/src/st_azimuth.rs
+++ b/rust/sedona-functions/src/st_azimuth.rs
@@ -40,13 +40,13 @@ pub fn st_azimuth_udf() -> SedonaScalarUDF {
 fn st_azimuth_doc() -> Documentation {
     Documentation::builder(
         DOC_SECTION_OTHER,
-        "Returns the azimuth in radians from geomA to geomB",
+        "Returns the azimuth (a clockwise angle measured from north) in radians from geomA to geomB",
         "ST_Azimuth (A: Geometry, B: Geometry)",
     )
     .with_argument("geomA", "geometry: Start point geometry")
     .with_argument("geomB", "geometry: End point geometry")
     .with_sql_example(
-        "SELECT ST_Azimuth(\n            ST_GeomFromText('POINT (0 0)'),\n            ST_GeomFromText('POINT (1 1)')\n        )",
+        "SELECT ST_Azimuth(\n            ST_Point(0, 0),\n            ST_Point(1, 1)\n        )",
     )
     .build()
 }
--- a/rust/sedona-functions/src/st_azimuth.rs
+++ b/rust/sedona-functions/src/st_azimuth.rs
@@ -1,0 +1,66 @@
+// Licensed to the Apache Software Foundation (ASF) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership.  The ASF licenses this file
+// to you under the Apache License, Version 2.0 (the
+// "License"); you may not use this file except in compliance
+// with the License.  You may obtain a copy of the License at
+//
+//   http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+use arrow_schema::DataType;
+use datafusion_expr::{scalar_doc_sections::DOC_SECTION_OTHER, Documentation, Volatility};
+use sedona_expr::scalar_udf::SedonaScalarUDF;
+use sedona_schema::{datatypes::SedonaType, matchers::ArgMatcher};
+
+/// ST_Azimuth() scalar UDF stub
+///
+/// Stub function for azimuth calculation between two points.
+pub fn st_azimuth_udf() -> SedonaScalarUDF {
+    SedonaScalarUDF::new_stub(
+        "st_azimuth",
+        ArgMatcher::new(
+            vec![
+                ArgMatcher::is_geometry_or_geography(),
+                ArgMatcher::is_geometry_or_geography(),
+            ],
+            SedonaType::Arrow(DataType::Float64),
+        ),
+        Volatility::Immutable,
+        Some(st_azimuth_doc()),
+    )
+}
+
+fn st_azimuth_doc() -> Documentation {
+    Documentation::builder(
+        DOC_SECTION_OTHER,
+        "Returns the azimuth in radians from geomA to geomB",
+        "ST_Azimuth (A: Geometry, B: Geometry)",
+    )
+    .with_argument("geomA", "geometry: Start point geometry")
+    .with_argument("geomB", "geometry: End point geometry")
+    .with_sql_example(
+        "SELECT ST_Azimuth(\n            ST_GeomFromText('POINT (0 0)'),\n            ST_GeomFromText('POINT (1 1)')\n        )",
+    )
+    .build()
+}
+
+#[cfg(test)]
+mod tests {
+    use datafusion_expr::ScalarUDF;
+
+    use super::*;
+
+    #[test]
+    fn udf_metadata() {
+        let udf: ScalarUDF = st_azimuth_udf().into();
+        assert_eq!(udf.name(), "st_azimuth");
+        assert!(udf.documentation().is_some());
+    }
+}
--- a/rust/sedona-geo/src/st_azimuth.rs
+++ b/rust/sedona-geo/src/st_azimuth.rs
@@ -1,0 +1,182 @@
+// Licensed to the Apache Software Foundation (ASF) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership.  The ASF licenses this file
+// to you under the Apache License, Version 2.0 (the
+// "License"); you may not use this file except in compliance
+// with the License.  You may obtain a copy of the License at
+//
+//   http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+use std::sync::Arc;
+
+use arrow_array::builder::Float64Builder;
+use arrow_schema::DataType;
+use datafusion_common::{error::Result, exec_datafusion_err};
+use datafusion_expr::ColumnarValue;
+use geo_traits::{CoordTrait, GeometryTrait, GeometryType, PointTrait};
+use sedona_expr::scalar_udf::{ScalarKernelRef, SedonaScalarKernel};
+use sedona_functions::executor::WkbExecutor;
+use sedona_schema::{datatypes::SedonaType, matchers::ArgMatcher};
+use wkb::reader::Wkb;
+
+/// ST_Azimuth() implementation following PostGIS semantics
+pub fn st_azimuth_impl() -> ScalarKernelRef {
+    Arc::new(STAzimuth {})
+}
+
+#[derive(Debug)]
+struct STAzimuth {}
+
+impl SedonaScalarKernel for STAzimuth {
+    fn return_type(&self, args: &[SedonaType]) -> Result<Option<SedonaType>> {
+        let matcher = ArgMatcher::new(
+            vec![ArgMatcher::is_geometry(), ArgMatcher::is_geometry()],
+            SedonaType::Arrow(DataType::Float64),
+        );
+
+        matcher.match_args(args)
+    }
+
+    fn invoke_batch(
+        &self,
+        arg_types: &[SedonaType],
+        args: &[ColumnarValue],
+    ) -> Result<ColumnarValue> {
+        let executor = WkbExecutor::new(arg_types, args);
+        let mut builder = Float64Builder::with_capacity(executor.num_iterations());
+        executor.execute_wkb_wkb_void(|maybe_start, maybe_end| {
+            match (maybe_start, maybe_end) {
+                (Some(start), Some(end)) => match invoke_scalar(start, end)? {
+                    Some(angle) => builder.append_value(angle),
+                    None => builder.append_null(),
+                },
+                _ => builder.append_null(),
+            }
+
+            Ok(())
+        })?;
+
+        executor.finish(Arc::new(builder.finish()))
+    }
+}
+
+fn invoke_scalar(start: &Wkb, end: &Wkb) -> Result<Option<f64>> {
+    let Some((start_x, start_y)) = point_xy(start)? else {
+        return Ok(None);
+    };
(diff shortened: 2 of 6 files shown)


rust/sedona-functions/src/st_azimuth.rs
```diff
@@ -0,0 +1,66 @@
...
+use arrow_schema::DataType;
+use datafusion_expr::{scalar_doc_sections::DOC_SECTION_OTHER, Documentation, Volatility};
+use sedona_expr::scalar_udf::SedonaScalarUDF;
+use sedona_schema::{datatypes::SedonaType, matchers::ArgMatcher};
+
+/// ST_Azimuth() scalar UDF stub
+///
+/// Stub function for azimuth calculation between two points.
+pub fn st_azimuth_udf() -> SedonaScalarUDF {
+    SedonaScalarUDF::new_stub(
```
petern48: Since `ST_Azimuth` is simple enough to be implemented manually, we should move the entire implementation that you wrote inside of `sedona-geo` to this file in `sedona-functions`. No need for this stub function. `sedona-geo` is really for functions where we need to use the `geo_generic_alg` package to call more complicated algorithms. You can see an example implementation in [st_isempty.rs](https://github.com/apache/sedona-db/blob/main/rust/sedona-functions/src/st_isempty.rs). It should be as simple as copy-pasting over the implementation and modifying this function a bit to use `SedonaScalarUDF::new` instead of `SedonaScalarUDF::new_stub`.

rust/sedona-functions/src/st_azimuth.rs
```diff
@@ -0,0 +1,66 @@
...
+        "st_azimuth",
+        ArgMatcher::new(
+            vec![
+                ArgMatcher::is_geometry_or_geography(),
+                ArgMatcher::is_geometry_or_geography(),
```
petern48:
```suggestion
                ArgMatcher::is_geometry(),
                ArgMatcher::is_geometry(),
```

Then, we can reduce this to geometry only for now, since we don't want to call the geometry implementation on geography objects.

rust/sedona-geo/src/st_azimuth.rs
```diff
@@ -0,0 +1,174 @@
...
+            }
+        }
+        _ => Err(datafusion_common::error::DataFusionError::Execution(
+            "ST_Azimuth expects both arguments to be POINT geometries".into(),
+        )),
+    }
+}
+
+// Note: When the two points are completely coincident, PostGIS's ST_Azimuth()
+//       returns NULL. However, this returns 0.0.
```
petern48: Let's definitely add a test for this case.

But first, we should confirm what the desired behavior is. I suspect this is a just something that was missed in the original Sedona, and maybe we should follow PostGIS behavior here instead. There doesn't seem to be any discussion about this in the [original Sedona PR](https://github.com/apache/sedona/pull/449). @jiayuasu was this difference intentional or maybe just something that was missed? Shall we fix it in Sedona as well?

rust/sedona-functions/src/st_azimuth.rs
```diff
@@ -0,0 +1,66 @@
...
+fn st_azimuth_doc() -> Documentation {
+    Documentation::builder(
+        DOC_SECTION_OTHER,
+        "Returns the azimuth (a clockwise angle measured from north) in radians from geomA to geomB",
+        "ST_Azimuth (A: Geometry, B: Geometry)",
+    )
+    .with_argument("geomA", "geometry: Start point geometry")
+    .with_argument("geomB", "geometry: End point geometry")
+    .with_sql_example(
+        "SELECT ST_Azimuth(\n            ST_Point(0, 0),\n            ST_Point(1, 1)\n        )",
```
petern48:
```suggestion
        "SELECT ST_Azimuth(ST_Point(0, 0), ST_Point(1, 1))",
```
Let's have the example all on one line, like we how it is in the rest of the other functions.

petern48:
Also, add a benchmark here in [native-functions.rs](https://github.com/apache/sedona-db/blob/7f911353335bd2fb51148aadb39a858a5ec12ffd/rust/sedona-functions/benches/native-functions.rs#L29) (they're in alphabetical order).

There's an example of a bench for a function with two geometries as inputs [here](https://github.com/apache/sedona-db/blob/7f911353335bd2fb51148aadb39a858a5ec12ffd/c/sedona-geos/benches/geos-functions.rs#L66-L72)

yutannihilation:
Thanks for reviewing!

yutannihilation:
I think I addressed all the comments. Thanks for the detailed explanation, it really helps!

I switch the implementation to return `NULL` for the coincident points case, but I'll wait for confirmation.

petern48:
Looks great. I just remembered, we should also add a new `test_st_azimuth` test in [test_functions.py](https://github.com/apache/sedona-db/blob/7f911353335bd2fb51148aadb39a858a5ec12ffd/python/sedonadb/tests/functions/test_functions.py#L120). We use this file for comparing the outputs directly with PostGIS results. There's also already a `numeric_epsilon` parameter that you can use to relax the precision (see `test_st_buffer` for an example). We often have a more comprehensive set of tests there since adding test cases is much more concise in Python.

You can follow these directions in the [contributors-guide.md](https://github.com/apache/sedona-db/blob/7f911353335bd2fb51148aadb39a858a5ec12ffd/docs/contributors-guide.md#python) for testing Python. It will also require a running instance of PostGIS, which you can spin up using Docker by following these [instructions](https://github.com/apache/sedona-db/pull/184) I'm adding.

yutannihilation:
Thanks, I'll try it.

I also wonder if I should update this page, but probably there's no tool for this yet, guessing from https://github.com/apache/sedona-db/issues/180?

https://github.com/apache/sedona-db/blob/main/docs/reference/sql.md

rust/sedona-functions/src/st_azimuth.rs
```diff
@@ -0,0 +1,221 @@
...
+            match (start_point.coord(), end_point.coord()) {
+                // If both geometries are non-empty points, calculate the angle
+                (Some(start_coord), Some(end_coord)) => Ok(calc_azimuth(
+                    start_coord.x(),
+                    start_coord.y(),
+                    end_coord.x(),
+                    end_coord.y(),
+                )),
+                // If either of the points is empty, the result is NULL
+                _ => Ok(None),
```
paleolimbot: Just checking: does PostGIS allow a MULITPOINT with a single child here?

rust/sedona-functions/src/st_azimuth.rs
```diff
@@ -0,0 +1,221 @@
...
+            }
+        }
+        _ => Err(datafusion_common::error::DataFusionError::Execution(
+            "ST_Azimuth expects both arguments to be POINT geometries".into(),
+        )),
```
paleolimbot: DataFusion has a helper for this one!

```suggestion
        _ => exec_err!("ST_Azimuth expects both arguments to be POINT geometries"),
```

rust/sedona-geo/src/st_azimuth.rs
paleolimbot: To date, SedonaDB tests against PostGIS for feature parity and we file bugs with Sedona when we notice something is inconsistent.

paleolimbot (commented):
Awesome! (And thanks Peter for the review!)

> Looks great. I just remembered, we should also add a new test_st_azimuth test

These will be similar to the ones for `st_distance()`:

https://github.com/apache/sedona-db/blob/7f911353335bd2fb51148aadb39a858a5ec12ffd/python/sedonadb/tests/functions/test_distance.py#L21-L47

...and could go here:

https://github.com/apache/sedona-db/blob/7f911353335bd2fb51148aadb39a858a5ec12ffd/python/sedonadb/tests/functions/test_functions.py#L120

> I also wonder if I should update this page

You're correct...don't worry about this, as long as the rust function documentation is there it will get updated.

(We'll hopefully do a better job documenting the process of adding a function in the future 😬 )

rust/sedona-functions/src/st_azimuth.rs
yutannihilation: It seems PostGIS also rejects MULTIPOINT.

```sh
postgres=# SELECT ST_Azimuth(ST_Point(0, 0), ST_GeomFromText('MULTIPOINT (1 1)'));
ERROR:  Argument must be POINT geometries
```

yutannihilation:
During adding these tests, I found PostGIS raise errors for these two cases where my implementation returns `NULL`.

The first one is two `NULL`s. I think test should not fail with this error because `NULL::geometry` works. However, casting to geometry doesn't work on SedonaDB yet.

```
postgres=# SELECT ST_Azimuth(NULL, NULL);
ERROR:  function st_azimuth(unknown, unknown) is not unique
LINE 1: SELECT ST_Azimuth(NULL, NULL);
               ^
HINT:  Could not choose a best candidate function. You might need to add explicit type casts.
```

The second one is empty point. I thought this returns `NULL` without errors, but it seems PostGIS doesn't allow empty point. I think we can follow this behavior.

```
postgres=# SELECT ST_Azimuth(ST_Point(0, 0), ST_GeomFromText('POINT EMPTY'));
NOTICE:  lwgeom_api.c [351] called with n=0 and npoints=0
ERROR:  Error extracting point
```

rust/sedona-functions/src/st_azimuth.rs
yutannihilation: Thanks!

yutannihilation:
Thanks for your help. I think I addressed your comments.

> You're correct...don't worry about this, as long as the rust function documentation is there it will get updated.

Good to know!

jiayuasu:
@yutannihilation does it make sense to show off your benchmark result (compared to DuckDB and PostGIS)? 😁

python/sedonadb/tests/functions/test_functions.py
```diff
@@ -119,6 +119,29 @@ def test_st_astext(eng, geom):
     eng.assert_query_result(f"SELECT ST_AsText({geom_or_null(geom)})", expected)
 
 
+@pytest.mark.parametrize("eng", [SedonaDB, PostGIS])
+@pytest.mark.parametrize(
+    ("geom1", "geom2", "expected"),
+    [
+        # TODO: PostGIS fails without explicit ::GEOMETRY type cast, but casting
+        # doesn't work on SedonaDB yet.
+        # (None, None, None),
```
paleolimbot, line 128: Thanks for catching this! We have a few cases where we're not sure exactly what this kind of thing should do (luckily people typing SQL NULLs for this kind of thing is pretty rare 🙂 )

paleolimbot (approved):
Thank you!

yutannihilation:
Thanks for reviewing!

@jiayuasu 
I'm not familiar with benchmarking. Are there any guidance to do this? Do you mean I should have added a case here?

https://github.com/apache/sedona-db/blob/main/benchmarks/test_functions.py

jiayuasu:
@yutannihilation we have a tool in SedonaDB to produce benchmark results like what's shown in this PR: https://github.com/apache/sedona-db/pull/171

@petern48 @paleolimbot can you help @yutannihilation figure out how to run it?

yutannihilation:
Thanks, I guess I can generate the benchmark result if I add a case to `benchmarks/` like that PR, and I can follow this README.

https://github.com/apache/sedona-db/blob/main/benchmarks/README.md

jiayuasu:
I think you are right. Looking forward to it!

yutannihilation:
Sure, I'll try it!

petern48:
Yep, that [README.md](https://github.com/apache/sedona-db/blob/main/benchmarks/README.md) is exactly it. I was reluctant to suggest adding a benchmark because there are still some issues to iron out about its purpose and validity (discussion in [this PR](https://github.com/apache/sedona-db/pull/24).

Regardless, it surely doesn't hurt to add it. Especially since this is a native (manual) implementation, I'd expect to see some appealing numbers compared to DuckDB.

yutannihilation:
Hmm, before adding the case, it seems all benchmark tests are skipped on my local. 

```
❯ pytest test_functions.py::TestBenchFunctions
======================================= test session starts =======================================
platform darwin -- Python 3.12.9, pytest-8.4.2, pluggy-1.6.0
benchmark: 5.1.0 (defaults: timer=time.perf_counter disable_gc=False min_rounds=5 min_time=0.000005 max_time=1.0 calibration_precision=10 warmup=False warmup_iterations=100000)
rootdir: /Users/yutani/repo/sedona-db/benchmarks
plugins: benchmark-5.1.0
collected 57 items

test_functions.py sssssssssssssssssssssssssssssssssssssssssssssssssssssssss                 [100%]

======================================= 57 skipped in 1.13s =======================================
```

Here's what I did. Am I missing some steps? Since I'm not very good at Python, I might make some silly mistake...

```sh
python3 -m venv .venv
source .venv/bin/activate
python -m pip install --upgrade pip
pip install "python/sedonadb[test]" pytest pytest-benchmark

cd benchmarks
pytest test_functions.py::TestBenchFunctions
```

> I'd expect to see some appealing numbers compared to DuckDB.

Since DuckDB's `ST_Azimuth()` is also implemented by me, you can blame me if DuckDB is not slow enough :) Anyway, looking forward to seeing the result!

petern48:
I'm not sure how you got actual `s`'s (skips), but I suspect it's because you tried running the entire file, which would take forever. I submitted a [PR](https://github.com/apache/sedona-db/pull/188) to do this myself, since it requires adding a new points-only table. It wouldn't have been fun for you to figure out and debug since our current setup just hangs for a while (minutes) and eventually outputs a sort of useful error message. Also updated the docs to mention running one bench at a time.

Here's the benchmark screenshot copied over.

image

> Since DuckDB's ST_Azimuth() is also implemented by me, you can blame me if DuckDB is not slow enough :) Anyway, looking forward to seeing the result!

I guess I'll instead thank you that DuckDB is slow enough 😉

yutannihilation:
> I suspect it's because you tried running the entire file

Sorry, it was turned out that I simply forgot to launch PostGIS docker image this time...

Anyway, thanks for adding benchmark! Probably I can do it next time.

> I guess I'll instead thank you that DuckDB is slow enough 😉

😉

petern48:
Got it, thanks for catching that. It helps smooth the process for everyone when we know to document these small details. It's not ideal that the whole thing is skipped when the container isn't running, but I added a note about it to that same PR for now.

PR #184, opened 2025-10-06
docs: Improve Python testing instructions in `contributors-guide.md`

petern48:
Adds instructions for starting Docker to run the Python integration tests with a running PostGIS instance. Plus a few other improvements.

### DIFF
--- a/docs/contributors-guide.md
+++ b/docs/contributors-guide.md
@@ -200,6 +200,7 @@ docker compose up -d
 ```
 
 You can later shut it down with
+
 ```shell
 docker compose down
 ```


docs/contributors-guide.md
```diff
@@ -155,6 +155,12 @@ native component with:
 maturin develop
 ```
 
+If you don't yet have maturin installed, you can install it using pip
+
+```shell
+pip install maturin
+```
+
```
petern48, line 163: When I first started, I was confused why maturin wasn't already included after running `pip install ".[test]"`, making me hesitate whether I installed sedonadb correctly. I'm not sure if it was intentionally left out of the `pyproject.toml`, but I think we should either add it to `pyproject.toml` or if that's not desirable for whatever reason, add these instructions here to pip install it. Which do we prefer?

paleolimbot: Typically build dependencies are not included as test dependencies (e.g., projects using Cython do not typically add Cython to their pyproject.toml's test dependencies that I'm aware of). Because maturin is a "build backend", pip installs it automatically if a build is required. If a build isn't required (e.g., somebody downloaded a nightly wheel), there's no need to have it installed.

All this to say that I think your documentation change is good here!

docs/contributors-guide.md
```diff
@@ -175,6 +181,39 @@ Rust, C, or C++ code can be debugged using the
...
+### Running Python tests
+
+A large number of the Python tests rely on a running PostGIS instance. You can spin one up by using the providied PostGIS docker compose file.
+
+```shell
+docker compose up -d
+```
+
+You can later shut it down with
+```shell
```
paleolimbot, line 204:
```suggestion
You can later shut it down with

```shell
```

paleolimbot (approved):

PR #188, opened 2025-10-07
feat: Add ST_Azimuth benchmark and update benchmarking docs

petern48:
Follow up to https://github.com/apache/sedona-db/pull/183, this PR adds the pytest-benchmarks for ST_Azimuth, which uses a new `points_simple` table. Additionally, it updates the docs to recommend only running one benchmark at a time.

Result of ST_Azimuth benchmark
image


paleolimbot (approved):
Thank you!

PR #197, opened 2025-10-09
fix: Support projection pushdown for RecordBatchReader provider (fixes #186)

### LINKED ISSUE #186: rust/sedona: Projecting a data frame derived from a record batch reder results in a projection error
Reprex:

```python
import pyarrow as pa
import sedona.db

sd = sedona.db.connect()

def batches():
    for i in range(10):
        yield pa.record_batch({"a":["a", "b", "c"], "b": [1, 2, 3]})

reader = pa.RecordBatchReader.from_batches(next(batches()).schema, batches())
sd.create_data_frame(reader).to_view("foofy", overwrite=True)
sd.sql("SELECT b as banana FROM foofy").show()
#> SedonaError: Input field name a does not match with the projection expression b.
```

Kontinuation:
This fixes #186.

The provider ignored the projection indices passed to TableProvider::scan(), so the physical plan schema ([a,b]) did not match the pushed-down logical projection ([b]). This PR implements projection pushdown for RecordBatchReaderProvider and adds a regression test.

### DIFF
--- a/rust/sedona/src/record_batch_reader_provider.rs
+++ b/rust/sedona/src/record_batch_reader_provider.rs
@@ -91,9 +91,9 @@ impl TableProvider for RecordBatchReaderProvider {
         let mut reader_guard = self.reader.lock();
         if let Some(reader) = reader_guard.take() {
             let projection = projection.cloned();
-            Ok(Arc::new(RecordBatchReaderExec::new(
+            Ok(Arc::new(RecordBatchReaderExec::try_new(
                 reader, limit, projection,
-            )))
+            )?))
         } else {
             sedona_internal_err!("Can't scan RecordBatchReader provider more than once")
         }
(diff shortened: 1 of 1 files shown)


rust/sedona/src/record_batch_reader_provider.rs
```diff
@@ -383,6 +419,49 @@ mod test {
         }
     }
 
+    #[tokio::test]
+    async fn test_projection_pushdown() {
+        use arrow_array::{RecordBatch, RecordBatchIterator};
+        use arrow_schema::{DataType, Field, Schema};
+        use datafusion::prelude::col;
+        use datafusion::prelude::SessionContext;
```
paleolimbot: Should these be module-level `use`s?

Kontinuation: Moved these to module level.

PR #208, opened 2025-10-10
fix(r/sedonadb): Fix build on Windows for path restriction

### LINKED ISSUE #189: r/sedonadb: R CMD INSTALL fails on Windows due to path length restriction
We've tried to solve this a few times, but we still need to find a solution here:

- https://github.com/apache/sedona-db/pull/161
- https://github.com/apache/sedona-db/pull/160
- https://github.com/apache/sedona-db/pull/152

Without this, we can't use R-Universe to publish binaries.

Jeroen kindly attempted to override TMPDIR on the R-Universe windows runner to make the path shorter, but apparently R CMD INSTALL overrides the temp directory 😮‍💨 .

paleolimbot:
This PR allows a developer to override the target path for Cargo, which is useful on Windows where most paths a user might build this under fail (e.g., I run into the path length restriction for CMake in my usual home directory/documents folder). This allows overriding that (for local development it one can put `SEDONADB_TARGET_DIR=C:/sedonadb_tmp` in their .Rprofile), and automatically overrides it on GitHub actions (where R Universe Windows packages are built).

If successful, closes #189.


yutannihilation:
Great! Confirmed this works on my Windows. This makes local development much easier.

> (for local development it one can put `SEDONADB_TARGET_DIR=C:/sedonadb_tmp` in their .Rprofile),

Btw, this made me wonder if this requires us to tweak the `rust-cache` setting as well, but it seems it currently doesn't cache R's `target` dir anyway? So, I think this can be addressed in a separate pull request.

https://github.com/apache/sedona-db/blob/f79d8219fdfad0356cedd56689ff380d4c540ed8/.github/workflows/r.yml#L83-L86

yutannihilation:
Congratulations, now the build on R-universe succeeds! 🎉 

https://github.com/r-universe/apache/actions/runs/18423385925/job/52501661384

I installed it and confirmed the README example works fine.

Issue #200, opened 2025-10-09
feat: Implement aliases for existing functions


petern48:
Adding aliases is trivial. Like a 5 line PR. Here's an example commit of adding an alias (two actually) https://github.com/apache/sedona-db/commit/5c75a5c678f8e9e7f7191fcf02bcd91cfdd53180. The syntax changed a bit in this recent [PR](https://github.com/apache/sedona-db/pull/178), but it's still very simple/concise. Took a brief skim through https://github.com/apache/sedona-db/issues/174, probably missed some

Here's a list of aliases we need to add. I will update as I come across more, let me know if you find more to add to this list.

- [x] `ST_GeometryFromText` alias for ST_GeomFromText function
- [ ] `ST_NPoints` alias for ST_NumPoints
- [ ] `ST_NRings` alias for ST_NumRings*

`*`-  not yet implemented at time of writing

ayushjariyal:
@petern48, I want to solve this issue under Hacktoberfest. Could you please assign it to me?

petern48:
@ayushjariyal Sure, I've created a new issue to assign you to add the `ST_GeometryFromText` for ST_GeomFromText, since this issue is more of an umbrella for multiple PRs. In the future, no need to wait to be assigned. Just a quick mention that you're working on something is enough. Or if the PR is short and quick, you can just submit one without waiting.

tomatotomata:
I checked the current `rust/sedona-functions` implementation before starting and found that `ST_NPoints` already exposes `ST_NumPoints` as an alias, with a dedicated alias test. That means this item is already covered on `main`, so I’m not opening a duplicate PR. The issue checklist may just need updating for this entry.

paleolimbot:
Thanks for noting!

I just checked and `ST_NumRings()` isn't a PostGIS function, so I think we can close this one.

Issue #212, opened 2025-10-11
chore: Add python integrations tests for aliases


petern48:
We didn't have python integration tests when we originally added aliases. We should add some sort of tests to ensure python can call them.

At the moment, we test only one version. In this case below, we actually test the alias and not the original function `ST_GeomFromWKT` anywhere.

https://github.com/apache/sedona-db/blob/73e38fc9701deaacb8c4deba15272efb6939ddc8/python/sedonadb/tests/functions/test_functions.py#L411

Options I see:

1) Parametrize the existing test functions so all aliases are called for the entire test suite of that function

```python
@pytest.mark.parametrize(func_name, ["ST_GeomFromText", "ST_GeomFromWKT"])
...
def test_st_geomfromwkt()
```

2) Add some separate tests, just make sure the aliases are working

```python
def test_st_geomfromtext_alias()
      # < a single query > 
```

paleolimbot:
I think (2) works (just make sure it's plugged in), as long as we're sure it's supposed to be just be an alias.

PR #213, opened 2025-10-11
feat: Add ST_GeometryFromText alias for ST_GeomFromWKT

### LINKED ISSUE #211: feat: Implement ST_GeometryFromText alias for ST_GeomFromWKT function
See parent #200

Here's where the original function is.

https://github.com/apache/sedona-db/blob/73e38fc9701deaacb8c4deba15272efb6939ddc8/rust/sedona-functions/src/st_geomfromwkt.rs#L42-L52

ayushjariyal:
Fixes #211 as a part of #200 

Adds `ST_GeometryFromText` as an additional alias for `ST_GeomFromWKT` to support all function names from the original Apache Sedona documentation.


petern48:
We'll need to add Python integration tests for this later, but haven't finalized how yet (https://github.com/apache/sedona-db/issues/212). For now, this is great!

PR #216, opened 2025-10-12
feat(rust/sedona-expr): Use Covers filter for ST_Equals for more Geoparquet pruning

petern48:
Previously, we've been using the Intersects filter for ST_Equals. This PR updates the SpatialFilter to use the Covers predicate for ST_Equals instead, which is more restrictive and should prune more data.

Continues to pass existing integration [test](https://github.com/petern48/sedona-db/blob/6fec5dcca25dfdf34636e800fc1ee6476f3c90b3/python/sedonadb/tests/io/test_parquet.py#L70-L74) that checks that we return the same non-empty result as the equivalent geopandas query.

### DIFF
--- a/python/sedonadb/src/reader.rs
+++ b/python/sedonadb/src/reader.rs
@@ -43,12 +43,21 @@ impl Iterator for PySedonaStreamReader {
     type Item = std::result::Result<RecordBatch, ArrowError>;
 
     fn next(&mut self) -> Option<Self::Item> {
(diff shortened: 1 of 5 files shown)


petern48:
I noticed this when implementing https://github.com/apache/sedona-db/pull/217. There will be a merge conflict that I will have to resolve. Doesn't matter to me which one is merged first.

PR #217, opened 2025-10-12
feat(rust/sedona-expr): Implement SpatialFilter for overlaps and crosses for geoparquet pruning

petern48:
Implement SpatialFilter for overlaps and crosses for geoparquet pruning


python/sedonadb/tests/io/test_parquet.py
petern48, line 207: A few lines down, there's this check to make sure these actually yield non-empty results

PR #219, opened 2025-10-14
dev: ST_IsClosed

### LINKED ISSUE #215: feat: Implement native ST_IsClosed function
The `geo_traits` library's LineString and MultiLineString structs both have an `is_closed()` method ([here](https://github.com/georust/geo/blob/8940db79aa6aa4ec8820d6328c68a2ae08ac8fdc/geo-types/src/geometry/line_string.rs#L361) and [here](https://github.com/georust/geo/blob/8940db79aa6aa4ec8820d6328c68a2ae08ac8fdc/geo-types/src/geometry/multi_line_string.rs#L72)). We can use this to implement ST_IsClosed in SedonaDB's `rust/native-functions` directory.

You can see this commit (https://github.com/apache/sedona-db/commit/3d8d50e49333f1e9e16fc45f6b32220c318d970f) that adds ST_IsEmpty to guide this implementation. You'll likely want to use a similar match statement to handle linestrings and multipolygons separately from other geometry types.

Behavior that we want: 
https://postgis.net/docs/ST_IsClosed.html

parent #174

Abeeujah:
closes #215

### DIFF
--- a/rust/sedona-functions/src/st_isclosed.rs
+++ b/rust/sedona-functions/src/st_isclosed.rs
@@ -21,6 +21,7 @@ use arrow_array::builder::BooleanBuilder;
 use arrow_schema::DataType;
 use datafusion_common::error::Result;
 use datafusion_expr::{scalar_doc_sections::DOC_SECTION_OTHER, Documentation, Volatility};
+use geo_traits::GeometryCollectionTrait;
 use geo_traits::{
     to_geo::{ToGeoLineString, ToGeoMultiLineString},
     GeometryTrait,
@@ -93,8 +94,6 @@ fn invoke_scalar(item: &Wkb) -> Result<bool> {
 }
 
 fn is_geometry_closed(item: &Wkb) -> Result<bool> {
-    use geo_traits::GeometryCollectionTrait;
-
     if is_geometry_empty(&item).map_err(|e| {
         datafusion_common::error::DataFusionError::Execution(format!(
             "Failed to check if geometry is empty: {e}"


Abeeujah:
Not completed, Because, I'm not particularly sure how to implement the match for `Multipoint`, `Multipolygon`, `GeometryCollection`, maybe my implementation for `Point` might be wrong as well

petern48:
> Not completed, Because, I'm not particularly sure how to implement the match for `Multipoint`, `Multipolygon`, `GeometryCollection`, maybe my implementation for `Point` might be wrong as well

Nice work so far! I think the next thing that would be helpful is for you to figure out what PostGIS outputs for these less obvious cases (e.g point, multipoint). Two ways I can think of

1. Just start writing some [python integration tests](https://github.com/apache/sedona-db/blob/f1571555ade1907ba9ffe2ea88d3ca4a92f1d937/python/sedonadb/tests/functions/test_functions.py#L4) like you did before and run the tests to see what PostGIS outputs.
2. Run queries directly against your running PostGIS container. After starting your container with `docker compose up -d`, you can get into the PostGIS shell with the following command and run SQL queries directly in it. Don't be afraid to ping, if you need help.

```bash
# I'm like 90% confident the container name of 'sedona-db-postgis-1' should be the same for you
```

rust/sedona-functions/src/st_isclosed.rs
```diff
@@ -0,0 +1,114 @@
...
+        geo_traits::GeometryType::LineString(linestring) => {
+            Ok(linestring.to_line_string().is_closed())
+        }
+        geo_traits::GeometryType::Polygon(polygon) => {
+            Ok(polygon.to_polygon().exterior().is_closed())
+        }
+        geo_traits::GeometryType::MultiLineString(multilinestring) => {
+            Ok(multilinestring.to_multi_line_string().is_closed())
+        }
+        _ => internal_err!("Invalid geometry type"),
```
petern48:
```suggestion
        _ => sedona_internal_err!("Invalid geometry type"),
```

minor nit while i'm here (I would've probably forgot later), we should instead use `sedona_internal_err!` here. You can see this [PR](https://github.com/apache/sedona-db/pull/2) for why we introduced that. tldr, it's error msg is misleading. Here's the import path `use sedona_common::sedona_internal_err;`

Abeeujah:
Hi @petern48 @paleolimbot I'm getting this weird error on the python end, Actually, It was the wild card in the pattern matching being triggered

python/sedonadb/.gitignore
```diff
@@ -1,88 +1,2 @@
-# Licensed to the Apache Software Foundation (ASF) under one
```
petern48, line 1: Let's undo this change to `.gitignore`

rust/sedona-functions/src/st_isclosed.rs
```diff
@@ -0,0 +1,175 @@
...
+            .try_fold(true, |acc, item| {
+                is_geometry_closed(item).map(|is_closed| acc && is_closed)
+            }),
+        geo_traits::GeometryType::Point(_)
+        | geo_traits::GeometryType::Line(_)
+        | geo_traits::GeometryType::MultiPoint(_)
+        | geo_traits::GeometryType::Polygon(_)
+        | geo_traits::GeometryType::MultiPolygon(_)
+        | geo_traits::GeometryType::Rect(_)
+        | geo_traits::GeometryType::Triangle(_) => Ok(true),
```
petern48:
```suggestion
        geo_traits::GeometryType::Point(_)
        | geo_traits::GeometryType::MultiPoint(_)
        | geo_traits::GeometryType::Polygon(_)
        | geo_traits::GeometryType::MultiPolygon(_) => Ok(true),
        _ => sedona_internal_err!("Invalid geometry type"),
```

SedonaDB doesn't support `Line`, `Rect`, or `Triangle` geometry types, so I think it's better to omit them and error on the default case (this is what we do in other functions). If we somehow got those geom types, I'd rather not hide that bug.

rust/sedona-functions/src/st_isclosed.rs
```diff
@@ -0,0 +1,175 @@
...
+        executor.finish(Arc::new(builder.finish()))
+    }
+}
+
+fn invoke_scalar(item: &Wkb) -> Result<bool> {
+    is_geometry_closed(item).map_err(|e| {
+        datafusion_common::error::DataFusionError::Execution(format!(
+            "Failed to check if geometry is closed: {e}"
+        ))
+    })
```
petern48:
```suggestion
    is_geometry_closed(item)
```

`.map_err()` is usually only needed to convert the Error type of the result from one error type to `DataFusionError` (e.g the `is_geometry_empty` returns `Result<_, SedonaGeometryError>`, so we need to call it for that below) In this case, both `is_geometry_closed` and `invoke_scalar` return the DataFusionError, so there's no need to convert

python/sedonadb/tests/functions/test_functions.py
```diff
@@ -573,6 +573,23 @@ def test_st_isempty(eng, geom, expected):
...
+@pytest.mark.parametrize("eng", [SedonaDB, PostGIS])
+@pytest.mark.parametrize(
+    ("geom", "expected"),
+    [
+        (None, None),
+        ("LINESTRING(0 0, 1 1)", False),
+        ("LINESTRING(0 0, 0 1, 1 1, 0 0)", True),
+        ("MULTILINESTRING((0 0, 0 1, 1 1, 0 0),(0 0, 1 1))", False),
+        ("POINT(0 0)", True),
+        ("MULTIPOINT((0 0), (1 1))", True),
```
petern48, line 585:
```suggestion
        ("MULTIPOINT((0 0), (1 1))", True),
        ("POLYGON ((0 0, 1 0, 1 1, 0 1, 0 0))", True),
        ("GEOMETRYCOLLECTION (LINESTRING(0 0, 0 1, 1 1, 0 0))", True),
        ("GEOMETRYCOLLECTION (LINESTRING(0 0, 0 1, 1 1, 0 0), LINESTRING(0 0, 1 1))", False),
        ("POINT EMPTY", False),
        ("LINESTRING EMPTY", False),
        ("POLYGON EMPTY", False),
        ("MULTIPOINT EMPTY", False),
        ("MULTILINESTRING EMPTY", False),
        ("MULTIPOLYGON EMPTY", False),
        ("GEOMETRYCOLLECTION EMPTY", False),
        ("GEOMETRYCOLLECTION (LINESTRING EMPTY)", False),
```

Since we're implementing more of this logic manually, we should test more comprehensively. Here are some more cases I suggest we add.

- One non-empty geometry for each of the 7 types
- GeometryCollection with a closed linestring and an open linestring (False)
- Empty geometry for each of the 7 types
- Non-empty GeometryCollection containing only empties

rust/sedona-functions/src/st_isclosed.rs
```diff
@@ -0,0 +1,175 @@
...
+}
+
+fn is_geometry_closed<G: GeometryTrait<T = f64>>(item: G) -> Result<bool> {
```
petern48:
```suggestion
fn is_geometry_closed<G: GeometryTrait<T = f64>>(item: G) -> Result<bool> {
    if is_geometry_empty(&item).map_err(|e| {
        datafusion_common::error::DataFusionError::Execution(format!(
            "Failed to check if geometry is empty: {e}"
        ))
    })? {
        return Ok(false);
    }
```

I just queried PostGIS locally, it seems like it returns `false` for any empty geometry, so we should use the Sedona's `is_geometry_empty` function. The import for it is below. Note: I originally placed this inside of the final match arm below, but it turns out `geo`'s `is_closed()` methods return `true` for empty linestrings (which is not what we want), so we need to have this check above the match statement. Otherwise, we'll return the wrong result for empty linestring, multilinestring, and geomcol.

```rust
use sedona_geometry::is_empty::is_geometry_empty;
```

rust/sedona-functions/src/st_isclosed.rs
```diff
@@ -0,0 +1,175 @@
...
+        let result = tester.invoke_wkb_scalar(None).unwrap();
+        tester.assert_scalar_result_equals(result, ScalarValue::Null);
+
+        let input_wkt = vec![
+            None,
+            Some("LINESTRING(0 0, 1 1)"),
+            Some("LINESTRING(0 0, 0 1, 1 1, 0 0)"),
+            Some("MULTILINESTRING((0 0, 0 1, 1 1, 0 0),(0 0, 1 1))"),
+            Some("POINT(0 0)"),
+            Some("MULTIPOINT((0 0), (1 1))"),
```
petern48, line 163:
```suggestion
            Some("MULTIPOINT((0 0), (1 1))"),
            Some("LINESTRING EMPTY"),
            Some("POINT EMPTY"),
```

optional nit: The Rust tests don't need to be nearly as comprehensive as the Python ones, but I think it would be nice to at least have one test for the empty cases, since it's a non-neglible part of the implementation.

petern48 (commented):
We've got to be a little more careful for this one since we're implementing some manual logic, but it's looking great. Nearly there.

rust/sedona-functions/src/st_isclosed.rs
```diff
@@ -0,0 +1,182 @@
...
+
+        executor.finish(Arc::new(builder.finish()))
+    }
+}
+
+fn invoke_scalar(item: &Wkb) -> Result<bool> {
+    is_geometry_closed(item)
+}
+
+fn is_geometry_closed<G: GeometryTrait<T = f64>>(item: G) -> Result<bool> {
```
petern48:
```suggestion
fn is_geometry_closed(item: &Wkb) -> Result<bool> {
```
I think this should fix the trait recursion problem.

rust/sedona-functions/src/st_isclosed.rs
```diff
@@ -0,0 +1,182 @@
...
+    match item.as_type() {
+        geo_traits::GeometryType::LineString(linestring) => {
+            Ok(linestring.to_line_string().is_closed())
+        }
+        geo_traits::GeometryType::MultiLineString(multilinestring) => {
+            Ok(multilinestring.to_multi_line_string().is_closed())
+        }
+        geo_traits::GeometryType::GeometryCollection(geometry_collection) => geometry_collection
+            .to_geometry_collection()
+            .iter()
```
petern48:
```suggestion
            .geometries()
```
optional nit: If you want, you can add use `geo_traits::GeometryCollectionTrait` to use this code instead. This change would've fixed one of the arms on its own, but I didn't find a way to avoid the `to_line_string()` and `to_multi_line_string()` calls (I didn't try for too long), so we need the fix I mentioned in the above comment.

edit: Actually, I prefer we do this because it's cheaper. `to_geometry_collection` creates a new vector to push everything to, while `.geometries()` iterates through the existing object.

petern48 (commented):
@Abeeujah I think there's something else that failed before you pushed that last commit. 

```
 error[E0275]: overflow evaluating the requirement `G: GeometryTrait`
    |
    = help: consider increasing the recursion limit by adding a `#![recursion_limit = "256"]` attribute to your crate (`sedona_functions`)
    = note: required for `<... as GeometryCollectionTrait>::GeometryType<'_>` to implement `ToGeoGeometry<f64>`
    = note: the full name for the type has been written to '/home/runner/work/sedona-db/sedona-db/target/release/deps/sedona_functions-440740f413ee93de.long-type-15992899749200598173.txt'
    = note: consider using `--verbose` to print the full type name to the console
```
from this action: https://github.com/apache/sedona-db/actions/runs/18559935149/job/52919073154?pr=219

I took a look, and (though I wasn't able to reproduce it locally, I think my suggestion below will fix it. Fingers crossed 🤞

rust/sedona-functions/src/st_isclosed.rs
```diff
@@ -92,7 +92,9 @@ fn invoke_scalar(item: &Wkb) -> Result<bool> {
     is_geometry_closed(item)
 }
 
-fn is_geometry_closed<G: GeometryTrait<T = f64>>(item: G) -> Result<bool> {
+fn is_geometry_closed(item: &Wkb) -> Result<bool> {
+    use geo_traits::GeometryCollectionTrait;
```
petern48: optional nit: We can move this import up with the rest

edit: actually it's whatever to me tbh (maybe dewey will disagree). i know it can be annoying to have to go through another CI run. Let's at least see if CI passes on this run.

petern48 (approved):
Nice! Just a minor nit.

rust/sedona-functions/src/st_isclosed.rs
Abeeujah: :sweat_smile:  It's fine on my end, Just felt like keeping the import there instead of atop with the other imports, but yeah, I'm fine with moving it up.

paleolimbot: This should be at the top of the file (we do have some lingering accidental precedent where we forgot to do this in other places 😬 )

paleolimbot (approved):
Thank you! (And thank you to Peter for the review!)

If you don't mind taking care of the import that would be great but it's not strictly harmful either 🙂

Abeeujah:
@paleolimbot I think the other placements might best be addressed by a follow up PR, to reduce the footprint of this PR, please do let me know what you think.

petern48:
Great PR! Thanks for handling all the feedback.

PR #222, opened 2025-10-14
chore(ci): Bump caches

### LINKED ISSUE #209: ci: R CI jobs may not caching be the right target directory
As noted in https://github.com/apache/sedona-db/pull/208#issuecomment-3392715555 , the cache is probably not getting the correct target directory for the R build. Windows R seems to take the longest of all the CI jobs at the moment and after https://github.com/apache/sedona-db/pull/208 has a special target directory to work around the path length restriction.

paleolimbot:
After the deforking PRs, CI jobs are mostly restoring useless caches and are taking a while.

Also, R's jobs are caching the wrong target directory and caching the correct one will possibly make those jobs faster. This doesn't seem to be effective on R/Windows although I think the configuration changes are still a step in the right direction.

Closes #209.

### DIFF
--- a/.github/workflows/r.yml
+++ b/.github/workflows/r.yml
@@ -67,7 +67,8 @@ jobs:
           rustup default stable
 
       # Set the default toolchain here so that rust-cache saves/restores
-      # the correct target. (The R package will use this target regardless)
+      # the correct target. (The R package will use this target regardless,
+      # this just in theory helps the cache)
       - name: Use stable Rust (windows)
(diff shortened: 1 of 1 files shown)


jiayuasu:
If I understand the problem correctly, you can drop the github cache yourself since you are a committer: https://github.com/apache/sedona-db/actions/caches

Github CLI allows you to drop all caches in 1 command: https://cli.github.com/manual/gh_cache_delete

PR #226, opened 2025-10-16
feat(python/sedonadb): Implement `DataFrame.columns`

p-vdp:
Also add `polars` to test dependencies in `pyproject.toml`

Following up with #176 and the addition suggested by [petern48](https://github.com/petern48)

### DIFF
--- a/python/sedonadb/pyproject.toml
+++ b/python/sedonadb/pyproject.toml
@@ -37,6 +37,7 @@ test = [
     "geoarrow-pyarrow",
     "geopandas",
     "pandas",
+    "polars",
     "pytest",
 ]
 geopandas = [
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -612,7 +612,7 @@ version = "0.32.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "a2b715a6010afb9e457ca2b7c9d2b9c344baa8baed7b38dc476034c171b32575"
 dependencies = [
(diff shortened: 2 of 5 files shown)


python/sedonadb/pyproject.toml
```diff
@@ -37,6 +37,7 @@ test = [
     "geoarrow-pyarrow",
     "geopandas",
     "pandas",
+    "polars",
```
petern48, line 40: Sorry, why is Polars being added as a dependency? The PR isn't introducing any changes that would require it, and SedonaDB is not built on Polars, if that was what you were thinking. Aside from that, `columns()` implementation looks good.

petern48: Unless there's still a reason to keep polars, how about we push a new commit to remove the dependency changes, and then we can look into merging the PR?

p-vdp: It's an optional dependency under the `test` group, it's part of the `test_dataframe` suite: https://github.com/apache/sedona-db/blob/df8f7205a0a99bd75195933084ae4216781f067c/python/sedonadb/tests/test_dataframe.py#L89

Seemed like an oversight that it wasn't there. Happy to take it out if it's a problem.

petern48: That makes a lot more sense. I just wasn't aware of that. Thanks for the context. 😅

paleolimbot (approved):
Thank you!

PR #227, opened 2025-10-17
chore: Sort imports to prefer module-level imports to function-level imports

Abeeujah:
Thi  Pull Request aims to address the issue where imports are duplicated, not properly sorted, Omitting only `rust/sedona-geo-generic-alg/src/algorithm/line_measures/metric_spaces/euclidean/utils.rs` which the import choices at the tail tests seems intentional.

### DIFF
--- a/rust/sedona-functions/src/barrier.rs
+++ b/rust/sedona-functions/src/barrier.rs
@@ -24,7 +24,6 @@ use datafusion_expr::{
 };
 use sedona_expr::scalar_udf::{SedonaScalarKernel, SedonaScalarUDF};
 use sedona_schema::datatypes::SedonaType;
-use ScalarValue::*;
 
 /// barrier() scalar UDF implementation
 ///
@@ -333,6 +332,7 @@ impl Barrier {
 
     /// Compare two scalar values using the given operator
     fn compare_values(left: &ScalarValue, op: &str, right: &ScalarValue) -> Result<bool> {
+        use ScalarValue::*;
         match (left, right) {
             (Int64(Some(l)), Int64(Some(r))) => match op {
                 "=" | "==" => Ok(l == r),
--- a/rust/sedona-geo-generic-alg/src/algorithm/area.rs
+++ b/rust/sedona-geo-generic-alg/src/algorithm/area.rs
@@ -19,7 +19,6 @@
 //! Ported (and contains copied code) from `geo::algorithm::area`:
 //! <https://github.com/georust/geo/blob/5d667f844716a3d0a17aa60bc0a58528cb5808c3/geo/src/algorithm/area.rs>.
 //! Original code is dual-licensed under Apache-2.0 or MIT; used here under Apache-2.0.
-use crate::MapCoords;
 use sedona_geo_traits_ext::*;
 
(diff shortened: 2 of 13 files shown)


rust/sedona-geo-generic-alg/src/algorithm/area.rs
paleolimbot, line 1: The files in sedona-geo-generic-alg are mostly copied from another source. I think moving around the imports may make it difficult to track updates to the upstream code.

rust/sedona-functions/src/barrier.rs
```diff
@@ -332,8 +333,6 @@ impl Barrier {
 
     /// Compare two scalar values using the given operator
     fn compare_values(left: &ScalarValue, op: &str, right: &ScalarValue) -> Result<bool> {
-        use ScalarValue::*;
-
```
paleolimbot, line 336: This should probably stay here...the wildcard imports from enums are often best kept local because they may bring a lot of common names into scope!

rust/sedona-geoparquet/src/metadata.rs
```diff
@@ -72,7 +74,6 @@ impl Default for GeoParquetColumnEncoding {
 
 impl Display for GeoParquetColumnEncoding {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
-        use GeoParquetColumnEncoding::*;
```
paleolimbot, line 75: This one should also probably stay here (it is a wildcard import of an enum that brings a lot of names into scope)

rust/sedona/src/object_storage.rs
```diff
@@ -819,8 +821,6 @@ mod tests {
     #[cfg(not(target_os = "windows"))]
     #[test]
     fn test_substitute_tilde() {
-        use std::env;
-        use std::path::MAIN_SEPARATOR;
```
paleolimbot, line 823: These should probably stay here: function level imports under a `[cfg(...)]` help prevent clippy errors about unused imports when compiling on a different platform.

paleolimbot (commented):
There are a few of these that should stay put but in general this is a helpful change...thank you for taking this on!

rust/sedona-geo-generic-alg/src/algorithm/area.rs
Abeeujah: Thought as much, I've reverted all changes, and restored it to its initial state

rust/sedona-functions/src/barrier.rs
Abeeujah: This has been restored as well

rust/sedona-geoparquet/src/metadata.rs
Abeeujah: Same here, I've restored this

rust/sedona/src/object_storage.rs
Abeeujah: Yeah, I've restored

Abeeujah:
My Bad, I didn't crosscheck `rust/sedona-geoparquet/src/metadata.rs`

PR #228, opened 2025-10-18
feat(python/sedonadb): Implement Python UDFs

paleolimbot:
This PR implements user-defined functions implemented in Python that can be called from DataFusion. This was supposed to be a Friday afternoon project that rather spiraled out of control.

I think there are a few design choices here and I've marked the interface as experimental since I think it will take some experimenting to feel out where the holes are. There is extensive documentation but the gist of it is:

```python
import pyarrow as pa
import sedona.db
from sedonadb import udf
import shapely
import geoarrow.pyarrow as ga

sd = sedona.db.connect()


@udf.arrow_udf(ga.wkb(), [udf.GEOMETRY, udf.NUMERIC])
def shapely_udf(geom, distance):
    geom_wkb = pa.array(geom.storage.to_array())
    distance = pa.array(distance.to_array())
    geom = shapely.from_wkb(geom_wkb)
    result_shapely = shapely.buffer(geom, distance)
    return pa.array(shapely.to_wkb(result_shapely))


sd.register_udf(shapely_udf)
sd.sql("SELECT shapely_udf(ST_Point(0, 0), 2.0) as col").show()
#> ┌──────────────────────────────────────────────────────────────────────────────────────────────────┐
#> │                                                col                                               │
#> │                                             geometry                                             │
#> ╞══════════════════════════════════════════════════════════════════════════════════════════════════╡
#> │ POLYGON((2 0,1.9615705608064609 -0.3901806440322565,1.8477590650225735 -0.7653668647301796,1.66… │
#> └──────────────────────────────────────────────────────────────────────────────────────────────────┘
```

...unfortunately pyproj's Python API is not thread safe which means that GeoPandas crashes when doing CRS related tasks; however, we can still use shapely.

### DIFF
--- a/python/sedonadb/python/sedonadb/udf.py
+++ b/python/sedonadb/python/sedonadb/udf.py
@@ -274,6 +274,10 @@ class ScalarUdfImpl:
         volatility: Literal["immutable", "stable", "volatile"] = "immutable",
         name: Optional[str] = None,
     ):
+        # If the input_types are None, the return_type must be callable when passed
+        # to the internals. In the Python API we allow a data type as the return type
+        # to the argument easier to understand, which means we may have to wrap
+        # it in a callable here.
         if input_types is None and not callable(return_type):
 
             def return_type_impl(*args, **kwargs):


python/sedonadb/python/sedonadb/udf.py
```diff
@@ -0,0 +1,284 @@
...
+            def func_wrapper(args, return_type, num_rows):
+                return func(*args, return_type=return_type)
+        elif "num_rows" in kwarg_names:
+
+            def func_wrapper(args, return_type, num_rows):
+                return func(*args, num_rows=num_rows)
+        else:
+
+            def func_wrapper(args, return_type, num_rows):
+                return func(*args)
```
Kontinuation, line 230: The decorated UDF function can accept kw-only args `num_rows` and `return_type`, and we pass special parameters for these kw args. However, this behavior is not documented. Is it because the API is still experimental, or is using these kw-only args not recommended?

python/sedonadb/python/sedonadb/udf.py
```diff
@@ -0,0 +1,284 @@
...
+        name: Optional[str] = None,
+    ):
+        if input_types is None and not callable(return_type):
+
+            def return_type_impl(*args, **kwargs):
+                return return_type
+
+            self._return_type = return_type_impl
+        else:
+            self._return_type = return_type
```
Kontinuation, line 288: Do we support UDFs with flexible args (`input_types=None`) while the return type is fixed? I believe that we need to wrap `return_type` as a function iff `return_type` is not callable, regardless of whether `input_types` is None or not.

python/sedonadb/src/udf.rs
```diff
@@ -0,0 +1,341 @@
...
+            let result_sedona_type = SedonaType::from_storage_field(&result_field)?;
+
+            if return_type != &result_sedona_type {
+                let return_type_storage = SedonaType::Arrow(return_type.storage_type().clone());
+                if return_type_storage != result_sedona_type {
+                    return Err(PySedonaError::SedonaPython(format!(
+                        "Expected result of user-defined function to return array of type {return_type} or its storage but got {result_sedona_type}"
+                    )));
+                }
+            }
```
Kontinuation: We allow Python UDFs to return arrays with different sedona_type but the same storage type. Will this trigger some runtime type-checking failure when the result of the Python UDF is fed into some other functions?

Kontinuation (approved):
Generally looks good.

python/sedonadb/python/sedonadb/udf.py
paleolimbot: I added a comment here...I *think* the current logic is correct and tested. Basically, we want to use a `ArgMatcher` internally to avoid calling into Python more than necessary, but we can't use an `ArgMatcher` if `input_types` is a function. On the Rust side we require the return type to be callable if the input types are None, but it felt nicer to allow fixed data type in Python for any case.

python/sedonadb/python/sedonadb/udf.py
paleolimbot: Thanks! (I added an example). I initially added them so that UDFs could return something with the correct CRS, but then just let the UDFs optionally return storage after realizing the pyproj crashes if you touch it from a UDF.

Issue #224, opened 2025-10-15
feat: Implement functions using geos library


petern48:
Many functions involve complex algorithms, so we use a library like as [geos](https://github.com/georust/geos) to implement the logic for us. Most of these should be pretty straight forward and easy to implement, since it's just calling a library function. These implementations should go in the `c/sedona-geos/` directory ("c/" instead of "rust/" because the library is using bindings for a C library). Most of the geos geometry methods are shown in this [geometry.rs](https://github.com/georust/geos/blob/master/src/geometry.rs) file (e.g `is_valid()` can be used to implement Sedona's ST_IsValid). This issue lists functions that we can leverage the `geos` library to implement the logic. All of these functions have a method in `geos` that we can use to implement these functions directly. Each function should be implemented in its own PR. Feel free to ping if you need more guidance.

- [x] ST_IsValid
- [x] ST_IsValidReason
- [x] ST_IsRing
- [x] ST_IsSimple
- [x] ST_SimplifyPreserveTopology
- [x] ST_UnaryUnion
- [x] ST_Boundary
- [x] ST_ConvexHull
- [x] ST_ConcaveHull (actually geos method not released yet)
- [x] ST_Simplify (before starting, see https://github.com/apache/sedona-db/issues/244)
- [x] ST_Snap
- [x] ST_MakeValid
- [x] ST_MinimumClearance
- [x] ST_MinimumClearanceLine
- [x] ~~ST_GeometryN~~ (done faster w/ existing `geo-traits` methods)
- [x] ST_Polygonize

Update 11/8
- [x] ST_NumInteriorRing (alias for below)
- [x] ST_NumInteriorRings
- [x] ST_NumRings (I think this would be num interior rings + 1)
- [x] ST_NumPoints
- [x] ~~ST_InteriorRingN~~ (geo-traits instead)
- [x] ST_BuildArea
- [x] ST_DelaunayTriangles
- [x] ST_Normalize
- [ ] ST_Project
- [x] ST_PointOnSurface (point_on_surface)
- [x] ST_ExteriorRing (get_exterior_ring)
- [x] ST_Reverse

... (there's more, but I'll keep the list short for now)

Here is an example commit: https://github.com/apache/sedona-db/commit/6ba858bb6ac1c28b6cb40a6e4eb4607bfcff8a67. It simply calls the `is_valid` method and returns the result as WKB.

[...]

PR #229, opened 2025-10-18
Implement ST_IsValid using geos library

Abeeujah:
This pull request references the discussions had in #224 Starting from the first, UnImplemented in the list, If I read correctly, there are integration tests available already, so I didn't write any integration tests, but if otherwise, please mention and I would implement the integration tests.

Edit: Going through the issue discussions again, I found out I misinterpreted the Integration tests system being set up, for the actual tests being available, I just pushed the integration tests for this function

### DIFF
--- a/python/sedonadb/pyproject.toml
+++ b/python/sedonadb/pyproject.toml
@@ -50,8 +50,3 @@ geopandas = [
 features = ["pyo3/extension-module"]
 module-name = "sedonadb._lib"
 python-source = "python"
-
-[dependency-groups]
-dev = [
-    "ruff>=0.14.1",
-]
--- a/python/sedonadb/tests/functions/test_functions.py
+++ b/python/sedonadb/tests/functions/test_functions.py
@@ -219,17 +219,37 @@ def test_st_centroid(eng, geom, expected):
         ("POINT (0 0)", True),
         ("POINT EMPTY", True),
         ("LINESTRING (0 0, 1 1)", True),
+        ("LINESTRING (0 0, 1 1, 1 0, 0 1)", True),
+        (
+            "MULTILINESTRING ((0 0, 1 1), (0 0, 1 1, 1 0, 0 1))",
+            True,
+        ),
         ("LINESTRING EMPTY", True),
+        # Invalid LineStrings
+        ("LINESTRING (0 0, 0 0)", False),  # Degenerate - both points identical
+        ("LINESTRING (0 0, 0 0, 0 0)", False),  # All points identical
+        # Invalid MultiLineStrings
+        ("MULTILINESTRING ((0 0, 0 0), (1 1, 2 2))", False),  # Degenerate component
+        (
+            "MULTILINESTRING ((0 0, 0 0), (1 1, 1 1))",
+            False,
+        ),  # Multiple degenerate components
         ("POLYGON ((0 0, 0 1, 1 1, 1 0, 0 0))", True),
         ("POLYGON EMPTY", True),
-        ("POLYGON ((0 0, 1 1, 0 1, 1 0, 0 0))", False),
         ("POLYGON ((0 0, 10 0, 10 10, 0 10, 0 0), (1 1, 1 2, 2 2, 2 1, 1 1))", True),
+        # Invalid Polygons
+        # Self-intersecting polygon (bowtie)
+        ("POLYGON ((0 0, 1 1, 0 1, 1 0, 0 0))", False),
+        # Inner ring shares an edge with the outer ring
         ("POLYGON ((0 0, 10 0, 10 10, 0 10, 0 0), (0 0, 0 1, 1 1, 1 0, 0 0))", False),
+        # Self-intersecting polygon (figure-8)
         ("Polygon((0 0, 2 0, 1 1, 2 2, 0 2, 1 1, 0 0))", False),
+        # Inner ring touches the outer ring at a point
         (
             "POLYGON ((0 0, 10 0, 10 10, 0 10, 0 0), (1 10, 1 9, 2 9, 2 10, 1 10))",
             False,
         ),
+        # Overlapping polygons in a multipolygon
         (
             "MULTIPOLYGON (((0 0, 2 0, 2 2, 0 2, 0 0)), ((1 1, 3 1, 3 3, 1 3, 1 1)))",
             False,
@@ -238,6 +258,7 @@ def test_st_centroid(eng, geom, expected):
             "MULTIPOLYGON (((0 0, 1 0, 1 1, 0 1, 0 0)), ((2 2, 3 2, 3 3, 2 3, 2 2)))",
             True,
         ),
+        # Geometry collection with an invalid polygon
         ("GEOMETRYCOLLECTION (POLYGON ((0 0, 1 1, 0 1, 1 0, 0 0)))", False),
         ("GEOMETRYCOLLECTION (POLYGON ((0 0, 0 1, 1 1, 1 0, 0 0)))", True),
     ],


c/sedona-geos/src/st_isvalid.rs
```diff
@@ -0,0 +1,123 @@
...
+        tester.assert_scalar_result_equals(result, false);
+
+        let result = tester.invoke_scalar(ScalarValue::Null).unwrap();
+        assert!(result.is_null());
+
+        let input_wkt = vec![
+            None,
+            Some("POLYGON ((0 0, 0 1, 1 1, 1 0, 0 0))"),
+            Some("POLYGON ((0 0, 1 1, 0 1, 1 0, 0 0))"),
+            Some("LINESTRING (0 0, 1 1)"),
```
petern48:
```suggestion
            Some("LINESTRING (0 0, 1 1)"),
            Some("Polygon((0 0, 2 0, 1 1, 2 2, 0 2, 1 1, 0 0))"),
```

Let's add self intersecting ring case.

petern48, line 112:
```suggestion
            Some("LINESTRING (0 0, 1 1)"),
            Some("Polygon((0 0, 2 0, 1 1, 2 2, 0 2, 1 1, 0 0))"),
```

Let's add self self-intersecting ring case. To here and the Python integration tests.

c/sedona-geos/benches/geos-functions.rs
```diff
@@ -295,6 +295,20 @@ fn criterion_benchmark(c: &mut Criterion) {
         "st_overlaps",
         ArrayScalar(Polygon(10), Polygon(500)),
     );
+    benchmark::scalar(
+        c,
+        &f,
+        "geos",
+        "st_isvalid",
+        ArrayScalar(Polygon(10), Polygon(10)),
```
petern48:
```suggestion
        Polygon(10),
```

These are like arguments to the function, since is_valid only takes one geometry, the value here is just a single Polygon. You can see `st_centroid` above as an example. This should fix the CI failure.

c/sedona-geos/benches/geos-functions.rs
```diff
@@ -295,6 +295,20 @@ fn criterion_benchmark(c: &mut Criterion) {
...
+    );
+    benchmark::scalar(
+        c,
+        &f,
+        "geos",
+        "st_isvalid",
+        ArrayScalar(Polygon(10), Polygon(500)),
```
petern48:
```suggestion
        Polygon(500),
```

c/sedona-geos/src/st_isvalid.rs
Abeeujah: Yeah, I've added this test case

c/sedona-geos/benches/geos-functions.rs
Abeeujah: I guess I should pay more attention, Is there like a way I could catch subtle errors like these before pushing? My current checks include, tests passing and pre-commit hook not failing

c/sedona-geos/benches/geos-functions.rs
Abeeujah: Thank you!

python/sedonadb/pyproject.toml
```diff
@@ -50,3 +50,8 @@ geopandas = [
 features = ["pyo3/extension-module"]
 module-name = "sedonadb._lib"
 python-source = "python"
+
+[dependency-groups]
+dev = [
+    "ruff>=0.14.1",
+]
```
paleolimbot: Can you remove this change from this PR? I agree it's useful for most users to have ruff installed, but it doesn't necessarily have to be in your environment (`pre-commit run -a` will run it for you)

python/sedonadb/tests/functions/test_functions.py
```diff
@@ -211,6 +211,42 @@ def test_st_centroid(eng, geom, expected):
     eng.assert_query_result(f"SELECT ST_Centroid({geom_or_null(geom)})", expected)
 
 
+@pytest.mark.parametrize("eng", [SedonaDB, PostGIS])
+@pytest.mark.parametrize(
+    ("geom", "expected"),
+    [
+        (None, None),
```
paleolimbot, line 218: In this particular case it might be nice to add a comment to the parameter combinations where this returns False to indicate why the example is invalid (I think they are mostly self-intersecting rings).

python/sedonadb/tests/functions/test_functions.py
```diff
@@ -211,6 +211,42 @@ def test_st_centroid(eng, geom, expected):
...
+        ("POINT (0 0)", True),
+        ("POINT EMPTY", True),
+        ("LINESTRING (0 0, 1 1)", True),
+        ("LINESTRING EMPTY", True),
```
paleolimbot, line 227:
```suggestion
        ("LINESTRING (0 0, 1 1)", True),
        ("LINESTRING (0 0, 1 1, 1 0, 0 1)", False),
        ("LINESTRING EMPTY", True),
```

...I believe this will also be flagged as invalid because the linestring intersects itself (but check!). If this does indeed return false, adding a multilinestring version would also be a good idea.

Abeeujah: This is considered valid, cause unlike polygons, self-intersection does not invalidate a LineString in most standard spatial models (like OGC SFA or ISO 19107) which I believe the geos library based their implementation upon

A **LINESTRING** is generally considered valid if it meets two main conditions, and the geometry you provided adheres to both:

1. A LineString must have at least two distinct points, or be empty. `LINESTRING (0 0, 1 1, 1 0, 0 1)` has four distinct points.
2. No two consecutive points can be identical (no "spikes"). The points in your LineString are distinct and non-consecutive.

python/sedonadb/tests/functions/test_functions.py
Abeeujah: Yeah, Done that now

python/sedonadb/pyproject.toml
Abeeujah: Oh, yeah, I've removed it, Currently using the command you suggested!

c/sedona-geos/benches/geos-functions.rs
petern48: Yep! For benches, you can try running you can run `cargo bench -- <name pattern>` while being in the correct directory.

For this case, you need to be in `cd c/sedona-geos/` and run `cargo bench -- st_isvalid`. This is documented [here](https://github.com/apache/sedona-db/blob/3d756644caccd33d7a2bba41e646924caddd3203/docs/contributors-guide.md#running-benchmarks) in the contributors guide (though tbh, they could be better since it doesn't explain how to interpret them).

Personally, I don't run these myself locally unless I'm actually working on optimizing performance. They generally work as long as you have the right number of args and arg types (I've never seen them fail for other reasons). It also takes over a minute just for one function.

python/sedonadb/tests/functions/test_functions.py
```diff
@@ -211,6 +211,63 @@ def test_st_centroid(eng, geom, expected):
...
+        # Inner ring shares an edge with the outer ring
+        ("POLYGON ((0 0, 10 0, 10 10, 0 10, 0 0), (0 0, 0 1, 1 1, 1 0, 0 0))", False),
+        # Self-intersecting polygon (figure-8)
+        ("Polygon((0 0, 2 0, 1 1, 2 2, 0 2, 1 1, 0 0))", False),
+        # Inner ring touches the outer ring at a point
+        (
+            "POLYGON ((0 0, 10 0, 10 10, 0 10, 0 0), (1 10, 1 9, 2 9, 2 10, 1 10))",
+            False,
+        ),
+        # Overlapping polygons in a multipolygon
```
petern48, line 252: Super helpful comments! I didn't know about these invalid geometry cases until now.

petern48 (approved):
I learned a lot more about what is and isn't considered an invalid geometry in this PR. Another great PR. Thanks!

PR #231, opened 2025-10-19
feat(c/sedona-geos): Implement ST_IsRing

joonaspessi:
Hello!

Thought that implementing something relatively simple from your [first issue proposal](https://github.com/apache/sedona-db/issues/224) would be a good way to understand how `sedona-db` works. Thanks for creating this project.

This PR implements `ST_IsRing` with GEOS. 

I did some research on how other spatial libraries handle this function when a non-LineString geometry is passed:

  | System | Non-LineString Behavior |
  |--------|------------------------|
  | **DuckDB spatial extensions** | Returns `false` |
  | **Apache Sedona** | Returns `null` |
  | **PostGIS** | Throws error  |
  | **This PR** | Returns `false`|

  I chose to follow the original Apache Sedona rather than PostGIS because it matches the existing Sedona ecosystem.

### DIFF
--- a/c/sedona-geos/src/st_isring.rs
+++ b/c/sedona-geos/src/st_isring.rs
@@ -1,0 +1,179 @@
+// Licensed to the Apache Software Foundation (ASF) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership.  The ASF licenses this file
+// to you under the Apache License, Version 2.0 (the
+// "License"); you may not use this file except in compliance
+// with the License.  You may obtain a copy of the License at
+//
+//   http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+use std::sync::Arc;
+
+use arrow_array::builder::BooleanBuilder;
+use arrow_schema::DataType;
+use datafusion_common::{error::Result, DataFusionError};
+use datafusion_expr::ColumnarValue;
+use geos::{Geom, GeometryTypes};
+use sedona_expr::scalar_udf::{ScalarKernelRef, SedonaScalarKernel};
+use sedona_schema::{datatypes::SedonaType, matchers::ArgMatcher};
+
+use crate::executor::GeosExecutor;
+
+/// ST_IsRing() implementation using the geos crate
+pub fn st_is_ring_impl() -> ScalarKernelRef {
+    Arc::new(STIsRing {})
+}
+
+#[derive(Debug)]
+struct STIsRing {}
+
+impl SedonaScalarKernel for STIsRing {
+    fn return_type(&self, args: &[SedonaType]) -> Result<Option<SedonaType>> {
+        let matcher = ArgMatcher::new(
+            vec![ArgMatcher::is_geometry()],
+            SedonaType::Arrow(DataType::Boolean),
+        );
+
+        matcher.match_args(args)
+    }
+
+    fn invoke_batch(
+        &self,
+        arg_types: &[SedonaType],
+        args: &[ColumnarValue],
+    ) -> Result<ColumnarValue> {
+        let executor = GeosExecutor::new(arg_types, args);
+        let mut builder = BooleanBuilder::with_capacity(executor.num_iterations());
+
+        executor.execute_wkb_void(|maybe_wkb| {
+            match maybe_wkb {
+                Some(wkb) => {
+                    builder.append_value(invoke_scalar(&wkb)?);
+                }
+                _ => builder.append_null(),
+            }
+            Ok(())
+        })?;
+
+        executor.finish(Arc::new(builder.finish()))
+    }
+}
+
+fn invoke_scalar(geos_geom: &geos::Geometry) -> Result<bool> {
+    // Check if geometry is empty - (PostGIS compatibility)
+    let is_empty = geos_geom.is_empty().map_err(|e| {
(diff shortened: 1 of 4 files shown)


c/sedona-geos/src/st_isring.rs
```diff
@@ -0,0 +1,178 @@
...
+    let is_closed = geos_geom.is_closed().map_err(|e| {
+        DataFusionError::Execution(format!("Failed to check if geometry is closed: {e}"))
+    })?;
+
+    // Check if the LineString is simple (no self-intersections)
+    let is_simple = geos_geom.is_simple().map_err(|e| {
+        DataFusionError::Execution(format!("Failed to check if geometry is simple: {e}"))
+    })?;
+
+    Ok(is_closed && is_simple)
```
petern48:
```suggestion
    Ok(geos_geom.is_ring().map_err(|e| {
        DataFusionError::Execution(format!("Failed to check if geometry is a ring: {e}"))
    })?)
```

We can actually simplify this code to simply call `geos`'s `.is_ring()` method [here](https://github.com/georust/geos/blob/47afbad2483e489911ddb456417808340e9342c3/src/geometry.rs#L252-L264). Seems to pass everything when I tried this locally.

python/sedonadb/tests/functions/test_functions.py
```diff
@@ -604,6 +604,52 @@ def test_st_isclosed(eng, geom, expected):
...
+        ("LINESTRING(0 0, 1 0, 1 1, 0 0)", True),
+        ("LINESTRING(0 0, 2 2, 1 2, 0 0)", True),
+        # Closed but self-intersecting - bowtie shape
+        ("LINESTRING(0 0, 0 1, 1 0, 1 1, 0 0)", False),
+        # Not closed
+        ("LINESTRING(0 0, 1 1)", False),
+        ("LINESTRING(2 0, 2 2, 3 3)", False),
+        ("LINESTRING(0 0, 2 2)", False),
+        # Empty LineString
+        ("LINESTRING EMPTY", False),
```
petern48, line 680:
```suggestion
        ("LINESTRING EMPTY", False),
        # Collections of linestrings that would have been considered rings
        ("MULTILINESTRING((0 0, 0 1, 1 1, 1 0, 0 0))", False),
        ("GEOMETRYCOLLECTION(LINESTRING(0 0, 0 1, 1 1, 1 0, 0 0))", False),
```

Let's add these two cases. Returning `False` is the natural behavior for the `geos`'s `is_ring()` method, which agrees with DuckDB

petern48 (commented):
Welcome! I left a few comments, but it looks good so far. Just need to agree on which behavior to follow.

petern48:
I personally agree with the proposed behavior. [PostGIS's implementation](https://postgis.net/docs/doxygen/3.5/d4/d06/postgis_2lwgeom__geos_8c_aee8112fca59e61cf6c5938ba03096865.html) is odd imo. For non-linestrings, it returns `ERROR:  ST_IsRing() should only be called on a linear feature` if it's non-empty, but returns False if it's empty. The thing I don't like about this is that if you have **any** non-empty non-linestring geometry in your column, your query will just fail.

```sql
CREATE TABLE geom (geom GEOMETRY);
INSERT INTO geom VALUES (ST_GeomFromText('LINESTRING EMPTY')), (ST_Point(3, 4));
SELECT st_isring(geom) FROM geom;
-- ERROR:  ST_IsRing() should only be called on a linear feature
```

DuckDB returns `False` instead of the error, which agrees with the natural behavior of the `is_ring()` function. Sedona actually returns `NULL` for any non-linestring geometry (empty or not).

WDYT @paleolimbot?

paleolimbot (commented):
Thank you!

> DuckDB returns False instead of the error, which agrees with the natural behavior of the is_ring() function. Sedona actually returns NULL for any non-linestring geometry (empty or not).

I think we should follow PostGIS here and file a bug in Sedona-spark to discuss if that decision was intentional. If a user is trying to check the ring status of something that isn't even the correct geometry type there is likely an error in the query or an error in the data.

In the next DataFusion release we'll have the ability to pipe options into scalar functions and so we can do things like enable a Sedona-spark compatibility mode. Here I think PostGIS compatibility is the right choice because it's the easiest to test and review (and in general has more intentional corner case behaviour.

jiayuasu:
Fine by me. PostGIS compatibility is more important

python/sedonadb/tests/functions/test_functions.py
joonaspessi: Makes sense. Added these under `test_st_isring_non_linestring` as PostGis is throwing error on these.

joonaspessi:
Thanks for the comments. I think the PostGis compatibility is a good choice to follow. Updated the PR to follow that approach. While doing it, also noticed that for empty geometries, PostGis will return `false`.

python/sedonadb/tests/functions/test_functions.py
```diff
@@ -661,6 +661,66 @@ def test_st_isclosed(eng, geom, expected):
...
+    """
+    eng = eng.create_or_skip()
+    with pytest.raises(Exception) as exc_info:
+        eng.assert_query_result(f"SELECT ST_IsRing(ST_GeomFromText('{geom}'))", None)
+
+    # Verify error message contains the expected text
+    error_msg = str(exc_info.value).lower()
+    assert (
+        "linear" in error_msg or "linestring" in error_msg
+    ), f"Expected error about linear feature, got: {exc_info.value}"
```
paleolimbot: I think this will work:

```suggestion
    with pytest.raises(Exception, match="linear|linestring") as exc_info:
        eng.assert_query_result(f"SELECT ST_IsRing(ST_GeomFromText('{geom}'))", None)
```

joonaspessi: Good idea! Updated to cleanly utilize pytest match pattern. Pre-commit passes locally

paleolimbot (approved):
Thank you!

PR #233, opened 2025-10-19
perf: Implement (2x) faster ST_Buffer kernel using `geo`

### LINKED ISSUE #232: perf: Investigate if `geo`'s `ST_Buffer` implementation is faster than current `geos`
`ST_Buffer` is an expensive function (and IMO commonly used enough) that it's worth investigating whether the `geo` implementation is faster early on

petern48:
closes #232

Performance is about 2x faster:

sedona-geos
```
geos-st_buffer-ArrayScalar(Polygon(10), Float64(1.0, 10.0))
                        time:   [1.2389 s 1.2621 s 1.2871 s]

geos-st_buffer-ArrayScalar(Polygon(50), Float64(1.0, 10.0))
                        time:   [2.3664 s 2.3742 s 2.3825 s]
```

sedona-geo (this PR)
```
geo-st_buffer-ArrayScalar(Polygon(10), Float64(1.0, 10.0))
                        time:   [597.68 ms 613.95 ms 632.17 ms]

geo-st_buffer-ArrayScalar(Polygon(50), Float64(1.0, 10.0))
                        time:   [1.1246 s 1.1285 s 1.1327 s]
```

### DIFF
--- a/python/sedonadb/tests/functions/test_functions.py
+++ b/python/sedonadb/tests/functions/test_functions.py
@@ -269,7 +269,7 @@ def test_st_buffer(eng, geom, dist, expected_area):
 
 @pytest.mark.parametrize("eng", [SedonaDB, PostGIS])
 @pytest.mark.parametrize(
-    ("geom", "dist", "expected"),
+    ("geom", "expected"),
     [
         ("POINT EMPTY", "POLYGON EMPTY"),
         ("LINESTRING EMPTY", "POLYGON EMPTY"),


rust/sedona-geo/src/st_buffer.rs
```diff
@@ -0,0 +1,224 @@
...
+        );
+        let buffer_result = tester
+            .invoke_wkb_array_scalar(input_wkt, input_dist)
+            .unwrap();
+        let envelope_result = envelope_tester.invoke_array(buffer_result).unwrap();
+        assert_array_equal(&envelope_result, &expected_envelope);
+    }
+
+    #[test]
+    fn test_empty_geometry() {
```
petern48, line 188: The test suite here is the same as the `geos` buffer test suite plus this new function, which I also copied over to `geos` st_buffer to be sure it works the same.

rust/sedona-geo/src/st_buffer.rs
```diff
@@ -0,0 +1,223 @@
...
+}
+
+fn invoke_scalar(
+    wkb: &Wkb,
+    params: BufferStyle<f64>,
+    writer: &mut impl std::io::Write,
+) -> Result<()> {
+    // PostGIS returns POLYGON EMPTY for all empty geometries
+    let is_empty = is_geometry_empty(wkb).map_err(|e| DataFusionError::External(Box::new(e)))?;
+    if is_empty {
```
petern48, line 107: I was running into an error here with `POINT EMPTY` since `geo` apparently doesn't support it.
https://github.com/apache/sedona-db/blob/3d756644caccd33d7a2bba41e646924caddd3203/rust/sedona-geo/src/to_geo.rs#L62-L66

My current workaround is to use `WKBExecutor` here instead of `GeoTypesExecutor` and use our native `is_geometry_empty` check.

Wondering if there's a better way we can handle empty points in our [item_to_geometry() method](https://github.com/apache/sedona-db/blob/3d756644caccd33d7a2bba41e646924caddd3203/rust/sedona-geo/src/to_geo.rs#L58C8-L58C24). The docstring for `geo`'s [try_to_point()](https://github.com/georust/geo/blob/7263968eb533028fd9be81eaeb9ca455e8a2aee6/geo-traits/src/to_geo.rs#L45-L48) function we are in there says returning `None` represents an empty point. Though returning `None` is not a safe option, so I can't think of anything better than this workaround atm.

petern48:
Are there any convenient existing ways to benchmark this against the `geos` implementation? I'm not aware of any. Not yet sure if this is faster or not

petern48:
Looks like this is indeed faster. I've updated the PR description with the benchmark results.

python/sedonadb/tests/functions/test_functions.py
```diff
@@ -262,7 +262,30 @@ def test_st_buffer(eng, geom, dist, expected_area):
...
+        # geos passes with 1e-9, but geo needs it as high as 1e-3
+        numeric_epsilon=1e-3,
+    )
+
+
+@pytest.mark.parametrize("eng", [SedonaDB, PostGIS])
+@pytest.mark.parametrize(
+    ("geom", "dist", "expected"),
+    [
+        ("POINT EMPTY", 2.0, "POLYGON EMPTY"),
```
paleolimbot: Given the distance doesn't matter here it might be easier for future us to mentally parse if the parameters were just `geom` and `expected` (or even just `geom` since the result is identical)

rust/sedona-geo/src/st_buffer.rs
paleolimbot: > Wondering if there's a better way we can handle empty points in item_geometry()

We could have it return something like `enum ItemToGeometryResult { Unsupported(Wkb), Supported(Geometry))`? (No need to do that here unless you're excited about it).

I suppose it might be more complicated because each algorithm might have different considerations.

petern48: still not the cleanest solution, but i do like the idea. though, i prefer to keep it simple for now. we can revisit this as we get a better idea of what we need to handle as we encounter this issue more for other functions.

paleolimbot (approved):

PR #234, opened 2025-10-21
Implement ST_UnaryUnion

Abeeujah:

### DIFF
--- a/python/sedonadb/python/sedonadb/testing.py
+++ b/python/sedonadb/python/sedonadb/testing.py
@@ -269,74 +269,11 @@ class DBEngine:
             result_pandas = self.result_to_pandas(result)
             pandas.testing.assert_frame_equal(result_pandas, expected, **kwargs)
         elif isinstance(expected, list):
-            result_table = self.result_to_table(result)
-            geom_cols_schema = _geometry_columns(result_table.schema)
-
-            if geom_cols_schema:
-                import shapely
-
-                result_tuples = self.result_to_tuples(result, **kwargs)
-                if len(result_tuples) != len(expected):
-                    raise AssertionError(
-                        f"Expected {len(expected)} rows, got {len(result_tuples)}"
-                    )
-
-                is_geom_col = [
-                    name in geom_cols_schema for name in result_table.schema.names
-                ]
-
-                for i, (row_result, row_expected) in enumerate(
-                    zip(result_tuples, expected)
-                ):
-                    if len(row_result) != len(row_expected):
-                        raise AssertionError(
-                            f"Expected {len(row_expected)} columns in row {i}, got {len(row_result)}"
-                        )
-                    for j, (val_result, val_expected) in enumerate(
-                        zip(row_result, row_expected)
-                    ):
-                        if not is_geom_col[j]:
-                            if val_result != val_expected:
-                                raise AssertionError(
-                                    f"Non-geometry value mismatch at row {i}, col {j}:\n"
-                                    f"Expected:\n  {val_expected}\n"
-                                    f"Got:\n  {val_result}"
-                                )
-                            continue
-
-                        if val_result is None and val_expected is None:
-                            continue
-                        elif val_result is None or val_expected is None:
-                            raise AssertionError(
-                                f"Null mismatch at row {i}, col {j}:\n"
-                                f"Expected:\n  {val_expected}\n"
-                                f"Got:\n  {val_result}"
-                            )
-
-                        if "EMPTY" in val_expected and "nan" in val_result:
-                            continue
-
-                        try:
-                            shape_result = shapely.from_wkt(val_result)
-                            shape_expected = shapely.from_wkt(val_expected)
-                        except Exception as e:
-                            raise AssertionError(
-                                f"Failed to parse WKT at row {i}, col {j}: {e}"
-                            )
-
-                        if not shape_result.equals(shape_expected):
-                            raise AssertionError(
-                                f"Geometry mismatch at row {i}, col {j}:\n"
-                                f"Expected:\n  {val_expected}\n"
-                                f"Got:\n  {val_result}"
-                            )
-
-            else:
-                result_tuples = self.result_to_tuples(result, **kwargs)
-                if result_tuples != expected:
-                    raise AssertionError(
-                        f"Expected:\n  {expected}\nGot:\n  {result_tuples}"
-                    )
+            result_tuples = self.result_to_tuples(result, **kwargs)
+            if result_tuples != expected:
+                raise AssertionError(
+                    f"Expected:\n  {expected}\nGot:\n  {result_tuples}"
+                )
         elif isinstance(expected, tuple):
             self.assert_result(result, [expected], **kwargs)
         elif isinstance(expected, str):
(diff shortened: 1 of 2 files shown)


petern48:
```
FAILED sedonadb/tests/functions/test_functions.py::test_st_unaryunion[MULTIPOLYGON (((0 0, 1 0, 1 1, 0 1, 0 0)), ((1 0, 2 0, 2 1, 1 1, 1 0)))-POLYGON ((0 0, 0 1, 1 1, 2 1, 2 0, 1 0, 0 0))-PostGIS] - AssertionError: Expected:
  [('POLYGON ((0 0, 0 1, 1 1, 2 1, 2 0, 1 0, 0 0))',)]
Got:
  [('POLYGON ((0 1, 1 1, 2 1, 2 0, 1 0, 0 0, 0 1))',)]
FAILED sedonadb/tests/functions/test_functions.py::test_st_unaryunion[GEOMETRYCOLLECTION (POLYGON ((0 0, 1 0, 1 1, 0 1, 0 0)), POLYGON ((1 0, 2 0, 2 1, 1 1, 1 0)))-POLYGON ((0 0, 0 1, 1 1, 2 1, 2 0, 1 0, 0 0))-PostGIS] - AssertionError: Expected:
  [('POLYGON ((0 0, 0 1, 1 1, 2 1, 2 0, 1 0, 0 0))',)]
Got:
  [('POLYGON ((0 1, 1 1, 2 1, 2 0, 1 0, 0 0, 0 1))',)]
```
https://github.com/apache/sedona-db/actions/runs/18684659490/job/53279430046?pr=234

Alright, looking at the CI failure, the expected and actual result are actually the **same geometry**. Their coordinates are just ordered differently. This fails because our current test framework compares using string comparison [here](https://github.com/apache/sedona-db/blob/04aff75e3e69e900bc4516ddd51719ddaea62631/python/sedonadb/python/sedonadb/testing.py#L279-L280).

We can solve this by modifying the test suite in some way to use shapely geom comparisons instead of string comparisons. Here's a script to explain what I mean. Comparing shapely geometries will result in True, but str comparison is False
```python
import shapely
expected = 'POLYGON ((0 0, 0 1, 1 1, 2 1, 2 0, 1 0, 0 0))'
actual = 'POLYGON ((0 1, 1 1, 2 1, 2 0, 1 0, 0 0, 0 1))'
print(shapely.from_wkt(actual).equals(shapely.from_wkt(expected)))
# True
print(expected == actual)
# False
```

Here's some (messy) starter code, from when I tried this myself. If you plop this under the str case [here](https://github.com/apache/sedona-db/blob/04aff75e3e69e900bc4516ddd51719ddaea62631/python/sedonadb/python/sedonadb/testing.py#L279-L280) in `testing.py`, the test passes. We still need the test framework to support string comparison functions that still output strings like `st_geomfromtext`, so we can't just overwrite that logic completely, but maybe we can add a boolean flag or something in `assert_query_results` to use compare shapely geometries instead. I'll leave it up to you to clean it up and propose how to configure it.

```python
        elif isinstance(expected, str):
            import shapely
            result_pandas = self.result_to_pandas(result)
            geom = result_pandas.iloc[0, 0]
            expected_geom = shapely.wkt.loads(expected)
            assert geom.equals(expected_geom), (
                f"Expected {expected}, got {geom}"
            )
```

python/sedonadb/python/sedonadb/testing.py
```diff
@@ -269,11 +269,74 @@ def assert_result(self, result, expected, **kwargs) -> "DBEngine":
...
-            result_tuples = self.result_to_tuples(result, **kwargs)
-            if result_tuples != expected:
-                raise AssertionError(
-                    f"Expected:\n  {expected}\nGot:\n  {result_tuples}"
-                )
+            result_table = self.result_to_table(result)
+            geom_cols_schema = _geometry_columns(result_table.schema)
+
+            if geom_cols_schema:
+                import shapely
```
petern48: Hmm, this logic is much more complicated than I had in mind. I assume it's all because tests were failing otherwise. I actually had a simpler short-term fix in mind. Is there a reason you didn't go with a simpler Boolean flag approach, that we only turn on when needed?

```python
def assert_result(sef result, expected, **kwargs):
      ...
      if kwargs["geom_compare"]:
             < the ~5 line shapely comparison code I had above >
      else:
            < old existing str comparison logic >
            
# then we add the new `geom_compare=True` flag only in test_st_unaryunion()
```

It might not support every case, but I think it should be enough for this PR. Future cases like CRS can be handled later (if they're needed). The reason I imagined a simpler boolean flag approach was to keep this change isolated (though something we could easily turn on/off) and keep the same testing logic for the rest of the functions. The reason was that I wonder if there's any reason we specifically would want to keep the string comparison approach for some functions (ie. returning False for different orderings). I can't think of any at the moment, though. From a reviewer's perspective, modifying this for all cases means we need to scrutinize your PR deeper since you're no longer just adding new functionality. Even if it passes, we need to make sure you're not doing anything negative to the existing code.

WDYT? If you feel strongly about this approach, let me know, and I will look to deeper understand it later. Also, feel free to move the shapely import to the top. I only put it inline in my example for conciseness.

Abeeujah: Hi, your suggestion is great, gets this out of the road, and keeps a very minimal footprint for this PR

I went with this, to help me better safeguard and prevent undefined situations from showing up 

The tests passed locally yesterday before I pushed, only for it to fail when ran on CI the reason I could come up with might be varying dependencies in my local and the CI environment 

To better prepare and prevent that from happening again, I went with this approach 

You can review deeply to ensure it's not creating a regression or if that would be a problem, I can go ahead with your suggestion

Either way, I'm cool with what you'd propose.

python/sedonadb/tests/functions/test_functions.py
```diff
@@ -284,6 +284,36 @@ def test_st_convexhull(eng, geom, expected):
...
+        ),
+        (
+            "GEOMETRYCOLLECTION (POLYGON ((0 0, 1 0, 1 1, 0 1, 0 0)), POLYGON ((1 0, 2 0, 2 1, 1 1, 1 0)))",
+            "POLYGON ((0 0, 0 1, 1 1, 2 1, 2 0, 1 0, 0 0))",
+        ),
+    ],
+)
+def test_st_unaryunion(eng, geom, expected):
+    eng = eng.create_or_skip()
+    eng.assert_query_result(f"SELECT ST_UnaryUnion({geom_or_null(geom)})", expected)
```
paleolimbot: Instead of modifying the assertion code, how about this:

```suggestion
    eng.assert_query_result(f"SELECT ST_Equals(ST_UnaryUnion({geom_or_null(geom)}), geom_or_null(expected))", True)
```

(ST_Equals should compare for topological equality which I think is what you're hoping for)

Abeeujah: This works, shows it is verifiable via other functions as well, I went with this approach, but also included edge cases for `EMPTY` and `None`

python/sedonadb/python/sedonadb/testing.py
Abeeujah: Hi @petern48  thank you so much for your suggestions as well, They were invaluable, made me remember and better understand that every line of code is technical debt

python/sedonadb/tests/functions/test_functions.py
petern48: Yeah, much cleaner. I like it.

PR #237, opened 2025-10-23
chore: Upgrade Datafusion (v50) and Arrow (v56) dependencies

zhangfengcdt:


zhangfengcdt:
> It might be good to go straight to [apache/datafusion#17888](https://github.com/apache/datafusion/pull/17888) (since we'll need it for Geometry/Geography support in Parquet), although if you need this version as an intermediary that's ok too!

I will need the dafafusion v50.2.0 to unblock merging upstream in a dependency project.

rust/sedona-geoparquet/src/writer.rs
```diff
@@ -230,6 +232,7 @@ fn project_bboxes(
                 bbox_udf.clone(),
                 vec![column],
                 Arc::new(Field::new("", bbox_type(), true)),
+                Arc::new(ConfigOptions::default()),
```
paleolimbot, line 237: We're calling a very specific UDF here and so I think this is OK, but worth at least adding a comment here that we could provide the real `ConfigOptions` by piping the `Session` through here (we have a reference a few levels up).

rust/sedona/src/show.rs
```diff
@@ -495,6 +496,7 @@ impl DisplayColumn {
                 arg_fields,
                 number_rows: array.len(),
                 return_field,
+                config_options: Arc::new(ConfigOptions::default()),
```
paleolimbot, line 500: This is a place where we should really pipe in the actual config options. I think we have a reference this a few levels up, but if it's too hard feel free to open up a follow-up ticket (and link to it in a comment here).

rust/sedona/src/show.rs
```diff
@@ -252,7 +252,8 @@ impl<'a> DisplayTable<'a> {
             }
         }
 
-        table.set_truncation_indicator(self.truncation_indicator());
+        // TODO: set_truncation_indicator method was removed in comfy_table 7.x
+        // table.set_truncation_indicator(self.truncation_indicator());
```
paleolimbot: Feel free to open up a follow-up ticket and link to it in this comment if it can't be solved here.

rust/sedona-geoparquet/Cargo.toml
```diff
@@ -46,6 +46,7 @@ chrono = { workspace = true }
 datafusion = { workspace = true, features = ["parquet"] }
 datafusion-catalog = { workspace = true }
 datafusion-common = { workspace = true }
+datafusion-datasource-parquet = "50.2.0"
```
paleolimbot:
```suggestion
datafusion-datasource-parquet = { workspace = true }
```

(...and add this to the top-level Cargo.toml)

paleolimbot (commented):
Thank you!

A few things that should be handled here but also feel free to open follow-up tickets for non-trivial updates.

rust/sedona-geoparquet/src/writer.rs
zhangfengcdt: Ok, I have added todo for these config options linked to this issue: https://github.com/apache/sedona-db/issues/248

rust/sedona/src/show.rs
zhangfengcdt: Upgrade it to 7.1.4 solves the issue now.

rust/sedona/src/show.rs
zhangfengcdt: Same as above, todo links to https://github.com/apache/sedona-db/issues/248

rust/sedona-geoparquet/src/writer.rs
Kontinuation: This API change is pretty interesting. This enables many use cases, such as selecting the spatial library to use by setting DataFusion configurations.

PR #238, opened 2025-10-23
docs: add delta lake documentation

MrPowers:
This adds a documentation page on how to create Delta tables from SedonaDB DataFrame and how to read Delta tables into SedonaDB DataFrames.

### DIFF
--- a/docs/delta-lake.ipynb
+++ b/docs/delta-lake.ipynb
@@ -342,10 +342,10 @@
     "dt = DeltaTable(table_path)\n",
     "dataset = dt.to_pyarrow_dataset()\n",
     "filter_expr = (\n",
-    "    (pc.field(\"bbox\", \"xmax\") >= -97.0) &\n",
-    "    (pc.field(\"bbox\", \"xmin\") <= -67.0) &\n",
-    "    (pc.field(\"bbox\", \"ymax\") >= 25.0) &\n",
-    "    (pc.field(\"bbox\", \"ymin\") <= 50.0)\n",
+    "    (pc.field(\"bbox\", \"xmax\") >= -97.0)\n",
+    "    & (pc.field(\"bbox\", \"xmin\") <= -67.0)\n",
+    "    & (pc.field(\"bbox\", \"ymax\") >= 25.0)\n",
+    "    & (pc.field(\"bbox\", \"ymin\") <= 50.0)\n",
     ")\n",
     "filtered_table = dataset.to_table(filter=filter_expr)"
    ]
--- a/docs/delta-lake.md
+++ b/docs/delta-lake.md
@@ -214,8 +214,8 @@ cities.to_view("cities", True)
 
 ```python
 df = sd.sql("""
-select 
-    name, 
+select
+    name,
     ST_AsBinary(geometry) as geometry_wkb,
     STRUCT(
         ST_XMin(geometry) as xmin,
@@ -277,4 +277,3 @@ sd.sql("select name, ST_GeomFromWKB(geometry_wkb) as geom from us_east_cities").
     ├╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌┤
     │ New York         ┆ POINT(-73.99571754361698 40.72156174972766)  │
     └──────────────────┴──────────────────────────────────────────────┘
-


docs/delta-lake.ipynb
```diff
@@ -65,13 +65,11 @@
...
+    "df = sd.sql(\n",
+    "    \"select name, continent, ST_AsText(geometry) as geometry_wkt from countries\"\n",
+    ")\n",
     "table_path = \"/tmp/delta_with_wkt\"\n",
-    "write_deltalake(\n",
-    "    table_path,\n",
-    "    df.to_pandas(),\n",
-    "    mode=\"overwrite\"\n",
-    ")"
+    "write_deltalake(table_path, df.to_pandas(), mode=\"overwrite\")"
```
paleolimbot, line 75: I think that `write_deltalake()` supports ArrowArrayStreamExportable input (which our `df` is!), so you can just do:

```python
write_deltalake(table_path, df, ...)
```

https://delta-io.github.io/delta-rs/api/delta_writer/#write-to-delta-tables

docs/delta-lake.ipynb
```diff
@@ -0,0 +1,242 @@
...
+      "│ Indonesia                   ┆ Asia          ┆ MULTIPOLYGON(((141.00021040259185 -2.600151055515… │\n",
+      "├╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌┤\n",
+      "│ Argentina                   ┆ South America ┆ MULTIPOLYGON(((-68.63401022758323 -52.63637045887… │\n",
+      "└─────────────────────────────┴───────────────┴────────────────────────────────────────────────────┘\n"
+     ]
+    }
+   ],
+   "source": [
+    "dt = DeltaTable(table_path)\n",
+    "arrow_table = dt.to_pyarrow_table()\n",
```
paleolimbot, line 127: Probably a user will want to select columns or filter with an expression here? (One of the cool things we could do here if we integrated a delta lake table provider would be to insert this query automatically based on the information DataFusion gives us).

docs/delta-lake.ipynb
MrPowers: Got this error when I changed from `df.to_pandas()` => `df`:

```
ValueError                                Traceback (most recent call last)
Cell In[7], line 6
      2 df = sd.sql(
      3     "select name, continent, ST_AsText(geometry) as geometry_wkt from countries"
      4 )
      5 table_path = "[/tmp/delta_with_wkt](http://localhost:8889/tmp/delta_with_wkt)"
----> 6 write_deltalake(table_path, df, mode="overwrite")

File [/opt/miniconda3/lib/python3.12/site-packages/deltalake/writer.py:306](http://localhost:8889/opt/miniconda3/lib/python3.12/site-packages/deltalake/writer.py#line=305), in write_deltalake(table_or_uri, data, schema, partition_by, mode, file_options, max_partitions, max_open_files, max_rows_per_file, min_rows_per_group, max_rows_per_group, name, description, configuration, schema_mode, storage_options, partition_filters, predicate, target_file_size, large_dtypes, engine, writer_properties, custom_metadata, post_commithook_properties)
    300 data, schema = _convert_data_and_schema(
    301     data=data,
    302     schema=schema,
    303     conversion_mode=ArrowSchemaConversionMode.PASSTHROUGH,
    304 )
    305 data = RecordBatchReader.from_batches(schema, (batch for batch in data))
--> 306 write_deltalake_rust(
    307     table_uri=table_uri,
    308     data=data,
    309     partition_by=partition_by,
    310     mode=mode,
    311     table=table._table if table is not None else None,
    312     schema_mode=schema_mode,
    313     predicate=predicate,
    314     target_file_size=target_file_size,
    315     name=name,
    316     description=description,
    317     configuration=configuration,
    318     storage_options=storage_options,
    319     writer_properties=writer_properties,
    320     custom_metadata=custom_metadata,
    321     post_commithook_properties=post_commithook_properties,
    322 )
    323 if table:
    324     table.update_incremental()

ValueError: C Data interface error: The datatype ""vu"" is still not supported in Rust implementation
```

docs/delta-lake.ipynb
MrPowers: Yea, for sure.  For DuckDB, the read Delta + apply filtering query pattern benefits from PyArrow Datasets vs PyArrow tables.  I can prepare a little benchmark if that'd be interesting.  

I added a filtering example to the notebook to show the benefits of the geometry data type.

docs/delta-lake.ipynb
paleolimbot: Ah, no string view support in the version of arrow being used by delta lake. That's a shame.

paleolimbot: It's worth trying this:

```python
sd.sql("SET datafusion.execution.parquet.schema_force_view_types = false").execute()
```

...and using the `df` object directly. You really want to be streaming the query result into DeltaLake so that you're not limited to the size of the memory.

docs/delta-lake.md
```diff
@@ -0,0 +1,202 @@
...
+
+Now write the DataFrame to a Delta Lake table.  Notice that you must convert the geometry column to Well-Known Text (WKT) before writing to the Delta table.
+
+Delta Lake does not support geometry columns.
+
+
+```python
+countries.to_view("countries", True)
+df = sd.sql(
+    "select name, continent, ST_AsText(geometry) as geometry_wkt from countries"
```
paleolimbot, line 56: In real life you'd probably want `ST_AsBinary()` (more compact for most usage)?

docs/delta-lake.ipynb
paleolimbot: Probably also worth demonstrating how you can push a query down into the DeltaLake scan using its `filter` argument (most real world usage would do that instead of load the entire table into memory and then issue a query on it?).

One of the things you could demo is adding bbox columns on write by using `ST_Xmin()` and friends. When you read, you can issue a Delta Lake filter like

```
bbox.xmin <= -73.11 AND
    bbox.ymin <= 44.03 AND
    bbox.xmax >= -73.21 AND
    bbox.ymax >= 43.97
```

...except in whatever syntax the delta lake `filter` argument uses. That should result in a more reasonable fetch from a large local or remote table. (Otherwise, users are better off just using Parquet because the pushdown is better).

docs/delta-lake.md
```diff
@@ -0,0 +1,202 @@
...
+
+```python
+res = sd.sql("""
+SELECT
+  name,
+  continent,
+  ST_GeomFromWKT(geometry_wkt) as geom
+FROM my_table
+WHERE ST_Intersects(
+  ST_GeomFromWKT(geometry_wkt),
```
martin-g, line 175: educational question: could we use `geom` here instead?

docs/delta-lake.md
```diff
@@ -0,0 +1,202 @@
...
+  KIND, either express or implied.  See the License for the
+  specific language governing permissions and limitations
+  under the License.
+-->
+
+# SedonaDB + Delta Lake
+
+This page shows how to read and write Delta Lake tables with SedonaDB.
+
+Make sure you run `pip install deltalake` to run the cells in this notebook.
```
martin-g:
```suggestion
Make sure you run `pip install deltalake` to run the code snippets below.
```

docs/delta-lake.md
```diff
@@ -0,0 +1,202 @@
...
+    SedonaSchema with 3 fields:
+      name: utf8<Utf8>
+      continent: utf8<Utf8>
+      geom: geometry<Wkb>
+
+
+
+## Filter countries in a particular geographic region
+
+Now, let's grab all the countries in the western half of South America using a polygon region.
```
martin-g: The result contains Panama (North America). Maybe you should update the polygon region ?!

docs/delta-lake.ipynb
MrPowers: I tried this, but it still has this error: "ValueError: C Data interface error: The datatype ""vu"" is still not supported in Rust implementation"

docs/delta-lake.md
MrPowers: That doesn't work actually, I just checked.

docs/delta-lake.md
MrPowers: Updated

docs/delta-lake.md
martin-g: It might need to be something like:
```sql
WITH geom_data AS (
  SELECT name, continent, ST_GeomFromWKT(geometry_wkt) as geom
  FROM my_table
)
SELECT * FROM geom_data
WHERE ST_Intersects(
  geom,
  ST_GeomFromWKT('POLYGON((-81 12, -58 12, -58 -55, -81 -55, -81 12))')
)
```
but it is getting more complex this way

docs/delta-lake.md
MrPowers: Yea, good point, updated this!

docs/delta-lake.ipynb
MrPowers: I will add this section, good call.

docs/delta-lake.md
MrPowers: I added a section with `ST_AsBinary()`.  Good idea.

MrPowers:
Thanks @paleolimbot!

Can you just take a quick look at the bbox code I added before merging?

docs/delta-lake.md
```diff
@@ -0,0 +1,279 @@
...
+
+
+```python
+dt = DeltaTable(table_path)
+dataset = dt.to_pyarrow_dataset()
+filter_expr = (
+    (pc.field("bbox", "xmax") >= -97.0) &
+    (pc.field("bbox", "xmin") <= -67.0) &
+    (pc.field("bbox", "ymax") >= 25.0) &
+    (pc.field("bbox", "ymin") <= 50.0)
```
paleolimbot, line 242: I always have to look up this math but this looks right to me (particularly if it's giving you the features that you think it's giving you)!

MrPowers:
Let me know if any additional changes are needed or if this can be merged!

PR #239, opened 2025-10-24
Implement ST_IsSimple

Abeeujah:


petern48:
@paleolimbot What are your thoughts on making it so failing `cargo clippy` doesn't short-circuit the rest of the Rust checks? It's especially easy to forget to do since it's not included in the pre-commit.

paleolimbot (approved):
Thank you!

PR #241, opened 2025-10-24
Plumb remaining parameters for ST_Buffer

### LINKED ISSUE #223: feat: Implement remaining parameters for ST_Buffer
Here's the original commit that implemented basic support for ST_Buffer: https://github.com/apache/sedona-db/commit/0de00cbf479e2b4f4e6e96b341f3ef58687ea032

Currently, it simply creates and uses the default parameters
https://github.com/apache/sedona-db/blob/df8f7205a0a99bd75195933084ae4216781f067c/c/sedona-geos/src/st_buffer.rs#L57-L62

We should support the buffer style parameters as user input (following PostGIS syntax, not Sedona)
~~https://sedona.apache.org/1.8.0/api/sql/Function/#st_buffer~~
https://postgis.net/docs/ST_Buffer.html

The wording can be a bit misleading / ambiguous in the original PostGIS docs, so it may be worth looking at this [bug fix](https://github.com/apache/sedona/pull/2184) in the original Sedona to understand the proper behavior.

Abeeujah:
closes #223 

- `geos` doesn't have a method that computes `buffer` with `useSpheroid` so I didn't implement that yet
- The `testers.rs` isn't really clear to me how to proceed with four args so I haven't been thorough with the unit tests

### DIFF
--- a/c/sedona-geos/src/st_buffer.rs
+++ b/c/sedona-geos/src/st_buffer.rs
@@ -119,29 +119,17 @@ fn invoke_batch_impl(arg_types: &[SedonaType], args: &[ColumnarValue]) -> Result
     // Build BufferParams based on style parameters
     let params = parse_buffer_params(buffer_style_params.as_deref())?;
 
-    let is_single_sided = buffer_style_params
-        .as_ref()
-        .map(|s| s.eq_ignore_ascii_case("side="))
-        .unwrap_or(false);
-
-    let is_left = buffer_style_params
-        .as_ref()
-        .map(|s| s.eq_ignore_ascii_case("left"))
-        .unwrap_or(false);
-
-    let is_right = buffer_style_params
-        .as_ref()
-        .map(|s| s.eq_ignore_ascii_case("right"))
-        .unwrap_or(false);
+    // Parse 'side' from the style parameters
+    let (is_left, is_right) = parse_buffer_side_style(buffer_style_params.as_deref());
 
     executor.execute_wkb_void(|wkb| {
         match (wkb, distance_iter.next().unwrap()) {
             (Some(wkb), Some(mut distance)) => {
-                if is_single_sided && ((is_left && distance < 0.0) || (is_right && distance > 0.0))
-                {
+                if (is_left && distance < 0.0) || (is_right && distance > 0.0) {
                     distance = -distance;
                 }
-                builder.append_value(invoke_scalar(&wkb, distance, &params)?);
+                invoke_scalar(&wkb, distance, &params, &mut builder)?;
+                builder.append_value([]);
             }
             _ => builder.append_null(),
         }
@@ -155,15 +143,18 @@ fn invoke_scalar(
     geos_geom: &geos::Geometry,
     distance: f64,
     params: &BufferParams,
-) -> Result<Vec<u8>> {
+    writer: &mut impl std::io::Write,
+) -> Result<()> {
     let geometry = geos_geom
         .buffer_with_params(distance, params)
         .map_err(|e| DataFusionError::External(Box::new(e)))?;
+
     let wkb = geometry
         .to_wkb()
         .map_err(|e| DataFusionError::Execution(format!("Failed to convert to wkb: {e}")))?;
 
-    Ok(wkb.into())
+    writer.write_all(wkb.as_ref())?;
+    Ok(())
 }
 
 fn extract_optional_string(arg: Option<&ColumnarValue>) -> Result<Option<String>> {
@@ -179,6 +170,29 @@ fn extract_optional_string(arg: Option<&ColumnarValue>) -> Result<Option<String>
             "Expected scalar bufferStyleParameters, got: {arg:?}",
         ))),
     }
+}
+
+fn parse_buffer_side_style(params: Option<&str>) -> (bool, bool) {
+    params
+        .map(|s| {
+            let mut left = false;
+            let mut right = false;
+            for tok in s.split_whitespace() {
+                if let Some((k, v)) = tok.split_once('=') {
+                    if k.eq_ignore_ascii_case("side") {
+                        if v.eq_ignore_ascii_case("left") {
+                            left = true;
+                            right = false;
+                        } else if v.eq_ignore_ascii_case("right") {
+                            right = true;
+                            left = false;
+                        }
+                    }
+                }
+            }
+            (left, right)
+        })
+        .unwrap_or((false, false))
 }
 
 fn parse_buffer_params(params_str: Option<&str>) -> Result<BufferParams> {
@@ -206,7 +220,7 @@ fn parse_buffer_params(params_str: Option<&str>) -> Result<BufferParams> {
         } else if key.eq_ignore_ascii_case("side") {
             let single_sided = is_single_sided(value)?;
             if single_sided && !end_cap_specified {
-                params_builder = params_builder.end_cap_style(CapStyle::Square)
+                params_builder = params_builder.end_cap_style(CapStyle::Square);
             }
             params_builder = params_builder.single_sided(single_sided);
         } else if key.eq_ignore_ascii_case("mitre_limit") || key.eq_ignore_ascii_case("miter_limit")
@@ -216,7 +230,7 @@ fn parse_buffer_params(params_str: Option<&str>) -> Result<BufferParams> {
         } else if key.eq_ignore_ascii_case("quad_segs")
             || key.eq_ignore_ascii_case("quadrant_segments")
         {
-            let segs = parse_number(value, "quadrant_segments")?;
+            let segs: i32 = parse_number(value, "quadrant_segments")?;
             params_builder = params_builder.quadrant_segments(segs);
         } else {
             return Err(DataFusionError::Execution(format!(
@@ -591,5 +605,94 @@ mod tests {
             .unwrap();
 
         assert!(result_buffer.equals_exact(&expected_buffer, 0.1).unwrap());
+    }
+
+    #[test]
+    fn test_side_right_geos_3_13() {
+        let wkt = "LINESTRING(50 50, 150 150, 150 50)";
+        let line = geos::Geometry::new_from_wkt(wkt).unwrap();
+        let distance = 100.0;
+
+        // Test single-sided buffer (GEOS 3.13+ removes artifacts, giving 12713.61)
+        // PostGIS with GEOS 3.9 returns 16285.08 due to including geometric artifacts
+        // GEOS 3.12+ improvements: https://github.com/libgeos/geos/commit/091f6d99
+        let params_single = BufferParams::builder().single_sided(true).build().unwrap();
+
+        let buffer_right = line.buffer_with_params(-distance, &params_single).unwrap();
+        let area_right = buffer_right.area().unwrap();
+
+        // Expected area with GEOS 3.13 (improved algorithm without artifacts)
+        assert!(
+            (area_right - 12713.605978550266).abs() < 0.1,
+            "Expected GEOS 3.13+ area ~12713.61, got {}",
+            area_right
+        );
+    }
+
+    #[test]
+    fn test_empty_and_invalid_input() {
+        assert_eq!(
+            parse_buffer_side_style(None),
+            (false, false),
+            "Should return (false, false) for None."
+        );
+        assert_eq!(
+            parse_buffer_side_style(Some("")),
+            (false, false),
+            "Should return (false, false) for an empty string."
+        );
+        assert_eq!(
+            parse_buffer_side_style(Some("mitre_limit=5.0")),
+            (false, false),
+            "Should return (false, false) for an invalid key."
+        );
+    }
+
+    #[test]
+    fn test_single_side_and_case_insensitivity() {
+        assert_eq!(
+            parse_buffer_side_style(Some("side=left")),
+            (true, false),
+            "Should detect 'left'."
+        );
+        assert_eq!(
+            parse_buffer_side_style(Some("side=RIGHT")),
+            (false, true),
+            "Should detect 'RIGHT' case-insensitively."
+        );
+        assert_eq!(
+            parse_buffer_side_style(Some("SiDe=LeFt")),
+            (true, false),
+            "Should handle mixed case key and value."
+        );
+        assert_eq!(
+            parse_buffer_side_style(Some("join=mitre SIDE=RIGHT mitre_limit=5.0")),
+            (false, true),
+            "Should ignore other params and detect 'RIGHT'."
+        );
+        assert_eq!(
+            parse_buffer_side_style(Some("side=center")),
+            (false, false),
+            "Should ignore invalid side values."
+        );
+    }
+
+    #[test]
+    fn test_both_sides_present() {
+        assert_eq!(
+            parse_buffer_side_style(Some("side=left side=right")),
+            (false, true),
+            "Should detect both left and right."
+        );
+        assert_eq!(
+            parse_buffer_side_style(Some("side=right side=left join=round")),
+            (true, false),
+            "Should detect both regardless of order."
+        );
+        assert_eq!(
+            parse_buffer_side_style(Some("SIDE=RIGHT endcap=round side=left")),
+            (true, false),
+            "Should handle complex string with both sides."
+        );
     }
 }
--- a/python/sedonadb/tests/functions/test_functions.py
+++ b/python/sedonadb/tests/functions/test_functions.py
@@ -272,18 +272,26 @@ def test_st_buffer(eng, geom, dist, expected_area):
             "'side=left'",
             1000.0,
         ),
-        ("LINESTRING (50 50, 150 150, 150 50)", 100, "'side=right'", 16285.07633336958),
+        # GEOS version difference: GEOS 3.9 (PostGIS) returns 16285.08 with artifacts
+        # GEOS 3.12+ (SedonaDB) returns 12713.61 without artifacts (more accurate)
+        # See: https://github.com/libgeos/geos/commit/091f6d99
+        (
+            "LINESTRING (50 50, 150 150, 150 50)",
+            100,
+            "'side=right'",
+            12713.605978550266,
+        ),
         (
             "POLYGON ((50 50, 50 150, 150 150, 150 50, 50 50))",
             20,
             "'side=left'",
-            19248.578060903223,
+            10000.0,  # GEOS 3.9 (PostGIS): 19248.58
         ),
         (
             "POLYGON ((50 50, 50 150, 150 150, 150 50, 50 50))",
             20,
             "'side=right endcap=flat'",
-            3600.0,
+            6400.0,  # GEOS 3.9 (PostGIS): 3600.0
         ),
         (
             "LINESTRING (50 50, 150 150, 150 50)",
--- a/compose.yml
+++ b/compose.yml
@@ -17,7 +17,7 @@
 services:
   postgis:
     platform: linux/amd64
-    image: postgis/postgis:latest
+    image: postgis/postgis:18-3.6
     ports:
       - "5432:5432"
     environment:


c/sedona-geos/src/st_buffer.rs
```diff
@@ -32,6 +32,19 @@ use sedona_schema::{
 use crate::executor::GeosExecutor;
 
 /// ST_Buffer() implementation using the geos crate
+///
+/// Supports three signatures:
+/// - ST_Buffer(geometry: Geometry, distance: Double)
+/// - ST_Buffer(geometry: Geometry, distance: Double, useSpheroid: Boolean)
+/// - ST_Buffer(geometry: Geometry, distance: Double, useSpheroid: Boolean, bufferStyleParameters: String)
```
petern48: I see you're following Sedona's buffer signatures.

I think we want to follow PostGIS's signatures, which don't include useSpheroid.  @paleolimbot Confirm?
https://postgis.net/docs/ST_Buffer.html

We already have this one:
`ST_Buffer(geometry g1, float radius_of_buffer);`

This is the one I had in mind in the issue
`ST_Buffer(geometry g1, float radius_of_buffer, text buffer_style_parameters = '');`

PostGIS also supports the below signature. To my understanding, it's just the quad_segs argument in buffer style params, so it's not new functionality. You're welcome to implement this too, though I mainly had the previous one in mind.
`ST_Buffer(geometry g1, float radius_of_buffer, integer num_seg_quarter_circle);`

petern48:
> The testers.rs isn't really clear to me how to proceed with four args so I haven't been thorough with the unit tests

Doesn't seem like this will be needed.

Let's also add lots more Python integration tests, so we can compare directly to PostGIS. Makes sense to create a new test function like `test_st_buffer_with_params` or something.

c/sedona-geos/src/st_buffer.rs
paleolimbot: Yes, I think we want PostGIS signatures for now. PostGIS and Sedona Spark implement the spheroidal buffer by reprojecting, doing a buffer, then projecting back, but PostGIS does this as `ST_Buffer(<GEOGRAPHY>)` instead of geometry + useSpheroid.

c/sedona-geos/src/st_buffer.rs
```diff
@@ -41,41 +54,37 @@ struct STBuffer {}
...
     fn return_type(&self, args: &[SedonaType]) -> Result<Option<SedonaType>> {
-        let matcher = ArgMatcher::new(
-            vec![ArgMatcher::is_geometry(), ArgMatcher::is_numeric()],
-            WKB_GEOMETRY,
-        );
+        if !(2..=4).contains(&args.len()) {
+            return Err(DataFusionError::Plan(format!(
+                "ST_Buffer expects 2-4 arguments, got {}",
+                args.len()
+            )));
```
paleolimbot: The kernel system is really intended have multiple `SedonaScalarKernel`s that handle the matching of the various numbers/types of arguments. For example:

```rust
struct STBuffer{};

impl SedonaScalarKernel for STBufferStyle {
    fn return_type(&self, args: &[SedonaType]) -> Result<Option<SedonaType>> {
        let matcher = ArgMatcher::new(
            vec![ArgMatcher::is_geometry(), ArgMatcher::is_numeric()],
            WKB_GEOMETRY,
        );
        matcher.match_args(args)
    }

    fn invoke_batch(...) { invoke_batch_impl(...) }
}

struct STBufferStyle{};

impl SedonaScalarKernel for STBuffer {
    fn return_type(&self, args: &[SedonaType]) -> Result<Option<SedonaType>> {
        let matcher = ArgMatcher::new(
            vec![ArgMatcher::is_geometry(), ArgMatcher::is_numeric(), ArgMatcher::is_string()],
            WKB_GEOMETRY,
        );
        matcher.match_args(args)
    }

    fn invoke_batch(...) { invoke_batch_impl(...) }
}

fn invoke_batch_impl(...) { ... }
```

c/sedona-geos/src/st_buffer.rs
```diff
@@ -41,41 +54,37 @@ struct STBuffer {}
...
 
     fn invoke_batch(
         &self,
         arg_types: &[SedonaType],
         args: &[ColumnarValue],
     ) -> Result<ColumnarValue> {
-        // Default params
-        let params_builder = BufferParams::builder();
+        // Extract Args
+        let distance: Option<f64> = extract_optional_f64(&args[1])?;
```
paleolimbot: For the distance parameter, I think that it is worth the effort to allow this to be an array (and I have confidence in your ability to do so 🙂 ). The pattern we use for this in other places is something like:

```rust
let distance_value = args[1].cast_to(&DataType::Float64)?.to_array(executor.num_iterations());
let distance_array = if let ColumnarValue::Array(array_ref) = distance_value {
  as_float64_array(array_ref)?
} else {
  return sedona_internal_err!("Unexpected scalar value");
};

let mut distance_iter = distance_array.iter();

// ...

executor.execute_wkb_void(|wkb| {
   match (wkb, distance_iter.next().unwrap() {
       (Some(wkb_item), Some(distance_item)) => { invoke_scalar(...) }
       _ => { /* append null */ }
   }
})
```

c/sedona-geos/src/st_buffer.rs
```diff
@@ -115,6 +124,136 @@ fn invoke_scalar(
...
+        ColumnarValue::Scalar(scalar) if scalar.is_null() => Ok(None),
+        ColumnarValue::Scalar(_) => Ok(None),
+        _ => Err(DataFusionError::Execution(format!(
+            "Expected scalar bufferStyleParameters, got: {:?}",
+            arg
+        ))),
+    }
+}
+
+fn parse_buffer_params(params_str: Option<&str>) -> Result<BufferParams> {
```
paleolimbot, line 198: Given the complexity here, I think it is worth writing dedicated tests for this function below, including the error messages. (In many places we do `let err = parse_buffer_params(...).unwrap_err(); assert_eq!(err.message(), "...")`).

Abeeujah:
Hi @petern48 @paleolimbot Please guys do let me know if this is in line with the suggestions dropped, So I can proceed or course correct, The reviews has been great so far!

c/sedona-geos/src/st_buffer.rs
Abeeujah: I think, seeing the Sedona reference in the issue description made me think it was the more favored one, but I just updated it to use the postgis signature, please do check and let me know what you think.

c/sedona-geos/src/st_buffer.rs
Abeeujah: So if I get you correctly, two kernels, I get to register the new one as well?

c/sedona-geos/src/st_buffer.rs
Abeeujah: I did convert it to an array :sweat_smile: thanks for the vote of confidence, please do let me know if the implementation is going as expected

c/sedona-geos/src/st_buffer.rs
Abeeujah: Sure, I would include tests for the functions!

c/sedona-geos/src/register.rs
```diff
@@ -37,6 +44,7 @@ pub fn scalar_kernels() -> Vec<(&'static str, ScalarKernelRef)> {
     vec![
         ("st_area", st_area_impl()),
         ("st_buffer", st_buffer_impl()),
+        ("st_buffer_style", st_buffer_style_impl()),
```
petern48:
```suggestion
        ("st_buffer", st_buffer_style_impl()),
```

This string here represents the actual string we call in SQL, so we actually want this to be `"st_buffer"` and not `"st_buffer_style"`. With this current code, we have to instead call it like below. You can verify using the cli (`cd sedona-cli/; cargo run`), or using python.

```sql
select st_buffer_style(st_geomfromtext('point (0 0)'), 1, '');
```

Aside: Skimming through the datafusion code, it looked like `"st_buffer"` would be overwritten, since it's just using a map of name -> udf [here](https://github.com/apache/datafusion/blob/2a828972d1d4950a688ba5b2525202193bd2cabd/datafusion/core/src/execution/session_state.rs#L1855), but when I tried changing the string, both kernels work simultaneously as desired. 🤷

c/sedona-geos/src/st_buffer.rs
```diff
@@ -163,4 +300,46 @@ mod tests {
...
+                    "LINESTRING (0 0, 10 0)",
+                )))),
+                ColumnarValue::Scalar(ScalarValue::Float64(Some(1.0))),
+                ColumnarValue::Scalar(ScalarValue::Utf8(Some(
+                    "endcap=flat join=mitre quad_segs=2".to_string(),
+                ))),
+            ])
+            .unwrap();
+
+        assert!(matches!(buffer_result, ColumnarValue::Scalar(_)));
```
paleolimbot: I think the envelope tester (like the one used in the above example) would work here to give a reasonable "is it plugged in" check. I think the Python integration tests use `ST_Equals()` or maybe compare `ST_Area()` which should work here, too.

c/sedona-geos/src/st_buffer.rs
```diff
@@ -54,50 +67,73 @@ impl SedonaScalarKernel for STBuffer {
...
+        executor.num_iterations(),
+        WKB_MIN_PROBABLE_BYTES * executor.num_iterations(),
+    );
+
+    // Extract Args
+    let distance_value = args[1]
+        .cast_to(&DataType::Float64, None)?
+        .to_array(executor.num_iterations())?;
+    let distance_array = as_float64_array(&distance_value)?;
+    let mut distance_iter = distance_array.iter();
```
paleolimbot, line 115: This looks good...thanks!

c/sedona-geos/src/st_buffer.rs
petern48: Yes, let's definitely add negative tests. Some ideas: invalid keys (e.g `key1=2`), invalid value types (e.g `mitre_limit=abcd`).

Add a test for empty string (which should work). We should have some at the Python level, as well in case we add new kernels from another library (though idk how likely that is, considering `geo` doesn't support all params).

c/sedona-geos/src/st_buffer.rs
petern48: Sorry, that's my bad 😬 . I updated the issue now. Hopefully you didn't spend too long figure out how to get `useSpheroid` to work.

c/sedona-geos/src/st_buffer.rs
petern48: Just providing more motivation on *why* multiple kernels are preferred.

If we discover the `geo` implementation of a basic `st_buffer(geom, dist)` (no extra params) is faster (), we could support that kernel for `st_buffer(geom, dist)`, while simultaneously supporting the geos kernel that you're adding here.

In the code below, we register all of the functions from different libraries. If a kernel exists for two libraries, the earlier one will be overwritten. e.g `geo` overwrites `geos` at the moment.
https://github.com/apache/sedona-db/blob/b8ac5efbe358ebb9344cf9177496447e233d2253/rust/sedona/src/context.rs#L128-L134

If it turns out the `geo` implementation is faster than `geos` ([issue](https://github.com/apache/sedona-db/issues/232)), adding these new params as a separate kernel allows us to keep a `geo` kernel for `st_buffer(geom, dist)` while also keeping a `geos` kernel for `st_geom(geom, dist, params)`. `geo` [doesn't seem to support](https://github.com/georust/geo/blob/7263968eb533028fd9be81eaeb9ca455e8a2aee6/geo/src/algorithm/buffer.rs#L141-L145) all of the params either way, so this `geos` implementation will be helpful for us regardless.

idk off the top of my head if things are hooked up so it would work that way, but that's the idea anyways.

petern48:
Looking good so far to me. The biggest thing we're missing is Python integration tests. In case you write these up and find the current logic is flawed in *some* way (e.g we forgot/overlooked something), I recommend referencing [Sedona's parseBufferParams() function](https://github.com/apache/sedona/blob/d6ea87cf9f2a27adee00ff0fb41ee92002b9a834/common/src/main/java/org/apache/sedona/common/Functions.java#L311) and my previous [PR](https://github.com/apache/sedona/pull/2184) to fix it. I recall PostGIS's docs don't do the best job of explaining how default params are handled when certain values are specified. For example, I think we're missing [this logic](https://github.com/apache/sedona/blob/d6ea87cf9f2a27adee00ff0fb41ee92002b9a834/common/src/main/java/org/apache/sedona/common/Functions.java#L393-L402) at the moment (use python tests to check).

c/sedona-geos/src/st_buffer.rs
```diff
@@ -115,6 +151,115 @@ fn invoke_scalar(
...
+        "mitre" | "miter" => Ok(JoinStyle::Mitre),
+        "bevel" => Ok(JoinStyle::Bevel),
+        _ => Err(DataFusionError::Execution(format!(
+            "Invalid join style: '{}'. Valid options: round, mitre, miter, bevel",
+            value
+        ))),
+    }
+}
+
+fn parse_side(value: &str) -> Result<bool> {
```
martin-g:
```suggestion
fn is_single_sided(value: &str) -> Result<bool> {
```
nit: I think `is_single_sided` would be a better name here.
`parse_side` would be more appropriate if it was returning a `Side` enum (with variants BOTH, LEFT and RIGHT) and this enum could have a method `is_single_sided`

c/sedona-geos/src/st_buffer.rs
```diff
@@ -115,6 +151,115 @@ fn invoke_scalar(
...
+                let limit: f64 = parse_number(value, "mitre_limit")?;
+                params_builder = params_builder.mitre_limit(limit);
+            }
+            "quad_segs" | "quadrant_segments" => {
+                let segs = parse_number(value, "quadrant_segments")?;
+                params_builder = params_builder.quadrant_segments(segs);
+            }
+            _ => {
+                return Err(DataFusionError::Execution(format!(
+                    "Invalid buffer parameter: {} (accept: 'endcap', 'join', 'mitre_limit', 'miter_limit', 'quad_segs' and 'side')",
```
martin-g:
```suggestion
                    "Invalid buffer parameter: {} (accept: 'endcap', 'join', 'mitre_limit', 'miter_limit', 'quad_segs', 'quadrant_segments' and 'side')",
```

c/sedona-geos/src/st_buffer.rs
```diff
@@ -54,50 +67,73 @@ impl SedonaScalarKernel for STBuffer {
...
+    let distance_array = as_float64_array(&distance_value)?;
+    let mut distance_iter = distance_array.iter();
+
+    let buffer_style_params = extract_optional_string(args.get(2))?;
+
+    // Build BufferParams based on style parameters
+    let params = parse_buffer_params(buffer_style_params.as_deref())?;
+
+    executor.execute_wkb_void(|wkb| {
+        match (wkb, distance_iter.next().unwrap()) {
```
martin-g, line 126: Is it safe to call `.unwrap()` here ?
I.e. do you want to panic when there is a problem or an Err could be returned and let the caller decide what to do with it?

c/sedona-geos/src/st_buffer.rs
```diff
@@ -54,50 +67,73 @@ impl SedonaScalarKernel for STBuffer {
...
-            return Err(DataFusionError::Execution(format!(
-                "Invalid distance: {:?}",
-                args[1]
-            )));
-        }
+pub fn st_buffer_style_impl() -> ScalarKernelRef {
+    Arc::new(STBufferStyle {})
+}
+#[derive(Debug)]
+struct STBufferStyle {}
```
martin-g, line 78: I wonder whether the second struct is really needed.
Could you use an optional ArgMatcher (https://github.com/apache/sedona-db/blob/8e6234c7401fc72895752558cc6aeaedd1ff470a/rust/sedona-schema/src/matchers.rs#L202) for the third parameter and keep all the logic in STBuffer ?!

c/sedona-geos/src/st_buffer.rs
```diff
@@ -115,6 +151,115 @@ fn invoke_scalar(
...
+
+    for param in params_str.split_whitespace() {
+        let Some((key, value)) = param.split_once('=') else {
+            return Err(DataFusionError::Execution(format!(
+                "Missing value for buffer parameter: {}",
+                param
+            )));
+        };
+
+        match key.to_lowercase().as_str() {
```
martin-g: nit: this will cause allocations.
You may want to use eq_ignore_ascii_case() instead.

c/sedona-geos/src/st_buffer.rs
```diff
@@ -115,6 +151,115 @@ fn invoke_scalar(
...
+                params_builder = params_builder.mitre_limit(limit);
+            }
+            "quad_segs" | "quadrant_segments" => {
+                let segs = parse_number(value, "quadrant_segments")?;
+                params_builder = params_builder.quadrant_segments(segs);
+            }
+            _ => {
+                return Err(DataFusionError::Execution(format!(
+                    "Invalid buffer parameter: {} (accept: 'endcap', 'join', 'mitre_limit', 'miter_limit', 'quad_segs' and 'side')",
+                    key.to_lowercase()
```
martin-g: IMO it would be clearer for the developer to see the original key here instead of its lowercase version.

c/sedona-geos/src/st_buffer.rs
```diff
@@ -115,6 +151,115 @@ fn invoke_scalar(
...
+        }
+    }
+
+    params_builder
+        .build()
+        .map_err(|e| DataFusionError::External(Box::new(e)))
+}
+
+fn parse_cap_style(value: &str) -> Result<CapStyle> {
+    match value.to_lowercase().as_str() {
```
martin-g: nit: this will cause allocations.
You may want to use eq_ignore_ascii_case() instead.

c/sedona-geos/src/st_buffer.rs
```diff
@@ -115,6 +151,115 @@ fn invoke_scalar(
...
+        "square" => Ok(CapStyle::Square),
+        _ => Err(DataFusionError::Execution(format!(
+            "Invalid endcap style: '{}'. Valid options: round, flat, butt, square",
+            value
+        ))),
+    }
+}
+
+fn parse_join_style(value: &str) -> Result<JoinStyle> {
+    match value.to_lowercase().as_str() {
```
martin-g: nit: this will cause allocations.
You may want to use eq_ignore_ascii_case() instead.

c/sedona-geos/src/st_buffer.rs
Abeeujah: Theoretically, it's infallible, cause there would always be an item for every iteration, the array was initialized to be the size of the executors number of iteration, that's why unwrap is called there.

c/sedona-geos/src/st_buffer.rs
Abeeujah: I did something similar initially, but this current approach is preferred, as reviewed above in https://github.com/apache/sedona-db/pull/241#discussion_r2462399367

c/sedona-geos/src/st_buffer.rs
paleolimbot: I'm the one that suggested the unwrap here...it's true that we usually prefer returning an internal error even for things we think are impossible, but here I think unwrap is OK choice because it's in a tight loop and we really did just initialize the array.

c/sedona-geos/src/st_buffer.rs
paleolimbot: It's a great point that we could use an optional matcher...I kind of like the one kernel per combination of arguments but if the other way can be done cleanly it's OK too.

c/sedona-geos/src/st_buffer.rs
```diff
@@ -163,4 +313,211 @@ mod tests {
...
+    #[test]
+    fn test_parse_buffer_params_missing_value() {
+        let err = parse_buffer_params(Some("endcap=round bare_param join=mitre"))
+            .err()
+            .unwrap();
+        assert_eq!(
+            err.message(),
+            "Missing value for buffer parameter: bare_param"
+        );
+    }
```
paleolimbot, line 525: Thank you for testing all of these!

Abeeujah:
A bit weird cause these pass locally, could be due to the append_value, and geos not being same version, I got 3.13 locally and I believe CI uses 3.14;

I just modified the invoke_scalar to return the bytes, and then write it using the binary builder can you check again?

c/sedona-geos/src/st_buffer.rs
```diff
@@ -163,4 +302,198 @@ mod tests {
...
+        // This test at least ensures that the buffer parameters are plugged in.
+        let buffer_result_flat = tester
+            .invoke_scalar_scalar_scalar("LINESTRING (0 0, 10 0)", 2.0, "endcap=flat".to_string())
+            .unwrap();
+
+        let buffer_result_square = tester
+            .invoke_scalar_scalar_scalar("LINESTRING (0 0, 10 0)", 1.0, "endcap=square".to_string())
+            .unwrap();
+
+        assert_ne!(buffer_result_flat, buffer_result_square);
```
paleolimbot, line 368: I pushed this to try to get CI passing since I wasn't sure myself exactly how to do it without trying a few things...I think this is OK given the integration tests. Buffer is just very hard one to test.

petern48:
don't merge yet

python/sedonadb/tests/functions/test_functions.py
```diff
@@ -176,6 +176,115 @@ def test_st_buffer(eng, geom, dist, expected_area):
...
+            10,
+            "'join=miter miter_limit=2'",
+            2312.1445152258043,
+        ),
+        (
+            "LINESTRING(0 0, 0 100)",
+            10,
+            "'side=left'",
+            1000.0,
+        ),
```
petern48, line 274: This one test case with `side` is too simple that it's not catching the edge case behavior.

Here are 4 cases I want to add. You're welcome to modify the exact test cases, but I think we should follow the comments.

```python
        # non-default side and should implicitly use square endcap
        (
            "LINESTRING (50 50, 150 150, 150 50)",
            100,
            "'side=right'",
            <TODO>
        ),
        # non-default side should implicitly use square endcap
        (
            "POLYGON ((50 50, 50 150, 150 150, 150 50, 50 50))",
            20,
            "'side=left'",
            <TODO>,
        ),
        # non-default side and non-default flat endcap should not use square endcap
        # Specifying flat here doesn't actually make a difference. I found it hard to find a good test case here
        (
            "POLYGON ((50 50, 50 150, 150 150, 150 50, 50 50))",
            20,
            "'side=right endcap=flat'",
            <TODO>,
        ),
        # explicitly specifying default side=both should not set endcap=square
        (
            "LINESTRING (50 50, 150 150, 150 50)",
            100,
            "'side=both'",
            <TODO>,
        ),
```

Here are the results I got, the current ode currently fails the first 3, while it passes the 4th. Currently, around ~2x off from the correct answer on the wrong ones.
1. Sedona: 35847.55494469865. PostGIS: 16285.07633336958
2. Sedona: 10000.0 PostGIS: 19248.578060903223
3. Sedona: 10000.0 PostGIS: 3600.0
4. Sedona: 69888.08929186598  PostGIS: 69888.089291866

The 3rd or 4th cases tests cases that are easy to overlook while implementing the logic (each for different reasons). Just make sure you follow the comments I left there. It's not hard, but it's easy to miss. This is the exact bug that the PR I mentioned above was fixing, so consider checking out the java implementation there.

c/sedona-geos/src/st_buffer.rs
```diff
@@ -54,65 +67,191 @@ impl SedonaScalarKernel for STBuffer {
...
+                "Missing value for buffer parameter: {param}",
+            )));
+        };
+
+        if key.eq_ignore_ascii_case("endcap") {
+            params_builder = params_builder.end_cap_style(parse_cap_style(value)?);
+        } else if key.eq_ignore_ascii_case("join") {
+            params_builder = params_builder.join_style(parse_join_style(value)?);
+        } else if key.eq_ignore_ascii_case("side") {
+            params_builder = params_builder.single_sided(is_single_sided(value)?);
```
petern48: We're still missing this logic:
https://github.com/apache/sedona/blame/d6ea87cf9f2a27adee00ff0fb41ee92002b9a834/common/src/main/java/org/apache/sedona/common/Functions.java#L399-L402

Here's a better docstring to follow because the PostGIS docs hasn't been updated yet.

- `side=both|left|right` : Defaults to `both`. Setting `left` or `right` enables a single-sided buffer operation on the geometry, with the buffered side aligned according to the direction of the line. This functionality is specific to LINESTRING geometry and has no impact on POINT or POLYGON geometries. By default, square end caps are applied when `left` or `right` are specified.

Also, when I was experimenting with the python tests, they didn't behave as I thought they would. While I expected `side=right` to fail, I thought `side=right endcap=square` would pass, but the result was the same. I think it's possible there's another bug somewhere. Though I wouldn't worry about this for now. Just implement the logic for `side`, and we'll see if there's still strange behavior.

c/sedona-geos/src/st_buffer.rs
```diff
@@ -163,4 +302,203 @@ mod tests {
...
+    fn test_parse_buffer_params_duplicate_params_no_error() {
+        let result = parse_buffer_params(Some("endcap=round endcap=flat"));
+        assert!(result.is_ok());
+    }
+
+    #[test]
+    fn test_parse_buffer_params_quad_segs_out_of_range() {
+        let result = parse_buffer_params(Some("quad_segs=-5"));
+        assert!(result.is_ok());
+    }
```
petern48, line 537: It would be nice to write tests that directly check the values set for the params object (rather than testing the resulting geometries). Particularly, it would test the cases I mentioned in the Python tests. Here they are again for convenience.

```
1. non-default side and should implicitly use square endcap
2. non-default side should implicitly use square endcap
3. non-default side and non-default flat endcap should not use square endcap
4. explicitly specifying default side=both should not set endcap=square
```

I know this PR already feels long, so if you're not up for it, feel free to create a new issue for you or someone to do later. As long as the correct behavior is implemented and we have some python tests for it, I'm good to merge it.

python/sedonadb/tests/functions/test_functions.py
Abeeujah: These test cases fail because of the following as identified

**GEOS Version Incompatibility:**

- PostGIS uses GEOS 3.9.0 (includes geometric artifacts in single-sided buffers)
- SedonaDB uses GEOS 3.13.1 (artifacts removed - more geometrically correct)

**Key Finding**
In GEOS 3.12.0 (May 2023), major improvements were made to single-sided buffers:

- [Commit 091f6d99](https://github.com/libgeos/geos/commit/091f6d99): "Only keep biggest polygon from single-sided buffer"
- This removes spurious artifacts that were incorrectly included in buffer results
- **Result:** GEOS 3.13 produces smaller, more accurate areas than GEOS 3.9

c/sedona-geos/src/st_buffer.rs
Abeeujah: This has now been added

c/sedona-geos/src/st_buffer.rs
Abeeujah: BufferParams don't expose getter methods, nor implement traits that makes them easily testable, so I went ahead with testing behaviors instead of the values

python/sedonadb/tests/functions/test_functions.py
petern48: Wow. You did an amazing job investigating. Glad to here you figured out what the other mystery was. I see [here](https://hub.docker.com/r/postgis/postgis) that there are indeed Docker images that support `geos` 13. I gave `postgis/postgis:18-3.6` a try locally, and the tests pass. It's the only one that falls under the ([recommended debian section](https://hub.docker.com/r/postgis/postgis#debian-based-recommended)). How about you just update it [here](https://github.com/apache/sedona-db/blob/0551583a675da1cf5cc61cb7c7a61352051f1fc6/compose.yml#L20).

```
image: postgis/postgis:18-3.6
```

WDYT @paleolimbot, is this a reasonable change? It ensures our `geos` version matches the container's. They say "This Docker-PostGIS image has a cautious release cycle to guarantee high stability", so it seems safe to use to me.

c/sedona-geos/src/st_buffer.rs
```diff
@@ -54,65 +67,216 @@ impl SedonaScalarKernel for STBuffer {
...
+    let buffer_style_params = extract_optional_string(args.get(2))?;
+
+    // Build BufferParams based on style parameters
+    let params = parse_buffer_params(buffer_style_params.as_deref())?;
+
+    let is_single_sided = buffer_style_params
+        .as_ref()
+        .map(|s| s.eq_ignore_ascii_case("side="))
+        .unwrap_or(false);
+
```
petern48:
```suggestion
```

This condition for `is_single_sided` isn't quite right. If we put `side=both`, this would incorrectly return true. Looking a few lines below. There's only one place where this variable is used, and it's unnecessary (the rest of the condition guarantees correctness for us), so let's remove this variable.

c/sedona-geos/src/st_buffer.rs
```diff
@@ -54,65 +67,216 @@ impl SedonaScalarKernel for STBuffer {
...
+    let is_right = buffer_style_params
+        .as_ref()
+        .map(|s| s.eq_ignore_ascii_case("right"))
+        .unwrap_or(false);
+
+    executor.execute_wkb_void(|wkb| {
+        match (wkb, distance_iter.next().unwrap()) {
+            (Some(wkb), Some(mut distance)) => {
+                if is_single_sided && ((is_left && distance < 0.0) || (is_right && distance > 0.0))
```
petern48:
```suggestion
                if (is_left && distance < 0.0) || (is_right && distance > 0.0)
```

`is_single_sided` here is unnecessary since it's guaranteed to be true if `is_left` or `is_right` here is true.

c/sedona-geos/src/st_buffer.rs
```diff
@@ -54,65 +67,216 @@ impl SedonaScalarKernel for STBuffer {
...
+) -> Result<Vec<u8>> {
     let geometry = geos_geom
         .buffer_with_params(distance, params)
         .map_err(|e| DataFusionError::External(Box::new(e)))?;
     let wkb = geometry
         .to_wkb()
         .map_err(|e| DataFusionError::Execution(format!("Failed to convert to wkb: {e}")))?;
 
-    writer.write_all(wkb.as_ref())?;
-    Ok(())
```
petern48, line 115: While it looks odd, we actually want to keep the pattern here where we write the bytes directly into the `writer` buffer instead of returning it from the function. It's faster this way. We follow this pattern for any function that returns a geometry object. See my other comment above in the file to see the other half of this to reverse. I originally did the same thing when I first wrote this function, but dewey caught me.

Here's what the caller code should look like. I'm referencing the existing code (before this PR) here.

https://github.com/apache/sedona-db/blob/0551583a675da1cf5cc61cb7c7a61352051f1fc6/c/sedona-geos/src/st_buffer.rs#L88-L89

python/sedonadb/tests/functions/test_functions.py
petern48: Wow. You did an amazing job investigating. Glad to here you figured out what the other mystery was. I see [here](https://hub.docker.com/r/postgis/postgis) that there are indeed Docker images that support `geos` 13. I gave `postgis/postgis:18-3.6` a try locally, and the tests pass. It's the only one that falls under the ([recommended debian section](https://hub.docker.com/r/postgis/postgis#debian-based-recommended)). How about you just update it [here](https://github.com/apache/sedona-db/blob/0551583a675da1cf5cc61cb7c7a61352051f1fc6/compose.yml#L20).

```
image: postgis/postgis:18-3.6
```

WDYT @paleolimbot, is this a reasonable change? It ensures our `geos` version matches the container's. They say "This Docker-PostGIS image has a cautious release cycle to guarantee high stability", so it seems safe to use to me.

c/sedona-geos/src/st_buffer.rs
```diff
@@ -592,4 +606,93 @@ mod tests {
...
+            parse_buffer_side_style(Some("side=center")),
+            (false, false),
+            "Should ignore invalid side values."
+        );
+    }
+
+    #[test]
+    fn test_both_sides_present() {
+        assert_eq!(
+            parse_buffer_side_style(Some("side=left side=right")),
```
petern48, line 683: nice test case!

Abeeujah: Thank you!!

Abeeujah:
Great job from you all with the review so far!!!

python/sedonadb/tests/functions/test_functions.py
paleolimbot: Thank you for investigating! We will have to remember to keep this up-to-date with PostGIS/Postgres releases but I think that is OK and helps predictability.

PR #242, opened 2025-10-25
chore(ci): Turn off fail-fast in rust.yml

petern48:
See discussion here: https://github.com/apache/sedona-db/pull/239#issuecomment-3443914706

tldr; It can be easy to forget to run clippy since it's not included in pre-commit. Contributors already have to wait for someone to trigger CI for them. We should avoid failing fast, so we don't have to wait an extra CI cycle to see CI results.


petern48:
Works 👍 

image

PR #243, opened 2025-10-25
feat(c/sedona-geos): Implement ST_SimplifyPreserveTopology using geos

joonaspessi:
Implements `ST_SimplifyPreserveTopology(geometry, tolerance)` function using geos.

**Notice about ST_Simplify**

Actually started the implementation of `ST_Simplify` but noticed that the PostGIS has different behavior for collapsed geometries and collection types in comparison to [GEOS implementation](https://libgeos.org/doxygen/classgeos_1_1simplify_1_1DouglasPeuckerSimplifier.html). 

Achieving full PostGIS compatibility (including all the params) would probably require custom Rust implementation.

- Both PostGIS and Sedona-DB uses now GEOS for ST_SimplifyPreserveTopology
    - https://github.com/postgis/postgis/blob/master/postgis/lwgeom_geos.c#L935 - Entry point
    - https://github.com/postgis/postgis/blob/master/postgis/lwgeom_geos.c#L966 - GEOS function call
- PostGIS uses custom implementation for ST_Simplify
   - https://github.com/postgis/postgis/blob/master/postgis/lwgeom_functions_analytic.c#L58 - Entry point
   - https://github.com/postgis/postgis/blob/master/liblwgeom/lwgeom.c#L1851 - Core custom implementation

This might be actually nice function to be ported into Rust.

### DIFF
--- a/c/sedona-geos/src/st_simplifypreservetopology.rs
+++ b/c/sedona-geos/src/st_simplifypreservetopology.rs
@@ -19,6 +19,7 @@ use std::sync::Arc;
 
 use arrow_array::builder::BinaryBuilder;
 use arrow_schema::DataType;
+use datafusion_common::cast::as_float64_array;
 use datafusion_common::error::Result;
 use datafusion_common::DataFusionError;
 use datafusion_expr::ColumnarValue;
@@ -55,35 +56,28 @@ impl SedonaScalarKernel for STSimplifyPreserveTopology {
         arg_types: &[SedonaType],
         args: &[ColumnarValue],
     ) -> Result<ColumnarValue> {
-        let tolerance: Option<f64>;
-        let arg1 = args[1].cast_to(&DataType::Float64, None)?;
-        if let ColumnarValue::Scalar(scalar_arg) = &arg1 {
-            if scalar_arg.is_null() {
-                tolerance = None;
-            } else {
-                tolerance = Some(f64::try_from(scalar_arg.clone())?);
-            }
-        } else {
-            return Err(DataFusionError::Execution(format!(
-                "Invalid tolerance: {:?}",
-                args[1]
-            )));
-        }
-
         let executor = GeosExecutor::new(arg_types, args);
+
+        let tolerance_value = args[1]
+            .cast_to(&DataType::Float64, None)?
+            .to_array(executor.num_iterations())?;
+        let tolerance_array = as_float64_array(&tolerance_value)?;
+
+        let mut tolerance_iter = tolerance_array.iter();
+
         let mut builder = BinaryBuilder::with_capacity(
             executor.num_iterations(),
             WKB_MIN_PROBABLE_BYTES * executor.num_iterations(),
         );
+
         executor.execute_wkb_void(|wkb| {
-            match (wkb, tolerance) {
+            match (wkb, tolerance_iter.next().unwrap()) {
                 (Some(wkb), Some(tolerance)) => {
                     invoke_scalar(&wkb, tolerance, &mut builder)?;
                     builder.append_value([]);
                 }
                 _ => builder.append_null(),
             }
-
             Ok(())
         })?;
 
@@ -198,6 +192,42 @@ mod tests {
         assert_array_equal(
             &tester.invoke_wkb_array_scalar(input_wkt_t20, 20.0).unwrap(),
             &expected_t20,
+        );
+
+        // Test array tolerance input - same geometry with different tolerance values
+        let input_wkt_array = vec![
+            Some("LINESTRING (0 0, 0 10, 0 51, 50 20, 30 20, 7 32)"),
+            Some("LINESTRING (0 0, 0 10, 0 51, 50 20, 30 20, 7 32)"),
+            Some("LINESTRING (0 0, 0 10, 0 51, 50 20, 30 20, 7 32)"),
+            None,
+        ];
+
+        use arrow_array::Float64Array;
+        let tolerance_array: Arc<Float64Array> = Arc::new(Float64Array::from(vec![
+            Some(2.0),
+            Some(10.0),
+            Some(50.0),
+            Some(5.0),
+        ]));
+
+        let expected_array = create_array(
+            &[
+                Some("LINESTRING (0 0, 0 51, 50 20, 30 20, 7 32)"), // Tolerance 2.0
+                Some("LINESTRING (0 0, 0 51, 50 20, 7 32)"),        // Tolerance 10.0
+                Some("LINESTRING (0 0, 7 32)"),                     // Tolerance 50.0
+                None,
+            ],
+            &WKB_GEOMETRY,
+        );
+
+        assert_array_equal(
+            &tester
+                .invoke_arrays(vec![
+                    create_array(&input_wkt_array, &sedona_type),
+                    tolerance_array,
+                ])
+                .unwrap(),
+            &expected_array,
         );
     }
 }


python/sedonadb/tests/functions/test_functions.py
```diff
@@ -1399,3 +1399,49 @@ def test_st_isvalidreason(eng, geom, expected):
...
+        ("LINESTRING (0 0, 0 10)", None, None),
+        # Empty geometries
+        ("LINESTRING EMPTY", 2, "LINESTRING EMPTY"),
+        ("POINT EMPTY", 2, "POINT EMPTY"),
+        ("POLYGON EMPTY", 2, "POLYGON EMPTY"),
+        # POLYGON with hole - topology preserved, both rings simplified
+        (
+            "POLYGON ((0 0, 10 0, 10 10, 0 10, 0 0),(5 5, 5 6, 6 6, 8 5, 5 5))",
+            20,
+            "POLYGON((0 0,10 0,10 10,0 10,0 0),(5 5,5 6,8 5,5 5))",
```
petern48: nit: The comment sounds misleading. The inner ring is simplified (removing the `6 6` coordinate), but the outer ring is left the same.

It would also be helpful if we kept the spacing consistent between the input and the expected string (this applies to all of these test cases). It visually makes it easier for readers to compare the geometries and see if there's a missing coordinate or not, since our eyes can just look up-down. I found myself "restarting" my comparison one or twice after losing track of which coordinate I was on, since my eyes had to track diagonally between the top and bottom.

python/sedonadb/tests/functions/test_functions.py
```diff
@@ -1399,3 +1399,49 @@ def test_st_isvalidreason(eng, geom, expected):
...
+            "MULTIPOLYGON (((100 100, 100 130, 130 130, 130 100, 100 100)),((0 0, 10 0, 10 10, 0 10, 0 0),(5 5, 5 6, 6 6, 8 5, 5 5)))",
+            20,
+            "MULTIPOLYGON(((100 100,100 130,130 130,130 100,100 100)),((0 0,10 0,10 10,0 10,0 0),(5 5,5 6,8 5,5 5)))",
+        ),
+    ],
+)
+def test_st_simplifypreservetopology(eng, geom, tolerance, expected):
+    eng = eng.create_or_skip()
+    eng.assert_query_result(
+        f"SELECT ST_AsText(ST_SimplifyPreserveTopology({geom_or_null(geom)}, {val_or_null(tolerance)}))",
```
petern48: optional: I was curious why `ST_AsText()` was there, so I tried removing it myself. We can actually get this to pass without it. After removing the `ST_AsText()`, the cases fail because your spacing is a bit off. It should pass if you add a space after each comma (`,`) and after the polygon name. For example, `LINESTRING(0 0,0 10)` would become `LINESTRING (0 0, 0 10)`. Another benefit is readability from having consistent spacing, as I mentioned above.

The one caveat is `POINT EMPTY`. In that case, we can set expected to `POINT (nan nan)` because that's how geoarrow-c renders it.

c/sedona-geos/src/register.rs
```diff
@@ -57,12 +59,14 @@ pub fn scalar_kernels() -> Vec<(&'static str, ScalarKernelRef)> {
...
+            "st_simplifypreservetopology",
+            st_simplify_preserve_topology_impl(),
+        ),
         ("st_symdifference", st_sym_difference_impl()),
         ("st_touches", st_touches_impl()),
         ("st_unaryunion", st_unary_union_impl()),
         ("st_union", st_union_impl()),
         ("st_within", st_within_impl()),
-        ("st_crosses", st_crosses_impl()),
-        ("st_overlaps", st_overlaps_impl()),
```
petern48, line 66: Let's add these function registrations back. We shouldn't be removing any existing functionality.

python/sedonadb/tests/functions/test_functions.py
joonaspessi: Thanks for the suggestion, the consistent spacing indeed made the test case much more easier to follow and understand. Also fixed that comment that was wrongly stating that both outer and inner rings would be simplified.

python/sedonadb/tests/functions/test_functions.py
joonaspessi: Noticed this inconsistency and thought that `ST_AsText` seems to normalize the test output. But I agree, it's better to keep all test cases aligned with similar approach.

c/sedona-geos/src/register.rs
joonaspessi: Noticed that these were actual duplicated.

c/sedona-geos/src/register.rs
```diff
@@ -57,12 +59,14 @@ pub fn scalar_kernels() -> Vec<(&'static str, ScalarKernelRef)> {
         ("st_length", st_length_impl()),
         ("st_overlaps", st_overlaps_impl()),
```
joonaspessi, line 60: Already defined here

c/sedona-geos/src/st_simplifypreservetopology.rs
```diff
@@ -0,0 +1,203 @@
...
+                tolerance = None;
+            } else {
+                tolerance = Some(f64::try_from(scalar_arg.clone())?);
+            }
+        } else {
+            return Err(DataFusionError::Execution(format!(
+                "Invalid tolerance: {:?}",
+                args[1]
+            )));
+        }
```
paleolimbot: While you're here, I think it would be good to support this as an array input as well (it's roughly the same amount of code). The example I wrote up for `ST_Buffer()` was:

```rust
let distance_value = args[1].cast_to(&DataType::Float64)?.to_array(executor.num_iterations());
let distance_array = if let ColumnarValue::Array(array_ref) = distance_value {
  as_float64_array(array_ref)?
} else {
  return sedona_internal_err!("Unexpected scalar value");
};

let mut distance_iter = distance_array.iter();

// ...

executor.execute_wkb_void(|wkb| {
   match (wkb, distance_iter.next().unwrap() {
       (Some(wkb_item), Some(distance_item)) => { invoke_scalar(...) }
       _ => { /* append null */ }
   }
})
```

paleolimbot (commented):
Thank you!

> Achieving full PostGIS compatibility (including all the params) would probably require custom Rust implementation.

Good call...our GEOS list is supposed to be functions that are easy to implement and we've tried hard so far to ensure that when we do add a function we do it all the way.

I wonder if there is a Rust `geo` implementation of Simplify that checks these boxes already...in any case I'll open up a dedicated issue for this since it seems like it will be more involved. Thank you for identifying!

c/sedona-geos/src/st_simplifypreservetopology.rs
joonaspessi: Thanks for pointing this out, nice that the array input was relatively easy to handle with ColumnarValue. Implementation is now updated with this change.

Btw. Does this introduce some room for performance/memory improvement? With this approach and using scalar value, we will copy it for all geometries, potentially quite many times. 

Thinking if these tracks should be separated, or might be just micro optimization...

```rust
let executor = GeosExecutor::new(arg_types, args);
let tolerance_arg = args[1].cast_to(&DataType::Float64, None)?;

let mut builder = BinaryBuilder::with_capacity(
    executor.num_iterations(),
);

match tolerance_arg {
    ColumnarValue::Scalar(scalar_val) => {
        // SCALAR PATH: Extract scalar once, reuse for all geometries
       ...
    }
    ColumnarValue::Array(_) => {
        // ARRAY PATH: Use iterator for per-row tolerances
       ...
    }
}

```

paleolimbot (approved):
Thank you!

Issue #247, opened 2025-10-27
Design and Prototype Memory Model


jesspav:
Hi folks,

Pulled together requirements and a design for a memory model for an efficient, flexible in-memory model for raster data that aligns with SedonaDB's architecture.  Also built a quick prototype with the designed schema, as well as some accessors and some functions that use them.  

Looking forward to your feedback.

### Raster Metadata
Standard raster metadata fields in Sedona, Havasu, and similar systems:
- **Width**: Pixel count along X axis
- **Height**: Pixel count along Y axis
- **UpperleftX** / **UpperleftY**: Upper-left corner coordinates (CRS units)
- **ScaleX** / **ScaleY**: Cell scaling factors
- **SkewX** / **SkewY**: Cell skew parameters

### Bounding Box
- Optional WGS84 bounding box essential for speeding up spatial queries

### CRS
- Full CRS info (including SRID)

### Raster Bands
Bands store metadata and, for in-db, the data itself.  We may want to expand this to include statistical data about the bands as well.

- **NoDataValue**: The no data value
- **Storage Types**:
  - OutDB Ref: External reference
  - InDB: In-memory values on the band
  - InDb Reference: Memory pool storage (outside Arrow array) - not in the initial release
- **Data Types**: Standard data types including UInt8, Int32, Float32, etc. Initial version will not include complex types, but we would like to include these in the future
- **Out Db Metadata:**  URL + band id
- **Compression Type:** The compression type of the band
- **Data** 

**Structure:**
```
Raster
├─ Metadata
├─ BBox
├─ CRS
└─ Bands
    └─ Band
        ├─ Metadata
        ├─ Statistics (optional)
        └─ Data
```

## Apache Arrow Arrays
SedonaDB leverages Apache Arrow for speed and efficiency:
- Immutable: Metadata updates require array copies
- Columnar: Fast metadata queries
- Typed: Runtime validation
- Zero-Copy: Language interoperability
- Null Handling: Efficient bitmaps
- Batch/Vectorized: SIMD-ready operations

**StructArrays:**
- Struct arrays separate fields as independent child arrays, enabling flexible queries and future-proofing.

See the schema prototype using StructArrays: [sedona-schema/src/datatypes.rs#L368-L481](https://github.com/jesspav/sedona-db/blob/prototype_raster/rust/sedona-schema/src/datatypes.rs#L368-L481)

### Access Patterns
[...]

PR #249, opened 2025-10-27
Prototype raster

jesspav:
Simple prototype code to demonstrate viability of a StructArray data type for rasters (issue: #247 ).

If we are happy with the basic ideas here, we will break this up into many small PRs so we can get down to the nitty gritty. :)

### DIFF
--- a/rust/sedona-gdal/src/readers.rs
+++ b/rust/sedona-gdal/src/readers.rs
@@ -26,14 +26,13 @@ use sedona_schema::datatypes::{
 };
 use std::sync::Arc;
 
-pub fn read_raster(
-    filepath: &str,
-) -> Result<Arc<StructArray>, ArrowError> {
+pub fn read_raster(filepath: &str) -> Result<Arc<StructArray>, ArrowError> {
     let dataset = Dataset::open(filepath.to_string())
         .map_err(|err| ArrowError::ParseError(err.to_string()))?;
 
     // Extract geotransform components
-    let (origin_x, origin_y, pixel_width, pixel_height, rotation_x, rotation_y) = geotransform_components(&dataset)?;   
+    let (origin_x, origin_y, pixel_width, pixel_height, rotation_x, rotation_y) =
+        geotransform_components(&dataset)?;
 
     let (raster_width, raster_height) = dataset.raster_size();
 
@@ -158,18 +157,18 @@ mod tests {
 
     #[test]
     fn test_load_raster() {
-       // TODO: Add proper tests here. 
-       // To load a raster and view contents
-       // for prototyping fun:
-       //
-       // let filepath = "<your_path>/test1.tiff";
-       // let result = read_raster(filepath);
-       // assert!(result.is_ok());
-       //
-       // To view loaded raster:
-       // let raster_array = result.unwrap();
-       // for raster in raster_iterator(&raster_array) {
-       //    println!("{}", pretty_print_indb(&raster, 1, 2).unwrap());
-       // }
+        // TODO: Add proper tests here.
+        // To load a raster and view contents
+        // for prototyping fun:
+        //
+        // let filepath = "<your_path>/test1.tiff";
+        // let result = read_raster(filepath);
+        // assert!(result.is_ok());
+        //
+        // To view loaded raster:
+        // let raster_array = result.unwrap();
+        // for raster in raster_iterator(&raster_array) {
+        //    println!("{}", pretty_print_indb(&raster, 1, 2).unwrap());
+        // }
     }
 }
--- a/rust/sedona-gdal/src/rs_value.rs
+++ b/rust/sedona-gdal/src/rs_value.rs
@@ -16,6 +16,7 @@
 // under the License.
 use std::sync::Arc;
 
+use crate::dataset::outdb_dataset;
 use arrow_array::builder::Float64Builder;
 use arrow_schema::{ArrowError, DataType};
 use datafusion_common::{error::Result, scalar::ScalarValue};
@@ -24,7 +25,6 @@ use sedona_expr::scalar_udf::{ScalarKernelRef, SedonaScalarKernel};
 use sedona_functions::executor::RasterExecutor;
 use sedona_raster::datatype_functions::{bytes_per_pixel, read_pixel_value};
 use sedona_schema::datatypes::{BandMetadataRef, BandRef, RasterRef, SedonaType, StorageType};
-use crate::dataset::outdb_dataset;
 
 /// RS_Value() implementation
 pub fn rs_value_impl() -> ScalarKernelRef {
@@ -147,8 +147,6 @@ fn outdb_pixel(
     _width: usize,
     _height: usize,
 ) -> Result<f64, ArrowError> {
-
-
     let dataset = outdb_dataset(metadata)?;
 
     let band_number = match metadata.outdb_band_id() {
--- a/rust/sedona-raster/src/display_functions.rs
+++ b/rust/sedona-raster/src/display_functions.rs
@@ -18,18 +18,19 @@
 use crate::datatype_functions::{bytes_per_pixel, read_pixel_value};
 use arrow::error::ArrowError;
 use sedona_schema::datatypes::{RasterRef, RasterRefImpl, StorageType};
+use std::fmt::Write;
 
-/// Pretty print a raster band to a string with specified precision
-pub fn pretty_print_indb(
+/// Write raster band matrix directly to a StringBuilder with specified precision
+pub fn write_band_to_builder(
     raster: &RasterRefImpl,
     band_number: usize,
-    precision: usize, // TODO: change this to an optional format string
-) -> Result<String, ArrowError> {
+    precision: usize,
+    out: &mut impl Write,
+) -> Result<(), ArrowError> {
     let band = raster.bands().band(band_number).unwrap();
     let metadata = raster.metadata();
     let height = metadata.height() as usize;
     let width = metadata.width() as usize;
-    let mut result = String::new();
 
     let slice = band.data() as &[u8];
     let data_type = band.metadata().data_type();
@@ -39,6 +40,7 @@ pub fn pretty_print_indb(
         ));
     }
     let bytes_per_pixel = bytes_per_pixel(data_type.clone()).unwrap_or(1);
+
     for row in 0..height {
         for col in 0..width {
             let start = (row * width + col) * bytes_per_pixel;
@@ -46,19 +48,23 @@ pub fn pretty_print_indb(
             let pixel_bytes = &slice[start..end];
 
             match read_pixel_value(pixel_bytes, data_type.clone()) {
-                Ok(value) => result.push_str(&format!("{:8.*} ", precision, value)),
-                Err(_) => result.push_str(&format!("{:>8} ", "?")), // Well-spaced question mark
+                Ok(value) => {
+                    out.write_fmt(format_args!("{:8.*} ", precision, value))
+                        .unwrap();
+                }
+                Err(_) => out.write_fmt(format_args!("{:>8} ", "?")).unwrap(),
             }
         }
-        result.push('\n');
+        out.write_fmt(format_args!("\n")).unwrap();
     }
 
-    Ok(result)
+    Ok(())
 }
 
 #[cfg(test)]
 mod tests {
     use super::*;
+    use arrow::array::StringBuilder;
     use sedona_schema::datatypes::{
         BandDataType, BandMetadata, RasterBuilder, RasterMetadata, StorageType,
     };
@@ -101,9 +107,12 @@ mod tests {
         let raster_struct = raster_builder.finish().unwrap();
         let raster = sedona_schema::datatypes::RasterRefImpl::new(&raster_struct, 0);
 
-        let pretty = pretty_print_indb(&raster, 0, 2).unwrap();
+        let builder = StringBuilder::new();
+        write_band_to_builder(&raster, 1, 2, &mut builder).unwrap();
 
+        let binding = builder.finish();
+        let result = binding.value(0);
         let expected = "    1.10     2.20     3.30 \n    4.40     5.50     6.61 \n";
-        assert_eq!(pretty, expected);
+        assert_eq!(result, expected);
     }
 }
--- a/rust/sedona-schema/src/datatypes.rs
+++ b/rust/sedona-schema/src/datatypes.rs
@@ -104,9 +104,12 @@ impl SedonaType {
 
     /// Given an [`ExtensionType`], construct a SedonaType
     pub fn from_extension_type(extension: ExtensionType) -> Result<SedonaType> {
-        let (edges, crs) = deserialize_edges_and_crs(&extension.extension_metadata)?;
         if extension.extension_name == "geoarrow.wkb" {
+            let (edges, crs) = deserialize_edges_and_crs(&extension.extension_metadata)?;
             sedona_type_wkb(edges, crs, extension.storage_type)
+        } else if extension.extension_name == "sedona.raster" {
+            // For raster extension types, return the RASTER constant
+            Ok(RASTER)
         } else {
             sedona_internal_err!(
                 "Extension type not implemented: <{}>:{}",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4967,6 +4967,7 @@ dependencies = [
  "sedona-common",
  "sedona-expr",
  "sedona-geometry",
+ "sedona-raster",
  "sedona-schema",
  "sedona-testing",
  "serde_json",
@@ -4982,6 +4983,7 @@ dependencies = [
  "arrow",
  "arrow-array",
  "arrow-schema",
+ "criterion",
  "datafusion-common",
  "datafusion-expr",
  "gdal",
@@ -4991,6 +4993,7 @@ dependencies = [
  "sedona-functions",
  "sedona-raster",
  "sedona-schema",
+ "sedona-testing",
 ]
 
 [[package]]
--- a/rust/sedona-functions/Cargo.toml
+++ b/rust/sedona-functions/Cargo.toml
@@ -44,6 +44,7 @@ geo-traits = { workspace = true }
 sedona-common = { path = "../sedona-common" }
 sedona-expr = { path = "../sedona-expr" }
 sedona-geometry = { path = "../sedona-geometry" }
+sedona-raster = { path = "../sedona-raster" }
 sedona-schema = { path = "../sedona-schema" }
 wkb = { workspace = true }
 wkt = { workspace = true }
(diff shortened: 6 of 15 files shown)


rust/sedona-raster/src/display_functions.rs
```diff
@@ -0,0 +1,118 @@
...
+    let band = raster.bands().band(band_number).unwrap();
+    let metadata = raster.metadata();
+    let height = metadata.height() as usize;
+    let width = metadata.width() as usize;
+
+    let slice = band.data() as &[u8];
+    let data_type = band.metadata().data_type();
+    if band.metadata().storage_type() != StorageType::InDb {
+        return Err(ArrowError::InvalidArgumentError(
+            "Pretty print indb not supported for non-InDb storage".to_string(),
```
Kontinuation, line 39: `Pretty print not supported for non-InDb storage`

rust/sedona-raster/src/datatype_functions.rs
```diff
@@ -0,0 +1,132 @@
...
+    match data_type {
+        BandDataType::UInt8 => Ok(1),
+        BandDataType::Int16 => Ok(2),
+        BandDataType::UInt16 => Ok(2),
+        BandDataType::Int32 => Ok(4),
+        BandDataType::UInt32 => Ok(4),
+        BandDataType::Float32 => Ok(4),
+        BandDataType::Float64 => Ok(8),
+    }
+}
```
Kontinuation, line 31: We don't need to wrap the return type with `Result` if the match is always exhaustive.

rust/sedona-raster/src/display_functions.rs
```diff
@@ -0,0 +1,118 @@
...
+    for row in 0..height {
+        for col in 0..width {
+            let start = (row * width + col) * bytes_per_pixel;
+            let end = start + bytes_per_pixel;
+            let pixel_bytes = &slice[start..end];
+
+            match read_pixel_value(pixel_bytes, data_type.clone()) {
+                Ok(value) => {
+                    out.write_fmt(format_args!("{:8.*} ", precision, value))
+                        .unwrap();
```
Kontinuation, line 53: The implementation of `Write` trait may fail writing bytes due to lots of reasons, and the caller may want to handle some of them. We may want to propagate this error using `?` instead of `.unwrap()`.

rust/sedona-schema/src/datatypes.rs
```diff
@@ -333,6 +365,1241 @@ fn deserialize_edges(edges: &Value) -> Result<Edges> {
...
+        buffer.resize(capacity, 0);
+
+        let commit = move |buffer: MutableBuffer| {
+            // Convert MutableBuffer to &[u8] and append to BinaryBuilder
+            let data = buffer.as_slice();
+            self.band_data_writer().append_value(data);
+        };
+
+        (buffer, commit)
+    }
```
Kontinuation, line 598: We already have another method `commit_band_buffer`, so we don't have to return a committer with the buffer.

rust/sedona-schema/src/datatypes.rs
```diff
@@ -333,6 +365,1241 @@ fn deserialize_edges(edges: &Value) -> Result<Edges> {
...
+        buffer.resize(size, 0);
+
+        // Get mutable slice that GDAL can write to
+        let slice = unsafe {
+            // This is safe because we just allocated the buffer with the exact size
+            std::slice::from_raw_parts_mut(buffer.as_mut_ptr(), size)
+        };
+
+        (buffer, slice)
+    }
```
Kontinuation, line 614: This method has incorrect lifetime notation: the lifetime of the returned &mut [u8] is actually related to the lifetime of the returned `MutableBuffer` object, and not related to `self` (RasterBuilder).

We can simply remove this method since we already have `create_band_buffer`. The holder of the `MutableBuffer` can simply call `as_slice_mut` if they want a `&mut [u8]` from it.

rust/sedona-schema/src/datatypes.rs
```diff
@@ -333,6 +365,1241 @@ fn deserialize_edges(edges: &Value) -> Result<Edges> {
...
+
+        metadata_builder
+            .field_builder::<Float64Builder>(metadata_indices::SKEW_X)
+            .unwrap()
+            .append_value(metadata.skew_x());
+
+        metadata_builder
+            .field_builder::<Float64Builder>(metadata_indices::SKEW_Y)
+            .unwrap()
+            .append_value(metadata.skew_y());
```
Kontinuation, line 749: Calling `field_builder` for every column when inserting a struct value has huge performance overhead, we hit this problem in the initial vectorized engine POC for geometry types. However, I think we can keep it this way, since raster operations could be heavy weight enough to make it negligible.

rust/sedona-schema/src/datatypes.rs
```diff
@@ -333,6 +365,1241 @@ fn deserialize_edges(edges: &Value) -> Result<Edges> {
...
+                let binary_builder = self.band_data_writer();
+                write_bands(band_index, binary_builder)?
+            };
+            self.finish_band(band_metadata)?;
+        }
+
+        self.finish_raster()?;
+        Ok(())
+    }
+}
```
Kontinuation, line 930: The method signature looks cubersome and it is currently unused in the codebase. How it will be used by other raster functions?

rust/sedona-gdal/src/readers.rs
```diff
@@ -0,0 +1,174 @@
...
+        for tile_x in 0..x_tile_count {
+            let x_offset = tile_x * tile_width;
+            let y_offset = tile_y * tile_height;
+
+            // Calculate geographic coordinates for this tile
+            // using the geotransform from the original raster
+            let tile_origin_x =
+                origin_x + (x_offset as f64) * pixel_width + (y_offset as f64) * rotation_x;
+            let tile_origin_y =
+                origin_y + (x_offset as f64) * rotation_y + (y_offset as f64) * pixel_height;
```
Kontinuation, line 57: We can define a struct for affine transformation parameters and some utility functions for performing affine transformations.

rust/sedona-gdal/src/readers.rs
```diff
@@ -0,0 +1,174 @@
...
+                scale_y: pixel_height,
+                skew_x: rotation_x,
+                skew_y: rotation_y,
+                bounding_box: None, // TODO: should we calculate bounding box here?
+            };
+
+            raster_builder.start_raster(&tile_metadata, None, None)?;
+
+            for band_number in 1..=band_count {
+                let band: RasterBand = dataset.rasterband(band_number).unwrap();
```
Kontinuation, line 75: We should propagate the error instead of calling `.unwrap()`.

rust/sedona-gdal/src/readers.rs
```diff
@@ -0,0 +1,174 @@
...
+            let tile_metadata = RasterMetadata {
+                width: tile_width as u64,
+                height: tile_height as u64,
+                upperleft_x: tile_origin_x,
+                upperleft_y: tile_origin_y,
+                scale_x: pixel_width,
+                scale_y: pixel_height,
+                skew_x: rotation_x,
+                skew_y: rotation_y,
+                bounding_box: None, // TODO: should we calculate bounding box here?
```
Kontinuation, line 69: How about computing bounding_box lazily in `RasterMetadata`? We can define a constructor that takes an affine transformation as parameter and initialize bounding_box as None.

rust/sedona-gdal/src/readers.rs
```diff
@@ -0,0 +1,174 @@
...
+                let data_type = gdaldatatype_to_banddatatype(band.band_type())?;
+                let data_type_bytes = bytes_per_pixel(data_type.clone())?;
+                let buffer_size_bytes = x_size * y_size * data_type_bytes.clone();
+
+                // Get a mutable buffer slice for GDAL to write directly into
+                let (buffer, slice) = raster_builder.get_band_buffer_slice(buffer_size_bytes);
+
+                // TODO: Do we need resampling? If so set buffer_size to different from window_size
+                //       and have a ResampleAlgorithm.
+                band.read_into_slice(
```
Kontinuation, line 96: I suggest starting with no resampling

rust/sedona-schema/src/datatypes.rs
```diff
@@ -333,6 +365,1241 @@ fn deserialize_edges(edges: &Value) -> Result<Edges> {
...
+        } else {
+            None
+        }
+    }
+
+    fn size_hint(&self) -> (usize, Option<usize>) {
+        let remaining = self.raster_array.len().saturating_sub(self.current_row);
+        (remaining, Some(remaining))
+    }
+}
```
Kontinuation, line 1485: I'm uncertain should we implement `Iterator` trait for `RasterStructIterator`, since the only actual usage of this type is random access (`get(&self, index: usize)`).

rust/sedona-schema/src/datatypes.rs
```diff
@@ -333,6 +365,1241 @@ fn deserialize_edges(edges: &Value) -> Result<Edges> {
...
+            .column(raster_indices::BBOX)
+            .as_any()
+            .downcast_ref::<StructArray>()
+            .unwrap();
+
+        let bands_list = raster_struct
+            .column(raster_indices::BANDS)
+            .as_any()
+            .downcast_ref::<ListArray>()
+            .unwrap();
```
Kontinuation, line 1353: Performing all these `.column`, `.as_any` and `downcast_ref` for each raster could be a huge performance overhead, especially for queries that only fetches the metadata.

rust/sedona-schema/src/datatypes.rs
```diff
@@ -333,6 +365,1241 @@ fn deserialize_edges(edges: &Value) -> Result<Edges> {
...
+}
+
+impl<'a> MetadataRef for MetadataRefImpl<'a> {
+    fn width(&self) -> u64 {
+        self.metadata_struct
+            .column(metadata_indices::WIDTH)
+            .as_any()
+            .downcast_ref::<UInt64Array>()
+            .unwrap()
+            .value(self.index)
```
Kontinuation, line 1046: Calling `.column`, `.as_any` and `.downcast_ref` each time we access the metadata fields could also be a huge performance overhead. We may observe this when benchmarking `rs_width`.

rust/sedona-schema/src/datatypes.rs
```diff
@@ -333,6 +365,1241 @@ fn deserialize_edges(edges: &Value) -> Result<Edges> {
...
+
+    /// Commit a MutableBuffer to the band data
+    pub fn commit_band_buffer(&mut self, buffer: MutableBuffer) {
+        let data = buffer.as_slice();
+        self.band_data_writer().append_value(data);
+    }
+
+    /// Finish writing the current band with its metadata
+    /// TODO: The band_metadata is in the finish in the band call, but in the
+    /// start in the raster call. Make it consistent.
```
Kontinuation, line 624: I don't think it is a problem. `start_raster` takes raster metadata as pararameter, while `finish_band` takes band metadata as parameter.

rust/sedona-schema/src/datatypes.rs
```diff
@@ -333,6 +365,1241 @@ fn deserialize_edges(edges: &Value) -> Result<Edges> {
...
+            Field::new(column::DATATYPE, DataType::UInt32, false),
+            // OutDb reference fields - only used when storage_type == OutDbRef
+            Field::new(column::OUTDB_URL, DataType::Utf8, true),
+            Field::new(column::OUTDB_BAND_ID, DataType::UInt32, true),
+        ]))
+    }
+
+    /// Band data schema (single binary blob)
+    pub fn band_data_type() -> DataType {
+        DataType::Binary // consider switching to BinaryView
```
Kontinuation, line 440: I believe that `BinaryView` is a better option. Growing the binary array when appending in-db band data could be less painful.

rust/sedona-gdal/src/readers.rs
```diff
@@ -0,0 +1,174 @@
...
+                    None => None,
+                };
+
+                let band_metadata = BandMetadata {
+                    nodata_value: nodata_value,
+                    storage_type: StorageType::InDb,
+                    datatype: data_type,
+                    outdb_url: None,
+                    outdb_band_id: None,
+                };
```
Kontinuation, line 120: We can add a parameter to control if the loaded tiles are in-db or out-db, so that we can test basic functions of out-db rasters.

rust/sedona-schema/src/datatypes.rs
```diff
@@ -333,6 +365,1241 @@ fn deserialize_edges(edges: &Value) -> Result<Edges> {
...
+    }
+
+    /// CRS schema to store json representation
+    pub fn crs_type() -> DataType {
+        DataType::Utf8 // TODO: Consider Utf8View
+    }
```
paleolimbot, line 446: I think a Utf8View here is a good choice since this value will be frequently repeated, less than 12 bytes, or both!

rust/sedona-schema/src/datatypes.rs
```diff
@@ -29,6 +39,7 @@ pub enum SedonaType {
     Arrow(DataType),
     Wkb(Edges, Crs),
     WkbView(Edges, Crs),
+    Raster(RasterSchema),
```
paleolimbot, line 42: I think this can probably be unparameterized (i.e., `SedonaType::Raster`). My reading of this PR is that the underlying Arrow schema is a constant (so it can be cloned from RASTER_DATATYPE when requested by `storage_type()`).

rust/sedona-schema/src/datatypes.rs
```diff
@@ -333,6 +365,1241 @@ fn deserialize_edges(edges: &Value) -> Result<Edges> {
     }
 }
 
+/// Schema for storing raster data in Apache Arrow format.
+/// Utilizing nested structs and lists to represent raster metadata and bands.
+#[derive(Debug, PartialEq, Clone)]
+pub struct RasterSchema;
+impl RasterSchema {
```
paleolimbot, line 372: I think it might be nice to have this live in a dedicated `sedona-schema/src/raster.rs` (arguably we should move the Wkb types into their own file as well in some future 😬 )

rust/sedona-schema/src/datatypes.rs
```diff
@@ -333,6 +365,1241 @@ fn deserialize_edges(edges: &Value) -> Result<Edges> {
...
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub enum StorageType {
+    InDb = 0,
+    OutDbRef = 1,
+}
+
+/// Builder for constructing raster arrays with zero-copy band data writing
+pub struct RasterBuilder {
+    main_builder: StructBuilder,
+}
```
paleolimbot, line 486: Nice! I think your sedona-raster crate would be a good home for this one. I know our existing sedona-geometry doesn't depend on Arrow but I think that's fine for sedona-raster (we can move these things around if there's good reason to do so later on).

rust/sedona-functions/src/executor.rs
```diff
@@ -75,6 +75,116 @@ pub struct GenericExecutor<'a, 'b, Factory0, Factory1> {
...
+/// Helper for writing raster kernel implementations
+///
+/// The [RasterExecutor] provides a simplified interface for executing functions
+/// on raster arrays, handling the common pattern of downcasting to StructArray,
+/// creating raster iterators, and handling null values.
+pub struct RasterExecutor<'a, 'b> {
+    pub arg_types: &'a [SedonaType],
+    pub args: &'b [ColumnarValue],
+    num_iterations: usize,
+}
```
paleolimbot, line 87: I think having a `sedona-raster-functions` crate would make sense! (sedona-functions is already a little crowded and might benefit from being split even in its current form)

rust/sedona-gdal/src/rs_value.rs
```diff
@@ -0,0 +1,443 @@
...
+use datafusion_expr::ColumnarValue;
+use sedona_expr::scalar_udf::{ScalarKernelRef, SedonaScalarKernel};
+use sedona_functions::executor::RasterExecutor;
+use sedona_raster::datatype_functions::{bytes_per_pixel, read_pixel_value};
+use sedona_schema::datatypes::{BandMetadataRef, BandRef, RasterRef, SedonaType, StorageType};
+
+/// RS_Value() implementation
+pub fn rs_value_impl() -> ScalarKernelRef {
+    Arc::new(RSValue {})
+}
```
paleolimbot, line 32: This is probably something we want to be an async scalar function (which we'd need to implement a Sedona wrapper for to get the kernel dispatch thing working): https://datafusion.apache.org/library-user-guide/functions/adding-udfs.html#adding-an-async-scalar-udf . That would provide a mechanism for doing the IO required in parallel and a way to tell DataFusion that operating in small batches is a good idea.

rust/sedona-schema/src/datatypes.rs
```diff
@@ -333,6 +365,1241 @@ fn deserialize_edges(edges: &Value) -> Result<Edges> {
...
+            Field::new(column::HEIGHT, DataType::UInt64, false),
+            // Geospatial transformation parameters
+            Field::new(column::UPPERLEFT_X, DataType::Float64, false),
+            Field::new(column::UPPERLEFT_Y, DataType::Float64, false),
+            Field::new(column::SCALE_X, DataType::Float64, false),
+            Field::new(column::SCALE_Y, DataType::Float64, false),
+            Field::new(column::SKEW_X, DataType::Float64, false),
+            Field::new(column::SKEW_Y, DataType::Float64, false),
+        ]))
+    }
```
paleolimbot, line 397: A potential future consideration, that it might be just as fast to do something like:

```rust
#[derive(Serialize, Deserialize)
struct RasterMetadata {
  shape: [u64; 2],
  upper_left: [f64; 2],
  geo_transform: [f64; 6],
}
```

...and use `serde_json` to read and write from JSON. (i.e., the `metadata` field would just be `Utf8`).

You've done a great job abstracting this so that nobody ever has to manually create or read this and so it should be easy to experiment with changing it in the future 🙂

rust/sedona-schema/src/datatypes.rs
```diff
@@ -333,6 +365,1241 @@ fn deserialize_edges(edges: &Value) -> Result<Edges> {
...
+    /// Band metadata schema
+    pub fn band_metadata_type() -> DataType {
+        DataType::Struct(Fields::from(vec![
+            Field::new(column::NODATAVALUE, DataType::Binary, true), // Allow null nodata values
+            Field::new(column::STORAGE_TYPE, DataType::UInt32, false),
+            Field::new(column::DATATYPE, DataType::UInt32, false),
+            // OutDb reference fields - only used when storage_type == OutDbRef
+            Field::new(column::OUTDB_URL, DataType::Utf8, true),
+            Field::new(column::OUTDB_BAND_ID, DataType::UInt32, true),
+        ]))
```
paleolimbot, line 435: With JSON storage this could also collapse to:

```rust
#[derive(Serialize, Deserialize)]
struct BandMetadata {
  nodata_value: Vec<u8>,
  storage_type: StorageType,
  data_type: BandDataType,
  #[serde(skip_serializing_if = None)]
  outdb_url: Option<String>,
  #[serde(skip_serializing_if = None)]
  outdb_band_id: Option<String>
}
```

(Again, this is abstracted nicely so we don't have to consider this now!)

jesspav:
Closing - was just a prototype.

PR #251, opened 2025-10-28
feat(rust/sedona-datasource): Implement generic RecordBatchReader-based format

paleolimbot:
This PR implements a simple "file format" aimed at making it easier to implement GDAL/OGR reading, although the boiler plate is applicable to reading various formats and/or implementing them in a higher-level language like Python.

This is basically a watered down version of the DataFusion FileFormat that's a bit easier to implement (at the expense of not supporting some features of the file format). The basic idea is that if you can answer these two questions:

```rust
#[async_trait]
pub trait RecordBatchReaderFormatSpec: Debug + Send + Sync {
    async fn infer_schema(&self, location: &Object) -> Result<Schema>;
    async fn open_reader(&self, args: &OpenReaderArgs)
        -> Result<Box<dyn RecordBatchReader + Send>>;
    // ... more options for advanced features
}
```

...you can implement a file format. This is significantly easier than the existing situation which involves chasing listing tables, file formats, file format factories, and file openers. The `OpenReaderArgs` passes along a filter and projection so that a format can push down a filter if it would like.

This might need some tweaks when we actually implement a file format with it, but I think this is ready for review!

### DIFF
--- a/python/sedonadb/python/sedonadb/dbapi.py
+++ b/python/sedonadb/python/sedonadb/dbapi.py
@@ -38,7 +38,7 @@ def connect(**kwargs: Mapping[str, Any]) -> "Connection":
 
         >>> con = sedona.dbapi.connect()
         >>> with con.cursor() as cur:
-        ...     cur.execute("SELECT 1 as one")
+        ...     _ = cur.execute("SELECT 1 as one")
         ...     cur.fetchall()
         [(1,)]
     """


rust/sedona-datasource/src/spec.rs
```diff
@@ -0,0 +1,193 @@
...
+    /// Given a single file, infer what schema [ExternalFormatSpec::open_reader]
+    /// would produce in the absence of any other guidance.
+    async fn infer_schema(&self, location: &Object) -> Result<Schema>;
+
+    /// Open a [RecordBatchReader] for a given file
+    ///
+    /// The implementation must handle the `file_projection`; however,
+    /// need not handle the `filters` (but may use them for pruning).
+    async fn open_reader(&self, args: &OpenReaderArgs)
+        -> Result<Box<dyn RecordBatchReader + Send>>;
```
paleolimbot, line 52: This is the user-facing API that would be used to implement readers

rust/sedona-datasource/src/provider.rs
```diff
@@ -0,0 +1,112 @@
...
+/// Create a [ListingTable] from an [ExternalFormatSpec] and one or more URLs
+///
+/// This can be used to resolve a format specification into a TableProvider that
+/// may be registered with a [SessionContext].
+pub async fn external_listing_table(
+    spec: Arc<dyn ExternalFormatSpec>,
+    context: &SessionContext,
+    table_paths: Vec<ListingTableUrl>,
+    check_extension: bool,
+) -> Result<ListingTable> {
```
paleolimbot, line 41: This is the wrapper we'd use to implement `read_xxxx()` in Python

rust/sedona-datasource/src/format.rs
```diff
@@ -0,0 +1,717 @@
...
+use crate::spec::{ExternalFormatSpec, Object, OpenReaderArgs, SupportsRepartition};
+
+/// Create a [FileFormatFactory] from a [ExternalFormatSpec]
+///
+/// The FileFormatFactory is the object that may be registered with a
+/// SessionStateBuilder to allow SQL queries to access this format.
+#[derive(Debug)]
+pub struct ExternalFormatFactory {
+    spec: Arc<dyn ExternalFormatSpec>,
+}
```
paleolimbot, line 53: This is what we'd use to register a format with the session so that things like `SELECT * FROM 'gpkg/*.gpkg'` work in SQL.

rust/sedona-datasource/Cargo.toml
```diff
@@ -0,0 +1,66 @@
...
+object_store = { workspace = true }
+parquet = { workspace = true }
+sedona-common = { path = "../sedona-common" }
+sedona-expr = { path = "../sedona-expr" }
+sedona-functions = { path = "../sedona-functions" }
+sedona-geometry = { path = "../sedona-geometry" }
+sedona-schema = { path = "../sedona-schema" }
+serde = { workspace = true }
+serde_json = { workspace = true }
+serde_with = { workspace = true }
```
Kontinuation: Are there unneeded dependencies? I believe that `chrono`, `float_next_after`, and `parquet` are not needed here.

rust/sedona-datasource/src/format.rs
```diff
@@ -0,0 +1,717 @@
...
+    }
+
+    async fn create_writer_physical_plan(
+        &self,
+        _input: Arc<dyn ExecutionPlan>,
+        _state: &dyn Session,
+        _conf: FileSinkConfig,
+        _order_requirements: Option<LexRequirement>,
+    ) -> Result<Arc<dyn ExecutionPlan>> {
+        not_impl_err!("writing not yet supported for SimpleSedonaFormat")
```
Kontinuation: Should this be `"ExternalFileFormat"`?

rust/sedona-datasource/src/format.rs
```diff
@@ -0,0 +1,717 @@
...
+struct ExternalFileOpener {
+    spec: Arc<dyn ExternalFormatSpec>,
+    args: OpenReaderArgs,
+}
+
+impl FileOpener for ExternalFileOpener {
+    fn open(&self, file_meta: FileMeta, _file: PartitionedFile) -> Result<FileOpenFuture> {
+        if file_meta.range.is_some() {
+            return sedona_internal_err!(
+                "Expected SimpleOpener to open a single partition per file"
```
Kontinuation: I believe that this should be `ExternalFileOpener`.

Kontinuation (commented):
The API looks good to me, I only have a few comments about some names in error messages.

rust/sedona-datasource/src/format.rs
```diff
@@ -0,0 +1,717 @@
...
+            projected_statistics: Some(statistics),
+            ..self.clone()
+        })
+    }
+
+    fn metrics(&self) -> &ExecutionPlanMetricsSet {
+        &self.metrics
+    }
+
+    fn statistics(&self) -> Result<Statistics> {
```
martin-g: Why Result ? You don't use the Err but panic with `.expect()`.
IMO it would be better to return `Option<Statistics>` instead.

rust/sedona-datasource/src/format.rs
```diff
@@ -0,0 +1,717 @@
...
+
+    fn get_ext(&self) -> String {
+        self.spec.extension().to_string()
+    }
+
+    fn get_ext_with_compression(
+        &self,
+        _file_compression_type: &FileCompressionType,
+    ) -> Result<String> {
+        not_impl_err!("extension with compression type")
```
martin-g:
```suggestion
        match _file_compression_type {
            FileCompressionType::UNCOMPRESSED => Ok(self.get_ext()),
            _ => not_impl_err!("extension with compression type"),
        }
```
a minor improvement for the uncompressed case.

Issue #252, opened 2025-10-28
Add Raster Schema


jesspav:
Create a schema for Raster support based on #247 / #249 including some basic ways of interacting with it.

PR #253, opened 2025-10-28
Add raster schema

jesspav:
#252 Create a basic schema for raster support as well as a builder and iterator to interact with the new schema.

### Testing
- unit tests
- lots of prototyping


rust/sedona-schema/src/datatypes.rs
```diff
@@ -208,6 +237,9 @@ impl Display for SedonaType {
             SedonaType::Arrow(data_type) => Display::fmt(data_type, f),
             SedonaType::Wkb(edges, crs) => display_geometry("Wkb", edges, crs, f),
             SedonaType::WkbView(edges, crs) => display_geometry("WkbView", edges, crs, f),
+            SedonaType::Raster() => {
+                unimplemented!("Display for Raster type not yet implemented")
```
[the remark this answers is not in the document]
jesspav: thanks - fixed

rust/sedona-raster/src/builder.rs
```diff
@@ -0,0 +1,1600 @@
...
+    }
+
+    fn storage_type(&self) -> StorageType {
+        match self.storage_type_array.value(self.band_index) {
+            0 => StorageType::InDb,
+            1 => StorageType::OutDbRef,
+            _ => panic!(
+                "Unknown storage type: {}",
+                self.storage_type_array.value(self.band_index)
+            ),
```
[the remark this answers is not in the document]
jesspav: @paleolimbot I'm curious about your take on this. My thinking is that if we're hitting unknown storage types or band data types here, something has gone seriously wrong.  Either we have corrupted data, a version mismatch, or a bug in our writing logic.

Returning a default or trying to handle it gracefully feels risky since we could end up with silent corruption or wrong results. The `panic!` makes it obvious that something's broken.

But I'm still getting familiar with Rust patterns and very open to suggestions.  Is an internal error enough to capture that?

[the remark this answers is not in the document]
paleolimbot: I think you have the right idea (let's make sure this is an error that is surfaced immediately!); however, I think copilot is correct here because a panic in some cases will just crash the process without displaying a message (and our Python package is one of those contexts!).

rust/sedona-raster/src/builder.rs
```diff
@@ -0,0 +1,1600 @@
...
+use datafusion_common::error::Result;
+use sedona_schema::raster::{
+    band_indices, band_metadata_indices, bounding_box_indices, column, metadata_indices,
+    raster_indices, BandDataType, RasterSchema, StorageType,
+};
+
+/// Builder for constructing raster arrays with zero-copy band data writing
+pub struct RasterBuilder {
+    main_builder: StructBuilder,
+}
```
paleolimbot: I think it would be nice to separate the data structures, the iterator, and the builder into separate modules for future readers (there is a lot of great documentation but it takes some scrolling to find it!)

rust/sedona-raster/src/builder.rs
```diff
@@ -0,0 +1,1600 @@
...
+
+/// Bounding box coordinates
+#[derive(Debug, Clone)]
+pub struct BoundingBox {
+    pub min_x: f64,
+    pub min_y: f64,
+    pub max_x: f64,
+    pub max_y: f64,
+}
```
paleolimbot: We have at least one, and possibly two BoundingBox structs somewhere. If possible we should use one of them here (or rename this to `RasterBoundingBox` and implement `From<sedona_geometry::BoundingBox> for RasterBoundingBox` and `From<RasterBoundingBox> for BoundingBox` to help convert between them.

rust/sedona-raster/src/builder.rs
```diff
@@ -0,0 +1,1600 @@
...
+
+impl RasterBuilder {
+    /// Create a new raster builder with the specified capacity
+    pub fn new(capacity: usize) -> Self {
+        let metadata_builder = StructBuilder::from_fields(
+            match RasterSchema::metadata_type() {
+                DataType::Struct(fields) => fields,
+                _ => panic!("Expected struct type for metadata"),
+            },
```
paleolimbot: Rather than panic, perhaps `RasterSchema` could just generate `fields` directly? (Also for the components below)

rust/sedona-raster/src/builder.rs
```diff
@@ -0,0 +1,1600 @@
...
+    /// Start a new raster with metadata, optional CRS, and optional bounding box
+    ///
+    /// This is the unified method for starting a raster with all optional parameters.
+    ///
+    /// # Arguments
+    /// * `metadata` - Raster metadata (dimensions, geotransform parameters)
+    /// * `crs` - Optional coordinate reference system as string
+    /// * `bbox` - Optional bounding box coordinates
+    ///
+    /// # Examples
```
paleolimbot: This is a nice example and might be easier for users to find if it were in the documentation for `RasterBuilder`!

rust/sedona-raster/src/builder.rs
```diff
@@ -0,0 +1,1600 @@
...
+        let bands_builder = self
+            .main_builder
+            .field_builder::<ListBuilder<StructBuilder>>(raster_indices::BANDS)
+            .unwrap();
+        let band_builder = bands_builder.values();
+
+        // Get the metadata builder and populate its fields
+        {
+            let metadata_builder = band_builder
+                .field_builder::<StructBuilder>(band_indices::METADATA)
```
paleolimbot: Kristin had a good point on the other PR that downcasting on a per-row basis is typically viewed as something that should be avoided. The alternative here would be to keep concrete builders for the individual columns in the RasterBuilder:

```rust
struct RasterBuilder {
  crs: StringViewBuilder,
  height: UInt64Builder,
  width: UInt64Builder,
  ...,
  band_offsets: OffsetBufferBuilder,
  band_nodata: BinaryBuilder,
  ...
}
```

...and then you would have to assemble the built arrays at the very end.

rust/sedona-raster/src/builder.rs
```diff
@@ -0,0 +1,1600 @@
...
+
+    /// Get a specific raster by index without consuming the iterator
+    pub fn get(&self, index: usize) -> Option<RasterRefImpl<'a>> {
+        if index >= self.raster_array.len() {
+            return None;
+        }
+
+        Some(RasterRefImpl::new(self.raster_array, index))
+    }
+}
```
Kontinuation: `RasterStructIterator` allows random access and I found no usage of its iterator interface in https://github.com/apache/sedona-db/pull/249/files. Does it make sense to rename it to `RasterStructArray` and remove impl for `Iterator`?

rust/sedona-schema/src/datatypes.rs
```diff
@@ -19,16 +19,19 @@ use datafusion_common::error::{DataFusionError, Result};
...
 use crate::extension_type::ExtensionType;
+use crate::raster::RasterSchema;
 
 /// Data types supported by Sedona that resolve to a concrete Arrow DataType
 #[derive(Debug, PartialEq, Clone)]
 pub enum SedonaType {
     Arrow(DataType),
     Wkb(Edges, Crs),
     WkbView(Edges, Crs),
+    Raster(),
```
Kontinuation: nit: We don't need the parentheses when no data is associated with this item.

rust/sedona-raster/src/builder.rs
```diff
@@ -0,0 +1,1600 @@
...
+
+impl<'a> BandsRef for BandsRefImpl<'a> {
+    fn len(&self) -> usize {
+        let start = self.bands_list.value_offsets()[self.raster_index] as usize;
+        let end = self.bands_list.value_offsets()[self.raster_index + 1] as usize;
+        end - start
+    }
+
+    /// Get a specific band by number (1-based index)
+    fn band(&self, number: usize) -> Result<Box<dyn BandRef + '_>, String> {
```
paleolimbot: This should probably be an `ArrowError` (which is the error class that is used in the rest of this file)

rust/sedona-raster/src/builder.rs
```diff
@@ -0,0 +1,1600 @@
...
+/// Each iteration yields a `RasterRefImpl` that provides access to both metadata and band data.
+pub struct RasterStructIterator<'a> {
+    raster_array: &'a StructArray,
+    current_row: usize,
+}
+
+/// Create a raster iterator for a StructArray containing raster data
+pub fn raster_iterator(raster_array: &StructArray) -> RasterStructIterator<'_> {
+    RasterStructIterator::new(raster_array)
+}
```
paleolimbot: `RasterStructIterator::new()` might be more Rust-ish (it's the name I was expecting)

rust/sedona-schema/src/datatypes.rs
jesspav: oh! thanks! removed

rust/sedona-raster/src/builder.rs
jesspav: In the prototype it was being used in this in the executor, but it probably doesn't need an entire iterator implementation.  I'll take a look to see how we can improve it.  I'm fine with the `RasterStructArray` rename.

rust/sedona-raster/src/builder.rs
```diff
@@ -0,0 +1,1600 @@
...
+use arrow_array::{
+    builder::{
+        BinaryBuilder, BinaryViewBuilder, Float64Builder, ListBuilder, StringBuilder,
+        StringViewBuilder, StructBuilder, UInt32Builder, UInt64Builder,
+    },
+    Array, BinaryArray, Float64Array, ListArray, StringArray, StringViewArray, StructArray,
+    UInt32Array, UInt64Array,
+};
+use arrow_schema::{ArrowError, DataType, Field};
+use datafusion_common::error::Result;
```
paleolimbot: We should use either `ArrowError` or `DataFusionError` consistently in this crate. Most usage is currently ArrowError so perhaps the DataFusion Result should move to `Result<..., ArrowError>`

rust/sedona-raster/Cargo.toml
```diff
@@ -0,0 +1,36 @@
...
+edition.workspace = true
+rust-version.workspace = true
+
+[lints.clippy]
+result_large_err = "allow"
+
+[dependencies]
+arrow-schema = { workspace = true }
+arrow-array = { workspace = true }
+arrow = { workspace = true }
```
paleolimbot: (I think this is not used?)

```suggestion
```

rust/sedona-raster/src/builder.rs
```diff
@@ -0,0 +1,1600 @@
...
+            .as_any()
+            .downcast_ref::<StructArray>()
+            .ok_or("Failed to downcast to StructArray")?;
+
+        // Get the metadata substructure from the band struct
+        let band_metadata_struct = bands_struct
+            .column(band_indices::METADATA)
+            .as_any()
+            .downcast_ref::<StructArray>()
+            .ok_or("Failed to downcast metadata to StructArray")?;
```
paleolimbot: It would be nice to avoid this downcast in a loop; however, it's nicely isolated as an implementation detail and I think this is OK for now.

rust/sedona-raster/src/builder.rs
```diff
@@ -0,0 +1,1600 @@
...
+    ) -> Result<(), ArrowError> {
+        self.append_metadata_from_ref(metadata)?;
+        self.append_crs(crs)?;
+        self.append_bounding_box(bbox)?;
+
+        Ok(())
+    }
+
+    /// Start a new band - this must be called before writing band data
+    pub fn start_band(&mut self, band_metadata: BandMetadata) -> Result<(), ArrowError> {
```
paleolimbot, line 165: Should the RasterBuilder be able to append a `RasterRef`?

rust/sedona-raster/src/builder.rs
Kontinuation: +1. I also favor this approach.

rust/sedona-raster/src/builder.rs
jesspav: I love the idea of unifying them - it'll make raster-geometry intersections much cleaner.

Looking at the potential differences:
1. CRS handling: ~~Raster bounding boxes are typically in EPSG:4326~~, while geometry bounding boxes inherit the CRS from their geometries.
2. Dimensionality: Geometries can have Z coordinates in their bounding boxes, while raster bounding boxes are typically 2D (though rasters may have elevation data, that's separate from the spatial bounds).

For 1, we could consider adding a CRS to the BoundingBox to avoid bugs.  Good utility functions will also help with this.
For 2, we are find as we only consider z-coordinates in the bounding box intersection if both have them have it defined ([ref](https://github.com/apache/sedona-db/blob/0551583a675da1cf5cc61cb7c7a61352051f1fc6/rust/sedona-geometry/src/bounding_box.rs#L100))

I'll explore making this switch today.

rust/sedona-raster/src/builder.rs
jesspav: Thanks, I'll look into improving this.

rust/sedona-raster/src/builder.rs
jesspav: Agree 100%

rust/sedona-raster/src/builder.rs
paleolimbot: > Raster bounding boxes are typically in EPSG:4326

I didn't know this! If this is the case that should be labelled every place it's accessed (possibly even in the name of the accessor/column name in the raster) because this is an easy mistake to make (or highlighted here if I missed it!). I suppose the native bounding box can be calculated from the top left point and affine transform.

We don't have a WGS84-only bounding box anywhere, so I think having this as its own `struct` is OK. Calling the fields `lon/lat` instead of `x/y` would be another way to avoid confusion.

rust/sedona-raster/src/builder.rs
jesspav: Done. Was excited to learn that it gets run as part of `cargo test`!

rust/sedona-raster/src/builder.rs
jesspav: done.

rust/sedona-raster/src/builder.rs
jesspav: Probably, but I don't have a use case at the moment. Can we add it when we need it?  it will be very easy to add.

rust/sedona-raster/Cargo.toml
jesspav: removed.

rust/sedona-raster/src/builder.rs
jesspav: done.

rust/sedona-raster/src/builder.rs
jesspav: done

rust/sedona-raster/src/builder.rs
```diff
@@ -0,0 +1,1600 @@
...
+            1 => BandDataType::UInt16,
+            2 => BandDataType::Int16,
+            3 => BandDataType::UInt32,
+            4 => BandDataType::Int32,
+            5 => BandDataType::Float32,
+            6 => BandDataType::Float64,
+            _ => panic!(
+                "Unknown band data type: {}",
+                self.datatype_array.value(self.band_index)
+            ),
```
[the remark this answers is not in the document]
jesspav: no longer panics. thanks

rust/sedona-raster/src/builder.rs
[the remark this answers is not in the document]
jesspav: oh! I didn't know that the error would be hidden with a panic.
Switched to errors.

rust/sedona-raster/src/builder.rs
jesspav: removed the function. using new. switched the name.

rust/sedona-raster/src/builder.rs
jesspav: Oops! Thanks!

rust/sedona-raster/src/builder.rs
jesspav: Agreed.. that file was way too long.
Done.

rust/sedona-raster/src/builder.rs
jesspav: This code has been reformatted and there are no more panics.

rust/sedona-raster/src/builder.rs
jesspav: moved.

rust/sedona-raster/src/traits.rs
```diff
@@ -0,0 +1,144 @@
...
+// specific language governing permissions and limitations
+// under the License.
+
+use arrow_schema::ArrowError;
+
+use sedona_schema::raster::{BandDataType, StorageType};
+
+/// Metadata for a raster
+#[derive(Debug, Clone)]
+pub struct RasterMetadata {
```
paleolimbot, line 24: This does a great job of highlighting the data model!

rust/sedona-raster/src/traits.rs
```diff
@@ -0,0 +1,144 @@
...
+    pub upperleft_y: f64,
+    pub scale_x: f64,
+    pub scale_y: f64,
+    pub skew_x: f64,
+    pub skew_y: f64,
+}
+
+/// Bounding box coordinates
+#[derive(Debug, Clone)]
+pub struct BoundingBox {
```
paleolimbot: It would be nice to clarify the meaning of this but the hard part is appending and iterating over four doubles, which this PR does nicely. It may be nice to rename this to `RasterBoundingBox` in the meantime to avoid accidentally using it for non-raster things (totally optional).

rust/sedona-raster/src/builder.rs
jesspav: Sorry, I think that was wrong. I am looking more into this.

jesspav: I've decided to remove the bbox for now.  We can add it back if we find we need it.

rust/sedona-raster/src/traits.rs
jesspav: Agreed, but I've decided to remove it for now.  We can add it back if we need it.

jesspav:
The clippy failure is from another commit outside of the packages that this PR changes:
```
error: this `impl` can be derived
  --> c/sedona-tg/src/tg.rs:38:1
   |
38 | / impl Default for IndexType {
39 | |     fn default() -> Self {
40 | |         Self::Default
41 | |     }
42 | | }
   | |_^
   |
   = help: for further information visit https://rust-lang.github.io/rust-clippy/rust-1.91.0/index.html#derivable_impls
   = note: `-D clippy::derivable-impls` implied by `-D warnings`
   = help: to override `-D warnings` add `#[allow(clippy::derivable_impls)]`
help: replace the manual implementation with a derive attribute and mark the default variant
   |
27 + #[derive(Default)]
28 ~ pub enum IndexType {
29 |     /// Do not build an index
30 |     Unindexed,
31 |     /// Use the statically set default
32 ~     #[default]
33 ~     Default,
   |

error: could not compile `sedona-tg` (lib) due to 1 previous error
warning: build failed, waiting for other jobs to finish...
Error: Process completed with exit code 101.
```

paleolimbot (approved):
🚀

PR #254, opened 2025-10-29
feat: Support plain Binary/BinaryView as geometry types for Datafusion Comet integration

zhangfengcdt:
This PR adds support for plain Arrow Binary and BinaryView datatypes as valid geometry types in sedona-db, enabling cleaner integration with Apache Comet's geometry handling.

- Enables geometry functions and aggregates to work with Comet's Binary columns
- Supports cleaner architecture using extension metadata instead of Struct wrapping

### DIFF
--- a/rust/sedona/tests/aggregate.rs
+++ b/rust/sedona/tests/aggregate.rs
@@ -259,6 +259,6 @@ mod aggregate_functions {
         assert_eq!(batch.num_columns(), 1);
 
         let result_array = batch.column(0);
-        assert!(result_array.len() > 0, "Expected non-empty result");
+        assert!(!result_array.is_empty(), "Expected non-empty result");
     }
 }


zhangfengcdt:
> Thank you for looking into this!
> 
> I would prefer to find the place in Comet that is dropping the type and fix it upstream (perhaps in one of its calls to `return_field()`, `accumlator()`, or `invoke_batch_with_args()`?).
> 
> If we absolutely must do it here, I think we should do it behind a feature flag...for all standard usage we want to error for this case to catch and fix dropped metadata as soon as we know it occurs.

This is actually an issue in Datafusion (not Comet) issue for aggregation function, which impact the four ST UADFs we added before: ST_Envelope_Aggr, ST_Intersection_Aggr, ST_Union_Aggr, and ST_Analyze_Aggr. 

In datafusion, datafusion/physical-plan/src/aggregates/mod.rs - PhysicalGroupBy::group_fields()

```
  Field::new(name, expr.data_type(input_schema)?,
             group_expr_nullable || expr.nullable(input_schema)?)
  .with_metadata(expr.return_field(input_schema)?.metadata().clone())
```
However, in datafusion/physical-expr-common/src/physical_expr.rs, it simply creates a new fileds without adding the metadata from original field. See: https://github.com/apache/datafusion/blob/f57da83aac35f0ea4506ccb6a4ddbd26a503c1c1/datafusion/physical-expr-common/src/physical_expr.rs#L86

This could be fixed by changing the Datafusion source code to check if expression references an input field, if so, get that field's metadata and call .with_metadata() to preserve it. However, I am not sure this is acceptable for all senarios because the aggregated fields may NOT be legit to actually inherit original field's metadata. For example, this is not the case for ST_Analyze_Aggr, which the output is not Geometry type anymore. So, not sure Datafusion commit will accept such changes.

Anyway, I can add a feature flag to only have changes in the PR, but I don't see risks of not having it under a feature flag.

zhangfengcdt:
> If this is a DataFusion issue it likely means that the aggregations in SedonaDB are broken (I am not sure we have integration tests for these and we probably should!).
> 
> I would be surprised if this is not solvable in DataFusion...in any case, the lack of field metadata propagation for our specific case is a bug (not necessarily the generic case, as you noted). Our return field in theory declares metadata properly!

I got your points!  I have added e2e tests (aggregate.rs) to test these aggregation functions and verified that without the "fix" in this PR, they will fail with the same errors we see earlier.

I understand that our aggregate_udf.rs has the correct return_field() method implemented that carries the metadata. However, it seems only be used in Datafusion logic plan planning. But in Datafusion's physical planning, as noted before, it does not honor the return_field(), instead, it creates a brand new field from scratch based only on the array's DataType, which is just Binary without extension metadata. The real fix is to make DataFusion's physical aggregate use the logical plan's field (with metadata) instead of creating a new one.

Until we have that fix in Datafusion, it does not seem to have an easy way to fully fix it in sedona db. The "fix" in this PR should be able to unblock us, so I would suggest we:
(1) add feature flag to control this fixing behavior
(2) add todo to remove such "fix" once the root cause in datafusion is resolved.

Please let me know if you'd be good with this or any other thoughts, and I can make the changes accordingly.

zhangfengcdt:
@paleolimbot I think I found the issue after using your example code that the issue is isolated to some SQL functions, not all. For example:

This does NOT preserve the metadata created in the ST_GeomFromText:
```
SELECT ST_AsText(ST_Envelope_Aggr(geom)) FROM (
        VALUES
            (ST_GeomFromText('POINT (1 2)')),
            (ST_GeomFromText('POINT (3 4)')),
            (ST_GeomFromText('POINT (5 6)'))
    ) AS t(geom)
```

BUT, this one does:
```
SELECT ST_AsText(ST_Envelope_Aggr(geom)) FROM (
        SELECT ST_GeomFromText('POINT (1 2)') as geom
        UNION ALL
        SELECT ST_GeomFromText('POINT (3 4)') as geom
        UNION ALL
        SELECT ST_GeomFromText('POINT (5 6)') as geom
    ) AS t(geom)
```

This surely shows the bug in Datafusion and I think I will try to find time fix it upstream there. For now,  I can change the unit test to use the UNION ALL or read from parquet. I don't think the bug of the "VALUES" syntax with metadata carry-over will impact a lot of real use cases at this stage.

zhangfengcdt:
Close the PR as issues have been addressed upstream.

Issue #198, opened 2025-10-09
feat: Implement fast implementations of functions using WKBHeader


petern48:
Once PR https://github.com/apache/sedona-db/pull/171 is merged, we'll have a handle WKBHeader struct for performing fast computations without parsing the whole geometry. This issue will track the functions we should use it for once it's merged. Just took a quick pass through https://github.com/apache/sedona-db/issues/174. Possibly missed some or picked some that are not possible. Just jotting down notes

New Functions:
- GeometryType https://github.com/apache/sedona-db/issues/259
- ST_Zmflag
- ST_NumGeometries
- ST_NumInteriorRing
- ST_NumInteriorRings
- ST_NumRings
- ST_NumPoints
- ST_IsCollection
- ST_GeometryN (this would need some more logic on top of WKBHeader to iterate through geoms)

Update 11/8:
- ST_NDims

Speedup Existing Functions:
- ST_X/ST_Y
- ST_IsEmpty
- ST_Dimension (the GeomCollection case is more complicated since it would need to traverse each nested geoms header)
  - alternatively, we can leverage this optimization only for the non-GeomCollection cases and leave the complex case for later

paleolimbot:
Thanks for creating this! ST_X/Y are probably in this category, too.

These are all good things to do and that the more we can do with lazy parsing the better! I do want to note that these functions aren't typically bottlenecks: the case where they're useful are where you have things like absolutely massive multipoints for each feature. Those workflows have other elements that make them slow (e.g., the IO to read them or the memory pressure from having huge rows) and so the case where you have a huge number of massive collections is actually pretty rare. The workflows that are working with huge numbers of features are typically points, linestrings, and single-ring polygons, which are adequately lazy when parsed by our existing parser.

petern48:
Yeah, this issue is more for *how* we should implement these functions, rather than optimizing existing functions as much as possible. Kinda just kept adding to the list, even if it seemed too complicated to be worth at the moment. 

Removed my assignment, which I only added originally when I had like the first two in mind.

petern48:
Hmm, the rest under the "New Functions" section seem non-trivial. e.g ST_NumRings applies for MULTIPOLYGONs too, so if we wanted to implement it with lazy reading, we'd need to add logic for lazily iterating over the nested polygons. The same idea applies to ST_NumPoints, and even ST_Dimension's geom col case.

[...]

Issue #201, opened 2025-10-09
Explore decomposed branchless loop WKB parsing


paleolimbot:
The `WkbHeader` will be a great optimization for a number of things we'd like to do! Another technique potentially worth exploring to optimize these very cheap functions is decomposing the parsing into a series of branchless loops. The compiler has more options to optimize a loop that contains absolutely zero `if` `match` or `||/&&`s (e.g. SIMD autovectorization). I've prototyped this in C++ before and found order-of-magnitude speed improvements for Binary arrays (probably less helpful for BinaryView arrays, where there is always an if statement to access the data).

```rust
let mut n_valid_size = 0;
for item in not_null_array {
  n_valid_size += item.len() >= 5;
}

if n_valid_size != not_null_array.len() {
  return do_the_slow_version();
}

let mut n_little_endian = 0;
for item in not_null_array {
  n_little_endian += item[0] == 0x01;
}

if n_little_endian != not_null_array.len() {
  return do_the_slow_version();
}

let mut geometry_types: u32 = 0;
let mut geometry_type_bytes: [u8; 4];
for item in not_null_array {
  geometry_type_bytes.copy_from_slice(&item[1..5]);
  let geometry_type_id = u32::from_be_bytes(&geometry_type_bytes);
  geometry_types |= 1_u32 << geometry_type_id & 0x07;
}

// Potentially do something faster if all we have are points
```

There comes a point where the multiple branchless loops become slower than a single loop with branching...I haven't experimented with this enough to know where that point is.

For functions that only operate on points, there is also a cool optimization you can do for Binary (not BinaryView) arrays: because the data for arrays with zero nulls are all lined up consecutively in memory (i.e., one WKB item after another all in the same buffer), you can loop through X values like this (once you've validated the input using the above):

```rust
let mut offset = first_x_offset;
let mut x_bytes: [u8; 8];
for i in 0..array.len() {
  x_bytes.copy_from_slice(data_buffer[offset..(offset + 8)]);
  let x = f64::from_le_bytes(x_bytes);
  offset += 21;
}
```

My initial explorations of this are here: https://github.com/geoarrow/geoarrow-c/blob/28eca0fea6f47c70113dc1719e7597e53bede461/dev/benchmarks/c/wkb_bounding_benchmark.cc#L408-L450 . Specifically for the bounding operation, the initial numbers suggested that this was able to trigger simd for at least the bounding operation.

petern48:
[...]

PR #260, opened 2025-10-30
feat(rust/sedona-functions): Implement native ST_ZMFlag using WKBHeader

petern48:
Part of #198. 

Quite literally copied `st_haszm` implementation (including tests) and modified everything to be the int outputs instead of the boolean outputs.

### DIFF
--- a/benchmarks/test_functions.py
+++ b/benchmarks/test_functions.py
@@ -275,3 +275,21 @@ class TestBenchFunctions(TestBenchBase):
             eng.execute_and_collect(f"SELECT ST_EndPoint(geom1) from {table}")
 
         benchmark(queries)
+
+    @pytest.mark.parametrize(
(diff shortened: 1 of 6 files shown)


petern48:
image

See https://github.com/apache/sedona-db/pull/171#issuecomment-3466172666, for why the difference isn't as large anymore.

paleolimbot (approved):
Thank you!

To really make these go fast I think we would need something like https://github.com/apache/sedona-db/issues/201 (i.e., allocate a `Vec<WkbHeader>`s and fill them in using a series of loops that do not contain `match`, `if`, or `?`). Another potential improvement would be eliminate (most) buffer bounds checking by doing it once based on the top-level geometry type (works for everything except geometrycollection).

(Also...these are really fast already!)

PR #265, opened 2025-10-31
feat(rust/sedona-functions): Implement native ST_IsCollection using WkbHeader

petern48:
Part of https://github.com/apache/sedona-db/issues/198


petern48:
At the very end, I ran benchmarks and learned DuckDB doesn't implement st_iscollection, so there's no comparison or insights to see 🙃.

image

PR #266, opened 2025-10-31
chore: Fix clippy errors from newer Rust and pin Python version

paleolimbot:
Totally optional fixes...my local Rust gives yellow squigglies for variables not inlined into the format string.

Also pins the Python version as 3.14 was just released and a bunch of dependencies don't have published wheels for that yet.

### DIFF
--- a/.github/workflows/packaging.yml
+++ b/.github/workflows/packaging.yml
@@ -45,7 +45,7 @@ jobs:
           submodules: 'true'
       - uses: actions/setup-python@v5
         with:
-          python-version: "3.x"
+          python-version: "3.13"
 
       - name: Prepare dev version
         shell: bash


.github/workflows/python.yml
```diff
@@ -63,7 +63,7 @@ jobs:
 
       - uses: actions/setup-python@v4
         with:
-          python-version: '3.x'
+          python-version: '3.13'
```
petern48, line 66: I never realized this until now that we don't test multiple Python versions. Is that intentional and planned to be kept that way? (i have zero complaints. makes my life as a dev much easier haha)

paleolimbot: We don't run multiple versions of Python in CI for PRs, but we build wheels for (and run tests on those wheels) for Python 3.9 to Python 3.14 (including the free-threaded versions of Python 3.13 and 3.14). Mostly our Python surface area/rate of change is not very high (much more Rust code than Python code!)

paleolimbot: (Those tests run only on commits to main because they take forever)

PR #267, opened 2025-10-31
feat(rust/sedona-functions): Add ST_Translate()

paleolimbot:
This PR implements `ST_Translate()`.

```
--------------------------------------- benchmark 'table=collections_complex': 3 tests --------------------------------------
Name (time in ms)                                                   Median                  Mean             StdDev          
-----------------------------------------------------------------------------------------------------------------------------
test_st_translate[collections_complex-SedonaDBSingleThread]       256.2327 (1.0)        277.4057 (1.0)      40.6925 (1.0)    
test_st_translate[collections_complex-DuckDBSingleThread]       1,277.4913 (4.99)     1,290.4847 (4.65)     46.4146 (1.14)   
test_st_translate[collections_complex-PostGISSingleThread]      6,999.6389 (27.32)    7,026.8960 (25.33)    64.5042 (1.59)   
-----------------------------------------------------------------------------------------------------------------------------

------------------------------------- benchmark 'table=points_simple': 3 tests ------------------------------------
Name (time in ms)                                          Median                Mean              StdDev          
-------------------------------------------------------------------------------------------------------------------
test_st_translate[points_simple-SedonaDBSingleThread]      3.6685 (1.0)        3.8089 (1.0)        1.1013 (1.0)    
test_st_translate[points_simple-DuckDBSingleThread]       10.4048 (2.84)      11.0317 (2.90)       1.2912 (1.17)   
test_st_translate[points_simple-PostGISSingleThread]      36.1355 (9.85)     117.9003 (30.95)    184.4294 (167.46) 
-------------------------------------------------------------------------------------------------------------------
```

### DIFF
--- a/python/sedonadb/tests/functions/test_transforms.py
+++ b/python/sedonadb/tests/functions/test_transforms.py
@@ -14,9 +14,9 @@
 # KIND, either express or implied.  See the License for the
 # specific language governing permissions and limitations
 # under the License.
-import pytest
 import pyproj
-from sedonadb.testing import geom_or_null, PostGIS, SedonaDB, val_or_null
+import pytest
+from sedonadb.testing import PostGIS, SedonaDB, geom_or_null, val_or_null
 
 
 @pytest.mark.parametrize("eng", [SedonaDB, PostGIS])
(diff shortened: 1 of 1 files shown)


python/sedonadb/tests/functions/test_transforms.py
```diff
@@ -93,3 +93,32 @@ def test_st_crs_sedonadb(eng):
...
+        (None, None, None, None),
+        ("POINT (0 1)", 1.0, None, None),
+        ("POINT (0 1)", None, 2.0, None),
+        # WKT output of geoarrow-c is causing this (both correctly output
+        # empties)
+        ("POINT EMPTY", 1.0, 2.0, "POINT (nan nan)"),
+        ("POINT Z EMPTY", 1.0, 2.0, "POINT Z (nan nan nan)"),
+        ("POINT (0 1)", 1.0, 2.0, "POINT (1 3)"),
+        ("POINT Z (0 1 2)", 1.0, 2.0, "POINT Z (1 3 2)"),
+        ("LINESTRING EMPTY", 1.0, 2.0, "LINESTRING EMPTY"),
```
petern48, line 135: optional: A few cases that aren't tested anywhere (in python or rust). I don't think any would fail, but I think some would be nice to test. Particularly non-point geoms, so it's clear it works on them too. Personally, I often look at tests to see how I can use a function when documentation isn't specific enough.

- geometry arg
  - non-empty geometries other than POINT (e.g polygon, multi-*)
  - any `M` dimension
- dx/dy arg
  - zero or negative
  - integers (inspired by this https://github.com/apache/sedona-db/pull/32)

python/sedonadb/tests/functions/test_transforms.py
```diff
@@ -93,3 +93,32 @@ def test_st_crs_sedonadb(eng):
         "SELECT ST_CRS(ST_SetCrs(ST_GeomFromText('POINT (1 1)'), NULL))",
         None,
     )
+
+
+@pytest.mark.parametrize("eng", [SedonaDB, PostGIS])
+@pytest.mark.parametrize(
+    ("geom", "dx", "dy", "expected"),
+    [
+        (None, None, None, None),
```
petern48, line 103:
```suggestion
        (None, None, None, None),
        (None, 1.0, 2.0, None),
```
nit: While I don't think it's necessary to test every single combination of nulls, I think it's nice to at least have a case for each arg where *only* that arg is null, so we know that arg on it's own is enough to trigger a null output. There are currently cases for only dx null, only dy null, but not "only geometry null" (which I'd guess is the most likely case).

petern48 (approved):
Looks good. Just suggested some optional nits about testing.

petern48 (approved):

Issue #262, opened 2025-10-30
First Function


jesspav:
Add the smallest possible `RS_Width` function, as the first function.

This will require:
- raster execute function
- sample test data
- native udf

Benchmarking raster functions is out of scope for this change.

PR #268, opened 2025-10-31
Add a first raster function - RS_Width

jesspav:
This PR introduces the first simple raster function (`RS_Width`) along with the supporting infrastructure required for raster functions.

#262 

### Included
The minimal amount of changes to introduce a first function:
- Raster-specific function package
- Raster executor for a raster function with no arguments
- Raster argument matcher
- `RS_Width` function itself

Out of scope

- Benchmarking (to be added in a follow-up PR)

### Testing
- Unit tests only

### DIFF
--- a/rust/sedona-raster-functions/src/executor.rs
+++ b/rust/sedona-raster-functions/src/executor.rs
@@ -19,6 +19,7 @@ use arrow_array::{Array, ArrayRef, StructArray};
 use datafusion_common::error::Result;
 use datafusion_common::{DataFusionError, ScalarValue};
 use datafusion_expr::ColumnarValue;
+use sedona_common::sedona_internal_err;
 use sedona_raster::array::{RasterRefImpl, RasterStructArray};
 use sedona_schema::datatypes::SedonaType;
 use sedona_schema::datatypes::RASTER;
@@ -61,9 +62,7 @@ impl<'a, 'b> RasterExecutor<'a, 'b> {
         F: FnMut(usize, Option<RasterRefImpl<'_>>) -> Result<()>,
     {
         if self.arg_types[0] != RASTER {
-            return Err(DataFusionError::Internal(
-                "First argument must be a raster type".to_string(),
-            ));
+            return sedona_internal_err!("First argument must be a raster type");
         }
         let raster_array = match &self.args[0] {
             ColumnarValue::Array(array) => array,


rust/sedona-raster-functions/Cargo.toml
```diff
@@ -0,0 +1,40 @@
...
+result_large_err = "allow"
+
+[dependencies]
+arrow-schema = { workspace = true }
+arrow-array = { workspace = true }
+arrow-buffer = { workspace = true }
+datafusion-common = { workspace = true }
+datafusion-expr = { workspace = true }
+sedona-common = { path = "../sedona-common" }
+sedona-expr = { path = "../../rust/sedona-expr" }
```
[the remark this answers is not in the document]
jesspav: Fixed. t hanks!

rust/sedona-raster-functions/src/executor.rs
```diff
@@ -0,0 +1,188 @@
...
+/// The [RasterExecutor] provides a simplified interface for executing functions
+/// on raster arrays, handling the common pattern of downcasting to StructArray,
+/// creating raster iterators, and handling null values.
+pub struct RasterExecutor<'a, 'b> {
+    pub arg_types: &'a [SedonaType],
+    pub args: &'b [ColumnarValue],
+    num_iterations: usize,
+}
+
+impl<'a, 'b> RasterExecutor<'a, 'b> {
```
paleolimbot, line 38: I wonder if we can simplify this a bit because not all the reasons the "visit" thing exists for Geometry/Geography apply here. For Wkb the items could be very small and the per-iteration cost matters a lot...also we encode geometry in potentially a lot of different ways (even though binary view and regular binary are currently the only two we care about). For raster the per-iteration cost is is less of a factor and we have exactly one representation of it in our engine, so things like the double dynamic function pointer calls of `Box<dyn Iterator<Item = &dyn RasterRef>>` aren't an issue like they might be for Wkb iteration.

I am not sure if the lifetimes will work out, but maybe something like:

```rust
pub trait IntoRasterIterator {
  fn into_raster_iterator(&self, num_iterations: usize) -> Box<dyn Iterator<Item = &dyn RasterRef>>;
}

impl IntoRasterIterator for ScalarValue { .. }
impl IntoRasterIterator for ArrayRef { .. } 
impl IntoRasterIterator for ColumnarValue { .. } 
```

...would be easier to mix and match with functions that iterate over geometry and raster together?

rust/sedona-raster-functions/src/register.rs
```diff
@@ -0,0 +1,51 @@
...
+
+    function_set
+}
+
+/// Functions whose implementations are registered independently
+///
+/// These functions are included in the default function set; however,
+/// it is useful to expose them individually for testing in crates that
+/// implement them.
+pub mod stubs {}
```
paleolimbot:
```suggestion
```

(We can probably punt on these until they're needed?)

rust/sedona-raster-functions/src/rs_size.rs
```diff
@@ -0,0 +1,128 @@
...
+}
+
+fn rs_width_doc() -> Documentation {
+    Documentation::builder(
+        DOC_SECTION_OTHER,
+        "Return the width component of a raster".to_string(),
+        "RS_Width(raster: Raster)".to_string(),
+    )
+    .with_argument("raster", "Raster: Input raster")
+    .with_sql_example("SELECT RS_Width(raster)".to_string())
```
paleolimbot, line 49: No need to do this right now, but a raster constructor would help here (I don't know if one exists, or maybe we can make a `RS_Example()` in the future to avoid typing extra long strings here).

rust/sedona-raster-functions/src/rs_size.rs
```diff
@@ -0,0 +1,128 @@
...
+    }
+
+    fn invoke_batch(
+        &self,
+        arg_types: &[SedonaType],
+        args: &[ColumnarValue],
+    ) -> Result<ColumnarValue> {
+        let executor = RasterExecutor::new(arg_types, args);
+        let mut builder = UInt64Builder::with_capacity(executor.num_iterations());
+
```
paleolimbot, line 73: Without the executor this might be something like:

```rust
let executor = WkbExecutor::new(arg_types, args);
let mut builder = UInt64Builder::with_capacity(executor.num_iterations());
for raster_ref in args[0].into_raster_iterator(executor.num_iterations()) {

}

executor.finish(Arc::new(builder.finish()))
```

(I know using the name `WkbExecutor` is a fishy...I'm open to better ideas 🙂 )

rust/sedona-raster/src/array.rs
```diff
@@ -621,4 +628,41 @@ mod tests {
...
+        let metadata = RasterMetadata {
+            width: 5,
+            height: 5,
+            upperleft_x: 0.0,
+            upperleft_y: 0.0,
+            scale_x: 1.0,
+            scale_y: -1.0,
+            skew_x: 0.0,
+            skew_y: 0.0,
+        };
```
paleolimbot: Can we use `generate_test_rasters()` here?

rust/sedona-raster-functions/src/register.rs
jesspav: Yes, I copied the structure from Sedona-functions, but do not have any use for it now.  will remove

rust/sedona-raster/src/array.rs
jesspav: Yes!

rust/sedona-raster-functions/src/executor.rs
jesspav: hmm... the primary reason I wrote it this way was so that the raster and geometry functions would look and feel the same, which is definitely not a requirement. Let me think a bit more about this.

Kontinuation: What will the executor API look like if we want to support raster functions such as RS_Clip? It takes both raster and geometry arguments.

If we decided to stick to the current `RasterExecutor` design, will the new `execute_raster_wkb_void` be added to `RasterExecutor`?

If we switch to `IntoRasterIterator`, we'll call next() in the closure passed into `execute_wkb_void` in order to keep the raster iterator and wkb iteration process in sync.

rust/sedona-raster-functions/src/executor.rs
```diff
@@ -0,0 +1,188 @@
...
+    /// 3. Iterating with null checks
+    /// 4. Calling the provided function with each raster
+    pub fn execute_raster_void<F>(&self, mut func: F) -> Result<()>
+    where
+        F: FnMut(usize, Option<RasterRefImpl<'_>>) -> Result<()>,
+    {
+        if self.arg_types[0] != RASTER {
+            return Err(DataFusionError::Internal(
+                "First argument must be a raster type".to_string(),
+            ));
```
Kontinuation: We had a better way to return sedona internal errors: `sedona_internal_err!`. It constructs a `DataFusionError::External` with proper message referring to https://github.com/apache/sedona-db/issues to avoid the message suggesting it's internal to DataFusion.

I believe that we need to do a clean-up for the entire code base as well, but in a separate PR.

rust/sedona-testing/src/rasters.rs
```diff
@@ -0,0 +1,118 @@
...
+        builder.start_raster(&raster_metadata, None)?;
+        builder.start_band(BandMetadata {
+            datatype: BandDataType::UInt16,
+            nodata_value: Some(vec![0u8; 2]),
+            storage_type: StorageType::InDb,
+            outdb_url: None,
+            outdb_band_id: None,
+        })?;
+
+        let pixel_count = i * (i + 1);
```
Kontinuation: `width` is i + 1, `height` is i + 2. `pixel_count` should be (i + 1) * (i + 2)

rust/sedona-raster-functions/src/executor.rs
paleolimbot: Great point...Peter did a great job converting uses of `internal_err!()` to `sedona_internal_err!()` but there are a lot of `Err(DataFusionError::Internal)`s around, too.

rust/sedona-raster-functions/src/executor.rs
paleolimbot: All good points...we'll also need to handle iterating over the CRS (resolving once at the top level and then iterating or looping over items) of the wkb as well and this is probably a great place to handle that. A `RasterExecutor` is definitely less fishy than the other thing I suggested 🙂

rust/sedona-raster-functions/src/rs_size.rs
jesspav: Yes! I was thinking the same thing, but wasn't sure how best to make it available.  `RS_Example` is perfect. will add in another PR

rust/sedona-testing/src/rasters.rs
jesspav: Thanks! I changed it to avoid zeros and then forgot to update it here.

rust/sedona-raster-functions/src/executor.rs
jesspav: Thanks!

rust/sedona-raster/src/array.rs
jesspav: fixed.

rust/sedona-testing/src/rasters.rs
jesspav: fixed

rust/sedona-raster-functions/src/rs_size.rs
```diff
@@ -0,0 +1,128 @@
...
+    fn udf_invoke() {
+        // 3 rasters, second one is null
+        let rasters = generate_test_rasters(3, Some(1)).unwrap();
+
+        // Create the UDF and invoke it
+        let kernel = RsWidth {};
+        let args = [ColumnarValue::Array(Arc::new(rasters))];
+        let arg_types = vec![RASTER];
+
+        let result = kernel.invoke_batch(&arg_types, &args).unwrap();
```
paleolimbot, line 114: No need to do it here, but we should add a parser like we have for Geometry to the `ScalarUdfTester` so that we can write things like `tester.invoke_scalar("[[1, 2, 3], [4, 5, 6], [7, 8, 9]]")` (or maybe we already have a convention in Sedona Java for this)

rust/sedona-raster-functions/src/executor.rs
jesspav: @Kontinuation yeah, make something like `execute_raster_wkb_void` and handle the geometries the same way we do in the other executor.

rust/sedona-raster-functions/src/rs_size.rs
jesspav: Good idea!

PR #269, opened 2025-11-01
Implement ST_Dump

yutannihilation:
Note: Sedona's `ST_Dump` returns a different result from PostGIS and DuckDB. This pull request implements the PostGIS's version (`[]Struct { path: [int], geom: GEOMETRY }`)

https://sedona.apache.org/latest/api/sql/Function/#st_dump
https://postgis.net/docs/en/ST_Dump.html

### DIFF
--- a/rust/sedona-functions/src/st_dump.rs
+++ b/rust/sedona-functions/src/st_dump.rs
@@ -79,6 +79,7 @@ struct STDumpBuilder {
     geom_builder: BinaryBuilder,
     struct_offsets_builder: OffsetBufferBuilder<i32>,
     null_builder: NullBufferBuilder,
+    parent_path: Vec<u32>,
 }
 
 impl STDumpBuilder {
@@ -96,24 +97,20 @@ impl STDumpBuilder {
             geom_builder,
             struct_offsets_builder,
             null_builder,
+            parent_path: Vec::new(), // Reusable buffer to avoid allocation per row
         }
     }
 
     // This appends both path and geom at once.
-    fn append_inner(
-        &mut self,
-        parent_path: &[u32],
-        cur_index: Option<u32>,
-        wkb: SingleWkb<'_>,
-    ) -> Result<()> {
-        self.path_array_builder.append_slice(parent_path);
+    fn append_single_struct(&mut self, cur_index: Option<u32>, wkb: SingleWkb<'_>) -> Result<()> {
+        self.path_array_builder.append_slice(&self.parent_path);
         if let Some(cur_index) = cur_index {
             self.path_array_builder.append_value(cur_index);
             self.path_array_offsets_builder
-                .push_length(parent_path.len() + 1);
+                .push_length(self.parent_path.len() + 1);
         } else {
             self.path_array_offsets_builder
-                .push_length(parent_path.len());
+                .push_length(self.parent_path.len());
         }
 
         let write_result = match wkb {
@@ -138,39 +135,29 @@ impl STDumpBuilder {
         Ok(())
     }
 
-    fn append(
-        &mut self,
-        // Note: in order to avoid allocation, this needs to be &mut Vec, not &mut [].
-        parent_path: &mut Vec<u32>,
-        wkb: &wkb::reader::Wkb<'_>,
-    ) -> Result<i32> {
+    fn append_structs(&mut self, wkb: &wkb::reader::Wkb<'_>) -> Result<i32> {
         match wkb.as_type() {
             GeometryType::Point(point) => {
-                self.append_inner(parent_path, None, SingleWkb::Point(point))?;
+                self.append_single_struct(None, SingleWkb::Point(point))?;
                 Ok(1)
             }
             GeometryType::LineString(line_string) => {
-                self.append_inner(parent_path, None, SingleWkb::LineString(line_string))?;
+                self.append_single_struct(None, SingleWkb::LineString(line_string))?;
                 Ok(1)
             }
             GeometryType::Polygon(polygon) => {
-                self.append_inner(parent_path, None, SingleWkb::Polygon(polygon))?;
+                self.append_single_struct(None, SingleWkb::Polygon(polygon))?;
                 Ok(1)
             }
             GeometryType::MultiPoint(multi_point) => {
                 for (index, point) in multi_point.points().enumerate() {
-                    self.append_inner(
-                        parent_path,
-                        Some((index + 1) as _),
-                        SingleWkb::Point(&point),
-                    )?;
+                    self.append_single_struct(Some((index + 1) as _), SingleWkb::Point(&point))?;
                 }
                 Ok(multi_point.num_points() as _)
             }
             GeometryType::MultiLineString(multi_line_string) => {
                 for (index, line_string) in multi_line_string.line_strings().enumerate() {
-                    self.append_inner(
-                        parent_path,
+                    self.append_single_struct(
                         Some((index + 1) as _),
                         SingleWkb::LineString(line_string),
                     )?;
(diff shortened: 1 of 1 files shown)


rust/sedona-functions/src/st_dump.rs
```diff
@@ -0,0 +1,163 @@
...
+}
+
+fn get_geom_list_type(
+    geom_type: Option<&SedonaType>,
+) -> std::result::Result<Option<SedonaType>, datafusion_common::DataFusionError> {
+    match geom_type {
+        Some(SedonaType::Wkb(edges, crs) | SedonaType::WkbView(edges, crs)) => {
+            let geom_type = SedonaType::Wkb(edges.clone(), crs.clone());
+            let geom_list_type =
+                DataType::List(Arc::new(geom_type.to_storage_field("geom", true)?));
```
paleolimbot:
```suggestion
                DataType::List(Arc::new(geom_type.to_storage_field("item", true)?));

```

rust/sedona-functions/src/st_dump.rs
```diff
@@ -0,0 +1,163 @@
...
+    .with_sql_example("SELECT ST_Dump(ST_GeomFromWKT('MULTIPOINT (0 1, 2 3, 4 5)'))")
+    .build()
+}
+
+#[derive(Debug)]
+struct STDump;
+
+impl SedonaScalarKernel for STDump {
+    fn return_type(&self, args: &[SedonaType]) -> Result<Option<SedonaType>> {
+        let matcher = ArgMatcher::new(vec![ArgMatcher::is_geometry()], WKB_GEOMETRY);
```
paleolimbot:
```suggestion
        let matcher = ArgMatcher::new(vec![ArgMatcher::is_geometry()], get_geom_list_type()?);

```

rust/sedona-functions/src/st_dump.rs
```diff
@@ -0,0 +1,163 @@
...
+    fn invoke_batch(
+        &self,
+        arg_types: &[SedonaType],
+        args: &[ColumnarValue],
+    ) -> Result<ColumnarValue> {
+        let executor = WkbExecutor::new(arg_types, args);
+
+        let mut builder = ListBuilder::with_capacity(
+            // TODO: what's the decent default value for the capacity?
+            BinaryBuilder::with_capacity(8, WKB_MIN_PROBABLE_BYTES * 8),
```
paleolimbot: `WKB_MIN_PROBABLE_BYTES * executor.num_iterations()` is probably fine. This will be very close to the input binary array in capacity and size.

rust/sedona-functions/src/st_dump.rs
```diff
@@ -0,0 +1,163 @@
...
+        assert!(st_dump_udf.documentation().is_some());
+    }
+
+    #[rstest]
+    fn udf(#[values(WKB_GEOMETRY, WKB_VIEW_GEOMETRY)] sedona_type: SedonaType) {
+        let tester = ScalarUdfTester::new(st_dump_udf().into(), vec![sedona_type.clone()]);
+
+        let input = create_array(&[Some("POINT (1 2)")], &sedona_type);
+
+        let expected = create_array(&[Some("POINT (1 2)")], &WKB_GEOMETRY);
```
paleolimbot: You're right that this will be pretty verbose to test. This is a unique one but luckily byte-for-byte equality shouldn't be an issue. You can probably use `create_array()` here and assemble a `ListArray` using offsets, then use `assert_eq!()`. The failure message will be terrible but it will get the job done.

rust/sedona-functions/src/st_dump.rs
paleolimbot: DuckDB and PostGIS both agree on the output type here (Sedona is the odd one out). Let's implement the PostGIS/DuckDB version because we can test those (I'll file a bug report in Sedona to find out if that behaviour was intentional).

paleolimbot (commented):
Awesome!

> This crashes R session, so probably I'm doing something wrong.

When the R session crashes, please assume that *I* am the one that caused it 🙂 . This one is probably nanoarrow having a hard time converting an list of an extension type. Try `|> as_nanoarrow_array_stream() |> as_arrow_table()` (the conversion of geoarrow.wkb to extension type attempts to do a zero-copy conversion and that is probably doing something very fishy).

yutannihilation:
Thanks so much for the tip!

rust/sedona-functions/src/st_dump.rs
yutannihilation: Ah, okay. Since I've never found the `path` useful and been tired of always typing `.geom`, I felt Sedona's `ST_Dump` is cool! But, probably you are right. Let's focus on the compatibility.

(I still think SedonaDB / Sedona can have better version of `ST_Dump`, but let's forget for now)

yutannihilation:
Thanks to your help, I think I figured out how to handle and test these complex types. I'll work on the actual implementation...

rust/sedona-functions/src/st_dump.rs
```diff
@@ -0,0 +1,424 @@
...
+    ) -> Result<ColumnarValue> {
+        let executor = WkbExecutor::new(arg_types, args);
+
+        let mut builder = STDumpBuilder::new(executor.num_iterations());
+
+        executor.execute_wkb_void(|maybe_wkb| {
+            if let Some(wkb) = maybe_wkb {
+                let mut struct_builder = builder.struct_builder();
+
+                let mut cur_path: Vec<u32> = Vec::new();
```
paleolimbot: It probably doesn't make much of a difference here but you could declare this outside the loop and reset the size to zero to reuse any previous heap allocation.

yutannihilation:
Thanks for the Python test! I hope I can get started soon with it.

yutannihilation:
Benchmark result:

```
------------------------------------------------------------------------------------------------------------- benchmark: 6 tests -------------------------------------------------------------------------------------------------------------
Name (time in ms)                                                  Min                    Max                   Mean              StdDev                 Median                 IQR            Outliers      OPS            Rounds  Iterations
----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
test_st_dump[collections_simple-SedonaDBSingleThread]          25.8595 (1.0)          28.0133 (1.0)          26.1220 (1.0)        0.4491 (1.0)          25.9570 (1.0)        0.1710 (1.0)           2;5  38.2819 (1.0)          26           1
test_st_dump[collections_simple-DuckDBSingleThread]            54.1694 (2.09)         61.4570 (2.19)         56.3276 (2.16)       2.9164 (6.49)         55.2865 (2.13)       2.1671 (12.68)         1;1  17.7533 (0.46)          5           1
test_st_dump[collections_complex-SedonaDBSingleThread]        382.6457 (14.80)       472.4213 (16.86)       402.0678 (15.39)     39.3457 (87.60)       385.3617 (14.85)     23.3447 (136.55)        1;1   2.4871 (0.06)          5           1
test_st_dump[collections_complex-DuckDBSingleThread]          749.3529 (28.98)       763.6902 (27.26)       757.9570 (29.02)      5.9036 (13.14)       760.3898 (29.29)      9.0438 (52.90)         1;0   1.3193 (0.03)          5           1
test_st_dump[collections_simple-PostGISSingleThread]        3,297.8760 (127.53)    3,372.9212 (120.40)    3,340.3870 (127.88)    32.3133 (71.94)     3,355.1089 (129.26)    53.6658 (313.91)        2;0   0.2994 (0.01)          5           1
test_st_dump[collections_complex-PostGISSingleThread]      37,008.9275 (>1000.0)  37,411.0606 (>1000.0)  37,149.3238 (>1000.0)  184.6306 (411.07)   37,029.6383 (>1000.0)  295.5638 (>1000.0)       1;0   0.0269 (0.00)          5           1
----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
```

rust/sedona-functions/src/st_dump.rs
```diff
@@ -0,0 +1,433 @@
...
+    fn append(
+        &mut self,
+        parent_path: &[u32],
+        cur_index: Option<u32>,
+        wkb: SingleWkb<'_>,
+    ) -> Result<()> {
+        let path_builder = self
+            .struct_builder
+            .field_builder::<ListBuilder<UInt32Builder>>(0)
+            .unwrap();
```
paleolimbot: I am not sure it matters here given that you've nicely demonstrated that this is pretty fast, but we have tried in other places to avoid downcasting in a loop. Something like:

```rust
struct STDumpBuilder {
    path_array_builder: UInt32Builder,
    path_array_offsets_builder: OffsetBufferBuilder,
    geom_builder: BinaryBuilder,
}
```

...and then implement a `finish() -> StructArray` that assembles at the end.

(optional!)

paleolimbot (approved):
This is so cool!

One optional suggestion that you should feel free to punt on (this is also great as is!)

yutannihilation:
Thanks, I simply didn't know! It looks cleaner. I'm trying to switch to that approach, but probably it takes some more time. Currently, I'm struggling to figure out how `append_null()` should be handled, but probably there are several more things that I still don't understand...

yutannihilation:
This should be ready for review again.

rust/sedona-functions/src/st_dump.rs
```diff
@@ -0,0 +1,437 @@
...
+        let struct_offsets = self.struct_offsets_builder.finish();
+        let struct_field = Arc::new(Field::new("item", DataType::Struct(fields), true));
+        let nulls = self.null_builder.finish();
+        ListArray::new(struct_field, struct_offsets, Arc::new(struct_array), nulls)
+    }
+}
+
+impl SedonaScalarKernel for STDump {
+    fn return_type(&self, args: &[SedonaType]) -> Result<Option<SedonaType>> {
+        let matcher = ArgMatcher::new(vec![ArgMatcher::is_geometry()], geometry_dump_type());
```
paleolimbot, line 242: I believe this should to propagate the CRS of args[0] into the `geometry_dump_type()`. The reason this isn't done in any other type matcher code is because the type matcher does this automatically.

yutannihilation: Ah, thanks for catching!

PR #270, opened 2025-11-01
feat(sedona-functions): Implement native ST_NumGeometries using WKBHeader

petern48:
Part of #198

### DIFF
--- a/rust/sedona-geometry/src/wkb_header.rs
+++ b/rust/sedona-geometry/src/wkb_header.rs
@@ -141,6 +141,7 @@ impl WkbHeader {
         self.first_geom_dimensions
     }
 
+    /// Returns true if this geometry is EMPTY or false otherwise
     pub fn is_empty(&self) -> Result<bool, SedonaGeometryError> {
         let geometry_type_id = self.geometry_type_id()?;
         if geometry_type_id == GeometryTypeId::Point {


petern48:
image

rust/sedona-functions/src/st_numgeometries.rs
```diff
@@ -0,0 +1,182 @@
...
+    }
+
+    let geometry_type = header
+        .geometry_type_id()
+        .map_err(|e| DataFusionError::External(Box::new(e)))?;
+    match geometry_type {
+        GeometryTypeId::Point => {
+            // If POINT EMPTY (represented as POINT (NaN, NaN)), return 0
+            let first_xy = header.first_xy();
+            if first_xy.0.is_nan() && first_xy.1.is_nan() {
```
petern48: Look! This came in handy 🤠 (I didn't have a great idea of what you wanted `first_xy` for, but this was neat)

petern48: I was also thinking about pulling out this logic into a new `is_empty` function, since there are other places we'd want this (e.g ST_IsEmpty). I was thinking about something like one of these in `sedona-geometry`

A new method in WkbHeader:
```rust
// WkbHeader
    pub fn is_empty(&self) -> Result<bool>, SedonaGeometryError> {
```

OR a new function in sedona_geometry::is_empty.rs that takes a `WkbHeader`
```rust
// proposed function
pub fn is_geometry_empty_header(
    header: WkbHeader,
) -> Result<bool, SedonaGeometryError> {

// The Existing function
pub fn is_geometry_empty<G: GeometryTrait<T = f64>>(
    geometry: &G,
) -> Result<bool, SedonaGeometryError> {
```

The former is cleaner imo, but I wasn't sure if you'd agree that it makes sense for `WkbHeader` to have that method.

paleolimbot: I think it makes sense for the `WkbHeader` to have an `is_empty()` method!

paleolimbot (approved):
Thank you!

Will be slightly cleaner with the `is_empty()` bit you suggested (feel free to do that here or some other time!)

rust/sedona-geometry/src/wkb_header.rs
```diff
@@ -140,6 +140,15 @@ impl WkbHeader {
     pub fn first_geom_dimensions(&self) -> Option<Dimensions> {
         self.first_geom_dimensions
     }
+
+    pub fn is_empty(&self) -> Result<bool, SedonaGeometryError> {
```
paleolimbot, line 145:
```suggestion
    /// Returns true if this geometry is EMPTY or false otherwise
    pub fn is_empty(&self) -> Result<bool, SedonaGeometryError> {
```

paleolimbot (approved):

Issue #126, opened 2025-09-22
rust/sedona-functions: Add SRID argument to ST_Point, ST_GeomFromText, and ST_GeomFromWkb


paleolimbot:
The "constructor" family of functions accepts SRID arguments in PostGIS; however, our kernels don't support that yet.

We might be able to do this generically with something like

```rust
struct SRIDifiedKernel {
  inner: ScalarKernelRef,
}

impl SedonaScalarKernel for SRIDifiedKernel {
   // Wrap the self.inner implementations but ensure we also accept the extra argument
}
```

We do have `ST_SetSRID()` as a workaround (where we could perhaps lift or integrate the input SRID handling).

PR #275, opened 2025-11-04
feat(rust/sedona-functions): Add SRID argument to ST_Point()

yutannihilation:
Part of https://github.com/apache/sedona-db/issues/126

This pull request implements the `SRIDifiedKernel` wrapper, which is suggested in #126, and applies it to `ST_Point()` to see if it works.

Currently, it seems to work. I need to figure out how to test this because `ScalarUdfTester` doesn't return `SedonaType`.

```
> sd_sql("select st_srid(st_point(1, 1, 4326))")
┌──────────────────────────────────────────────────┐
│ st_srid(st_point(Int64(1),Int64(1),Int64(4326))) │
│                      uint32                      │
╞══════════════════════════════════════════════════╡
│                                             4326 │
└──────────────────────────────────────────────────┘
Preview of up to 6 row(s)
```

### DIFF
--- a/rust/sedona-functions/src/st_setsrid.rs
+++ b/rust/sedona-functions/src/st_setsrid.rs
@@ -27,8 +27,6 @@ use sedona_expr::scalar_udf::{ScalarKernelRef, SedonaScalarKernel, SedonaScalarU
 use sedona_geometry::transform::CrsEngine;
 use sedona_schema::{crs::deserialize_crs, datatypes::SedonaType, matchers::ArgMatcher};
 
-use crate::executor::WkbExecutor;
-
 /// ST_SetSRID() scalar UDF implementation
 ///
 /// An implementation of ST_SetSRID providing a scalar function implementation


rust/sedona-functions/src/st_setsrid.rs
```diff
@@ -227,6 +227,79 @@ fn determine_return_type(
...
+                } else {
+                    validate_crs(&crs, None)?;
+                    deserialize_crs(&serde_json::Value::String(crs))?
+                }
+            }
+            Ok(ScalarValue::Utf8(None)) => None,
+            Ok(_) => {
+                // TODO: This branch is not really the "invalid CRS value" case.
+                //       If it can be cast to Utf-8, it falls into the first branch.
+                return sedona_internal_err!("Invalid CRS value");
```
paleolimbot: Maybe `Can't cast Crs {crs:?} to Utf8`?

rust/sedona-functions/src/st_setsrid.rs
```diff
@@ -227,6 +227,79 @@ fn determine_return_type(
     sedona_internal_err!("Unexpected argument types: {}, {}", args[0], args[1])
 }
 
+#[derive(Debug)]
+pub(crate) struct SRIDifiedKernel {
```
paleolimbot, line 233:
```suggestion
/// [SedonaScalarKernel] wrapper that handles the SRID argument for constructors like ST_Point
#[derive(Debug)]
pub(crate) struct SRIDifiedKernel {
```

(or something better)

rust/sedona-functions/src/st_point.rs
```diff
@@ -247,6 +254,23 @@ mod tests {
...
+        let tester = ScalarUdfTester::new(
+            udf.into(),
+            vec![
+                SedonaType::Arrow(DataType::Float64),
+                SedonaType::Arrow(DataType::Float64),
+                SedonaType::Arrow(DataType::UInt16),
+            ],
+        );
+
+        tester.assert_return_type(WKB_GEOMETRY);
```
paleolimbot: It's a great point that the UDF tester doesn't propagate CRSes because it doesn't consider scalar arguments. We could add `return_type_with_with_scalars()` to the `ScalarUdfTester`?

yutannihilation:
> We could add `return_type_with_with_scalars()` to the `ScalarUdfTester`?

Thanks for the hint! I tried it, but currently I'm seeing this error. I hope if this is just my implementation is not good, but `ScalarUdfTester::return_type()` goes a different code path than actually invoking the function...?

``called `Result::unwrap()` on an `Err` value: NotImplemented("st_point([Arrow(Float64), Arrow(Float64), Arrow(UInt16)]): No kernel matching arguments")``

edit: answer to self. `return_field_from_args()` calls `return_type()`, not `return_type_from_args_and_scalars()`.

https://github.com/apache/datafusion/blob/f32984b2dbf9e5a193c20643ce624167295fbd61/datafusion/expr/src/udf.rs#L628-L637

rust/sedona-functions/src/st_setsrid.rs
```diff
@@ -227,6 +227,78 @@ fn determine_return_type(
...
+
+impl SedonaScalarKernel for SRIDifiedKernel {
+    fn return_type_from_args_and_scalars(
+        &self,
+        args: &[SedonaType],
+        scalar_args: &[Option<&ScalarValue>],
+    ) -> Result<Option<SedonaType>> {
+        let orig_args_len = args.len() - 1;
+        let orig_args = &args[..orig_args_len];
+        let orig_scalar_args = &scalar_args[..orig_args_len];
```
paleolimbot, line 262: I think we may need to check the number of arguments here (return `None` if it's not the expected number) to avoid a panic

rust/sedona-testing/src/testers.rs
```diff
@@ -192,21 +194,84 @@ impl ScalarUdfTester {
...
-            scalar_arguments: &scalar_arguments,
+            scalar_arguments,
         };
-        let return_field = self.udf.return_field_from_args(args)?;
+        let return_field = if from_scalar {
+            let scalar_udf: ScalarUDF = self.udf.clone().into();
+            scalar_udf.return_field_from_args(args)
+        } else {
+            self.udf.return_field_from_args(args)
+        }?;
```
yutannihilation: `ScalarUDF::return_field_from_args()` uses `return_type()` while `SedonaScalarUDF::return_field_from_args()` uses `return_type_from_args_and_scalars()`. So, in theory, something like should would fix the problem, but I'm not sure if I can finish this. This change broke benchmark_utils.rs...

yutannihilation:
> I'm pretty the `SedonaScalarFunction`/`SedonaKernel` handles that correctly.

Yes. It seems the problem was that `ScalarUdfTester` holds `ScalarUDF`, not `SedonaScalarUDF`.

yutannihilation:
Ah, thanks. Got it at last...

yutannihilation:
Okay, I think I figured out. I see two problems. (Sorry I was a bit confused. You are right in that functions around `return_type()` itself works correctly).

First, `ScalarUdfTester::invoke()` calls `ScalarUdfTester::return_type()`, which doesn't consider scalar arguments. So, we need some hatch to specify the return type calculated from the scalar arguments. Fortunately, we have the actual scalar arguments in these `invoke_*()` variants, so this is easy to fix, but a code looks a bit complicated.

https://github.com/apache/sedona-db/blob/964c4dd867f72b723417f63e1660462e0015d497/rust/sedona-testing/src/testers.rs#L406

Second, `ScalarUdfTester::assert_scalar_result_equals()` also calls `return_type()`. In this case, we don't have clue to infer the result type, so probably it needs to be provided from outside.

https://github.com/apache/sedona-db/blob/964c4dd867f72b723417f63e1660462e0015d497/rust/sedona-testing/src/testers.rs#L187

yutannihilation:
I think this is ready for review now, but with one caveat about the differences from PostGIS.

First, it seems PostGIS treats `NULL` CRS and unknown (`0`) CRS differently. I'm not sure where we should tweak in SedonaDB to match this behavior, but I decided not to address it in this pull request. So, I commented out the failing test with some comments.

```
postgres=# SELECT ST_SRID(ST_POINT(1, 1));
 st_srid
---------
       0
(1 row)

postgres=# SELECT ST_SRID(ST_POINT(1, 1, null));
 st_srid
---------

(1 row)
```

Also, another difference I found is that PostGIS doesn't accept CRS with authority e.g. `EPSG:4326` while SedonaDB accepts. I don't think this difference is a problem, but I'm not sure if I should include this in the test cases.

python/sedonadb/tests/functions/test_functions.py
```diff
@@ -1313,6 +1313,31 @@ def test_st_point(eng, x, y, expected):
...
+    ("x", "y", "srid", "expected"),
+    [
+        (None, None, None, None),
+        # TODO: This is a bit tricky, but in PostGIS, NULL and unknown CRS are distinguished.
+        #
+        # - ST_SRID(ST_POINT(x, y, NULL)) returns NULL
+        # - ST_SRID(ST_POINT(x, y, 0)) returns 0
+        # - ST_SRID(ST_POINT(x, y)) returns 0
+        #
+        # (1, 1, None, None),
```
paleolimbot: I think that `ST_SetSRID()` handles this in the same way as PostGIS and it would be helpful to handle that here (I'll leave a suggestion below about how we might do that)

rust/sedona-functions/src/st_setsrid.rs
```diff
@@ -227,6 +227,89 @@ fn determine_return_type(
...
+
+    fn invoke_batch(
+        &self,
+        arg_types: &[SedonaType],
+        args: &[ColumnarValue],
+    ) -> Result<ColumnarValue> {
+        let orig_args_len = arg_types.len() - 1;
+        self.inner
+            .invoke_batch(&arg_types[..orig_args_len], &args[..orig_args_len])
+    }
```
paleolimbot, line 335: I think this is the place where we'd have to check `if let ColumnarValue::Scalar(sc) = args[orig_args_len] { if sc.is_null()`, and perhaps modify the validity buffer of the `inner.invoke_batch().to_array()`. Feel free to punt on that and file a follow-on ticket 🙂

yutannihilation: Ahh, thanks! I didn't notice `ST_POINT(x, y, NULL)` should return `NULL`... I'll fix.

yutannihilation:
> perhaps modify the validity buffer of the `inner.invoke_batch().to_array()`

I tried this approach, but I couldn't find any API that exposes the actual buffer as mutable. So, I chose a different way that skips `invoke_batch()`.

yutannihilation:
(Just a side note)
After I commented above, I started to wonder if it's really fine to skip invoking the actual logic. For example, should `ST_GeomFromText()` reject invalid WKT inputs when the SRID is `NULL`? But, it seems PostGIS also skips any checks, so it should be fine.

```
postgres=# SELECT ST_GeomFromText('point (1 1', null);
 st_geomfromtext 
-----------------

(1 row)

postgres=# SELECT ST_GeomFromText('point (1 1');
ERROR:  parse error - invalid geometry
HINT:  "point (1 1" <-- parse error at position 12 within geometry
```

rust/sedona-functions/src/st_setsrid.rs
```diff
@@ -227,6 +230,105 @@ fn determine_return_type(
...
+        arg_types: &[SedonaType],
+        args: &[ColumnarValue],
+    ) -> Result<ColumnarValue> {
+        let orig_args_len = arg_types.len() - 1;
+        let orig_arg_types = &arg_types[..orig_args_len];
+        let orig_args = &args[..orig_args_len];
+
+        // If the specified SRID is NULL, the result is also NULL. So, return
+        // NULL early and doesn't run `invoke_batch()`.
+        if let ColumnarValue::Scalar(sc) = &args[orig_args_len] {
```
paleolimbot: I think your last point is a good one...we should probably `invoke_batch` first no matter what (to propagate any errors). I don't think that anybody is relying the performance of returning a column full of nulls because they probably made a mistake 🙂

rust/sedona-functions/src/st_setsrid.rs
```diff
@@ -227,6 +230,105 @@ fn determine_return_type(
...
+        //  ^^^^^^^^^^^^^^^
+        //     orig_args
+        let orig_args_len = match (args.len(), scalar_args.len()) {
+            (0, 0) => return Ok(None),
+            (l1, l2) if l1 == l2 => l1 - 1,
+            _ => return sedona_internal_err!("Arg types and arg values have different lengths"),
+        };
+
+        let orig_args = &args[..orig_args_len];
+        let orig_scalar_args = &scalar_args[..orig_args_len];
```
paleolimbot, line 262: I think I understand this better now...this works, although probably `if args.len() == 0 { return Ok(None) }` is sufficient (there are a lot of places in our code where we rely on DataFusion passing us the right number of things). I was worried before that if somebody passed (e.g.,) a single argument to `ST_Point()` something funny would happen here, but I see now that the call to the `inner.return_type_from_args_and_scalars()` will return correctly return `Ok(None)` for that case.

Totally optional, but the error message for something like `ST_Point('gazornenplat')` would probably be better if you moved the call to `inner.return_type_from_args_and_scalars()` before the CRS parsing.

paleolimbot (approved):
Thank you!

I think you're right about propagating the error and there's one small potential improvement; however, this is great and those are unlikely corner cases I'm happy to punt into a future follow-on when we have time.

rust/sedona-functions/src/st_setsrid.rs
yutannihilation: Thanks, sounds good to me.

yutannihilation:
Done! I hope I got what you meant.

paleolimbot (approved):
Thank you!

A quick note that hopefully in the next week or so we'll have item-level CRSes (i.e., for the "srid is an array" case we'll have a separate return type). I don't think that's much code change on top of this but thought I'd put it out there in case it affects anything you're working on!

yutannihilation:
Thanks, I saw the issue about item-level CRSes and was wondering how it relates to here. Looking forward to seeing how it is implemented!

yutannihilation:
Just curious. I think this doesn't happen yet. Was it that you were simply too busy (I guess releasing is a tough job!), or you found some technical difficulty?

> A quick note that hopefully in the next week or so we'll have item-level CRSes

yutannihilation:
Thanks, good to know!

PR #277, opened 2025-11-05
chore: Refactor spatial join code for easier implementing new features

Kontinuation:
This PR refactors the structure of spatial join code to make it easier to add new features, such as out-of-core spatial joins. The main changes are:

* Breaks the largest source file index.rs in rust/sedona-spatial-join into several smaller modules:
  * `evaluated_batch`: Add types for representing evaluated record batches as streams. This is for working with larger-than-memory collected batches later.
  * `index`: Split into spatial index, spatial index builder, build side batches collector, and a knn_adapter module for geometry distance computation.
* Move top-level utility modules into `utils` module
* Get rid of mod.rs and switch to Rust 2018 style modules.

### DIFF
--- a/rust/sedona-spatial-join/src/exec.rs
+++ b/rust/sedona-spatial-join/src/exec.rs
@@ -548,7 +548,7 @@ impl SpatialJoinExec {
                         build_streams.push(stream);
                     }
 
-                    let probe_thread_count = self.right.output_partitioning().partition_count();
+                    let probe_thread_count = probe_plan.output_partitioning().partition_count();
 
                     Ok(build_index(
                         Arc::clone(&context),


rust/sedona-spatial-join/src/index/spatial_index.rs
```diff
@@ -0,0 +1,1487 @@
...
+            geom_idx_vec: Vec::new(),
+            visited_left_side: None,
+            probe_threads_counter,
+            knn_components: KnnComponents::new(0, &[], memory_pool.clone()).unwrap(), // Empty index has no cache
+            reservation,
+        }
+    }
+
+    #[allow(clippy::too_many_arguments)]
+    pub fn new(
```
paleolimbot: It probably makes sense just to drop `new()` at this point (to force construction with named arguments)

rust/sedona-spatial-join/src/index/spatial_index_builder.rs
```diff
@@ -0,0 +1,304 @@
...
+        Self {
+            build_time: MetricBuilder::new(metrics).subset_time("build_time", partition),
+            build_mem_used: MetricBuilder::new(metrics).gauge("build_mem_used", partition),
+        }
+    }
+}
+
+impl SpatialIndexBuilder {
+    /// Create a new builder with the given configuration.
+    pub fn new(
```
paleolimbot, line 102: This is also probably a difficult `new()` call to read

rust/sedona-spatial-join/src/refine/geos.rs
```diff
@@ -321,6 +319,12 @@ impl GeosRefiner {
     }
 }
 
+fn estimate_prep_geom_in_mem_size(wkb: &Wkb<'_>) -> usize {
+    // TODO: This is a rough estimate of the memory usage of the prepared geometry and
+    // may not be accurate.
```
paleolimbot, line 324: Let's track this if it is not considered in this PR:

```suggestion
    // TODO: This is a rough estimate of the memory usage of the prepared geometry and
    // may not be accurate.
    // https://github.com/apache/sedona-db/issues/xxx
```

Kontinuation: Filed #281 and updated comment

rust/sedona-spatial-join/src/index/spatial_index.rs
Kontinuation: Dropped this new() call and marked all fields as `pub(crate)` to be accessible in spatial_index_builder.rs.

rust/sedona-spatial-join/src/index/spatial_index_builder.rs
Kontinuation: I decided to leave it as-is. Two fields of it need to be initialized with default values. I'd like to keep this new method to initialize the fields consistently. I've tried switching to builder pattern. However, applying the builder pattern to a builder also looks strange.

python/sedonadb/python/sedonadb/dbapi.py
```diff
@@ -38,7 +38,7 @@ def connect(**kwargs: Mapping[str, Any]) -> "Connection":
 
         >>> con = sedona.dbapi.connect()
         >>> with con.cursor() as cur:
-        ...     cur.execute("SELECT 1 as one")
+        ...     _ = cur.execute("SELECT 1 as one")
```
Kontinuation, line 41: Fix a doctest failure:

```
=================================== FAILURES ===================================
_______________________ [doctest] sedonadb.dbapi.connect _______________________
031 DBAPI is experimental.
032 
033 Args:
034     kwargs: Extra keyword arguments passed to
035         `adbc_driver_manager.dbapi.Connection()`.
036 
037 Examples:
038 
039     >>> con = sedona.dbapi.connect()
040     >>> with con.cursor() as cur:
Expected:
    [(1,)]
Got:
    <adbc_driver_manager.dbapi.Cursor object at 0x7fd4e20c1e80>
    [(1,)]
```

This is caused by a recent `adbc_driver_manager.dbapi` update: https://github.com/apache/arrow-adbc/commit/7076bee4fc1f30a49668723bcccb7ea25bb19fc8#diff-aac0b59df381ba9a1f9232bd15d9f12610bb80beb238eb51901febba5fabc066R742

The `execute` method now returns Self to enable chaining. This will print an extra message in interactive mode. We suppress this message by assigning the return value of `execute` to `_`.

rust/sedona-spatial-join/src/exec.rs
```diff
@@ -546,24 +543,21 @@ impl SpatialJoinExec {
...
                     let num_partitions = build_side.output_partitioning().partition_count();
                     let mut build_streams = Vec::with_capacity(num_partitions);
-                    let mut build_metrics = Vec::with_capacity(num_partitions);
                     for k in 0..num_partitions {
                         let stream = build_side.execute(k, Arc::clone(&context))?;
                         build_streams.push(stream);
-                        build_metrics.push(SpatialJoinBuildMetrics::new(k, &self.metrics));
                     }
 
                     let probe_thread_count = self.right.output_partitioning().partition_count();
```
martin-g:
```suggestion
                    let probe_thread_count = probe_plan.output_partitioning().partition_count();
```
Isn't this more correct ?

Kontinuation: Thank you for catching this.

Kontinuation: I also found that the properties of SpatialJoinExec may be incorrect for KNN join. For instance, [`maintains_input_order`](https://github.com/apache/sedona-db/blob/2939063fbac28dc06a03e1d18743fa327fbba7ed/rust/sedona-spatial-join/src/exec.rs#L215-L223) assumed that the right side is the probe side, which may not be the case for KNN join. We also need to revise the correctness of `compute_properties` of SpatialJoinExec as well.

python/sedonadb/python/sedonadb/dbapi.py
paleolimbot: Thanks!

rust/sedona-spatial-join/src/exec.rs
paleolimbot: > We also need to revise the correctness of compute_properties of SpatialJoinExec as well.

Should we file an issue?

Kontinuation: I filed https://github.com/apache/sedona-db/issues/282 to track this issue.

PR #283, opened 2025-11-07
feat(python/sedonadb): Implement GDAL/OGR formats via pyogrio

### LINKED ISSUE #137: python/sedonadb: Add pyogrio integration for read/write support for GDAL/OGR
Currently our non-Parquet IO depends on GeoPandas or DuckDB. These are great workarounds but they don't leverage the generic pushdown/pruning capability that DataFusion gives us.

While we could hook directly into GDAL, building, linking, and packaging GDAL isn't something I'd like to do specifically for vector support if there are any alternatives. It's possible that specifically for OGR support we might be able to do something similar to our PROJ support (dynamically pull symbols), but that won't scale beyond a very limited set of operations.

Since GDAL 3.6, an ArrowArrayStream interface has been provided for reading OGR layers. Our DataFrames support arbitrary ArrowArrayStream input, although we would need to modify it to be more resiliant to multiple collects (e.g., today if you try to create a data frame, `.show()` it twice, it will fail the second time because the array stream has already been pulled). Also, we need to wire in pushdown support which GDAL does very well (e.g., using embedded shapefile/fgb/gpkg spatial index).

Proof of concept:

```python
import pyogrio.raw
import sedona.db

sd = sedona.db.connect()

url = "https://raw.githubusercontent.com/geoarrow/geoarrow-data/v0.2.0/natural-earth/files/natural-earth_cities.fgb"
with pyogrio.raw.ogr_open_arrow(f"/vsicurl/{url}", {}) as info:
    meta, reader = info
    print(meta)
    df = sd.create_data_frame(reader).to_memtable()

df.show(5)
#> {'crs': 'EPSG:4326', 'encoding': 'UTF-8', 'fields': array(['name'], dtype=object), 'geometry_type': 'Point', 'geometry_name': '', 'fid_column': 'OGC_FID'}
#> ┌──────────────┬───────────────────────────────┐
#> │     name     ┆          wkb_geometry         │
#> │     utf8     ┆            geometry           │
#> ╞══════════════╪═══════════════════════════════╡
#> │ Vatican City ┆ POINT(12.4533865 41.9032822)  │
#> ├╌╌╌╌╌╌╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌┤
#> │ San Marino   ┆ POINT(12.4417702 43.9360958)  │
#> ├╌╌╌╌╌╌╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌┤
#> │ Vaduz        ┆ POINT(9.5166695 47.1337238)   │
#> ├╌╌╌╌╌╌╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌┤
#> │ Lobamba      ┆ POINT(31.1999971 -26.4666675) │
#> ├╌╌╌╌╌╌╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌┤
#> │ Luxembourg   ┆ POINT(6.1300028 49.6116604)   │
#> └──────────────┴───────────────────────────────┘
```

paleolimbot:
This PR implements GDAL/OGR format reading, which also required some infrastructure. I tried to put as much infrastructure as possible in Rust so that we can reuse it when we implement this in Rust for the CLI/maybe other bindings.

Closes #137

There were a few missing pieces I discovered here (extracting a bounding box from a filter expression, which in turn required the ability to intersect intervals/bounding boxes which we hadn't implemented yet. Anybody should feel free to say the word and I'm happy to PR these separately to help with review.

```python
import sedona.db

sd = sedona.db.connect()
sd.options.interactive = True

url = "https://github.com/geoarrow/geoarrow-data/releases/download/v0.2.0/ns-water_water-line.fgb"
sd.read_pyogrio(url).to_view("water_line")

sd.view("water_line").head(5)
```

Some cool features we support here:

- Bounding box pushdown: the (spatial) query filters are pushed into the datasource. We could potentially push some non-spatial filters but it's less clear if that would actually help (even for spatial filters we might be able to evaluate them faster than GDAL if there's no spatial index on the source or if it's a wide window because we can do it in parallel).
- Projection pushdown: skip reading unused columns
- `https://` and/or `.zip`ped sources using `/vsicurl/` and `/vsizip`/, respectively.
- Multi file support. You can `read_pyogrio("some/dir/*.fgb")` and it should work (and is tested).
- Infrastructure to wire up other file formats via Python. We probably need to extend the implementation a bit to support other formats but it is mostly there. While we don't really want to implement the actual reader in Python (== holding the GIL while reading stuff), using Python to obtain the C ArrowArrayStream should be OK.


petern48:
Closes https://github.com/apache/sedona-db/issues/137 I assume? Just creating a link

rust/sedona-datasource/src/spec.rs
```diff
@@ -182,16 +187,145 @@ impl Object {
...
+                } else if object_store_debug.contains("localfilesystem") {
                     Some(format!("file:///{}", meta.location))
                 } else {
                     None
                 }
             }
-            (Some(url), None) => Some(url.to_string()),
-            (Some(url), Some(meta)) => Some(format!("{url}/{}", meta.location)),
-            (None, None) => None,
+            (Some(url), Some(meta)) => Some(format!("{url}{}", meta.location)),
```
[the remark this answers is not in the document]
paleolimbot: This is tested and as far as I can tell shouldn't have a path separator when it generates the url like this

rust/sedona-expr/src/spatial_filter.rs
```diff
@@ -60,6 +64,44 @@ pub enum SpatialFilter {
...
+                return bounds;
+            }
+            SpatialFilter::LiteralFalse => {
+                return BoundingBox::xy(Interval::empty(), Interval::empty())
+            }
+            SpatialFilter::HasZ(_) | SpatialFilter::Unknown => {}
+        }
+
+        BoundingBox::xy(Interval::full(), Interval::full())
+    }
```
paleolimbot, line 103: @Kontinuation can you sanity check this?

rust/sedona-geometry/src/interval.rs
```diff
@@ -486,6 +506,53 @@ impl IntervalTrait for WraparoundInterval {
...
+                    // The intersection is the complement of the union of excluded regions
+                    Ok(WraparoundInterval::new(
+                        excluded_union.hi(),
+                        excluded_union.lo(),
+                    ))
+                } else {
+                    Err(SedonaGeometryError::Invalid(format!("Can't represent the intersection of {self:?} and {other:?} as a single WraparoundInterval")))
+                }
+            }
+        }
```
paleolimbot, line 558: This is also a bit of non-trivial math that could use a sanity check (the tests below have some ASCII art to help visualize the various cases)

rust/sedona-geometry/src/interval.rs
```diff
@@ -255,6 +265,12 @@ impl IntervalTrait for Interval {
 
         Self::new(self.lo - distance, self.hi + distance)
     }
+
+    fn intersection(&self, other: &Self) -> Result<Self, SedonaGeometryError> {
+        let new_lo = self.lo.max(other.lo);
+        let new_hi = self.hi.min(other.hi);
+        Ok(Self::new(new_lo, new_hi))
+    }
```
Kontinuation, line 277: The result of intersecting two disjoint intervals has `res.is_empty() == false`, if I interpret the code correctly. This could be a problem.

python/sedonadb/python/sedonadb/datasource.py
```diff
@@ -0,0 +1,194 @@
...
+            ogr_src = url.removeprefix("file://")
+        else:
+            raise ValueError(f"Can't open {url} with OGR")
+
+        if ogr_src.endswith(".zip"):
+            ogr_src = f"/vsizip/{ogr_src}"
+
+        if args.is_projected():
+            file_names = args.file_schema.names
+            columns = [file_names[i] for i in args.file_projection]
```
Kontinuation: I believe that `file_names` are field/column names in the file schema. Maybe we can name it `columns`?

PR #284, opened 2025-11-07
refactor(rust/sedona-functions): Use new .buf() method of the wkb crate

yutannihilation:
Thanks to @Kontinuation (cf. https://github.com/georust/wkb/pull/88), the wkb crate now has `.buf()` on all geometry types not only `Wkb`!

This pull request uses it to slightly improve `ST_Dump`. When I implemented `ST_Dump` in https://github.com/apache/sedona-db/pull/269, `.buf()` was not usable for this, so I used `wkb::writer::*`. But, it's probably more efficient to copy a raw WKB rather than newly constructing a WKB.


paleolimbot (approved):
Thank you!

PR #285, opened 2025-11-08
Add an efficient GeometryTypeAndDimensionsSet to improve the performance of geo statistics collection and ST_Analyze_Aggr

Kontinuation:
The idea is pretty simple: we use a 32-bit bitset to represent geometry type and dimensions. This replaces usages of HashSet<GeometryTypeAndDimensions> in `GeoStatistics` and reduces the overhead of updating geo statistics in `AnalyzeAccumulator`. We care about the performance geo statistics analyzer because it is applied to all geometries on the indexed side when running a spatial join. `AnalyzeAccumulator` and ST_Analyze_Aggr can be useful in some other places as well, so we'd like them to have minimal performance overhead.

One of the reasons why `AnalyzeAccumulator` is slow is that `GeoStatistics` has immutable interfaces. We cannot update `GeoStatistics` object directly, but can only obtain a new `GeoStatistics` object that incorporates the change. This introduces lots of clones when updating statistics for batches of geometry objects. The `HashSet<GeometryTypeAndDimensions>` inside `GeoStatistics` is the main performance killer of clone, as it involves memory allocations and deallocations. Switching to GeometryTypeAndDimensionsSet makes the clone more lightweight and does not change the original immutable interfaces.

Here is the benchmark result of st_analyze_aggr after applying this patch:

```
Gnuplot not found, using plotters backend
native-st_analyze_aggr-Array(Point)
                        time:   [4.1267 ms 4.2026 ms 4.3423 ms]
                        change: [-87.458% -87.216% -86.808%] (p = 0.00 < 0.05)
                        Performance has improved.
  6 (6.00%) high mild
  2 (2.00%) high severe

native-st_analyze_aggr-Array(LineString(10))
                        time:   [5.6607 ms 5.6728 ms 5.6868 ms]
                        change: [-83.578% -83.529% -83.482%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 1 outliers among 100 measurements (1.00%)
  1 (1.00%) high severe
```

### DIFF
--- a/rust/sedona-geometry/src/types.rs
+++ b/rust/sedona-geometry/src/types.rs
@@ -293,6 +293,14 @@ impl FromStr for GeometryTypeAndDimensions {
 }
 
 /// A set containing [`GeometryTypeAndDimensions`] values
+///
+/// This set is conceptually similar to `HashSet<GeometryTypeAndDimensions>` but
+/// uses a compact bitset representation for efficiency.
+///
+/// This set only supports the standard dimensions: XY, XYZ, XYM, and XYZM.
+/// Unknown dimensions (other than these four standard types) are not supported
+/// and will be rejected by [`insert`](Self::insert) or silently ignored by
+/// [`insert_or_ignore`](Self::insert_or_ignore).
 #[derive(Clone, Debug, Default, PartialEq, Eq)]
 pub struct GeometryTypeAndDimensionsSet {
     /// Bitset encoding geometry types and dimensions.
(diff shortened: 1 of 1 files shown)


paleolimbot (approved):
Thank you!

For what it's worth, it looks like the last time I did this I used a 64-bit integer and used some of the bits to handle the "unknown" case. Probably erroring like you do here makes more sense.

https://github.com/geoarrow/geoarrow-c/blob/a5c9b0ebb8babb5f719023c86a5c05ece0c36802/src/geoarrow/kernel.c#L359-L366

PR #286, opened 2025-11-08
feat(c/sedona-geos): Add ST_Polygonize() aggregate version

joonaspessi:
Implements `ST_Polygonize` aggregate function following PostGIS behavior. Creates polygons from a collection of linestrings that form closed rings using the GEOS polygonize algorithm.

This is the first GEOS-based aggregate function in the codebase and asking for opinions about the implementation as I'm quite unsure if this is the correct approach.

## How to support `geometry[]` array input?

PostGIS supports two function signatures:
  ```sql
  -- Aggregate form (implemented)
  SELECT ST_Polygonize(geom) FROM table;

  -- Array form (NOT implemented)
  SELECT ST_Polygonize(ARRAY[geom1, geom2, geom3]);
```

Any ideas how to implement this with DafaFusion and keep support for both input styles?
Should we separate the "array form" to another PR?

### DIFF
--- a/c/sedona-geos/src/st_polygonize_agg.rs
+++ b/c/sedona-geos/src/st_polygonize_agg.rs
@@ -33,15 +33,15 @@ use sedona_schema::{
 };
 use wkb::reader::read_wkb;
 
-/// ST_Polygonize() aggregate implementation using GEOS
-pub fn st_polygonize_impl() -> SedonaAccumulatorRef {
-    Arc::new(STPolygonize {})
+/// ST_Polygonize_Agg() aggregate implementation using GEOS
+pub fn st_polygonize_agg_impl() -> SedonaAccumulatorRef {
+    Arc::new(STPolygonizeAgg {})
 }
 
 #[derive(Debug)]
-struct STPolygonize {}
+struct STPolygonizeAgg {}
 
-impl SedonaAccumulator for STPolygonize {
+impl SedonaAccumulator for STPolygonizeAgg {
     fn return_type(&self, args: &[SedonaType]) -> Result<Option<SedonaType>> {
         let matcher = ArgMatcher::new(vec![ArgMatcher::is_geometry()], WKB_GEOMETRY);
         matcher.match_args(args)
@@ -214,8 +214,8 @@ mod tests {
 
     fn create_udf() -> SedonaAggregateUDF {
         SedonaAggregateUDF::new(
-            "st_polygonize",
-            vec![st_polygonize_impl()],
+            "st_polygonize_agg",
+            vec![st_polygonize_agg_impl()],
             datafusion_expr::Volatility::Immutable,
             None,
         )
@@ -225,7 +225,7 @@ mod tests {
     fn udf_metadata() {
         let udf = create_udf();
         let aggregate_udf: AggregateUDF = udf.into();
-        assert_eq!(aggregate_udf.name(), "st_polygonize");
+        assert_eq!(aggregate_udf.name(), "st_polygonize_agg");
     }
 
     #[rstest]
--- a/python/sedonadb/tests/functions/test_aggregate.py
+++ b/python/sedonadb/tests/functions/test_aggregate.py
@@ -19,6 +19,11 @@ import pytest
 from sedonadb.testing import PostGIS, SedonaDB
 
 
+def polygonize_fn_suffix(eng):
+    """Return the appropriate suffix for the polygonize function for the given engine."""
+    return "" if isinstance(eng, PostGIS) else "_Agg"
+
+
 @pytest.mark.parametrize("eng", [SedonaDB, PostGIS])
 def test_st_collect_points(eng):
     eng = eng.create_or_skip()
@@ -120,8 +125,9 @@ def test_st_collect_zero_input(eng):
 @pytest.mark.parametrize("eng", [SedonaDB, PostGIS])
 def test_st_polygonize_basic_triangle(eng):
     eng = eng.create_or_skip()
+    suffix = polygonize_fn_suffix(eng)
     eng.assert_query_result(
-        """SELECT ST_Polygonize(ST_GeomFromText(geom)) FROM (
+        f"""SELECT ST_Polygonize{suffix}(ST_GeomFromText(geom)) FROM (
             VALUES
                 ('LINESTRING (0 0, 10 0)'),
                 ('LINESTRING (10 0, 10 10)'),
@@ -134,8 +140,9 @@ def test_st_polygonize_basic_triangle(eng):
 @pytest.mark.parametrize("eng", [SedonaDB, PostGIS])
 def test_st_polygonize_with_nulls(eng):
     eng = eng.create_or_skip()
+    suffix = polygonize_fn_suffix(eng)
     eng.assert_query_result(
-        """SELECT ST_Polygonize(ST_GeomFromText(geom)) FROM (
+        f"""SELECT ST_Polygonize{suffix}(ST_GeomFromText(geom)) FROM (
             VALUES
                 ('LINESTRING (0 0, 10 0)'),
                 (NULL),
@@ -150,8 +157,9 @@ def test_st_polygonize_with_nulls(eng):
 @pytest.mark.parametrize("eng", [SedonaDB, PostGIS])
 def test_st_polygonize_no_polygons_formed(eng):
     eng = eng.create_or_skip()
+    suffix = polygonize_fn_suffix(eng)
     eng.assert_query_result(
-        """SELECT ST_Polygonize(ST_GeomFromText(geom)) FROM (
+        f"""SELECT ST_Polygonize{suffix}(ST_GeomFromText(geom)) FROM (
             VALUES
                 ('LINESTRING (0 0, 10 0)'),
                 ('LINESTRING (20 0, 30 0)')
@@ -163,8 +171,9 @@ def test_st_polygonize_no_polygons_formed(eng):
 @pytest.mark.parametrize("eng", [SedonaDB, PostGIS])
 def test_st_polygonize_multiple_polygons(eng):
     eng = eng.create_or_skip()
+    suffix = polygonize_fn_suffix(eng)
     eng.assert_query_result(
-        """SELECT ST_Polygonize(ST_GeomFromText(geom)) FROM (
+        f"""SELECT ST_Polygonize{suffix}(ST_GeomFromText(geom)) FROM (
             VALUES
                 ('LINESTRING (0 0, 10 0)'),
                 ('LINESTRING (10 0, 5 10)'),
@@ -205,8 +214,9 @@ def test_st_polygonize_multiple_polygons(eng):
 )
 def test_st_polygonize_single_geom(eng, geom, expected):
     eng = eng.create_or_skip()
+    suffix = polygonize_fn_suffix(eng)
     eng.assert_query_result(
-        f"""SELECT ST_Polygonize(ST_GeomFromText(geom)) FROM (
+        f"""SELECT ST_Polygonize{suffix}(ST_GeomFromText(geom)) FROM (
             VALUES ('{geom}')
         ) AS t(geom)""",
         expected,
--- a/c/sedona-geos/src/lib.rs
+++ b/c/sedona-geos/src/lib.rs
@@ -31,7 +31,7 @@ mod st_isvalid;
 mod st_isvalidreason;
 mod st_length;
 mod st_perimeter;
-mod st_polygonize;
+mod st_polygonize_agg;
 mod st_reverse;
 mod st_simplifypreservetopology;
 mod st_unaryunion;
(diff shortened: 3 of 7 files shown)

### CODE REVIEW

File: python/sedonadb/tests/functions/test_aggregate.py [hunk 1]
```diff
@@ -115,3 +115,63 @@ def test_st_collect_zero_input(eng):
...
+            VALUES
+                ('LINESTRING (0 0, 10 0)'),
+                ('LINESTRING (10 0, 5 10)'),
+                ('LINESTRING (5 10, 0 0)'),
+                ('LINESTRING (20 0, 30 0)'),
+                ('LINESTRING (30 0, 25 10)'),
+                ('LINESTRING (25 10, 20 0)')
+        ) AS t(geom)""",
+        "GEOMETRYCOLLECTION (POLYGON ((10 0, 0 0, 5 10, 10 0)), POLYGON ((30 0, 20 0, 25 10, 30 0)))",
+    )
```
petern48, line 186: I think we should also add these cases too.

```sql
# a single polygon
select st_astext(st_polygonize(array[st_geomfromtext('POLYGON ((10 0, 0 0, 10 10, 10 0))')]));
-- GEOMETRYCOLLECTION(POLYGON((10 0,0 0,10 10,10 0)))

one of each of the other geom types (e.g points, multipoint, etc) and linestring empty
-- GEOMETRYCOLLECTION EMPTY

# A single linestring that's really a LinearRing
select st_astext(st_polygonize(array[st_geomfromtext('linestring(0 0, 0 1, 1 1, 1 0, 0 0)')]));
-- GEOMETRYCOLLECTION(POLYGON((0 0,0 1,1 1,1 0,0 0)))
```

These are using Postgis array syntax. Since we don't have array syntax yet, we could do something like this to avoid having to write a bunch of those polygon:

```python
@pytest.mark.parametrize("eng", [SedonaDB, PostGIS])
@pytest.mark.parametrize(
    ("geom", "expected"),  
def test_st_polygonize_single_geom(eng, geom, expected):  # feel free to come up with a better name
    eng = eng.create_or_skip()
    eng.assert_query_result(
        f"""SELECT ST_Polygonize(ST_GeomFromText(geom)) FROM (
            VALUES
                ('{geom}')
        ) AS t(geom)""",
        expected,
```

Optional: It would be nice to have a few of these cases on the Rust level. If someone makes changes during development, it's nice to have `cargo test` catch buggy changes rather than waiting to occasionally `pytest`.

c/sedona-geos/src/st_polygonize.rs
```diff
@@ -0,0 +1,311 @@
...
+    fn update_batch(&mut self, values: &[ArrayRef]) -> Result<()> {
+        if values.is_empty() {
+            return Err(DataFusionError::Internal(
+                "No input arrays provided to accumulator in update_batch".to_string(),
+            ));
+        }
+
+        let arg_types = [self.input_type.clone()];
+        let args = [ColumnarValue::Array(values[0].clone())];
+        let executor = sedona_functions::executor::WkbExecutor::new(&arg_types, &args);
```
petern48:
```suggestion
        let arg_types = std::slice::from_ref(&self.input_type);
        let args = [ColumnarValue::Array(values[0].clone())];
        let executor = sedona_functions::executor::WkbExecutor::new(arg_types, &args);
```
We can avoid this clone.

c/sedona-geos/src/st_polygonize.rs
```diff
@@ -0,0 +1,311 @@
...
+
+        executor.execute_wkb_void(|maybe_item| {
+            if let Some(item) = maybe_item {
+                self.geometries.push(item.buf().to_vec());
```
petern48: I dug in and found a way to avoid some of these `to_vec()` calls that copy the buffer. It's scattered across the file (and thus hard to convey here in a concise way), so I'm tried something new today and submitted a PR on your fork: https://github.com/joonaspessi/sedona-db/pull/1. Merge if it looks good to you, and it should reflect here in your PR.

paleolimbot: This is a similar idea to ST_Collect()...it may help here (or improvements you find here might be useful there!)

https://github.com/apache/sedona-db/blob/94ce7db5287b39218b803eae057a4eb92a50c397/rust/sedona-functions/src/st_collect.rs#L98-L105

c/sedona-geos/src/st_polygonize.rs
```diff
@@ -0,0 +1,311 @@
...
+            return Ok(None);
+        }
+
+        let mut geos_geoms = Vec::with_capacity(self.geometries.len());
+        for wkb in &self.geometries {
+            let geom = geos::Geometry::new_from_wkb(wkb).map_err(|e| {
+                DataFusionError::Execution(format!("Failed to convert WKB to GEOS: {e}"))
+            })?;
+            geos_geoms.push(geom);
+        }
```
paleolimbot: We should probably use a `GEOSWkbFactory` to create all of these (it's faster than `new_from_wkb()`)

https://github.com/apache/sedona-db/blob/main/c/sedona-geos/src/wkb_to_geos.rs#L24-L29

c/sedona-geos/src/st_polygonize.rs
```diff
@@ -0,0 +1,311 @@
...
+            DataType::List(Arc::new(Field::new("item", DataType::Binary, true))),
+            false,
+        ))])
+    }
+}
+
+#[derive(Debug)]
+struct PolygonizeAccumulator {
+    input_type: SedonaType,
+    geometries: Vec<Vec<u8>>,
```
paleolimbot: Totally optional, but it might simplify our `state()` implementation (and make it more similar to `ST_Collect()`) to use `Vec<u8>` here (in other words, accumulate a geometrycollection).

paleolimbot (commented):
Aggregate functions are hard...thank you!

I am not sure how well DataFusion handles aggregate functions and scalar function that have the same name...we might have to rename this to `ST_Polygonize_Agg()` (but we can evaluate when we get to the scalar functions).

I agree that the scalar function is a separate PR. We don't have any existing functions that accept an array of geometries and so it will probably be verbose to test and you'll have to write your own `TypeMatcher`.

python/sedonadb/tests/functions/test_aggregate.py
joonaspessi: Thanks for the idea of having test cases for single geometries. Tests are now updated for both rust and python suites.

c/sedona-geos/src/st_polygonize.rs
joonaspessi: Thanks for creating the PR, merged it into the branch and I hope that most of the unnecessary copy operations are now gone.

c/sedona-geos/src/st_polygonize.rs
joonaspessi: Tried to think if this would work out. Using single buffer like in `ST_Collect`, would it require then to keep track of offsets and then re-slice the buffer so that we can pass them individually `geos::Geometry::polygonize(&[geom1, geom2, ...])`. If I got it right, it feels little complicated

c/sedona-geos/src/st_polygonize.rs
```diff
@@ -0,0 +1,439 @@
...
+use sedona_expr::aggregate_udf::{SedonaAccumulator, SedonaAccumulatorRef};
+use sedona_schema::{
+    datatypes::{SedonaType, WKB_GEOMETRY},
+    matchers::ArgMatcher,
+};
+use wkb::reader::{read_wkb, Wkb};
+
+thread_local! {
+    static GEOS_WKB_FACTORY: GEOSWkbFactory = GEOSWkbFactory::new();
+}
```
paleolimbot: I don't think you need a thread local factory...we do that in at least one performance-critical place in SedonaDB but here I think it is sufficient to declare it once as a function-level variable before you loop over WKB geometries.

c/sedona-geos/src/st_polygonize.rs
```diff
@@ -0,0 +1,439 @@
...
+        }
+    }
+
+    fn make_wkb_result(&self) -> Result<Option<Vec<u8>>> {
+        if self.geometries.is_empty() {
+            return Ok(None);
+        }
+
+        let mut geos_geoms = Vec::with_capacity(self.geometries.len());
+        for wkb_bytes in &self.geometries {
```
paleolimbot: Right before this loop is where you could declare the function-level `GEOSWkbFactory`. If we run into performance issues with this function we can reevlauate whether a thread local is appropriate.

c/sedona-geos/src/st_polygonize.rs
paleolimbot: You'd have to write it as a valid GEOMETRYCOLLECTION (like ST_Collect()), use GEOS to parse the geometrycollection (once), then pass its children to `polygonize()`. It should be faster for a few reasons (e.g., a `Vec<Vec<u8>>` will perform heap allocations for each element, whereas `Vec<u8>` can reuse the same heap allocation for multiple elements). Optional to handle that here but it would be nice as a precedent for future aggregate function writers.

c/sedona-geos/src/st_polygonize_agg.rs
```diff
@@ -0,0 +1,459 @@
...
+
+        let mut geos_geoms = Vec::with_capacity(num_geoms);
+        for i in 0..num_geoms {
+            let geom = collection.get_geometry_n(i).map_err(|e| {
+                DataFusionError::Execution(format!("Failed to get geometry {}: {e}", i))
+            })?;
+            // Clone is necessary: get_geometry_n() returns ConstGeometry<'_> which doesn't
+            // implement Borrow<Geometry>. The GEOS polygonize() function signature requires
+            // T: Borrow<Geometry>, so we must clone to get owned Geometry instances.
+            geos_geoms.push(geom.clone());
```
joonaspessi, line 116: Couldn't get rid of this :(

c/sedona-geos/src/st_polygonize.rs
```diff
@@ -0,0 +1,457 @@
...
+        Ok(Some(wkb.into()))
+    }
+}
+
+impl Accumulator for PolygonizeAccumulator {
+    fn update_batch(&mut self, values: &[ArrayRef]) -> Result<()> {
+        if values.is_empty() {
+            return Err(DataFusionError::Internal(
+                "No input arrays provided to accumulator in update_batch".to_string(),
+            ));
```
petern48:
```suggestion
            return sedona_internal_err!("No input arrays provided to accumulator in update_batch");
```

nit: We try to avoid using `Datafusion::Internal` because it throws a misleading error message. See here for more detail: https://github.com/apache/sedona-db/pull/2. This is a handy macro we typically use instead.

Here's the import `use sedona_common::sedona_internal_err;`

c/sedona-geos/src/st_polygonize.rs
```diff
@@ -0,0 +1,457 @@
...
+
+        Ok(vec![serialized_count, serialized_item])
+    }
+
+    fn merge_batch(&mut self, states: &[ArrayRef]) -> Result<()> {
+        if states.len() != 2 {
+            return Err(DataFusionError::Internal(format!(
+                "Unexpected number of state fields for st_polygonize() (expected 2, got {})",
+                states.len()
+            )));
```
petern48:
```suggestion
            return sedona_internal_err!(
                "Unexpected number of state fields for st_polygonize() (expected 2, got {:?})",
                states.len()
            );
```

c/sedona-geos/src/st_polygonize.rs
```diff
@@ -0,0 +1,457 @@
...
+        }
+
+        let item_ref = self.item.as_mut().ok_or_else(|| {
+            DataFusionError::Internal("Unexpected internal state in ST_Polygonize()".to_string())
```
petern48:
```suggestion
            sedona_internal_err!("Unexpected internal state in ST_Polygonize()")
```

c/sedona-geos/src/st_polygonize_agg.rs
paleolimbot: This is OK! A heap allocation per output item is better than a heap allocation per input item 🙂

jiayuasu:
We should rename this to `ST_Polygonize_Agg`. Sedona, Snowflake, Databricks all use `_agg` to separate their Spatial SQL aggregate functions from scalar functions.

https://sedona.apache.org/latest/api/sql/AggregateFunction/
https://docs.snowflake.com/en/sql-reference/functions-geospatial
https://docs.databricks.com/aws/en/sql/language-manual/sql-ref-st-geospatial-functions

c/sedona-geos/src/st_polygonize.rs
joonaspessi: Thanks for the suggestion. Should be fixed now

joonaspessi:
@jiayuasu As suggested, renamed to `ST_Polygonize_Agg`. Maybe this makes most sense for both user and implementation perspective.

yutannihilation:
Sorry to nitpick. I think Sedona uses `_Aggr` instead of `_Agg` (although I feel `_Agg` is more familiar...)

jiayuasu:
Yes, I am aware. I plan to change them back to `agg`...

yutannihilation:
Oh, cool. Good to know!

PR #287, opened 2025-11-08
feat(c/sedona-geos): Add ST_ConcaveHull

LakshmiSowmya04:
According to my understanding , which provides native geometric operations via Rust bindings. 

lib.rs -  root that declares the modules compiled into the native library.
register.rs - central registry that maps function names (e.g., "st_convexhull,st_area"...) to their implementation.
test_fucntions.py - we will add a test query to test our file
src/file_name(st_concavehull) -core implementation using GEOS for geometric computation.

it should be like
SQL query -> register.rs -> st_concavehull.rs -> GEOS -> result


LakshmiSowmya04:
@petern48 can you review my PR and let me know if I have to change anything

petern48:
> lib.rs - root that declares the modules compiled into the native library.
register.rs - central registry that maps function names (e.g., "st_convexhull,st_area"...) to their implementation.
test_fucntions.py - we will add a test query to test our file
src/file_name(st_concavehull) -core implementation using GEOS for geometric computation.

> it should be like
SQL query -> register.rs -> st_concavehull.rs -> GEOS -> result

Yes! Everything you've said here looks right. The overall idea of the code looks right at a glance, though I think we need to backtrack a bit to guide you to build up the code iteratively so we avoid all these compile errors. This way, you'll be better set up to contribute more functions with less help in the future. I'll reach back out to you through Discord. It'll go faster / smoother that way.

petern48:
@LakshmiSowmya04 Also, it turns out, our current version of `geos` doesn't have the `concave_hull` methods yet... sorry. Let's leave this PR alone for now, I'd like you to take a shot at ST_NumInteriorRings instead. I confirmed the method already exists, and I think it's a simpler function to understand. I created you a separate issue [here](https://github.com/apache/sedona-db/issues/290).

Also, instead of explaining things through discord, I decided to make a step-by-step PR guide to explain how things work, so that it can potentially help others. It's linked in the issue.

FYI: the note at the bottom of [this comment](https://github.com/apache/sedona-db/pull/288#discussion_r2507384212) is exactly what makes ST_ConcaveHull not a great option for now.

PR #288, opened 2025-11-08
feat(c/sedona-geos): Implement ST_Reverse using `geos`

petern48:
Building a step-by-step walkthrough of how I go about implementing a new `geos` function. For readers, I encourage you to follow me on every step here, including clicking on the links I put and clicking on each commit to see the changes I make at that time.

This guide assumes you have already gone through the developer environment setup [here](https://github.com/apache/sedona-db/blob/94ce7db5287b39218b803eae057a4eb92a50c397/docs/contributors-guide.md).

I strongly suggest you set up the `pre-commit` hook in the [linting section](https://github.com/apache/sedona-db/blob/94ce7db5287b39218b803eae057a4eb92a50c397/docs/contributors-guide.md#linting) in the contributing guide. It's a one-time setup, and it will automatically format your code after each commit. This is a pretty standard tool you'll see in many repos. You shouldn't need to handle formatting nits manually.

### DIFF
--- a/python/sedonadb/tests/functions/test_functions.py
+++ b/python/sedonadb/tests/functions/test_functions.py
@@ -1271,6 +1271,32 @@ def test_st_perimeter(eng, geom, expected):
 
 @pytest.mark.parametrize("eng", [SedonaDB, PostGIS])
 @pytest.mark.parametrize(
+    ("geom", "expected"),
+    [
+        (None, None),
+        ("LINESTRING EMPTY", "LINESTRING EMPTY"),
+        ("LINESTRING(0 0, 1 1, 2 2)", "LINESTRING (2 2, 1 1, 0 0)"),
+        ("POINT (1 2)", "POINT (1 2)"),
+        ("POLYGON ((0 0, 1 0, 2 2, 1 2, 0 0))", "POLYGON ((0 0, 1 2, 2 2, 1 0, 0 0))"),
+        # Note MultiPoints don't change since each point is separate (e.g not a line string)
+        ("MULTIPOINT (1 2, 3 4)", "MULTIPOINT (1 2, 3 4)"),
+        (
+            "MULTIPOLYGON (((0 0, 1 0, 1 1, 0 2, 0 0)), ((5 5, 6 0, 7 1, 0 1, 5 5)))",
+            "MULTIPOLYGON (((0 0, 0 2, 1 1, 1 0, 0 0)), ((5 5, 0 1, 7 1, 6 0, 5 5)))",
+        ),
+        (
+            "GEOMETRYCOLLECTION (POINT (1 2), LINESTRING (3 4, 5 6), POLYGON ((0 0, 0 1, 1 1, 1 0, 0 0)))",
+            "GEOMETRYCOLLECTION (POINT (1 2), LINESTRING (5 6, 3 4), POLYGON ((0 0, 1 0, 1 1, 0 1, 0 0)))",
+        ),
+    ],
+)
+def test_st_reverse(eng, geom, expected):
+    eng = eng.create_or_skip()
+    eng.assert_query_result(f"SELECT ST_Reverse({geom_or_null(geom)})", expected)
+
+
+@pytest.mark.parametrize("eng", [SedonaDB, PostGIS])
+@pytest.mark.parametrize(
     ("x", "y", "expected"),
     [
         (None, None, None),


petern48:
Alright, so first, I pick a function from the corresponding `geos` issue [here](https://github.com/apache/sedona-db/issues/224). In this case, I picked `ST_Reverse`.

petern48:
The first thing I'm going to do is check the corresponding PostGIS docs for ST_Reverse [here](https://postgis.net/docs/ST_Reverse.html), since this is the behavior I want to follow. I found this by simply searching "ST_Reverse PostGIS" on Google.

Looking at the top, we see the function signature.

image

`geometry ST_Reverse(geometry g1);`

The first "geometry" indicates the return type, the second "geometry" indicates the arguments to the function. So in whole, this function takes a single geometry as input and outputs a geometry. Now the best way to go about implementing this whole thing is to find a similar function to model it as. Luckily, I noticed ST_Centroid has the same argument and return types (`geometry ST_Centroid(geometry g1);`), and it's already implemented in SedonaDB here [st_centroid.rs](https://github.com/apache/sedona-db/blob/94ce7db5287b39218b803eae057a4eb92a50c397/c/sedona-geos/src/st_centroid.rs). Now I'll create my new `st_reverse.rs` file in the location `c/sedona-geos/src/` along with the other geos functions. AND I'm going to **COPY-PASTE** the contents of [st_centroid.rs](https://github.com/apache/sedona-db/blob/94ce7db5287b39218b803eae057a4eb92a50c397/c/sedona-geos/src/st_centroid.rs) into it because I know most of the code will be the same. Click on the following commit to see it.

petern48:
Now, this file isn't actually known to the project yet. In fact, my IDE (Cursor/VS Code) and its Rust extension is warning me with white swiggly lines to tell me exactly that.

image

To fix this, I add this new file to the corresponding `lib.rs` inside of the `c/sedona-geos/src` directory. Click the next commit to see the change. After which, the white lines in my IDE are immediately gone.

c/sedona-geos/src/lib.rs
```diff
@@ -31,6 +31,7 @@ mod st_isvalid;
 mod st_isvalidreason;
 mod st_length;
 mod st_perimeter;
+mod st_reverse;
 mod st_simplifypreservetopology;
```
petern48, line 35: Also note, the convention is to put these in alphabetical order. Hence why it's not just at the bottom.

petern48:
In this next commit, I simply change most all "st_centroid" to "st_reverse". Though I haven't yet touched the `invoke_scalar` function, since that's where the core logic is.

c/sedona-geos/src/st_reverse.rs
```diff
@@ -0,0 +1,124 @@
...
+
+        executor.finish(Arc::new(builder.finish()))
+    }
+}
+
+fn invoke_scalar(geos_geom: &geos::Geometry, writer: &mut impl std::io::Write) -> Result<()> {
+    let geometry = geos_geom
+        .get_centroid()
+        .map_err(|e| DataFusionError::Execution(format!("Failed to calculate centroid: {e}")))?;
+
```
petern48, line 77: Alright, **here's the main change of the entire PR**. `invoke_scalar` here is called above to perform the actual operation on each of the inputs rows. It still uses `get_centroid`. I need to figure out how to perform the reverse operation. Because of VS Code's great Rust extension, I can `Ctrl + click` the `get_centroid()` function and my IDE will automatically open the library file this function is located in, which is [here](https://github.com/georust/geos/blob/47afbad2483e489911ddb456417808340e9342c3/src/geometry.rs#L905). If your IDE doesn't do that you can use that link above to go to the file. Now I `Ctrl + f` and search for a reverse function. I find it [here](https://github.com/georust/geos/blob/47afbad2483e489911ddb456417808340e9342c3/src/geometry.rs#L2829). Hooray! there's an easy method that exists that will do all the work for me. In the next commit, I change the `.get_centroid()` call to `.reverse()`.

Notice: STCentroid's function wasn't called `.centroid()`, it was called `.get_centroid()`. Keep that in mind when you're searching for the function: it might not be named exactly what you'd guess. It will likely at least be in `snake_case` rather than `CamelCase`.

**NOTE**: If you don't find a convenient function (which is entirely possible because `geos` Rust might not have been implemented yet), you should probably STOP and try a different function. Unless you're particularly interested in that function, maybe try asking for help.

c/sedona-geos/src/st_reverse.rs
```diff
@@ -104,19 +104,29 @@ mod tests {
...
+            Some("POINT (1 2)"),
+            Some("LINESTRING (1 2, 1 10)"),
+            Some("GEOMETRYCOLLECTION (MULTIPOINT (3 4, 1 2, 7 8, 5 6), LINESTRING (1 10, 1 2))"),
             None,
         ];
 
         let expected = create_array(
-            &[Some("POINT (2.5 2.5)"), Some("POINT EMPTY"), None],
+            &[
+                Some("POLYGON ((2 2, 3 2, 3 3, 2 3, 2 2))"),
```
petern48: In the last commit, I implemented the actual rust tests for ST_Reverse, since previously the tests were left over from when we copy-pasted from ST_Centroid.

Here, the top variable represents the inputs and the bottom represents the expected output. e.g Reading the tests, this means inputting the polygon at line 113 (the first geom of the input array) is expected to return the polygon at 123. This makes sense since the expected result is the same as the input except with it's xy, coords are reversed.

I'm testing this using `cargo test`. I keep going until they pass, and I feel like I've written a few decent cases.

The rest of this is **optional**, but a useful tool to check:  
Oftentimes, I'm running queries against a PostGIS docker container to check what the expected results are for this function (e.g testing edge cases). I'm spinning up a PostGIS docker container using 
```
docker compose up -d
```

And then enter the PostGIS shell (where I can run SQL queries) with this. There's a pretty high chance it's the exact same command for you.

```
```
Now I try some SQL queries (e.g this is the one for that first test case I was talking about above.
```sql
-- next line is my Spatial SQL query
SELECT ST_AsText(ST_Reverse(ST_GeomFromText('POLYGON ((2 2, 2 3, 3 3, 3 2, 2 2))')));
-- Below is the output
           st_astext            
--------------------------------
 POLYGON((2 2,3 2,3 3,2 3,2 2))
```

petern48:
Now our function exists, and the tests pass, but if you actually try it out, you'll see that the function doesn't exist because it's not registered yet.

A convenient way we can try it out is to run SedonaDB's CLI:
```bash
# cd back to the root directory, since I'm still in `c/sedona-geos/src`
cd ../../..
# cd into the cli directory
cd sedona-cli
# Run the CLI
cargo run
```

Now I run the same query I did in PostGIS:
```sql
Sedona CLI v0.2.0
> SELECT ST_AsText(ST_Reverse(ST_GeomFromText('POLYGON ((2 2, 2 3, 3 3, 3 2, 2 2))')));
-- The output: This is a sign our function isn't registered.
Error during planning: Invalid function 'st_reverse'.
Did you mean 'list_reverse'?
```

Registering is simple, just import and add the function to the [register.rs](https://github.com/apache/sedona-db/blob/94ce7db5287b39218b803eae057a4eb92a50c397/c/sedona-geos/src/register.rs#L4) file in `c/sedona-geos/src/` (again the convention is alphabetical order). See the next commit. Afterwards, running the query in the CLI works, and the warnings are gone. This is also needed for the function to work in Python.

Small note: I've been getting small warnings about my code not being used this whole time up until now (e.g see below). Registering the function fixes this too.
image

petern48:
The last step is to write Python integration tests. Here I'm trying to be more comprehensive than the rust tests, covering cases that I think are edge cases. Don't worry too much if you don't know what good tests are. Just write some, push it to your PR, and us reviewers can help you here during code review. Adjust the tests until they pass. Look at surrounding code and potentially similar functions for inspiration of how to write your tests.

To run the tests, I followed the [python instructions](https://github.com/apache/sedona-db/blob/94ce7db5287b39218b803eae057a4eb92a50c397/docs/contributors-guide.md#python) in the contributor guide and run this pytest command to test my exact function. It's based on the file name and test function name.

```
pytest tests/functions/test_functions.py::test_st_reverse
```

If I make changes to rust code, I run this command to update the python environment with the new changes. THIS IS IMPORTANT. If you don't run this command after making change to your rust implementation, the changes won't reflect in python.
```
maturin develop
```
Both of these commands assume you're in the `python/sedonadb/` directory.

petern48:
Once I'm done with my code, I run `cargo clippy` locally to see if there are any minor issues with my code (e.g unused imports), CI will fail and remind you otherwise. I didn't see any, so it looks like I'm good to go.

At this point, I just wait for someone to review my code. Please know, it is totally normal to get a review that suggests some changes. It's typical if you're new to the project, and that's okay! It's how you grow and learn from the reviewer. Hope this helps some people learn enough to contribute their first PR, and grow into contributing more.

paleolimbot (approved):
Thank you for the write-up and the implementation!

PR #291, opened 2025-11-09
docs: Update contributors guide about Windows DLL path

yutannihilation:
This tweaks the contributors guide in two minor points about the Windows case.

First one is that, with the current `PATH`, I get this error when I run `cargo run -p sedona-cli`. I mean, it successfully compiles, but fails to run. It seems this is because I forgot to add a path to vcpkg's DLLs. So, I added one.

```
     Running `target\debug\sedona-cli.exe`
error: process didn't exit successfully: `target\debug\sedona-cli.exe` (exit code: 0xc0000135, STATUS_DLL_NOT_FOUND)
```

Second, `PKG_CONFIG_PATH` refers to `VCPKG_ROOT/installed/x64-windows-dynamic-release/lib/pkgconfig/`, but I guess the triple should be `x64-windows` instead of `x64-windows-dynamic-release` because the guide doesn't specify `$env:VCPKG_DEFAULT_TRIPLET = 'x64-windows-dynamic-release'` (Guessing from the CI setting, this should be used only when building Python?) It seems either works fine, though.

Note that, while this allows me to run the CLI, I'm still failing to load the sedonadb module on Python. I hope we can eventually find how to make it work...

### DIFF
--- a/docs/contributors-guide.md
+++ b/docs/contributors-guide.md
@@ -119,10 +119,13 @@ Configure environment variables (PowerShell example — update paths as needed):
 $env:VCPKG_ROOT = 'C:\dev\vcpkg'
 $env:CMAKE_TOOLCHAIN_FILE = "${env:VCPKG_ROOT}/scripts/buildsystems/vcpkg.cmake"
 
-# pkg-config/ msys path (hash may vary)
-$env:PATH = "${env:VCPKG_ROOT}/downloads/tools/msys2/<msys-hash>/mingw64/bin/;$env:PATH"
+# Add pkg-config/ msys path (hash may vary) for using pkg-config command
+$env:PATH = "${env:VCPKG_ROOT}/downloads/tools/msys2/<msys-hash>/mingw64/bin/;${env:PATH}"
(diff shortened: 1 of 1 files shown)


paleolimbot (approved):
Thank you! I will try to revisit my own Windows build soon to try to sort some of this out. My vague memory from getting the Python wheel setup running was that I basically just ran

https://github.com/apache/sedona-db/blob/94ce7db5287b39218b803eae057a4eb92a50c397/ci/scripts/wheels-build-windows.ps1

...with the Python version set to my local Python version and installed the wheel. (Obviously that is not sustainable for anything except the occasional Windows usage!)

> Guessing from the CI setting, this should be used only when building Python?

We have custom "overlay triplets" that set some specific things related to vcpkg/cmake (concept borrowed from pyogrio).

yutannihilation:
Thanks for the details! Yeah, I first roughly followed that script, and later found I can use vcpkg without the overlay triplets. One difference might be that I use `uv run maturin develop`, but I don't think this matters much.

Anyway, I can use macOS machine for Python tests, so you don't need to worry about me too much :)

PR #292, opened 2025-11-09
added st_num_of_interior_rings file

LakshmiSowmya04:


LakshmiSowmya04:
@petern48 can you review and lmk for changes, i actually added properly but due to git issue i happened to commit lib.rs file seperately. I tested the test cases answer both manually and through postgis through docker.

petern48:
Looks like there's a compile issue in your function (i.e. your code can't run). To see that, here's what I did.

Clicking on the three dots on the right of one of the failed CI runs, you'll see a very helpful error message.

image

Though there are a few more issues than just the one it mentioned. I can tell you now that after fixing the compile error, I still see both the rust and python tests fail. For you to get this over the finish line effectively, you'll need to **run the tests locally on your laptop**, and fix up your errors one by one until they pass. At a courtesy, you should generally at least make sure to run the tests locally and make sure it compiles. (It's totally fine now that you didn't do that past time since you're new and learning.

Let's just start with getting the rust tests to pass, first the rust tests locally with `cargo test` (while being in the `c/sedona-geos/` directory) to see the error message, and figure out how to fix it. Then push that commit. It's a trivial one-line change, but you need to learn how to recognize how to fix it yourself based on the error message. The tests will still fail here, since there are issues in your rust tests too. We'll handle that next.

PR #294, opened 2025-11-09
perf: Use file_metadata_cache in geoparquet

### LINKED ISSUE #250: Use metadata cache when fetching metadata in geoparquet reader
In https://github.com/apache/sedona-db/pull/237 we updated to DataFusion 50, which includes some new features for caching metadata. In our GeoParquet reader we do one or two extra metadata fetches that would benefit greatly from using this! Some of the code we had to copy over doesn't pipe through the global cache.

Upstream:

```rust
DFParquetMetadata::new(store.as_ref(), object)
                    .with_metadata_size_hint(self.metadata_size_hint())
                    .with_decryption_properties(file_decryption_properties.as_ref())
                    .with_file_metadata_cache(Some(Arc::clone(&file_metadata_cache)))
                    .with_coerce_int96(coerce_int96)
                    .fetch_schema_with_location()
                    .await?;
```

Our code:

https://github.com/apache/sedona-db/blob/c67484c88d9b839b96bd7e81219db696c8f64de1/rust/sedona-geoparquet/src/format.rs#L187-L190

petern48:
Use file metadata cache for geoparquet

closes #250

### DIFF
--- a/rust/sedona-geoparquet/src/format.rs
+++ b/rust/sedona-geoparquet/src/format.rs
@@ -35,6 +35,7 @@ use datafusion::{
 use datafusion_catalog::{memory::DataSourceExec, Session};
 use datafusion_common::{plan_err, GetExt, Result, Statistics};
 use datafusion_datasource_parquet::metadata::DFParquetMetadata;
+use datafusion_execution::cache::cache_manager::FileMetadataCache;
 use datafusion_physical_expr::{LexRequirement, PhysicalExpr};
 use datafusion_physical_plan::{
     filter_pushdown::FilterPushdownPropagation, metrics::ExecutionPlanMetricsSet, ExecutionPlan,
@@ -275,7 +276,7 @@ impl FileFormat for GeoParquetFormat {
 
     async fn create_physical_plan(
         &self,
-        _state: &dyn Session,
+        state: &dyn Session,
         config: FileScanConfig,
     ) -> Result<Arc<dyn ExecutionPlan>> {
         // A copy of ParquetSource::create_physical_plan() that ensures the underlying
@@ -291,6 +292,9 @@ impl FileFormat for GeoParquetFormat {
         if let Some(metadata_size_hint) = metadata_size_hint {
             source = source.with_metadata_size_hint(metadata_size_hint)
         }
+
+        source = source
+            .with_file_metadata_cache(state.runtime_env().cache_manager.get_file_metadata_cache());
 
         let conf = FileScanConfigBuilder::from(config)
             .with_source(Arc::new(source))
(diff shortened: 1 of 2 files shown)


petern48:
Well that was suspiciously simple...

rust/sedona-geoparquet/src/format.rs
```diff
@@ -183,11 +183,16 @@ impl FileFormat for GeoParquetFormat {
...
         // store level is the way to go here.
         let metadatas: Vec<_> = futures::stream::iter(objects)
-            .map(|object| async move {
-                DFParquetMetadata::new(store.as_ref(), object)
-                    .with_metadata_size_hint(self.inner().metadata_size_hint())
-                    .fetch_metadata()
-                    .await
+            .map(|object| {
+                let file_metadata_cache =
+                    state.runtime_env().cache_manager.get_file_metadata_cache();
```
petern48, line 195: `with_file_metadata_cache()` is called for each iteration of the loop (`.map()`), we need a clone for each separate iteration. `get_file_metadata_cache()` returns a cloned `Arc`, already, so no need to call another `.clone()`.

https://github.com/apache/datafusion/blob/28755b1d7eb5222a8f5fb5417134dd6865ac1311/datafusion/execution/src/cache/cache_manager.rs#L174-L176

petern48:
It would be nice to see how effective this is, with some benchmarks involving geoparquet reading. That doesn't seem to exist yet, does it?

petern48:
Seems like you're right. Here's what I got in terms of numbers (4 trials each). (Before adding the extra changes)
```
# On branch:
20.807693004608154 seconds. 19.300298929214478 seconds. 30.610652208328247. 20.58704400062561

# On main:
19.879024744033813 seconds. 17.96437692642212 seconds. 21.73752498626709. 17.080259084701538.
```

Need more time to investigate.

petern48:
Found this neat datafusion feature for checking the metadata cache. I'll give it a try once I find the time to circle back to this.
https://github.com/apache/datafusion/pull/17126

petern48:
> Can you also the with_file_metadata_cache() line here?

Tried this before, and ran into some complications when trying to pass the state variable I need there. Datafusion's [opener](https://github.com/apache/datafusion/blob/fb14d7cb1a5eea5b2b2999c1213493d6385447c5/datafusion/datasource-parquet/src/opener.rs#L117) doesn't seem to use the metadata cache state either. I haven't yet looked deeper into what they're doing, but it's probably something.

(As a note to myself) I need to next look into what's happening [here](https://github.com/apache/datafusion/blob/fb14d7cb1a5eea5b2b2999c1213493d6385447c5/datafusion/datasource-parquet/src/opener.rs#L223-L225) with `ArrowReaderMetadata` and `async_file_reader`.

petern48:
this was a while ago, but i propagated the cache to `file_opener` and tried setting the metadata hint, and it continued to result in a decrease in performance. don't quite have time to revist this anytime soon. feel free to take it over if you manage to figure out the problem.

petern48:
Completed in https://github.com/apache/sedona-db/pull/646. Closing

PR #295, opened 2025-11-10
feat(c/sedona-geos): Implement ST_Simplify

### LINKED ISSUE #244: Implement ST_Simplify()
@joonaspessi kindly did some research on the PostGIS implementation of ST_Simplify(), which is not implemented as a thin wrapper around GEOS:

Actually started the implementation of `ST_Simplify` but noticed that the PostGIS has different behavior for collapsed geometries and collection types in comparison to [GEOS implementation](https://libgeos.org/doxygen/classgeos_1_1simplify_1_1DouglasPeuckerSimplifier.html). 

Achieving full PostGIS compatibility (including all the params) would probably require custom Rust implementation.

- Both PostGIS and Sedona-DB uses now GEOS for ST_SimplifyPreserveTopology
    - https://github.com/postgis/postgis/blob/master/postgis/lwgeom_geos.c#L935 - Entry point
    - https://github.com/postgis/postgis/blob/master/postgis/lwgeom_geos.c#L966 - GEOS function call
- PostGIS uses custom implementation for ST_Simplify
   - https://github.com/postgis/postgis/blob/master/postgis/lwgeom_functions_analytic.c#L58 - Entry point
   - https://github.com/postgis/postgis/blob/master/liblwgeom/lwgeom.c#L1851 - Core custom implementation

I wonder if the `geo` implementation of Simplify (if it exists) already ticks some of these boxes...since it is a more involved function to implement, it is probably worth going straight for a custom or geo-generic-alg based approach.

Abeeujah:
Having read both Issue #244 and PR #243 I'm quite interested in knowing or having the test cases that SedonaDB and PostGIS diverge, before I proceed to do a custom implementation, but at the moment, It seems to look fine on my end;
Everything is up for review.

closes #244

### DIFF
--- a/c/sedona-geos/src/st_simplify.rs
+++ b/c/sedona-geos/src/st_simplify.rs
@@ -91,17 +91,30 @@ fn invoke_scalar(
         .simplify(tolerance)
         .map_err(|e| DataFusionError::Execution(format!("Failed to simplify geometry: {e}")))?;
 
+    // GEOS inherently "promotes" a multi-geometry type (e.g., MultiPolygon, MultiLineString)
+    // to its simpler, single-component counterpart (Polygon, LineString) if the
+    // simplification process reduces the number of resulting components to exactly one.
+    //
+    // However, to ensure that the geometry type remains invariant after ST_Simplify
+    // (e.g., an initial MultiPolygon must always return a MultiPolygon), we revert
+    // this promotion by re-wrapping the single component back into its original
+    // multi-geometry container.
     let geometry = match (initial_type, geometry.geometry_type()) {
+        // If the original was a MultiPolygon but GEOS returned a Polygon (promotion),
+        // wrap the Polygon back into a MultiPolygon.
         (GeometryTypes::MultiPolygon, GeometryTypes::Polygon) => {
             Geometry::create_multipolygon(vec![geometry]).map_err(|e| {
                 DataFusionError::Execution(format!("Failed to revert geometry promotion: {e}"))
             })?
         }
+        // If the original was a MultiLineString but GEOS returned a LineString (promotion),
+        // wrap the LineString back into a MultiLineString.
         (GeometryTypes::MultiLineString, GeometryTypes::LineString) => {
             Geometry::create_multiline_string(vec![geometry]).map_err(|e| {
                 DataFusionError::Execution(format!("Failed to revert geometry promotion: {e}"))
             })?
         }
+        // Keep as is (type is correct)
         _ => geometry,
     };
 


c/sedona-geos/src/st_simplify.rs
```diff
@@ -0,0 +1,604 @@
...
+    geos_geom: &geos::Geometry,
+    tolerance: f64,
+    writer: &mut impl std::io::Write,
+) -> Result<()> {
+    let initial_type = geos_geom.geometry_type();
+    let geometry = geos_geom
+        .simplify(tolerance)
+        .map_err(|e| DataFusionError::Execution(format!("Failed to simplify geometry: {e}")))?;
+
+    let geometry = match (initial_type, geometry.geometry_type()) {
```
paleolimbot, line 102:
```suggestion

    // GEOS simplifies MultiPolygon/MultiLineString to Polygon/Linestring; however, we
    // attempt to keep the original geometry type for PostGIS compatibility.
    let geometry = match (initial_type, geometry.geometry_type()) {
```

c/sedona-geos/src/st_simplify.rs
```diff
@@ -0,0 +1,604 @@
...
+    use datafusion_common::ScalarValue;
+    use rstest::rstest;
+    use sedona_expr::scalar_udf::SedonaScalarUDF;
+    use sedona_schema::datatypes::{WKB_GEOMETRY, WKB_VIEW_GEOMETRY};
+    use sedona_testing::{
+        compare::assert_array_equal, create::create_array, testers::ScalarUdfTester,
+    };
+
+    #[rstest]
+    fn udf(#[values(WKB_GEOMETRY, WKB_VIEW_GEOMETRY)] sedona_type: SedonaType) {
```
paleolimbot, line 142: Thank you for all of these test cases! Are there any sources we should acknowledge for them? (May be useful to update our own tests at some point if there are sources you consulted!)

c/sedona-geos/src/st_simplify.rs
paleolimbot: optional: It might be nice to explain why we need the wrapper!

c/sedona-geos/src/st_simplify.rs
petern48: This is a lot of rust tests. FYI, for the future, you can be a little light on Rust tests, since they're kind of verbose. The python tests (which are very concise) is where we try to be comprehensive. e.g I'll often include a single empty geom in Rust tests, while in Python I'll test empty geoms for all types.

There's no need to delete these. Just mentioning it for the future because I know it can be kind of annoying to write so many of these.

petern48 (approved):

c/sedona-geos/src/st_simplify.rs
Abeeujah: > Thank you for all of these test cases! Are there any sources we should acknowledge for them? (May be useful to update our own tests at some point if there are sources you consulted!)

The implementation is based on the Spec defined in [POSTGIS](https://postgis.net/docs/ST_Simplify.html) I run the test cases specified there manually on the docker container, then proceed to assert in code that the behavior is compliant.

Another place I consult for the test cases is Geos test suite [ST_Simplify](https://github.com/libgeos/geos/tree/main/tests/unit/simplify)

Abeeujah: > This is a lot of rust tests. FYI, for the future, you can be a little light on Rust tests, since they're kind of verbose. The python tests (which are very concise) is where we try to be comprehensive. e.g I'll often include a single empty geom in Rust tests, while in Python I'll test empty geoms for all types.

I find it easier to test in Rust, especially as it's strongly typed, for Instance, POSTGIS amd Sedona tend to handle `EMPTY` differently when testing with python, Sedona returns `EMPTY` while POSTGIS returns `None`, would require me including conditional checks in test function to get them to coexist happily, but Rust Serializes and Deserializes it Just fine.

But yeah, I'd be a little light on the Rust tests, and focus testing energy on Python's

paleolimbot: It's useful for us to have the Rust tests because we'd like to implement a faster version at some point, and when we have Rust code we'll need the Rust tests to cover the behaviour. (But yes, the minimum requirement for wrapping external library behaviour is a "is it plugged in" check at the Rust level).

PR #299, opened 2025-11-13
feat(c/sedona-geos): Add ST_Snap()

Abeeujah:
This PR implements `ST_Snap` using the `geos` crate as mentioned in #224


paleolimbot (approved):
Thank you!

PR #300, opened 2025-11-13
Add random tiled raster test data

jesspav:
Separated out the PR for generating tiled random rasters because it was overlapping in multiple tasks (cropping, reading/writing, benchmarking).

Screenshot of the result made by writing the three bands as UInt8 with geotiff and opening with Preview. Note the no-data values in the 4 corners, which will be important for testing.
Screenshot 2025-11-13 at 09 20 57

### DIFF
--- a/rust/sedona-testing/src/rasters.rs
+++ b/rust/sedona-testing/src/rasters.rs
@@ -16,6 +16,7 @@
 // under the License.
 use arrow_array::StructArray;
 use datafusion_common::Result;
+use fastrand::Rng;
 use sedona_raster::array::RasterStructArray;
 use sedona_raster::builder::RasterBuilder;
 use sedona_raster::traits::{BandMetadata, RasterMetadata, RasterRef};
@@ -78,7 +79,12 @@ pub fn generate_tiled_rasters(
     tile_size: (usize, usize),
     number_of_tiles: (usize, usize),
     data_type: BandDataType,
+    seed: Option<u64>,
 ) -> Result<StructArray> {
+    let mut rng = match seed {
+        Some(s) => Rng::with_seed(s),
+        None => Rng::new(),
+    };
     let (tile_width, tile_height) = tile_size;
     let (x_tiles, y_tiles) = number_of_tiles;
     let mut raster_builder = RasterBuilder::new(x_tiles * y_tiles);
@@ -126,6 +132,7 @@ pub fn generate_tiled_rasters(
                     &data_type,
                     nodata_value.as_deref(),
                     corner_position,
+                    &mut rng,
                 );
 
                 raster_builder.band_data_writer().append_value(&band_data);
@@ -173,10 +180,11 @@ fn generate_random_band_data(
     data_type: &BandDataType,
     nodata_bytes: Option<&[u8]>,
     corner_position: Option<usize>,
+    rng: &mut Rng,
 ) -> Vec<u8> {
     match data_type {
         BandDataType::UInt8 => {
-            let mut data: Vec<u8> = (0..pixel_count).map(|_| fastrand::u8(..)).collect();
+            let mut data: Vec<u8> = (0..pixel_count).map(|_| rng.u8(..)).collect();
             // Set corner pixel to nodata value if this tile contains a corner
             if let (Some(nodata), Some(pos)) = (nodata_bytes, corner_position) {
                 if !nodata.is_empty() && pos < data.len() {
@@ -188,7 +196,7 @@ fn generate_random_band_data(
         BandDataType::UInt16 => {
             let mut data = Vec::with_capacity(pixel_count * 2);
             for _ in 0..pixel_count {
-                data.extend_from_slice(&fastrand::u16(..).to_ne_bytes());
+                data.extend_from_slice(&rng.u16(..).to_ne_bytes());
             }
             // Set corner pixel to nodata value if this tile contains a corner
             if let (Some(nodata), Some(pos)) = (nodata_bytes, corner_position) {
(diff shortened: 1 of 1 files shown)


rust/sedona-testing/src/rasters.rs
```diff
@@ -68,6 +69,287 @@ pub fn generate_test_rasters(
...
+            data
+        }
+        BandDataType::UInt16 => {
+            let mut data = Vec::with_capacity(pixel_count * 2);
+            for _ in 0..pixel_count {
+                data.extend_from_slice(&fastrand::u16(..).to_ne_bytes());
+            }
+            // Set corner pixel to nodata value if this tile contains a corner
+            if let (Some(nodata), Some(pos)) = (nodata_bytes, corner_position) {
+                if nodata.len() >= 2 && pos * 2 + 1 < data.len() {
```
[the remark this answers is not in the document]
jesspav: fixed

rust/sedona-testing/src/rasters.rs
```diff
@@ -68,6 +69,287 @@ pub fn generate_test_rasters(
     builder.finish()
 }
 
+/// Generates a set of tiled rasters arranged in a grid
+/// - Each raster tile has specified dimensions and random pixel values
+/// - Each raster has 3 bands which can be interpreted as RGB values
+///   and the result can be visualized as a mosaic of tiles.
+/// - There are nodata values at the 4 corners of the overall mosaic.
+/// - Note that this function is NOT being careful about ensuring that the
+///   tile widths and heights align perfectly with the provided raster size.
```
paleolimbot: We should perhaps expand on this or fix it (maybe check `(raster_size % tile_size) == 0` if this matters? Should it be parameterized as `tile_size` and `num_tiles`?

rust/sedona-testing/src/rasters.rs
```diff
@@ -68,6 +69,287 @@ pub fn generate_test_rasters(
...
+    true
+}
+
+/// Compare two rasters for equality
+/// This compares metadata and band data for equality
+pub fn raster_equal(raster1: &impl RasterRef, raster2: &impl RasterRef) -> bool {
+    // Compare metadata
+    let meta1 = raster1.metadata();
+    let meta2 = raster2.metadata();
+    if meta1.width() != meta2.width()
```
paleolimbot: I think we will be glad if we make this return `false` with a reasonable error message. If this is `assert_raster_equal()`, we could do `assert_eq!(expected_metadta.width(), actual_metadata.width())` which would probably help debug future test failures.

rust/sedona-testing/src/rasters.rs
```diff
@@ -68,6 +69,287 @@ pub fn generate_test_rasters(
...
+/// - Each raster has 3 bands which can be interpreted as RGB values
+///   and the result can be visualized as a mosaic of tiles.
+/// - There are nodata values at the 4 corners of the overall mosaic.
+/// - Note that this function is NOT being careful about ensuring that the
+///   tile widths and heights align perfectly with the provided raster size.
+pub fn generate_tiled_rasters(
+    raster_size: (usize, usize),
+    tile_size: (usize, usize),
+    data_type: BandDataType,
+) -> Result<StructArray, ArrowError> {
```
paleolimbot: This should probably be a `datafusion_common::Result`, which is what we use everywhere else in this crate (your existing `?`s should continue to work, as DataFusion implements `From<ArrowError> for DataFusionError`.

jesspav: thanks. fixed.

rust/sedona-testing/src/rasters.rs
jesspav: agreed. switched.

rust/sedona-testing/src/rasters.rs
jesspav: that works, too. switched.

rust/sedona-testing/src/rasters.rs
```diff
@@ -65,7 +66,317 @@ pub fn generate_test_rasters(
...
+/// Generates a set of tiled rasters arranged in a grid
+/// - Each raster tile has specified dimensions and random pixel values
+/// - Each raster has 3 bands which can be interpreted as RGB values
+///   and the result can be visualized as a mosaic of tiles.
+/// - There are nodata values at the 4 corners of the overall mosaic.
+pub fn generate_tiled_rasters(
+    tile_size: (usize, usize),
+    number_of_tiles: (usize, usize),
+    data_type: BandDataType,
+) -> Result<StructArray> {
```
paleolimbot, line 83: Sorry for missing this in the last round, but there should probably be a `seed` so that the data run through our tests is deterministic.

rust/sedona-testing/src/rasters.rs
```diff
@@ -115,4 +426,84 @@ mod tests {
...
+    #[test]
+    fn test_raster_arrays_equal() {
+        let raster_array1 = generate_test_rasters(3, None).unwrap();
+        let raster_struct_array1 = RasterStructArray::new(&raster_array1);
+        // Test that identical arrays are equal
+        assert_raster_arrays_equal(&raster_struct_array1, &raster_struct_array1);
+    }
+
+    #[test]
+    #[should_panic]
```
paleolimbot:
```suggestion
    #[should_panic = "Raster array lengths do not match"]
```

If this works we should do it (to make sure this test doesn't panic for an unrelated reason). You should be able to paste whatever the panic error is here (I forget if the specified string has to be exact or whether it can be a partial match).

jesspav: good point. fixed

rust/sedona-testing/src/rasters.rs
jesspav: good point. added it.

PR #302, opened 2025-11-14
feat(rust/sedona-raster-functions): Add RS_Height

jesspav:
Add the function `RS_Height`.

Sample run:
```
> Select RS_Height(rs_example());
┌─────────────────────────┐
│ rs_height(rs_example()) │
│          uint64         │
╞═════════════════════════╡
│                      32 │
└─────────────────────────┘
```

Bench:
```
     Running benches/native-raster-functions.rs (target/release/deps/native_raster_functions-b764d86a72aed70f)
native-rs_height-Array(Raster(64, 64))
                        time:   [545.69 µs 546.02 µs 546.33 µs]
  1 (1.00%) low mild
  4 (4.00%) high mild
```

### DIFF
--- a/rust/sedona-raster-functions/Cargo.toml
+++ b/rust/sedona-raster-functions/Cargo.toml
@@ -41,6 +41,7 @@ sedona-schema = { path = "../sedona-schema" }
 [dev-dependencies]
 criterion = { workspace = true }
 sedona-testing = { path = "../sedona-testing", features = ["criterion"] }
+rstest = { workspace = true }
 
 [[bench]]
 harness = false
--- a/rust/sedona-raster-functions/benches/native-raster-functions.rs
+++ b/rust/sedona-raster-functions/benches/native-raster-functions.rs
@@ -20,6 +20,7 @@ use sedona_testing::benchmark_util::{benchmark, BenchmarkArgSpec::*};
 fn criterion_benchmark(c: &mut Criterion) {
     let f = sedona_raster_functions::register::default_function_set();
 
+    benchmark::scalar(c, &f, "native", "rs_height", Raster(64, 64));
     benchmark::scalar(c, &f, "native", "rs_width", Raster(64, 64));
 }
 


rust/sedona/Cargo.toml
```diff
@@ -70,6 +70,7 @@ sedona-geometry = { path = "../sedona-geometry" }
 sedona-geoparquet = { path = "../sedona-geoparquet" }
 sedona-geos = { path = "../../c/sedona-geos", optional = true }
 sedona-proj = { path = "../../c/sedona-proj", default-features = false }
+sedona-raster-functions = { path = "../../rust/sedona-raster-functions", default-features = true }
```
paleolimbot:
```suggestion
sedona-raster-functions = { path = "../../rust/sedona-raster-functions" }
```

(I think default features are always included if this is unspecified?)

rust/sedona-raster-functions/src/rs_size.rs
```diff
@@ -99,28 +141,41 @@ mod tests {
...
+        assert!(udf.documentation().is_some());
     }
 
-    #[test]
-    fn udf_invoke() {
+    #[rstest]
+    fn udf_invoke(#[values(SizeType::Width, SizeType::Height)] st: SizeType) {
+        let kernel = RsSize {
+            size_type: st.clone(),
+        };
```
paleolimbot, line 154: No need to do it in this PR, but before we get too far we should ensure that the `ScalarUdfTester` can handle raster inputs and results (this would cut down quite a bit on the verbosity of this and future tests)

paleolimbot (approved):
A few things to consider now or later!

rust/sedona/Cargo.toml
jesspav: tested and you are correct. thanks. removed.

rust/sedona-raster-functions/src/rs_size.rs
jesspav: Agreed, but I'll do it in a future PR.

paleolimbot (approved):

PR #297, opened 2025-11-12
feat(rust/sedona-testing): Add Raster Benchmarking

jesspav:
Adds raster bench functionality and a first benchmark for the `rs_width`

Issue: #263 

Sample run:
```
     Running benches/native-raster-functions.rs (target/release/deps/native_raster_functions-5e941e84741ac20f)
native-rs_width-Array(Raster(64, 64))
                        time:   [375.11 µs 375.71 µs 376.32 µs]
                        change: [-2.7842% -1.4353% +1.7299%] (p = 0.20 > 0.05)
                        No change in performance detected.
```

PR #306, opened 2025-11-14
perf: Optimize `RasterStructArray::get` to avoid repeatedly extracting arrays from struct array

Kontinuation:
This optimization simply hoists per-column arrays in `RasterStructArray`, so that we don't need to extract individual arrays from the struct array each time we construct a `RasterRefImpl` value in `RasterStructArray::get`. We also marked short getters as inline so that the compiler could optimize away the construction of intermediate objects in simple metadata getter functions such as RS_Width.

I've benchmarked this using the `rs_width` bench in PR https://github.com/apache/sedona-db/pull/297, the performance improvement is quite significant.

Baseline:

```
native-rs_width-Array(Raster(64, 64))
                        time:   [7.3810 ms 7.3912 ms 7.4034 ms]
  5 (5.00%) high mild
  3 (3.00%) high severe
```

This PR:

```
native-rs_width-Array(Raster(64, 64))
                        time:   [474.81 µs 475.81 µs 476.75 µs]
                        change: [-93.593% -93.576% -93.560%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 2 outliers among 100 measurements (2.00%)
  2 (2.00%) high severe
```

### DIFF
--- a/rust/sedona-raster/src/array.rs
+++ b/rust/sedona-raster/src/array.rs
@@ -335,7 +335,11 @@ pub struct RasterRefImpl<'a> {
 }
 
 impl<'a> RasterRefImpl<'a> {
-    /// Create a new RasterRefImpl from a struct array and index using hard-coded indices
+    /// Creates a new RasterRefImpl that provides zero-copy access to the raster at the specified index.
+    ///
+    /// # Arguments
+    /// * `raster_struct_array` - The Arrow StructArray containing raster data
+    /// * `raster_index` - The zero-based index of the raster to access
     #[inline(always)]
     pub fn new(raster_struct_array: &RasterStructArray<'a>, raster_index: usize) -> Self {
         let metadata = MetadataRefImpl {


paleolimbot (approved):
Thank you!

> We marked short getters as inline

For my education, is this always necessary? I had assumed that like C or C++ that if the compiler has access to the sources it will do that inlining automatically, but I truly have no idea.

Kontinuation:
For calling non-generic functions in the same crate, adding `#[inline]` or not does not matter too much, since LLVM inlines code aggressively. For calling non-generic functions in other crates and LTO is not enabled, adding `#[inline]` will enable cross-crate inline for annotated functions; adding `#[inline]` or not will result in a measurable performance difference in this case. Here is the benchmark result I got before adding `#[inline]`:

```
native-rs_width-Array(Raster(64, 64))
                        time:   [1.2776 ms 1.2867 ms 1.3022 ms]
                        change: [-82.709% -82.587% -82.409%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 14 outliers among 100 measurements (14.00%)
  6 (6.00%) high mild
  8 (8.00%) high severe
```

PR #312, opened 2025-11-15
feat(c/sedona-geos): Implement ST_MakeValid()

Abeeujah:
`geos` crate doesn't support `make_valid` recieving additional parms so only the `geometry ST_MakeValid(geometry input);` version is currently implemented

This PR references the discussion in #224

### DIFF
--- a/c/sedona-geos/src/st_makevalid.rs
+++ b/c/sedona-geos/src/st_makevalid.rs
@@ -84,3 +84,78 @@ fn invoke_scalar(geos_geom: &geos::Geometry, writer: &mut impl std::io::Write) -
     writer.write_all(wkb.as_ref())?;
     Ok(())
 }
+
+#[cfg(test)]
+mod tests {
+    use rstest::rstest;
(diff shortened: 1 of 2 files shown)


petern48 (approved):
Adding rust [benches](https://github.com/apache/sedona-db/blob/faa3ed722d7505533cdaca323c3c20d2ff1d6606/c/sedona-geos/benches/geos-functions.rs#L4) would be nice. Looks like you missed it in [ST_Snap](https://github.com/apache/sedona-db/pull/299) too. Adding them is usually straightforward enough that they'll work and CI will pass even if you don't run them yourself.

In reality, they wouldn't be used until someone tries implementing a new kernel for the same function, but it's still a nice-to-have. I don't mind merging without them. 

@paleolimbot Do you still prefer to add them now, or would it make sense to push off adding them if/when a new kernel is implemented?

jiayuasu:
My preference is to add it together with each PR so we don't forget.

PR #317, opened 2025-11-18
feat(rust/sedona-functions): Implement ST_GeometryN

Abeeujah:
This PR implements `ST_GeometryN` using `geo-traits` as discussed in #224 

Benchmarking was not included in this PR due to `BenchmarkArgSpec` not having support for `Int64`, Adding benchmark for this function would require me implementing `Int64` support on `BenchMarkArgSpec` I plan to keep this PR footprint as little as possible, would implement `Int64` support in a follow up PR


c/sedona-geos/src/st_geometryn.rs
```diff
@@ -0,0 +1,201 @@
...
+            .to_array(executor.num_iterations())?;
+        let index_array = as_int64_array(&integer_value)?;
+        let mut index_iter = index_array.iter();
+
+        executor.execute_wkb_void(|maybe_wkb| {
+            match (maybe_wkb, index_iter.next().unwrap()) {
+                (Some(wkb), Some(index)) => {
+                    if invoke_scalar(&wkb, (index - 1) as usize, &mut builder).is_err() {
+                        // Unsupported Geometry Type, Invalid index encountered
+                        builder.append_null();
```
paleolimbot: Instead of using `is_err()` here, maybe `invoke_scalar()` could return something (`true/false`?) to indicate the out-of-bounds case so you can `append_null()` here. (If we swallow the error we might miss something else!)

c/sedona-geos/src/lib.rs
```diff
@@ -26,6 +26,7 @@ mod st_buffer;
 mod st_centroid;
 mod st_convexhull;
 mod st_dwithin;
+mod st_geometryn;
```
paleolimbot: You've very nicely implemented this using geo-traits/Wkb so it can live in `sedona-functions`!

python/sedonadb/tests/functions/test_functions.py
```diff
@@ -1064,6 +1064,79 @@ def test_st_geomfromwkb(eng, geom):
...
 
+@pytest.mark.parametrize("eng", [SedonaDB, PostGIS])
+@pytest.mark.parametrize(
+    ("geom", "index", "expected"),
+    [
+        (
+            "GEOMETRYCOLLECTION(POINT(1 1),MULTIPOLYGON(((0 2,1 1,0 0,0 2)),((2 0,1 1,2 2,2 0))))",
+            1,
+            "POINT (1 1)",
+        ),
```
paleolimbot: It's a bit hard to parse what each of these cases is trying to do. I think grouping them differently would be helpful:

- Cases that result in null output (using compact geometries/n values)
- Test that the n value is handled correctly (the same input geometry but multiple n values)
- Test different geometry types (same n value, different collections)

c/sedona-geos/src/st_geometryn.rs
```diff
@@ -0,0 +1,201 @@
...
+                None,
+                None,
+            ],
+            &WKB_GEOMETRY,
+        );
+
+        let nog = &tester
+            .invoke_arrays(vec![input_wkt.clone(), integers.clone()])
+            .unwrap();
+        println!("{:?}", nog);
```
petern48:
```suggestion
```

python/sedonadb/tests/functions/test_functions.py
```diff
@@ -1064,6 +1064,79 @@ def test_st_geomfromwkb(eng, geom):
...
+        (
+            None,
+            1,
+            None,
+        ),
+        (
+            "MULTIPOINT((0 0))",
+            None,
+            None,
+        ),
```
```suggestion
        ),
        (
            "MULTIPOINT((0 0))",
            -1,
            None
        ),
```
negative index would be nice to test

petern48: Also, nested geom collection would be nice
```
GEOMETRYCOLLECTION (GEOMETRYCOLLECTION(MULTIPOINT (1 2, 3 4)))
```

PR #326, opened 2025-11-21
fix(python/sedonadb): Remove one other problematic test case for wheel builds

paleolimbot:
This PR removes one more case that I missed in #324.


zhangfengcdt:
This is an unambiguous test that could have multiple correct results. I have changed the inputs with this PR (together with another fix) so it is not unambiguous any more. 

https://github.com/apache/sedona-db/pull/330

PR #329, opened 2025-11-22
chore: Update wkb crate dependency to v0.9.2

zhangfengcdt:
This PR upgrade the wkb dependency to version 0.9.2 and also remove the patch.crates-io to it because it is not necessary anymore.


paleolimbot (approved):
Thank you!

PR #354, opened 2025-11-23
chore(ci): ci checks support cargo fmt

caicancai:


paleolimbot (commented):
There's nothing particularly wrong with this, but we already run `cargo fmt` in pre-commit. Is there an advantage to run it in this CI job rather than running it in pre-commit?

caicancai:
> There's nothing particularly wrong with this, but we already run `cargo fmt` in pre-commit. Is there an advantage to run it in this CI job rather than running it in pre-commit?

Does the current CI process include any pre-commit checks? I recall that pre-commit checks usually need to be run locally by the user.

caicancai:
> > Does the current CI process include any pre-commit checks?
> 
> It does! (Happy to keep this if there's some advantage to the other version, though!)

Thank you for your answer. If pre-commit checks for `cargo fmt`, then it seems my CI check is unnecessary.

caicancai:
My approach to setting up this CI pipeline is mainly based on the configurations of some Rust projects I've previously worked on.  Those projects, however, did not include pre-commit checks in their CI pipelines.

PR #360, opened 2025-11-24
feat(rust/sedona-functions): Implement ordering framework that includes geometry/geography

paleolimbot:
This PR adds a function `sd_order()` that returns a proxy whose order is meaningful for a given type. Functionally this just returns its input except for geometry or geography, where it computes the s2 cell ID of the first coordinate. This is cheap and works for anything...we can also add/explicitly order on other reducers like xz2 or geohash but those are potentially more expensive to compute and have more compatibility considerations than I'd like to tackle just this second.

```python
import sedona.db

sd = sedona.db.connect()

f = "https://github.com/geoarrow/geoarrow-data/releases/download/v0.2.0/ns-water_water-junc.fgb"
sd.read_pyogrio(f).to_view("junc", overwrite=True)
sorted = sd.sql(
    "SELECT sd_order(wkb_geometry) as ord, ST_Transform(wkb_geometry, 4326) AS geometry FROM junc"
).to_pandas()

sorted.plot(column='ord', cmap='viridis', markersize=1, scheme='quantiles', k=10)
```

output


c/sedona-geos/src/st_minimumclearance_line.rs
```diff
@@ -113,7 +113,6 @@ mod tests {
             Some("POLYGON((0 0,0 3,3 3,3 0,0 0),(1 1,1 2,2 2,2 1,1 1))"),
             Some("POLYGON((0 0,0 1,0 1,1 1,1 0,0 0,0 0))"),
             Some("LINESTRING (0 0, 1 1, 2 2)"),
-            Some("MULTIPOLYGON(((0.5 0.5,0 0,0 1,0.5 0.5)),((0.5 0.5,1 1,1 0,0.5 0.5)),((2.5 2.5,2 2,2 3,2.5 2.5)),((2.5 2.5,3 3,3 2,2.5 2.5)))"),
```
paleolimbot, line 116: We already have a multipolygon test case here and this test fails for me locally (we've also removed these from Python)

c/sedona-proj/src/st_transform.rs
```diff
@@ -293,7 +293,7 @@ fn invoke_scalar(wkb: &Wkb, trans: &dyn CrsTransform, builder: &mut BinaryBuilde
 fn parse_source_crs(source_type: &SedonaType) -> Result<Option<String>> {
     match source_type {
         SedonaType::Wkb(_, Some(crs)) | SedonaType::WkbView(_, Some(crs)) => {
-            crs.to_authority_code()
+            Ok(Some(crs.to_crs_string()))
```
paleolimbot, line 296: This fixes a bug in `st_transform()`: previously any CRS that was unrepresentable with a single authority/code (e.g., the ns-water test data) would have returned `None` here. Because we have `lenient`, `st_transform()` would have assumed the source data to be lon/lat even though it was correctly annotated.

rust/sedona-functions/src/sd_order.rs
```diff
@@ -0,0 +1,118 @@
...
+use std::{fmt::Debug, sync::Arc};
+
+/// SD_Order() scalar UDF implementation
+///
+/// This function is invoked to obtain a proxy array whose order may be used
+/// to sort based on the value. The default implementation returns the value
+/// and a utility is provided to order geometry and/or geographies based on
+/// the first coordinate. More sophisticated sorting (e.g., XZ2) may be added
+/// in the future.
+pub fn sd_order_udf() -> SedonaScalarUDF {
```
paleolimbot, line 33: This is the main change: we add `sd_order()` whose registered kernels are the means by which to order stuff (that is possibly not meaningfully orderable with the default implementation).

rust/sedona-functions/src/sd_order.rs
```diff
@@ -0,0 +1,118 @@
...
+    .with_sql_example("SELECT SD_Order()")
+    .build()
+}
+
+/// Default implementation that returns its input (i.e., by default, just
+/// do whatever DataFusion would have done with the value)
+#[derive(Debug)]
+struct SDOrderDefault {}
+
+impl SedonaScalarKernel for SDOrderDefault {
```
paleolimbot, line 58: ...where, by default, we just fall back on the storage ordering

c/sedona-proj/src/sd_order_lnglat.rs
```diff
@@ -0,0 +1,205 @@
...
+/// Generic scalar kernel for sd_order based on the first coordinate
+/// of a geometry projected to lon/lat
+///
+/// This [SedonaScalarKernel] requires the actual function (e.g., S2, H3,
+/// or A5 cell identifier) to be provided but takes care of the extraction
+/// of the first coordinate and projecting to lon/lat space. The provided
+/// function must return a `u64`.
+pub struct OrderLngLat<F> {
+    order_fn: F,
+}
```
paleolimbot, line 40: This is the actual ordering implementation from geometry/geography. It's a little awkward because we have to pull in two somewhat disjoint dependencies (something to project to lon/lat, something to compute a value). The projecting needs the global engine so I put it here and parameterized it on the thing that's easier to mock (the actual ordering function).

There are definitely more effective orderings that consider the full geometry (e.g., xz2); however, this one is good as a default because it's fast (doesn't parse or reproject more than one coordinate) and works for geometry, geography, and the forthcoming row-level CRS type. Adding other orderable hashes (e.g., geohash, xz2) is harder because the compatibility with existing behaviour could take some time to replicate.

rust/sedona-schema/src/crs.rs
```diff
@@ -97,10 +97,37 @@ impl PartialEq<dyn CoordinateReferenceSystem + Send + Sync>
 /// A trait defining the minimum required properties of a concrete coordinate
 /// reference system, allowing the details of this to be implemented elsewhere.
 pub trait CoordinateReferenceSystem: Debug {
+    /// Compute the representation of this Crs in the form required for JSON output
+    ///
+    /// The output must be valid JSON (e.g., arbitrary strings must be quoted).
     fn to_json(&self) -> String;
```
paleolimbot, line 103: I noticed these weren't documented so I added some

rust/sedona-schema/src/crs.rs
```diff
@@ -97,10 +97,37 @@ impl PartialEq<dyn CoordinateReferenceSystem + Send + Sync>
...
+    /// (e.g., GeoArrow, Parquet GEOMETRY/GEOGRAPHY) do not make any guarantees
+    /// that an SRID comes from the EPSG authority.
     fn srid(&self) -> Result<Option<u32>>;
+
+    /// Compute a CRS string representation
+    ///
+    /// Unlike `to_json()`, arbitrary string values returned by this method should
+    /// not be escaped. This is the representation expected as input to PROJ, GDAL,
+    /// and Parquet GEOMETRY/GEOGRAPHY representations of CRS.
+    fn to_crs_string(&self) -> String;
```
paleolimbot, line 130: I added this one because we need it for Parquet GEOMETRY/GEOGRAPHY too (in addition to throwing this Crs object into PROJ).

c/sedona-proj/src/st_transform.rs
paleolimbot: cc @jesspav (did I get this part right!?)

c/sedona-geos/src/st_minimumclearance_line.rs
zhangfengcdt: I think i've fixed the python function test by removing the ambiguity of the result:

```
        (
            "MULTIPOLYGON(((0.5 0.5,0 0,0 1,0.5 0.5)),((0.5 0.5,1 1,1 0,0.5 0.5)),((2.5 2.5,2 2,2 3,2.5 2.5)),((2.5 2.5,3.5 3.5,3.5 1.5,2.5 2.5)))",
            "LINESTRING (2.5 2.5, 2 2.5)",
        ),
```

c/sedona-proj/src/sd_order_lnglat.rs
```diff
@@ -0,0 +1,205 @@
...
+        // Extract the source CRS, checking for lon/lat to see if we can avoid
+        // a transformation. If the CRS is missing we also skip any particular
+        // transform (although the resulting sort may not be effective).
+        let maybe_src_crs = match &arg_types[0] {
+            SedonaType::Wkb(_, maybe_crs) | SedonaType::WkbView(_, maybe_crs)
+                if maybe_crs != &lnglat() =>
+            {
+                maybe_crs.as_ref().map(|crs| crs.to_crs_string())
+            }
+            _ => None,
```
zhangfengcdt, line 78: Do empty and null both sort to the same sort position?

paleolimbot: That's a good question...EMPTY sorts to `u64::MAX`, and NULLs sort to NULL, which is potentially configurable.

c/sedona-proj/src/st_transform.rs
jesspav: Returning some here makes sense.  Lenient is false by default, so it should not be relevant unless it is being specified.   I'll double check the code path here.

rust/sedona-schema/src/crs.rs
jesspav: Thanks for adding this, those escapes were annoying. :)

jiayuasu:
1. Why do we call it SD_Order? If this is a UDF, why not ST?
2. Can we add docs together with the PR if possible?

c/sedona-proj/src/st_transform.rs
jesspav: In `st_transform` it would have err'd with `Source CRS is required when transforming to a CRS`, unless somebody had explicitly specified lenient. in either case it's a good fix. thanks.

rust/sedona-functions/src/sd_order.rs
jesspav: Maybe it's just me, but this feels a little awkward.  I would definitely expect the default ordering of `sd_order` to have meaning.

c/sedona-proj/src/st_transform.rs
paleolimbot: Thanks for investigating! I'm glad the failure mode here is an error 👍

rust/sedona-functions/src/sd_order.rs
paleolimbot: That's a great point...I added https://github.com/apache/sedona-db/issues/370 for some improvements. For the most part the ordering that will happen in our release builds for most input will be meaningful (exceptions: `None` crs, missing s2geography, missing proj config) but there are some ways we can improve this!

Issue #305, opened 2025-11-14
Implement ST_Reverse() using geo-traits


paleolimbot:
In https://github.com/apache/sedona-db/pull/288 we implemented ST_Reverse(), which may be faster if we implement it using geo-traits. We should be able to write output incrementally here using `sedona_geometry::wkb_factory`!

petern48:
Closing  as completed by #363. since we have the initial implementation done. I filed a follow-up based on the comments: https://github.com/apache/sedona-db/issues/377

PR #363, opened 2025-11-25
feat(rust/sedona-functions): Implement ST_Reverse using geo-traits

Abeeujah:
This PR implements `ST_Reverse` using `geo-traits` as mentioned in #305 
```sh
Gnuplot not found, using plotters backend
native-st_reverse-Array(Polygon(10))
                        time:   [21.723 ms 21.827 ms 21.946 ms]
Found 13 outliers among 100 measurements (13.00%)
  1 (1.00%) low severe
  2 (2.00%) low mild
  7 (7.00%) high mild
  3 (3.00%) high severe

native-st_reverse-Array(MultiPoint(10))
                        time:   [36.271 ms 36.349 ms 36.429 ms]
  5 (5.00%) low mild
  2 (2.00%) high mild
  1 (1.00%) high severe
  ```

### DIFF
--- a/rust/sedona-functions/src/st_reverse.rs
+++ b/rust/sedona-functions/src/st_reverse.rs
@@ -23,12 +23,12 @@ use datafusion_common::{DataFusionError, Result};
 use datafusion_expr::ColumnarValue;
 use datafusion_expr::{scalar_doc_sections::DOC_SECTION_OTHER, Documentation, Volatility};
 use geo_traits::{
-    CoordTrait, Dimensions, GeometryCollectionTrait, GeometryTrait, LineStringTrait,
-    MultiLineStringTrait, MultiPointTrait, MultiPolygonTrait, PointTrait, PolygonTrait,
+    CoordTrait, GeometryCollectionTrait, GeometryTrait, LineStringTrait, MultiLineStringTrait,
+    MultiPointTrait, MultiPolygonTrait, PointTrait, PolygonTrait,
 };
 use sedona_expr::scalar_udf::{SedonaScalarKernel, SedonaScalarUDF};
 use sedona_geometry::wkb_factory::{
-    write_wkb_coord, write_wkb_empty_point, write_wkb_geometrycollection_header,
+    write_wkb_coord_trait, write_wkb_empty_point, write_wkb_geometrycollection_header,
     write_wkb_linestring_header, write_wkb_multilinestring_header, write_wkb_multipoint_header,
     write_wkb_multipolygon_header, write_wkb_point_header, write_wkb_polygon_header,
     write_wkb_polygon_ring_header, WKB_MIN_PROBABLE_BYTES,
@@ -106,7 +106,8 @@ fn invoke_scalar(geom: &impl GeometryTrait<T = f64>, writer: &mut impl Write) ->
             if pt.coord().is_some() {
                 write_wkb_point_header(writer, dims)
                     .map_err(|e| DataFusionError::Execution(e.to_string()))?;
-                write_coord(writer, pt.coord().unwrap())?;
+                write_wkb_coord_trait(writer, &pt.coord().unwrap())
+                    .map_err(|e| DataFusionError::Execution(e.to_string()))?;
             } else {
                 write_wkb_empty_point(writer, dims)
                     .map_err(|e| DataFusionError::Execution(e.to_string()))?;
@@ -122,12 +123,9 @@ fn invoke_scalar(geom: &impl GeometryTrait<T = f64>, writer: &mut impl Write) ->
         }
 
         geo_traits::GeometryType::LineString(ls) => {
-            let coords: Vec<_> = ls.coords().collect();
-            write_wkb_linestring_header(writer, dims, coords.len())
+            write_wkb_linestring_header(writer, dims, ls.coords().count())
                 .map_err(|e| DataFusionError::Execution(e.to_string()))?;
-            for coord in coords.into_iter().rev() {
-                write_coord(writer, coord)?;
-            }
+            write_reversed_coords(writer, ls.coords())?;
         }
 
         geo_traits::GeometryType::Polygon(pgn) => {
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: c/sedona-geos/src/lib.rs [hunk 1]
```diff
@@ -37,7 +37,6 @@ mod st_minimumclearance_line;
 mod st_perimeter;
 mod st_polygonize;
 mod st_polygonize_agg;
-mod st_reverse;
```
petern48, line 40: I actually think it's fine to leave the `geos` implementation instead of deleting it entirely. We should register things in order so that `sedona-functions` takes precedence. Maybe it would be nice to keep things around for benchmarking purposes? Let's see what Dewey prefers. I can imagine adding a feature one day to allow us to pick which implementation to use easily, which could be nice for benchmarking purposes or maybe even dynamically choosing which implementation to use if it's not as simple as one always being faster. For this case, I do think `geo-traits` is always faster than `geos`.

Taking a look at the code here,

https://github.com/apache/sedona-db/blob/240898d4f48b18821c4190abbeaa1d7c71a4253a/rust/sedona/src/context.rs#L130-L135

It looks like `sedona-functions` is registered before `geos`, making `geos`  implementations take precedence. Shouldn't we move the `sedona-functions` to the bottom, since they are generally the fastest, @paleolimbot?

rust/sedona-functions/benches/native-functions.rs
```diff
@@ -147,6 +147,9 @@ fn criterion_benchmark(c: &mut Criterion) {
         ),
     );
 
+    benchmark::scalar(c, &f, "native", "st_reverse", Polygon(10));
+    benchmark::scalar(c, &f, "native", "st_reverse", MultiPoint(10));
```
petern48, line 151: I noticed your benchmark result in your description doesn't actually compare itself to the old `geos` implementation. Here's what it usually looks like for a performance boost (notice the `-` values).

Image

I ran into this issue before [here](https://github.com/apache/sedona-db/pull/233#issuecomment-3420056856), and pushed off getting back to benchmarking it.

Looking at the [source code](https://github.com/apache/sedona-db/blob/240898d4f48b18821c4190abbeaa1d7c71a4253a/rust/sedona-testing/src/benchmark_util.rs#L88) in benchmark_utils.rs, I think using the lib name is causing the bench util to not compare them. As a quick workaround, I think changing the lib names to be the same would enable the bench tool to group them together properly

```
cd c/sedona-geos/`
cargo bench -- st_reverse

# Manually modify the `"native"` string there to say `"geos"`

cd ../../rust/sedona-functions
cargo bench -- st_reverse
```

Don't actually push that string change. Maybe we can decide how to modify the bench helpers later.

petern48 (commented):
Just some higher-level comments. I haven't yet gone through the implementation, but overall seems good.

This would be a great PR to add some python benchmarks [here](https://github.com/apache/sedona-db/blob/240898d4f48b18821c4190abbeaa1d7c71a4253a/benchmarks/test_functions.py#L4) too to show off even more how fast you're making things 🤠

c/sedona-geos/src/lib.rs
Abeeujah: Thank you for mentioning, wasn't sure if I was to leave it, as the last time I implemented a function using `geo-traits` I had to move it from `sedona-geos` to `sedona-functions` this made me remove the `geos` implementation after I got the `geo-traits` version working, but I'm happy to add it back.

paleolimbot: Peter is right that for all the other ones we kept the original around. I think for this one either is OK because I'm fairly sure that the geo-traits version will be faster and I am not worried about correctness issues popping up (i.e., I'm confident our tests can cover the full range of expected input).

> Shouldn't we move the sedona-functions to the bottom, since they are generally the fastest

That's a great point. I don't think we've had a kernel yet that we implemented in both sedona-functions and sedona-geos, but I think we have to add sedona-functions first for the Documentation (i.e., the functions live there and in theory everything else is just kernels). We could split up sedona-functions so that sedona-geo-traits has the actual kernels and then we register it last; however, I think for today maybe we just keep what this PR did (remove the st_reverse we're pretty sure is slower).

rust/sedona-functions/benches/native-functions.rs
paleolimbot: >  I think using the lib name is causing the bench util to not compare them

That is intentional...if you have the benchmark name be the same then the benchmark harness gets confused and it's easy to make a mistake/hard to know what exactly you just benchmarked. I don't think it's too bad to do the math ourselves.

rust/sedona-functions/src/st_reverse.rs
```diff
@@ -0,0 +1,355 @@
...
+            }
+        }
+
+        geo_traits::GeometryType::LineString(ls) => {
+            let coords: Vec<_> = ls.coords().collect();
+            write_wkb_linestring_header(writer, dims, coords.len())
+                .map_err(|e| DataFusionError::Execution(e.to_string()))?;
+            for coord in coords.into_iter().rev() {
+                write_coord(writer, coord)?;
+            }
```
paleolimbot: Totally optional, but ideally we would query the dimensions exactly once before looping. Maybe:

```rust
pub fn write_wkb_coords_trait<C, I>(buf: &mut impl Write, dims: Dimensions, coords: I) -> Result<(), SedonaGeometryError>
where
    C: CoordTrait<T = f64>,
    I: Iterator<Item = C>,
{
    match dims.size() {
        2 => {
            for coord in coords {
              let coord_tuple = coord.x_y();
              write_wkb_coord(buf, coord_tuple)?;
            }
        }
        3 => {
            for coord in coords {
            let coord_tuple: (<C as CoordTrait>::T, _, _) =
                (coord.x(), coord.y(), coord.nth_or_panic(2));
            write_wkb_coord(buf, coord_tuple)?;
            }
        }
        4 => {
            for coord in coords {
            let coord_tuple = (
                coord.x(),
                coord.y(),
                coord.nth_or_panic(2),
                coord.nth_or_panic(3),
            );
            write_wkb_coord(buf, coord_tuple)?;
            }
        }
        _ => Err(SedonaGeometryError::Invalid(
            "Unsupported number of dimensions".to_string(),
        )),
    }
}
```

Maybe here: https://github.com/apache/sedona-db/blob/240898d4f48b18821c4190abbeaa1d7c71a4253a/rust/sedona-geometry/src/wkb_factory.rs#L395-L401

(Feel free to punt this one...we do exactly what you do here in a few other places as well!)

rust/sedona-functions/src/st_reverse.rs
```diff
@@ -0,0 +1,355 @@
...
+            return Err(DataFusionError::Execution(
+                "Unsupported geometry type for reversal operation".to_string(),
+            ));
+        }
+    }
+    Ok(())
+}
+
+fn write_reversed_ring(writer: &mut impl Write, ring: impl LineStringTrait<T = f64>) -> Result<()> {
+    let coords: Vec<_> = ring.coords().collect();
```
paleolimbot: You should (1) be able to reuse this the LineString case above and (2) you shouldn't have to `.collect()` here (in theory `ring.coords()` is required to implement `DoubleEndedIterator` and `.rev()` should work on it.

rust/sedona-functions/src/st_reverse.rs
```diff
@@ -0,0 +1,355 @@
...
+    write_wkb_polygon_ring_header(writer, coords.len())
+        .map_err(|e| DataFusionError::Execution(e.to_string()))?;
+    for coord in coords.into_iter().rev() {
+        write_coord(writer, coord)?;
+    }
+    Ok(())
+}
+
+fn write_coord(writer: &mut impl Write, coord: impl CoordTrait<T = f64>) -> Result<()> {
+    match coord.dim() {
```
paleolimbot: See the suggestion above (or use `write_wkb_coord_trait()`, which I think is almost identical to this)

rust/sedona-functions/src/st_reverse.rs
```diff
@@ -0,0 +1,355 @@
...
+
+    use super::*;
+
+    #[rstest]
+    fn udf(#[values(WKB_GEOMETRY, WKB_VIEW_GEOMETRY)] sedona_type: SedonaType) {
+        let tester = ScalarUdfTester::new(st_reverse_udf().into(), vec![sedona_type]);
+        tester.assert_return_type(WKB_GEOMETRY);
+
+        let result = tester.invoke_scalar("POINT EMPTY").unwrap();
+        tester.assert_scalar_result_equals(result, "POINT EMPTY");
```
paleolimbot, line 211: We're taking on more responsibility for this implementation, so I think we probably want to be a bit more intentional with these test cases. Maybe:

- Check the LINESTRING scalar case (i.e., check that a scalar call works)
- Check an `array` containing empties of all types (including Z, M, and ZM variants)
- Check an `array` containing XY examples (you have these here!)
- Check an `array` containing XYZ examples
- Check an `array` containing XYM examples
- Check an `array` containing XYZM examples

PR #368, opened 2025-11-26
docs: sql v0.2

MrPowers:
This PR adds API docs for the SQL functions that are added in v0.2.

I still need to get the library setup locally and double-check all the code snippets.  Also need to add some more functions.

### DIFF
--- a/docs/reference/sql.md
+++ b/docs/reference/sql.md
@@ -139,6 +139,8 @@ Returns a geometry that represents all points whose distance from the input geom
 
 Since: v0.1.
 
+Changed in version v0.2: Support buffer parameters argument
+
 #### Format
 
 `ST_Buffer (A: Geometry, distance: Double)`
@@ -608,7 +610,7 @@ SELECT ST_AsText(
 
 #### Description
 
-Construct a Geometry from WKT. This function also has the alias **ST_GeomFromText**.
+Construct a Geometry from WKT. This function also has the aliases **ST_GeomFromText** and ** ST_GeometryFromText**
 
 Since: v0.1.
 
@@ -785,7 +787,7 @@ SELECT ST_IsClosed(ST_GeomFromText('LINESTRING(0 0, 1 1, 1 0)'));
 
 Returns TRUE if the geometry type of the input is a geometry collection type. Collection types are the following:
 
-```sql
+```
 GEOMETRYCOLLECTION
 MULTI{POINT, POLYGON, LINESTRING}
 ```


docs/reference/sql.md
```diff
@@ -110,12 +111,38 @@ Return the Well-Known Text string representation of a geometry or geography.
...
+SELECT ST_Boundary(ST_GeomFromWKT('POLYGON((1 1,0 0, -1 1, 1 1))'))
+```
+
 ## ST_Buffer
 
 #### Description
 
 Returns a geometry that represents all points whose distance from the input geometry is less than or equal to a specified distance.
 
+Since: v0.1.
```
petern48, line 140:
```suggestion
Returns a geometry that represents all points whose distance from the input geometry is less than or equal to a specified distance.

Since: v0.1.

Changed in version v0.2: Support buffer parameters argument
```

Something that might be easy to miss is that we added support for buffer params: https://github.com/apache/sedona-db/pull/241

I always appreciate docs that mention versions when functions were added or received notable updates. More effort on us, but super helpful for developers. Pyspark does "versionchanged" along w/ a short comment like this [example](https://spark.apache.org/docs/latest/api/python/reference/pyspark.sql/api/pyspark.sql.DataFrame.html). I proposed a similar way to do it in the suggestion, but feel free to modify.

Here's a helpful docstring for updating the rest of the buffer docs
https://github.com/apache/sedona-db/blob/79cada0cae30b63ffd26592d8f671adf169fbe2a/c/sedona-geos/src/st_buffer.rs#L37-L47

docs/reference/sql.md
```diff
@@ -516,6 +687,8 @@ SELECT ST_AsText(ST_GeomFromWKB(FROM_HEX('0101000000000000000000F03F000000000000
 
 Construct a Geometry from WKT. This function also has the alias **ST_GeomFromText**.
```
petern48:
```suggestion
Construct a Geometry from WKT. This function also has the aliases **ST_GeomFromText** and ** ST_GeometryFromText**
```
Another thing easy to miss: new alias https://github.com/apache/sedona-db/pull/213

docs/reference/sql.md
MrPowers: Good call, I like "changed in version..."!

MrPowers:
@paleolimbot - this is ready for your review now!

docs/reference/sql.md
paleolimbot: Did you want to apply this?

docs/reference/sql.md
paleolimbot: Did you want to apply this?

docs/reference/sql.md
```diff
@@ -676,7 +766,40 @@ Return true if geomA intersects geomB.
...
+
+```sql
+SELECT ST_IsClosed(ST_GeomFromText('LINESTRING(0 0, 1 1, 1 0)'));
+```
+
+## ST_IsCollection
+
+Returns TRUE if the geometry type of the input is a geometry collection type. Collection types are the following:
+
+```sql
```
paleolimbot:
````suggestion
```
````

paleolimbot (approved):
Thank you!

PR #369, opened 2025-11-26
feat(c/sedona-geos): Add ST_NumInteriorRings

### LINKED ISSUE #290: Implement ST_NumInteriorRings
PostGIS docs: https://postgis.net/docs/ST_NumInteriorRings.html

Step-by-step guide to build the code up and learn along the way. It's pretty comprehensive and wordy, but it explains a lot about how things work: https://github.com/apache/sedona-db/pull/288

Parent Issue: https://github.com/apache/sedona-db/issues/224

LakshmiSowmya04:
closes #290[](https://github.com/apache/sedona-db/issues/290)

This issue is part of [224](https://github.com/apache/sedona-db/issues/224)


c/sedona-geos/src/st_numinteriorrings.rs
```diff
@@ -0,0 +1,131 @@
...
+        let tester = ScalarUdfTester::new(udf.into(), vec![sedona_type.clone()]);
+        tester.assert_return_type(DataType::Int32);
+
+        // scalar-scalar: polygon with two interior rings -> 2
+        let result = tester
+            .invoke_scalar_scalar(
+                "POLYGON((0 0,10 0,10 6,0 6,0 0),(1 1,2 1,2 5,1 5,1 1),(8 5,8 4,9 4,9 5,8 5))", // returns 2
+                "POLYGON((0 0,10 0,10 6,0 6,0 0))", // second arg ignored, only single-arg UDF tester will pass one value, but keep pattern consistent
+            )
+            .unwrap();
```
petern48, line 116: To help point you in the right direction, We have other methods that we could use here instead of [.invoke_scalar_scalar()](https://github.com/apache/sedona-db/blob/90f29892315b593dc4b6efd98e6b337ce5a271a4/rust/sedona-testing/src/testers.rs#L310). That method is meant for passing in two arguments to the SQL function. For ST_NumInteriorRings, we just want to pass in one.

Currently, this is like calling the function like this:
```
ST_NumInteriorRings(
     POLYGON((0 0,10 0,10 6,0 6,0 0),(1 1,2 1,2 5,1 5,1 1),(8 5,8 4,9 4,9 5,8 5))
     POLYGON((0 0,10 0,10 6,0 6,0 0))
)
```
whereas we should just call it with the first arg. (notice how you currently have a comment saying the second arg is ignore.
ST_NumInteriorRings(
     POLYGON((0 0,10 0,10 6,0 6,0 0),(1 1,2 1,2 5,1 5,1 1),(8 5,8 4,9 4,9 5,8 5))
)

We should instead just use a different method than `invoke_scalar_scalar` that only takes one argument.

Try taking a look at the rust tests in `st_area.rs` to model your tests after:
https://github.com/apache/sedona-db/blob/90f29892315b593dc4b6efd98e6b337ce5a271a4/c/sedona-geos/src/st_area.rs#L87-L106

Similar to ST_NumInteriorRings, ST_Area takes a single geometry as input and returns a number.

c/sedona-geos/src/st_numinteriorrings.rs
```diff
@@ -0,0 +1,131 @@
...
+            &[
+                Some("POLYGON((0 0,10 0,10 6,0 6,0 0),(1 1,2 1,2 5,1 5,1 1))"), //returns 1
+                Some("POINT (5 5)"), //returns 0 only out circle is there
+                None,
+            ],
+            &WKB_GEOMETRY,
+        );
+
+        let expected: ArrayRef = arrow_array!(Int32, [Some(1), Some(0), None]);
+        assert_array_equal(&tester.invoke_array_array(lhs).unwrap(), &expected);
```
petern48: Similar issue here I can quickly see in the following error in the [failed cargo test CI](https://github.com/apache/sedona-db/actions/runs/19710765299/job/56470443347?pr=369)
```
error[E0061]: this method takes 2 arguments but 1 argument was supplied
   --> c/sedona-geos/src/st_numinteriorrings.rs:129:36
    |
129 |         assert_array_equal(&tester.invoke_array_array(lhs).unwrap(), &expected);
```

This means we gave `invoke_array_array()` one argument, but it's a method that takes two. When you see an error like this, your IDE probably shows red swiggly lines showing you aren't calling that method properly. You can see the definition [here](https://github.com/apache/sedona-db/blob/90f29892315b593dc4b6efd98e6b337ce5a271a4/rust/sedona-testing/src/testers.rs#L407) (you should be able to figure this out by `Ctrl + click`-ing this on VsCode), it takes two arguments. Similar to above, we only want to pass in one argument, so we actually want to use a different method.

Look around in that [same file](https://github.com/apache/sedona-db/blob/90f29892315b593dc4b6efd98e6b337ce5a271a4/rust/sedona-testing/src/testers.rs#L407C12-L407C30) that `invoke_array_array` is in, and look for a similar function you can use that only takes one (array) argument as input instead of two. Your IDE should stop showing red swiggly lines when you're passing in the right number of arguments to the function.

c/sedona-geos/src/st_numinteriorrings.rs
LakshmiSowmya04: Thank you for review....isn't 
``` POLYGON((0 0,10 0,10 6,0 6,0 0),(1 1,2 1,2 5,1 5,1 1),(8 5,8 4,9 4,9 5,8 5))``` means it taking multiple polygons...its like square inside square and the output came correct when i drew this graph on paper. i checked postGIS docuemntation and likewise you said its showing single polygon as input...

**Comment by petern48** [comment]
Nice, looks like you're making good progress. Fixing some of the rust issues. Now we need to fix the python tests.

Here's a useful tip to help with your local development: Here's how you can execute queries on PostGIS Docker container

1. Start the postgis docker container (for you knowledge it runs this [compose.yml](https://github.com/apache/sedona-db/blob/16d232be0f28441b85801936b270fa43f7edceb4/compose.yml#L4) file.
```
docker compose up -d

# output:
[+] Running 1/1
 ✔ Container sedona-db-postgis-1  Running0.0s
```

2. Assuming it spun up properly, Now you can exec into that to enter the Postgres / PostGIS shell using the container name. It's usually the same for everyone `sedona-db-postgis-1`.
```

# output:
psql (18.0 (Debian 18.0-1.pgdg13+3))
Type "help" for help.

postgres=# 
```

Now you can execute queries locally against PostGIS (which is a correct solution we model SedonaDB after).
```
postgres=# SELECT ST_NumInteriorRings(ST_GeomFromText('POLYGON ((0 0,6 0,6 6,0 6,0 0), (2 2,4 2,4 4,2 4,2 2))'));

-- output:
 st_numinteriorrings 
---------------------
                   1
```

To help with testing you can copy-paste the geometries you're using in your tests to replace that Polygon in my example query. Based on the error in the [last Python CI run](https://github.com/apache/sedona-db/actions/runs/19741057767/job/56564766173?pr=369), it looks like at least one of the geometries is formatted improperly. You can use PostGIS to iterate and fix it faster.

```
FAILED sedonadb/tests/functions/test_functions.py::test_st_numinteriorrings_basic[POLYGON (                (0 0,10 0,10 6,0 6,0 0)                (1 1,2 1,2 5,1 5,1 1),                (8 5,8 4,9 4,9 5,8 5))-2-SedonaDB] - pyarrow.lib.ArrowInvalid: External error: Internal error: WKT parse error: Missing closing parenthesis for type.
```

You can see which polygon is problematic there in the error message. What I would do, is spin up the postgis container, and try running the SQL query I wrote above on each of the polygons in your tests. You're probably missing a parentheses somewhere or have an extra one.

File: python/sedonadb/tests/functions/test_functions.py [hunk 3]
```diff
@@ -2698,3 +2698,49 @@ def test_st_snap(eng, input, reference, tolerance, expected):
...
+        (None, None),  # none
+        ("POINT (1 2)", None),  # none
+        ("LINESTRING (0 0, 1 1, 2 2)", None),  # non-polygon -> NULL
+        ("POLYGON EMPTY", 0),  # empty polygon has no interior rings
+        ("POLYGON ((0 0, 4 0, 4 4, 0 4, 0 0))", 0),  # simple polygon, no holes
+        # example query - SELECT ST_NumInteriorRings(ST_GeomFromText('POLYGON((0 0,4 0,4 4,0 4,0 0))')) AS num_rings;
+        (
+            "POLYGON (\
+                (0 0,6 0,6 6,0 6,0 0),\
+                (2 2,4 2,4 4,2 4,2 2))",
```
```suggestion
            "POLYGON ((0 0,6 0,6 6,0 6,0 0), (2 2,4 2,4 4,2 4,2 2))"
```

It's slightly more convenient to write geometries on a single line. It makes it easier to copy-paste for testing on the PostGIS container I mentioned above. I suggest doing this with your other python tests to help you debug easier. This way, you also won't have to worry about backslashes and issues with multiline python strings.

c/sedona-geos/src/st_numinteriorrings.rs
```diff
@@ -0,0 +1,174 @@
...
+        let executor = GeosExecutor::new(arg_types, args);
+        let mut builder = Int32Builder::with_capacity(executor.num_iterations());
+        executor.execute_wkb_void(|maybe_geom| {
+            match maybe_geom {
+                None => builder.append_null(),
+                Some(geom) => {
+                    let res = invoke_scalar(&geom)?;
+                    match res {
+                        Some(n) => builder.append_value(n),
+                        // non-polygon / unsupported -> NULL (matches PostGIS + py tests)
```
petern48:
```suggestion
```
Just clean up. It's not wrong. I just don't think it's necessary to say it passes the postgis python tests. There's already comments below explaining the non-polygon case.

python/sedonadb/tests/functions/test_functions.py
```diff
@@ -2698,3 +2698,45 @@ def test_st_snap(eng, input, reference, tolerance, expected):
...
+        (
+            "POLYGON ((0 0,6 0,6 6,0 6,0 0),(2 2,4 2,4 4,2 4,2 2))",
+            1,
+        ),
+        (
+            "POLYGON ("
+            "(0 0,10 0,10 6,0 6,0 0),"
+            "(1 1,2 1,2 5,1 5,1 1),"
+            "(8 5,8 4,9 4,9 5,8 5)"
+            ")",
```
```suggestion
            "POLYGON ((0 0,10 0,10 6,0 6,0 0), (1 1,2 1,2 5,1 5,1 1),(8 5,8 4,9 4,9 5,8 5))",
```
Again, I slightly prefer to have these all on one line for the ease of copy-pasting. There's one more multiline case that's right below this to fix.

python/sedonadb/tests/functions/test_functions.py
```diff
@@ -2698,3 +2698,45 @@ def test_st_snap(eng, input, reference, tolerance, expected):
...
+            2,
+        ),
+        (
+            "MULTIPOLYGON ("
+            "((0 0, 5 0, 5 5, 0 5, 0 0), (1 1, 2 1, 2 2, 1 2, 1 1)),"
+            "((10 10, 14 10, 14 14, 10 14, 10 10))"
+            ")",
```
```suggestion
            "MULTIPOLYGON (((0 0, 5 0, 5 5, 0 5, 0 0), (1 1, 2 1, 2 2, 1 2, 1 1)), ((10 10, 14 10, 14 14, 10 14, 10 10)))",
```

python/sedonadb/tests/functions/test_functions.py
```diff
@@ -2698,3 +2698,45 @@ def test_st_snap(eng, input, reference, tolerance, expected):
...
+
+
+@pytest.mark.parametrize("eng", [SedonaDB, PostGIS])
+@pytest.mark.parametrize(
+    ("geom", "expected"),
+    [
+        (None, None),
+        ("POINT (1 2)", None),
+        ("LINESTRING (0 0, 1 1, 2 2)", None),
+        ("POLYGON EMPTY", 0),
```
petern48: Great job with the tests so far 👏. I'd like the Python tests to be a bit more comprehensive because it's very possible we implement ways to speed up ST_NumInteriorRings in the future. When we do, so it's important that we have good tests to make sure we don't introduce bugs.

Here are the tests I'd like to have:
- None (done)
- Polygon with one ring (zero interior rings) (done)
- Polygon with more rings (done)
- Empty geom for each geometry type (not done)
- One of each geometry type (not done)
  - All 7 of the geometry types are: `POINT`, `LINESTRING`, `POLYGON`, `MULTIPOINT`, `MULTILINESTRING`, `MULTIPOLYGON`, `GEOMETRYCOLLECTION`. You have most of these, but not quite all of them

You cover most of these (indicated by "done"), but not the last two bullets. Could you add cases for the last two bullets? Here's an example of how it's done for another function. You can use similar (or even the same) test cases, except your "expected" value will be different than the True/False values in that function.

https://github.com/apache/sedona-db/blob/16d232be0f28441b85801936b270fa43f7edceb4/python/sedonadb/tests/functions/test_functions.py#L1276-L1290

optional: It's also nice to group the tests by bullet (e.g. all empty cases together), for readability.

c/sedona-geos/src/st_numinteriorrings.rs
```diff
@@ -0,0 +1,174 @@
...
+            Some("POLYGON EMPTY"),
+            Some("POLYGON ((0 0, 4 0, 4 4, 0 4, 0 0))"),
+            Some("POLYGON ((0 0,6 0,6 6,0 6,0 0),(2 2,4 2,4 4,2 4,2 2))"),
+            Some(
+                "POLYGON (
+                    (0 0,10 0,10 6,0 6,0 0),
+                    (1 1,2 1,2 5,1 5,1 1),
+                    (8 5,8 4,9 4,9 5,8 5)
+                )",
+            ),
```
petern48, line 131: For copy-paste ease (and conciseness), I prefer to collapse these rust ones to a single line too. Two more cases below this and one more at the beginning of the function

petern48:
Could you also modify your PR description to say "closes <issue link>". GitHub has a feature to auto-close the issue once the PR is merged when you put that.

"closes https://github.com/apache/sedona-db/issues/290"

LakshmiSowmya04:
@petern48 my mkdocs.yml is

python/sedonadb/tests/functions/test_functions.py
```diff
@@ -2698,3 +2698,50 @@ def test_st_snap(eng, input, reference, tolerance, expected):
...
+        ),
+        (
+            "POLYGON ((0 0,10 0,10 6,0 6,0 0), (1 1,2 1,2 5,1 5,1 1),(8 5,8 4,9 4,9 5,8 5))",
+            2,
+        ),
+        (
+            "MULTIPOLYGON ("
+            "((0 0, 5 0, 5 5, 0 5, 0 0), (1 1, 2 1, 2 2, 1 2, 1 1)),"
+            "((10 10, 14 10, 14 14, 10 14, 10 10))"
+            ")",
```
```suggestion
            "MULTIPOLYGON (((0 0, 5 0, 5 5, 0 5, 0 0), (1 1, 2 1, 2 2, 1 2, 1 1)),((10 10, 14 10, 14 14, 10 14, 10 10)))",
```

c/sedona-geos/src/st_numinteriorrings.rs
```diff
@@ -0,0 +1,168 @@
...
+            Some("LINESTRING (0 0, 1 1, 2 2)"),
+            Some("POLYGON EMPTY"),
+            Some("POLYGON ((0 0, 4 0, 4 4, 0 4, 0 0))"),
+            Some("POLYGON ((0 0,6 0,6 6,0 6,0 0),(2 2,4 2,4 4,2 4,2 2))"),
+            Some(
+                "POLYGON (
+                    (0 0,10 0,10 6,0 6,0 0),
+                    (1 1,2 1,2 5,1 5,1 1),
+                    (8 5,8 4,9 4,9 5,8 5)
+                )",
```
petern48:
```suggestion
                "POLYGON ((0 0,10 0,10 6,0 6,0 0),(1 1,2 1,2 5,1 5,1 1),(8 5,8 4,9 4,9 5,8 5))",
```

c/sedona-geos/src/st_numinteriorrings.rs
```diff
@@ -0,0 +1,168 @@
...
+            ),
+            Some(
+                "MULTIPOLYGON (
+                    ((0 0,5 0,5 5,0 5,0 0),(1 1,2 1,2 2,1 2,1 1)),
+                    ((10 10,14 10,14 14,10 14,10 10))
+                )",
```
petern48:
```suggestion
                "MULTIPOLYGON (((0 0,5 0,5 5,0 5,0 0),(1 1,2 1,2 2,1 2,1 1)),((10 10,14 10,14 14,10 14,10 10)))",
```

c/sedona-geos/src/st_numinteriorrings.rs
```diff
@@ -0,0 +1,168 @@
...
+            ),
+            Some(
+                "GEOMETRYCOLLECTION (
+                    POINT (1 2),
+                    POLYGON ((0 0,3 0,3 3,0 3,0 0))
+                )",
```
petern48:
```suggestion
                "GEOMETRYCOLLECTION (POINT (1 2),POLYGON ((0 0,3 0,3 3,0 3,0 0)))",
```

c/sedona-geos/src/st_numinteriorrings.rs
```diff
@@ -0,0 +1,168 @@
...
+        tester.assert_return_type(DataType::Int32);
+
+        // Polygon with 2 interior rings -> 2
+        let result = tester
+            .invoke_scalar(
+                "POLYGON(
+                    (0 0,10 0,10 6,0 6,0 0),
+                    (1 1,2 1,2 5,1 5,1 1),
+                    (8 5,8 4,9 4,9 5,8 5)
+                )",
```
petern48:
```suggestion
                "POLYGON((0 0,10 0,10 6,0 6,0 0),(1 1,2 1,2 5,1 5,1 1),(8 5,8 4,9 4,9 5,8 5))",
```

petern48: nit: There are a few more multiline geometries to collapse to one line. I made a suggestion for each one, if you want to handle that before we merge.

petern48:
Merged. Thanks for sticking through it @LakshmiSowmya04!

LakshmiSowmya04:
> Merged. Thanks for sticking through it @LakshmiSowmya04!

Thank you so much for helping me and encouraging me, even with the small things. I really appreciate it a lot. @petern48

PR #375, opened 2025-11-26
chore(dev/release): Ensure release verification passes for 0.2.0

paleolimbot:
This PR updates the verification instructions and fixes one problem that arose in testing (s2geography is a tricky dependency to link into the Python package on an aribtrary computer by default so we skip it by default for verification).

### DIFF
--- a/dev/release/README.md
+++ b/dev/release/README.md
@@ -57,6 +57,14 @@ number of parallel jobs to avoid running out of memory:
 export CARGO_BUILD_JOBS=4
 ```
 
+Verifiers may opt in to additional features normally tested in CI with dedicated tooling.
+For example, to verify with specific Python build-time features, the `MATURIN_PEP517_ARGS`
+environment variable may be set.
+
+```shell
+export MATURIN_PEP517_ARGS="--features s2geography"
+```
+
 ## Creating a release
 
 Create a release branch on the corresponding remote pointing to the official Apache


dev/release/verify-release-candidate.sh
```diff
@@ -262,8 +262,6 @@ test_python() {
   show_info "Installing Python package"
   rm -rf "${SEDONADB_TMPDIR}/python"
 
-  # Keep this export in sync with the export in .github/workflows/python.yml
-  export MATURIN_PEP517_ARGS="--features s2geography"
```
paleolimbot, line 266: @petern48 I moved this into the verification instructions instead of the default behaviour (i.e., MacOS users or exceptionally above-and-beyond voters can try this) because in some of my testing on Linux the Python build failed. We use CI to ensure this works where we need it to (arbitrary computers are hard!)

python/sedonadb/tests/functions/test_order.py
```diff
@@ -17,9 +17,14 @@
 
 import geopandas
 import geopandas.testing
+import pytest
+import sedonadb
 
 
 def test_order_sql(con):
+    if "s2geography" not in sedonadb.__features__:
+        pytest.skip("Ordering currently requires build with feature s2geography")
```
paleolimbot, line 26: This is needed so that the tests pass without the s2geography feature, which is currently required for ordering

PR #381, opened 2025-11-28
Implement ST_InteriorRingN

Abeeujah:
This PR implements `ST_InteriorRingN` using `geo-traits` as described in #224 And adds support for `Int64` in `BenchmarkArgs`

### DIFF
--- a/python/sedonadb/tests/functions/test_functions.py
+++ b/python/sedonadb/tests/functions/test_functions.py
@@ -1216,6 +1216,12 @@ def test_st_hasz(eng, geom, expected):
             1,
             "LINESTRING (1 1, 1 2, 2 2, 2 1, 1 1)",
         ),
+        # Single hole, negative index=-1
+        (
+            "POLYGON ((0 0, 4 0, 4 4, 0 4, 0 0), (1 1, 1 2, 2 2, 2 1, 1 1))",
+            -1,
+            None,
+        ),
         # Single hole, index=2 (index too high)
         ("POLYGON ((0 0, 4 0, 4 4, 0 4, 0 0), (1 1, 1 2, 2 2, 2 1, 1 1))", 2, None),
         # Polygon: ((0 0, 6 0, 6 6, 0 6, 0 0), (1 1, 1 2, 2 2, 2 1, 1 1), (4 4, 4 5, 5 5, 5 4, 4 4))
--- a/rust/sedona-functions/src/st_interiorringn.rs
+++ b/rust/sedona-functions/src/st_interiorringn.rs
@@ -233,6 +233,7 @@ mod tests {
         let input_wkt = create_array(
             &[
                 Some("POLYGON ((0 0, 4 0, 4 4, 0 4, 0 0), (1 1, 1 2, 2 2, 2 1, 1 1))"),                                  // Single hole, n=1
+                Some("POLYGON ((0 0, 4 0, 4 4, 0 4, 0 0), (1 1, 1 2, 2 2, 2 1, 1 1))"),                                  // Single hole, n=1
                 Some("POLYGON ((0 0, 4 0, 4 4, 0 4, 0 0), (1 1, 1 2, 2 2, 2 1, 1 1))"),                                  // Single hole, n=2 (too high)
                 Some("POLYGON ((0 0, 6 0, 6 6, 0 6, 0 0), (1 1, 1 2, 2 2, 2 1, 1 1), (4 4, 4 5, 5 5, 5 4, 4 4))"),       // Two holes, n=1
                 Some("POLYGON ((0 0, 6 0, 6 6, 0 6, 0 0), (1 1, 1 2, 2 2, 2 1, 1 1), (4 4, 4 5, 5 5, 5 4, 4 4))"),       // Two holes, n=2
@@ -240,11 +241,14 @@ mod tests {
             ],
             &WKB_GEOMETRY,
         );
-        let integers =
-            arrow_array::create_array!(Int64, [Some(1), Some(2), Some(1), Some(2), Some(3)]);
+        let integers = arrow_array::create_array!(
+            Int64,
+            [Some(1), Some(-1), Some(2), Some(1), Some(2), Some(3)]
+        );
         let expected = create_array(
             &[
                 Some("LINESTRING (1 1, 1 2, 2 2, 2 1, 1 1)"),
+                None,
                 None,
                 Some("LINESTRING (1 1, 1 2, 2 2, 2 1, 1 1)"),
                 Some("LINESTRING (4 4, 4 5, 5 5, 5 4, 4 4)"),
--- a/rust/sedona-testing/src/benchmark_util.rs
+++ b/rust/sedona-testing/src/benchmark_util.rs
@@ -456,6 +456,7 @@ impl BenchmarkArgSpec {
         }
     }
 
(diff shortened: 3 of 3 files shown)


rust/sedona-functions/src/st_interiorringn.rs
```diff
@@ -0,0 +1,243 @@
...
+    }
+}
+
+fn invoke_scalar(geom: &Wkb, index: usize, writer: &mut impl std::io::Write) -> Result<bool> {
+    let geometry = match geom.as_type() {
+        geo_traits::GeometryType::Polygon(pgn) => pgn.interior(index),
+        _ => None,
+    };
+
+    if let Some(buf) = geometry {
```
paleolimbot: I think the writing of the output should be more like the output writing you added for ST_Reverse:

https://github.com/apache/sedona-db/blob/16d232be0f28441b85801936b270fa43f7edceb4/rust/sedona-functions/src/st_reverse.rs#L189-L190

...in particular, I am not sure that `to_line_string()` and `x_y()` are what we want here because we need to support Z, M, and ZM coordinate types.

Also, `wkb` might be a better name than `buf` here since I don't think buf is actually a buffer here

rust/sedona-functions/src/st_interiorringn.rs
```diff
@@ -0,0 +1,243 @@
...
+            vec![
+                sedona_type.clone(),
+                SedonaType::Arrow(arrow_schema::DataType::Int64),
+            ],
+        );
+        tester.assert_return_type(WKB_GEOMETRY);
+
+        let input_wkt = create_array(
+            &[
+                // I. Null/Empty/Non-Polygon Inputs
```
paleolimbot: Thank you for all of these!

I think the missing cases here are Z, M, and ZM. It may be helpful to break these up into a few related input_wkt/inters/expected triplets to keep the whole test case roughly on your screen as you're scrolling.

rust/sedona-testing/src/benchmark_util.rs
```diff
@@ -277,6 +277,8 @@ pub enum BenchmarkArgSpec {
     /// Randomly generated linestring input with a specified number of vertices
     MultiPoint(usize),
     /// Randomly generated floating point input with a given range of values
+    Int64(i64, i64),
+    /// Randomly generated floating point input with a given range of values
```
paleolimbot, line 283:
```suggestion
    /// Randomly generated integer input with a given range of values
```

rust/sedona-functions/benches/native-functions.rs
```diff
@@ -82,6 +82,21 @@ fn criterion_benchmark(c: &mut Criterion) {
     benchmark::scalar(c, &f, "native", "st_hasm", Point);
     benchmark::scalar(c, &f, "native", "st_hasm", LineString(10));
 
+    benchmark::scalar(
+        c,
+        &f,
+        "native",
+        "st_interiorringn",
+        BenchmarkArgs::ArrayArray(Polygon(10), Int64(1, 10)),
+    );
```
paleolimbot, line 91: I'm not sure this is a very representative benchmark because it sould return NULL for every input (`Polygon(10)` doesn't have a hole). It may be less confusing to omit it here (rather than include a version that isn't measuring what it appears to be measuring).

Abeeujah: Is there a way to guage and know which would have a hole? say `Polygon(500)`?

rust/sedona-testing/src/benchmark_util.rs
Abeeujah: Would fix :sweat_smile:

rust/sedona-functions/src/st_interiorringn.rs
Abeeujah: Would add them, and separate into groups :+1:

rust/sedona-functions/src/st_interiorringn.rs
Abeeujah: `polygon.interior()` returns a `LinearRing` which is `Wkb` coded differently from `Linestring`, though the coords are the same, the test tends to fail since it returns a linear ring instead of a linestring, hence the reason for the `to_linestring()` there.

Would update the logic there. :+1: and use `write_wkb_coord_trait` instead :+1:

rust/sedona-functions/benches/native-functions.rs
paleolimbot: We'd need a new `BenchmarkArgSpec` (maybe `PolygonWithHole`) here:

https://github.com/apache/sedona-db/blob/16d232be0f28441b85801936b270fa43f7edceb4/rust/sedona-testing/src/benchmark_util.rs#L275-L276

...and it would need to be added to `build_geometry()`:

https://github.com/apache/sedona-db/blob/16d232be0f28441b85801936b270fa43f7edceb4/rust/sedona-testing/src/benchmark_util.rs#L421-L429

(the polygon hole rate is something you can specify to the random partitioned data builder)

python/sedonadb/tests/functions/test_functions.py
```diff
@@ -1178,6 +1178,87 @@ def test_st_hasz(eng, geom, expected):
     eng.assert_query_result(f"SELECT ST_HasZ({geom_or_null(geom)})", expected)
 
 
+@pytest.mark.parametrize("eng", [SedonaDB, PostGIS])
+@pytest.mark.parametrize(
+    ("geom", "index", "expected"),
+    [
+        # I. Null/Empty/Non-Polygon Inputs
```
paleolimbot, line 1185: Can you add the Z, M, and ZM cases you generated below to this list as well?

rust/sedona-testing/src/benchmark_util.rs
```diff
@@ -426,6 +464,7 @@ impl BenchmarkArgSpec {
         vertex_count: usize,
         num_parts_count: usize,
         rows_per_batch: usize,
+        polygon_hole_rate: Option<f64>,
```
paleolimbot, line 468: This caused `clippy` to complain about too many arguments. While clippy *is* right here (the call to `build_geometry` is getting a bit unwieldy/unreadable), it's an internal function and it's well tested. You can add `#[allow(clippy::too_many_arguments)]` right above `fn build_geometry(` to fix the error.

paleolimbot (commented):
Thanks!

There's one clippy error we need to fix to keep CI happy (I put a suggestion inline).

python/sedonadb/tests/functions/test_functions.py
```diff
@@ -1178,6 +1178,115 @@ def test_st_hasz(eng, geom, expected):
...
+        ("MULTIPOLYGON (((1 1, 1 3, 3 3, 3 1, 1 1)))", 1, None),
+        # GEOMETRYCOLLECTION
+        ("GEOMETRYCOLLECTION (POINT(1 1))", 1, None),
+        # II. Polygon Edge Cases
+        # POLYGON EMPTY
+        ("POLYGON EMPTY", 1, None),
+        # Polygon with NO interior rings, index=1
+        ("POLYGON ((0 0, 1 0, 1 1, 0 1, 0 0))", 1, None),
+        # Invalid index n=0 (Assuming 1-based indexing means n=0 is invalid/out of range)
+        ("POLYGON ((0 0, 1 0, 1 1, 0 1, 0 0))", 0, None),
```
petern48, line 1208: Testing a negative index case would be nice

PR #384, opened 2025-11-28
fix(rust/sedona-expr): Resolve filter expression bounding box by name and not by index

paleolimbot:
Closes #380. The issue was that we were passed a column index of `0` in some circumstances when the geometry column index was `2` (see the debug output on the issue for details).

I think the fact that we are passed an incorrect `Column { ..., index: 0}` is a DataFusion bug (I am guessing there is an `unwrap_or(0)` in some `try_pushdown_filter()`); however, in the meantime using the column name seems to work.

```python
# pip install "apache-sedona[db]"
import sedona.db

sd = sedona.db.connect()

# 12 GB file
url = "https://flatgeobuf.septima.dk/population_areas.fgb"
sd.read_pyogrio(url).to_view("population_areas")

wkt = "POLYGON ((-73.978329 40.767412, -73.950005 40.767412, -73.950005 40.795098, -73.978329 40.795098, -73.978329 40.767412))"
sd.sql(f"""
SELECT sum(population::INTEGER) FROM population_areas
WHERE ST_Intersects(wkb_geometry, ST_SetSRID(ST_GeomFromWKT('{wkt}'), 4326))
""").show()
#> ┌──────────────────────────────────┐
#> │ sum(population_areas.population) │
#> │               int64              │
#> ╞══════════════════════════════════╡
#> │                           256251 │
#> └──────────────────────────────────┘
```

Issue #380, opened 2025-11-27
Query bounding box not being pushed down into GDAL/OGR read


paleolimbot:
Discovered when writing the release post:

```python
# pip install "apache-sedona[db]"
import sedona.db

sd = sedona.db.connect()

url = "https://flatgeobuf.septima.dk/population_areas.fgb"
sd.read_pyogrio(url).to_view("population_areas")

# Cancelled after 30s
wkt = "POLYGON ((-73.978329 40.767412, -73.950005 40.767412, -73.950005 40.795098, -73.978329 40.795098, -73.978329 40.767412))"
sd.sql(f"""
SELECT sum(population::INTEGER) FROM population_areas
WHERE ST_Intersects(wkb_geometry, ST_SetSRID(ST_GeomFromWKT('{wkt}'), 4326))
""").show()

# 0.5s
import geopandas
import shapely

wkt = "POLYGON ((-73.978329 40.767412, -73.950005 40.767412, -73.950005 40.795098, -73.978329 40.795098, -73.978329 40.767412))"
geopandas.read_file(url, bbox=shapely.from_wkt(wkt).bounds, use_arrow=True)
```

petern48:
@paleolimbot Something else is wrong. The same query except using `SELECT *` instead of `SUM()` executes instantly (`0.6205670833587646 seconds`).

```
sd.sql(f"""
SELECT * FROM population_areas WHERE ST_Intersects(wkb_geometry, ST_SetSRID(ST_GeomFromWKT('{wkt}'), 4326))
""").show()
```

Seeing that, I knew the issue was in the sum. Then I looked more carefully. I noticed the population field is a utf8 instead of a number.

```
┌────────────┬────────────────────┬──────────────────────────────────────────────────────────────────────────────────────────────────────────────┐
│ population ┆       GISJOIN      ┆                                                 wkb_geometry                                                 │
│    utf8    ┆        utf8        ┆                                                   geometry                                                   │
```

I tried various casts, and I never got it to complete. Either the cast isn't going smoothly or the whole query is just taking so long because of the casting.

Oddly enough, taking any sort of `count(*)` instead of `*` gets stuck similarly to the sum an cast. For reference, the result is only length 387 (based on geopandas).

```
SELECT count(*) FROM population_areas WHERE ST_Intersects(wkb_geometry, ST_SetSRID(ST_GeomFromWKT('{wkt}'), 4326))
SELECT count(*) FROM population_areas
```

[...]

PR #385, opened 2025-11-28
fix(rust/sedona-expr): Fix GeoParquet pruning when number of final columns is less than the geometry column index

paleolimbot:
The issue identified by the GDAL filter pushdown in https://github.com/apache/sedona-db/pull/384 / https://github.com/apache/sedona-db/issues/380 also affects GeoParquet reads. The following query wouldn't complete for me before this change (after this change it completes in ~40s on a debug build).

```python
import sedona.db

sd = sedona.db.connect()

buildings_url = (
    "s3://overturemaps-us-west-2/release/2025-11-19.0/theme=buildings/type=building/"
)

target_wkt = (
    "POLYGON ((-73.21 44.03, -73.21 43.98, -73.11 43.97, -73.12 44.03, -73.21 44.03))"
)

sd.read_parquet(
    buildings_url,
    options={"aws.skip_signature": True, "aws.region": "us-west-2"},
).to_view("buildings")

sd.sql(f"""
SELECT * FROM buildings
WHERE ST_Intersects(geometry, ST_SetSRID(ST_GeomFromText('{target_wkt}'), 4326))
""").count()
#> 2845
```

The changes I tried to mostly isolate to `sedona-expr` to ensure this change is more tightly scoped. I opened https://github.com/apache/sedona-db/issues/389 to more holistically solve this since we're about to revisit that code shortly when we update DataFusion to provide Geometry/Geography parquet support (unless there's consensus that we should just do that now).

### DIFF
--- a/rust/sedona-expr/src/spatial_filter.rs
+++ b/rust/sedona-expr/src/spatial_filter.rs
@@ -465,7 +465,7 @@ impl TableGeoStatistics {
 // Useful for testing (create from a single GeoStatistics)
 impl From<GeoStatistics> for TableGeoStatistics {
     fn from(value: GeoStatistics) -> Self {
-        TableGeoStatistics::ByPosition(vec![value.clone()])
+        TableGeoStatistics::ByPosition(vec![value])
     }
 }
 


rust/sedona-expr/src/spatial_filter.rs
```diff
@@ -389,6 +394,82 @@ impl SpatialFilter {
     }
 }
 
+/// Table GeoStatistics
+///
+/// Enables providing a collection of GeoStatistics to [SpatialFilter::evaluate]
+/// such that attemmpts to access out-of-bounds values results in an readable
```
petern48:
```suggestion
/// such that attempts to access out-of-bounds values results in a readable
```
nit: spelling and grammar :)

rust/sedona-expr/src/spatial_filter.rs
```diff
@@ -389,6 +394,82 @@ impl SpatialFilter {
...
+    ByPosition(Vec<GeoStatistics>),
+
+    /// Provide statistics for specific named columns. Columns not included
+    /// are treated as [GeoStatistics::unspecified].
+    ///
+    /// These are resolved using [Column::name]. While the column name is
+    /// typically intended for debugging, at least some DataFusion versions
+    /// pass a filter to the data source with an incorrect index but a correct
+    /// name <https://github.com/apache/sedona-db/pull/385>. This option may
+    /// be removed if the incorrect index can be resolved upstream.
```
petern48: If you have any (and I mean any) hunches of what the exact behavior causing this in DataFusion is, now would be a good time to mention it while you're fresh off your investigation. Not top priority, but I'd eventually like to narrow it down enough to submit an issue at least.

paleolimbot: Thank you for the continued prods to investigate this more deeply...this particular prod I think resulted in me finding the root cause here (not a datafusion issue 😬 ). I'll push a (hopefully) fix shortly.

paleolimbot: https://github.com/apache/sedona-db/issues/389

PR #386, opened 2025-11-29
feat(c/sedona-geos): Implement ST_NumPoints

LakshmiSowmya04:
added st_numpoints file which will check number of vertices of a polygon. Acording to documentation this only works with linestrings.
This issue is a part of [issue](https://github.com/apache/sedona-db/issues/224)

### DIFF
--- a/r/sedonadb/DESCRIPTION
+++ b/r/sedonadb/DESCRIPTION
@@ -8,7 +8,7 @@ Description: Provides bindings for Apache SedonaDB, a lightweight
 License: Apache License (>= 2)
 Encoding: UTF-8
 Roxygen: list(markdown = TRUE)
-RoxygenNote: 7.3.2
+RoxygenNote: 7.3.3
 SystemRequirements: Cargo (Rust's package manager), rustc
 Depends: R (>= 4.1.0)
 Suggests:
--- a/r/sedonadb/R/dataframe.R
+++ b/r/sedonadb/R/dataframe.R
@@ -214,7 +214,7 @@ sd_preview <- function(.data, n = NULL, ascii = NULL, width = NULL) {
 #'   for every geometry column in the output: some readers can use these columns
 #'   to prune row groups when files contain an effective spatial ordering.
 #'   The extra columns will appear just before their geometry column and
-#'   will be named "[geom_col_name]_bbox" for all geometry columns except
+#'   will be named "\[geom_col_name\]_bbox" for all geometry columns except
 #'   "geometry", whose bounding box column name is just "bbox"
 #' @param overwrite_bbox_columns Use TRUE to overwrite any bounding box columns
 #'   that already exist in the input. This is useful in a read -> modify
--- a/r/sedonadb/man/sd_write_parquet.Rd
+++ b/r/sedonadb/man/sd_write_parquet.Rd
@@ -38,7 +38,7 @@ Use "1.1" to compute an additional bounding box column
 for every geometry column in the output: some readers can use these columns
 to prune row groups when files contain an effective spatial ordering.
 The extra columns will appear just before their geometry column and
-will be named "\link{geom_col_name}_bbox" for all geometry columns except
+will be named "[geom_col_name]_bbox" for all geometry columns except
 "geometry", whose bounding box column name is just "bbox"}
 
 \item{overwrite_bbox_columns}{Use TRUE to overwrite any bounding box columns


LakshmiSowmya04:
@petern48 I tried to add circularstring as it is mentioned in [document](https://postgis.net/docs/ST_NumPoints.html).But i got a build fialure saying "Your test fails because CIRCULARSTRING is not supported." Is there anyway I can add this or we can skip this ?

c/sedona-geos/src/st_numpoints.rs
```diff
@@ -0,0 +1,142 @@
...
+            Ok(())
+        })?;
+
+        executor.finish(Arc::new(builder.finish()))
+    }
+}
+
+fn invoke_scalar(geom: &Geometry) -> Result<Option<i32>> {
+    match geom.geometry_type() {
+        GeometryTypes::LineString | GeometryTypes::LinearRing => {
```
petern48:
```suggestion
        GeometryTypes::LineString => {
```

Sedona converts `LinearRing`s into LineStrings before it gets here, since they are equivalent.

e.g. `select st_geomfromtext('LINEARRING EMPTY');` returns `LINESTRING EMPTY`.

python/sedonadb/tests/functions/test_functions.py
```diff
@@ -2742,3 +2742,32 @@ def test_st_numinteriorrings_basic(eng, geom, expected):
...
+        ("POINT EMPTY", None),
+        ("LINESTRING EMPTY", 0),
+        ("POLYGON EMPTY", None),
+        ("MULTIPOINT EMPTY", None),
+        ("MULTILINESTRING EMPTY", None),
+        ("MULTIPOLYGON EMPTY", None),
+        ("GEOMETRYCOLLECTION EMPTY", None),
+        ("POINT (1 2)", None),
+        ("LINESTRING (0 0, 1 1, 2 2)", 3),
+        ("LINESTRING (0 0, 1 1, 0 0)", 3),
```
petern48, line 2761:
```suggestion
        ("LINESTRING (0 0, 1 1, 0 0)", 3),
        ("LINESTRING Z (0 0 0, 1 1 1, 2 2 2, 3 3 3)", 4),
        ("LINESTRING M (0 0 0, 1 1 1, 2 2 2, 3 3 3)", 4),
        ("LINESTRING ZM (0 0 0 2, 1 1 1 4)", 2),
```
Let's also test Z/M coordinates for this function to be sure they don't mess anything up.

c/sedona-geos/src/st_numpoints.rs
```diff
@@ -0,0 +1,142 @@
...
+        let result = tester
+            .invoke_scalar(
+                "POLYGON(
+                    (0 0,10 0,10 6,0 6,0 0),
+                    (1 1,2 1,2 5,1 5,1 1),
+                    (8 5,8 4,9 4,9 5,8 5)
+                )",
+            )
+            .unwrap();
+        assert!(result.is_null());
```
petern48:
```suggestion
        let result = tester
            .invoke_scalar("LINESTRING (1 2, 3 4)")
            .unwrap();
        tester.assert_scalar_result_equals(result, 2_i32);
```
optional nit: For the one scalar test, it's nice to test a case that highlights the main functionality, since this is more readable than the array tests below. Since st_numpoints really only operates on LineStrings (returning Null for everything else), let's test one here instead of a polygon that returns null. You already have tests below to make sure we return null properly. (This is more of a style thing, not a requirement)

petern48 (commented):
Looks great. Just a few minor requests.

petern48:
> But i got a build fialure saying "Your test fails because CIRCULARSTRING is not supported." Is there anyway I can add this or we can skip this ?

Oh yeah, Sedona doesn't support `CIRCULARSTRING`. It's equivalent to a LineString. The same can be said about LinearRing. There's no need to test it.

petern48:
Last two things. The first is good practice for all repos. The second is specific to this SedonaDB repo.

- Could you mention the [issue](https://github.com/apache/sedona-db/issues/224) that this is a part of inside the PR description?
  - We don't want to say "closes" since the issue isn't done, but I like to do something like `"part of <issue link>"`
  - It's important to mention it, so that GitHub creates a link in that issue to this PR (e.g. shown below). When others view that issue, they can see that you have already started on implementing this function. When you don't do this, it increases the odds of someone trying to implement it too, since they might not see that you opened a PR for it already.
- Could you also update the PR title to follow this convention we've been following for this repo:
  - `feat(c/sedona-geos): Implement ST_NumPoints`
     - "feat" is short for feature. "c/sedona-geos" is the directory the change is in. It doesn't matter if you say "Add" or "Implement"

Also, for the future, try to do both of these when you first open the PR, especially the first one, to avoid chances of duplicate PRs.

LakshmiSowmya04:
> Last two things. The first is good practice for all repos. The second is specific to this SedonaDB repo.
> 
> * Could you mention the [issue](https://github.com/apache/sedona-db/issues/224) that this is a part of inside the PR description?
>   
>   * We don't want to say "closes" since the issue isn't done, but I like to do something like `"part of <issue link>"`
>   * It's important to mention it, so that GitHub creates a link in that issue to this PR (e.g. shown below). When others view that issue, they can see that you have already started on implementing this function. When you don't do this, it increases the odds of someone trying to implement it too, since they might not see that you opened a PR for it already.
> * Could you also update the PR title to follow this convention we've been following for this repo:
>   
>   * `feat(c/sedona-geos): Implement ST_NumPoints`
>     
>     * "feat" is short for feature. "c/sedona-geos" is the directory the change is in. It doesn't matter if you say "Add" or "Implement"
> 
> Also, for the future, try to do both of these when you first open the PR, especially the first one, to avoid chances of duplicate PRs.

okay  I will make sure follow these rules.

PR #387, opened 2025-11-29
feat(c/sedona-geos): Implement ST_NRings 

LakshmiSowmya04:
PostGIS used to have a function called ST_NumRings, but it seems to have been removed or renamed. The current function ST_NRings does the same thing—it counts all rings in a polygon, including the outer ring. So we’re using ST_NRings instead of ST_NumRings.This issue is a part of [issue](https://github.com/apache/sedona-db/issues/224).


LakshmiSowmya04:
@petern48 can you please review my PR and let me know if I should make any changes

python/sedonadb/tests/functions/test_functions.py
```diff
@@ -2774,3 +2774,40 @@ def test_st_numpoints(eng, geom, expected):
...
+@pytest.mark.parametrize(
+    ("geom", "expected"),
+    [
+        (None, None),
+        ("POINT (1 2)", 0),
+        ("LINESTRING (0 0, 1 1, 2 2)", 0),
+        ("MULTIPOINT ((0 0), (1 1))", 0),
+        ("MULTILINESTRING ((0 0, 1 1), (2 2, 3 3))", 0),
+        ("POLYGON EMPTY", 0),
+        ("MULTIPOLYGON EMPTY", 0),
```
```suggestion
        ("POINT EMPTY", 0),
        ("MULTIPOINT EMPTY", 0),
        ("LINESTRING EMPTY", 0),
        ("MULTILINESTRING EMPTY", 0),
        ("POLYGON EMPTY", 0),
        ("MULTIPOLYGON EMPTY", 0),
        ("GEOMETRYCOLLECTION EMPTY", 0),
```

c/sedona-geos/src/st_nrings.rs
```diff
@@ -0,0 +1,202 @@
...
+        let mut builder = Int32Builder::with_capacity(executor.num_iterations());
+        executor.execute_wkb_void(|maybe_geom| {
+            match maybe_geom {
+                None => builder.append_null(),
+                Some(geom) => {
+                    let res = invoke_scalar(&geom)?;
+                    match res {
+                        Some(n) => builder.append_value(n),
+                        None => builder.append_null(),
+                    }
```
petern48:
```suggestion
                    builder.append_value(invoke_scalar(&geom)?;
```
Looking at your code below, there aren't any cases that return `None`, so we should go ahead and avoid returning an Option here, so we don't have to handle `None`.

In other words, you'll need to update `invoke_scalar` to return `Result<i32>` instead of `Result<Option<i32>>`. This will happen naturally if you follow the suggestion I made below about using the `count_rings_recursive` function instead.

c/sedona-geos/src/st_nrings.rs
```diff
@@ -0,0 +1,202 @@
...
+                return Ok(Some(0));
+            }
+            let total = count_rings_recursive(geom)?;
+            Ok(Some(total))
+        }
+        _ => Ok(Some(0)),
+    }
+}
+
+fn count_rings_recursive<G: Geom>(geom: &G) -> Result<i32> {
```
petern48: I'm seeing a lot of duplicated logic between this function, `count_rings_recursive`, and `invoke_scalar`. The function signatures are *basically* the same. I think it's very possible to reduce this to one function.

`count_rings_recursive` seems to be the one we want here. 1) It doesn't return an Option<i32>. 2) It already has the recursive call we need here.

How about we delete the original `invoke_scalar` above and rename this function to `invoke_scalar`.

petern48 (commented):
Great job getting this to a working solution 👏 . Having to deal with some recursion is definitely more complicated than your previous PRs. There's a bit of clean up we can do, but I think we can merge after that.

c/sedona-geos/src/st_nrings.rs
```diff
@@ -0,0 +1,177 @@
...
+        args: &[ColumnarValue],
+    ) -> Result<ColumnarValue> {
+        let executor = GeosExecutor::new(arg_types, args);
+        let mut builder = Int32Builder::with_capacity(executor.num_iterations());
+        executor.execute_wkb_void(|maybe_geom| {
+            match maybe_geom {
+                None => builder.append_null(),
+                Some(geom) => {
+                    // Fixed: Removed the unused 'res' variable assignment
+                    // and directly appended the result.
```
petern48:
```suggestion
```

c/sedona-geos/src/st_nrings.rs
```diff
@@ -0,0 +1,177 @@
...
+// under the License.
+
+use std::sync::Arc;
+
+use crate::executor::GeosExecutor;
+use arrow_array::builder::Int32Builder;
+use arrow_schema::DataType;
+use datafusion_common::{error::Result, DataFusionError};
+use datafusion_expr::ColumnarValue;
+use geos::{Geom, GeometryTypes}; // Removed unused 'Geometry'
```
petern48:
```suggestion
use geos::{Geom, GeometryTypes};
```

petern48 (approved):
Looks good. Just some minor cleanup.

python/sedonadb/tests/functions/test_functions.py
```diff
@@ -2889,3 +2889,49 @@ def test_st_numpoints(eng, geom, expected):
...
+        ("POINT EMPTY", 0),
+        ("MULTIPOINT EMPTY", 0),
+        ("LINESTRING EMPTY", 0),
+        ("MULTILINESTRING EMPTY", 0),
+        ("MULTIPOINT ((0 0), (1 1))", 0),
+        ("MULTILINESTRING ((0 0, 1 1), (2 2, 3 3))", 0),
+        ("POINT EMPTY", 0),
+        ("MULTIPOINT EMPTY", 0),
+        ("LINESTRING EMPTY", 0),
+        ("MULTILINESTRING EMPTY", 0),
```
petern48, line 2912:
```suggestion
        ("POLYGON EMPTY", 0),
        ("MULTIPOLYGON EMPTY", 0),
```
Let's undo that [last commit](https://github.com/apache/sedona-db/pull/387/commits/51297de7dee41dfa8fccf988f55560c2890403cb). Looks like it removed these empty polygon tests and replaced it with duplicate versions of existing tests that are still there.

paleolimbot (approved):
Thank you!

c/sedona-geos/src/st_nrings.rs
petern48: Just as an FYI @LakshmiSowmya04, usually when I suggest removing comments like these. It means it's one I think we *should* remove. But it also generally means that other comments that I don't suggest on are *fine* to leave in. I'm just pointing this out since I've noticed you often delete more than I ask (e.g. in here https://github.com/apache/sedona-db/pull/387/commits/02ddb50e17a9c0b2b3b04974632faa064caba0a1). Deleting more is totally okay (I probably would delete some of these as well), but I just wanted to let you know it's generally fine to leave other comments in, if you prefer to / think they're helpful.

LakshmiSowmya04:
@petern48 i saw in other files and dint see any comments and its clean ....sometimes I wrote few lines  for my rough use. so before committing I remove most of the comments .

PR #399, opened 2025-12-02
St exterior ring

LakshmiSowmya04:
The current function ST_ExteriorRing—it gives the exterior Ring in a polygon and this function does not support MULTIPOLYGONs .please refer this [documentation](https://postgis.net/docs/ST_ExteriorRing.html).This issue is a part of https://github.com/apache/sedona-db/issues/224.
 
 
image

### DIFF
--- a/rust/sedona-raster-functions/benches/native-raster-functions.rs
+++ b/rust/sedona-raster-functions/benches/native-raster-functions.rs
@@ -25,6 +25,13 @@ fn criterion_benchmark(c: &mut Criterion) {
         c,
         &f,
         "native-raster",
+        "rs_rastertoworldcoord",
+        BenchmarkArgs::ArrayScalarScalar(Raster(64, 64), Int32(0, 63), Int32(0, 63)),
+    );
+    benchmark::scalar(
+        c,
+        &f,
+        "native-raster",
         "rs_rastertoworldcoordx",
         BenchmarkArgs::ArrayScalarScalar(Raster(64, 64), Int32(0, 63), Int32(0, 63)),
     );
--- a/rust/sedona-raster-functions/src/register.rs
+++ b/rust/sedona-raster-functions/src/register.rs
@@ -47,6 +47,7 @@ pub fn default_function_set() -> FunctionSet {
         crate::rs_geotransform::rs_upperlefty_udf,
         crate::rs_size::rs_height_udf,
         crate::rs_size::rs_width_udf,
+        crate::rs_worldcoordinate::rs_rastertoworldcoord_udf,
         crate::rs_worldcoordinate::rs_rastertoworldcoordx_udf,
         crate::rs_worldcoordinate::rs_rastertoworldcoordy_udf,
     );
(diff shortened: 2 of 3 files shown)


LakshmiSowmya04:
@petern48 , anyhow we will give input seperately as MULTIPOLYGON's , in this exterior ring we can return an array of exterior rings too if you are fine.
image

petern48:
> and this function does not support MULTIPOLYGONs

> anyhow we will give input seperately as MULTIPOLYGON's , in this exterior ring we can return an array of exterior rings too if you are fine.

In general, we try to follow PostGIS behavior, so let's follow the same behavior of returning `Null`. As mentioned in the PostGIS docs, users can get around this by using this function along with `ST_GeometryN` to get the same behavior.

c/sedona-geos/src/st_exteriorring.rs
```diff
@@ -0,0 +1,162 @@
...
+}
+
+#[derive(Debug)]
+struct STExteriorRing {}
+
+impl SedonaScalarKernel for STExteriorRing {
+    fn return_type(&self, args: &[SedonaType]) -> Result<Option<SedonaType>> {
+        let matcher = ArgMatcher::new(
+            vec![ArgMatcher::is_geometry()],
+            SedonaType::Arrow(DataType::Binary),
```
petern48: While yes, we represent geometries as binary datatypes, we have a special geometry datatype touse for convenience. See the reference file.

c/sedona-geos/src/st_exteriorring.rs
```diff
@@ -0,0 +1,162 @@
...
+    fn invoke_batch(
+        &self,
+        arg_types: &[SedonaType],
+        args: &[ColumnarValue],
+    ) -> Result<ColumnarValue> {
+        let executor = GeosExecutor::new(arg_types, args);
+        let mut builder = GenericBinaryBuilder::<i32>::with_capacity(
+            executor.num_iterations(),
+            executor.num_iterations() * 100,
+        );
```
petern48: You'll also want to udpate this too.

c/sedona-geos/src/st_exteriorring.rs
```diff
@@ -0,0 +1,162 @@
...
+
+    fn wkt_to_wkb(wkt: &str) -> Vec<u8> {
+        geos::Geometry::new_from_wkt(wkt)
+            .expect("Invalid WKT in test expectation")
+            .to_wkb()
+            .expect("Failed to convert to WKB")
+            .into()
+    }
+
+    fn build_expected_geometry_array(wkts: Vec<Option<&str>>) -> ArrayRef {
```
petern48: If you reference the test code in the reference file I mentioned above, you'll probably find some useful methods that you can use to avoid trying to do this manual logic yourself.

petern48:
Could you also pull main and apply this change (https://github.com/apache/sedona-db/pull/387#discussion_r2581685826) I asked for in your previous PR. The PR got merged before I had the chance to comment on it.

PR #400, opened 2025-12-02
chore(dependabot): Add groups for datafusion, wasm-bindgen, and savvy

yutannihilation:
Just like the `arrow-parquet` group, there are several more crates that are likely to be updated together. This pull request adds the following three groups in the dependabot configuration.

- datafusion (Maybe we can just let this ignored by dependabot...? I guess the update of DataFusion is well-tracked anyway)
- wasm-bindgen (I guess it would be really rare that wasm-bindgen bumps the minor or major version, though...)
- savvy

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -31,7 +31,7 @@ updates:
       day: "sunday"
     open-pull-requests-limit: 50
     ignore:
-       # For all packages, ignore all patch updates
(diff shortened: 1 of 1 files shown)


yutannihilation:
I see, sounds good to me! Then, we might want more tweaks on the setting to reduce noise, but let's think about it after the release.

PR #310, opened 2025-11-15
feat(c/sedona-libgpuspatial): Add GPU-accelerated spatial join library

pwrliang:
This PR adds a GPU-accelerated spatial join library called `libgpuspatial` under `c/sedona-libgpuspatial/libgpuspatial` to support SedonaDB's GPU-acceleration feature. The GPU execution is automatically enabled when available and provides significant performance improvements for large-scale spatial joins. A Rust package called `sedona-libgpuspatial` for building and using the library is included under `c/sedona-libgpuspatial`. Currently, there's no GitHub action runner that has a GPU, so we only build the library and tests without actually running them.

PR #407, opened 2025-12-03
feat(c/sedona-extension): Move FFI scalar kernel definitions to c/sedona-extension and write them in C

paleolimbot:
We haven't used the `ffi` scalar kernels in the `sedona` crate yet, which were based on the rust/DataFusion versions. To properly get geography working everywhere I'd like to avoid the Rust wrapper around s2geography, which means being able to fully express a scalar kernel in C. Luckily the existing C definition we'd been using is fairly close...this PR removes the previous FFI scalar kernel implementation in favour of a C one. Once this is merged I'll follow-up with a change to s2geography to actually use it (which will remove almost all of the wrapper cod there, too).

This PR also implements exporting any of our implementations as C-callable implementations, which would let any of these functions be called from something like Java given some glue code on the other side.

This PR has an eye towards other types of C extensions like the C GPU joiner in https://github.com/apache/sedona-db/pull/310, where we can simplify the development for both libgpuspatial and sedonadb by completely separating the builds.

### DIFF
--- a/c/sedona-extension/src/scalar_kernel.rs
+++ b/c/sedona-extension/src/scalar_kernel.rs
@@ -56,13 +56,8 @@ impl TryFrom<SedonaCScalarKernel> for ImportedScalarKernel {
     type Error = DataFusionError;
 
     fn try_from(value: SedonaCScalarKernel) -> Result<Self> {
-        match (
-            &value.function_name,
-            &value.new_impl,
-            &value.release,
-            value.private_data.is_null(),
-        ) {
-            (Some(function_name), Some(_), Some(_), false) => {
+        match (&value.function_name, &value.new_impl, &value.release) {
+            (Some(function_name), Some(_), Some(_)) => {
                 let name_ptr = unsafe { function_name(&value) };
                 let name = if name_ptr.is_null() {
                     None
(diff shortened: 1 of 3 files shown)


c/sedona-extension/src/sedona_extension.h
```diff
@@ -0,0 +1,208 @@
...
+  /// Implementations MAY take ownership over the elements of scalar_args but
+  /// are not required to do so (i.e., caller must check if these elements were
+  /// released, and must release them if needed).
+  /// \param n_args Number of elements in the arg_types and/or scalar_args arrays.
+  /// \param out Will be populated with the return type on success, or initialized
+  /// to a released value if this implementation does not apply to the arguments
+  /// passed.
+  ///
+  /// \return An errno-compatible error code, or zero on success.
+  int (*init)(struct SedonaCScalarKernelImpl* self, const struct ArrowSchema** arg_types,
```
Kontinuation: Should the type of `arg_types` be `const struct ArrowSchema* const*`?

c/sedona-extension/src/scalar_kernel.rs
```diff
@@ -0,0 +1,883 @@
...
+
+impl TryFrom<SedonaCScalarKernel> for ImportedScalarKernel {
+    type Error = DataFusionError;
+
+    fn try_from(value: SedonaCScalarKernel) -> Result<Self> {
+        match (
+            &value.function_name,
+            &value.new_impl,
+            &value.release,
+            value.private_data.is_null(),
```
Kontinuation: Is it possible that an implementation of `SedonaCScalarKernel` does not need private data at all?

paleolimbot: That's a great point, thanks!

PR #409, opened 2025-12-04
feat(c/sedona-geos): Implement ST_ExteriorRing

LakshmiSowmya04:
Introduced the st_exteriorfile function, which returns the outer boundary line of a given geometry.
This issue is a part of https://github.com/apache/sedona-db/issues/224


LakshmiSowmya04:
@petern48 can you please review my PR,thank you

python/sedonadb/tests/functions/test_functions.py
```diff
@@ -2935,3 +2929,32 @@ def test_st_NRings(eng, geom, expected):
...
+            "LINESTRING (0 0, 10 0, 10 10, 0 10, 0 0)",
+        ),
+        (
+            "MULTIPOLYGON (((0 0, 1 0, 1 1, 0 1, 0 0)), ((10 10, 20 10, 20 20, 10 20, 10 10)))",
+            None,
+        ),
+        ("GEOMETRYCOLLECTION(POINT(1 1), POLYGON((0 0, 1 0, 1 1, 0 0)))", None),
+        (
+            "POLYGON Z ((0 0 1, 1 0 1, 1 1 1, 0 1 1, 0 0 1))",
+            "LINESTRING Z (0 0 1, 1 0 1, 1 1 1, 0 1 1, 0 0 1)",
```
petern48: Can you also add `POLYGON M` and `POLYGON ZM` cases? I think your current code might fail these, at the moment, since I see logic for `has_z` instead of using our usual pattern.

You can `Ctrl + f` in this `test_functions.py` file and search for `POLYGON M` and `POLYGON ZM` to find examples of test cases.

python/sedonadb/tests/functions/test_functions.py
```diff
@@ -2904,12 +2904,6 @@ def test_st_numpoints(eng, geom, expected):
         ("MULTIPOINT EMPTY", 0),
         ("LINESTRING EMPTY", 0),
         ("MULTILINESTRING EMPTY", 0),
-        ("MULTIPOINT ((0 0), (1 1))", 0),
-        ("MULTILINESTRING ((0 0, 1 1), (2 2, 3 3))", 0),
-        ("POINT EMPTY", 0),
-        ("MULTIPOINT EMPTY", 0),
-        ("LINESTRING EMPTY", 0),
-        ("MULTILINESTRING EMPTY", 0),
```
petern48, line 3342: Thanks for doing this. For other readers, this is just clean up from a previous PR, as explained [here](https://github.com/apache/sedona-db/pull/387#discussion_r2581685826)

python/sedonadb/tests/functions/test_functions.py
```diff
@@ -2935,3 +2929,32 @@ def test_st_NRings(eng, geom, expected):
         f"SELECT ST_NRings({geom_or_null(geom)})",
         expected,
     )
+
+
+@pytest.mark.parametrize("eng", [SedonaDB, PostGIS])
+@pytest.mark.parametrize(
+    ("geom", "expected"),
+    [
+        (None, None),
```
```suggestion
        (None, None),
        ("POINT EMPTY", None),
        ("LINESTRING EMPTY", None),
        ("POLYGON EMPTY", 0),
        ("MULTIPOINT EMPTY", None),
        ("MULTILINESTRING EMPTY", None),
        ("MULTIPOLYGON EMPTY", None),
        ("GEOMETRYCOLLECTION EMPTY", None),
```

As always, I want to test empty versions of each of the geometry types

python/sedonadb/tests/functions/test_functions.py
```diff
@@ -2935,3 +2929,32 @@ def test_st_NRings(eng, geom, expected):
...
+        ("POINT (1 2)", None),
```
```suggestion
        ("POINT (1 2)", None),
        ("MULTIPOINT ((0 0), (1 1))", None),
```

I prefer to also test a non-empty version of each geometry type as well. You have all except multipoint already.

c/sedona-geos/src/st_exteriorring.rs
```diff
@@ -0,0 +1,165 @@
...
+
+    fn invoke_batch(
+        &self,
+        arg_types: &[SedonaType],
+        args: &[ColumnarValue],
+    ) -> Result<ColumnarValue> {
+        let executor = GeosExecutor::new(arg_types, args);
+        let mut builder = BinaryBuilder::with_capacity(
+            executor.num_iterations(),
+            100 * executor.num_iterations(),
```
petern48:
```suggestion
            WKB_MIN_PROBABLE_BYTES * executor.num_iterations(),
```

To figure out how to set values like this, I recommend looking at other functions that return a geometry like `st_centroid.rs`. e.g. [here](https://github.com/apache/sedona-db/blob/c48dfaa06858b5fe3ddf8d68115a890fe222cab1/c/sedona-geos/src/st_centroid.rs#L53-L56)

You'll need to add the import at the top too:
```rust
use sedona_geometry::wkb_factory::WKB_MIN_PROBABLE_BYTES;
```

c/sedona-geos/src/st_exteriorring.rs
```diff
@@ -0,0 +1,165 @@
...
+
+            if geos_geom.has_z().unwrap_or(false) {
+                writer.set_output_dimension(OutputDimension::ThreeD);
+            }
+
+            let wkb = writer.write_wkb(&ring).map_err(|e| {
+                DataFusionError::Execution(format!("Failed to convert to wkb: {e}"))
+            })?;
+
+            builder.append_value(wkb);
```
petern48: While it seems like this works (though I think it would fail for `M` and `ZM` cases), we should follow existing code patterns instead of introducing this less concise and somewhat more complex logic.

I encourage you to look at `st_centroid.rs` to see how you can implement the wkb writing more cleanly. Specifically look [here](https://github.com/apache/sedona-db/blob/c48dfaa06858b5fe3ddf8d68115a890fe222cab1/c/sedona-geos/src/st_centroid.rs#L78-L82) and how that's called [here](https://github.com/apache/sedona-db/blob/c48dfaa06858b5fe3ddf8d68115a890fe222cab1/c/sedona-geos/src/st_centroid.rs#L59-L63)

c/sedona-geos/src/st_exteriorring.rs
```diff
@@ -0,0 +1,165 @@
...
+        match result_z {
+            ScalarValue::Binary(Some(bytes)) | ScalarValue::LargeBinary(Some(bytes)) => {
+                let expected_bytes = geos_wkt_to_wkb(z_wkt_out);
+                assert_eq!(
+                    bytes, expected_bytes,
+                    "Z-coordinates were not preserved correctly"
+                );
+            }
+            _ => panic!("Expected Binary result for 3D test"),
+        }
```
petern48: (Genuine question) Do we really need this extra logic for testing `POLYGON Z` here? I would've thought we could use it in the `input_wkt` and `expected` arrays below.

Let's remove this complexity by moving this test case into the arrays below. If that doesn't work. Let me know.

petern48 (commented):
Nice. While it seems to work, we need to add some other test cases and do a bit of revising to clean up some unnecessary complexity. It's close, not far off from where it needs to be for us to merge it.

petern48:
Could you also fix the PR title and mention the issue in the PR description, according to my previous comment before: https://github.com/apache/sedona-db/pull/386#issuecomment-3591890245. You should generally try to do it whenever you first open the PR. If you don't do so, it increases the odds of someone else taking up that work too, which would eventually lead to someone (either you or them) going through the work just for their PR to not be merged.

If you're afraid of linking the issue earlier because you're not sure you can 100% commit, it's totally okay to link it anyways and later decide to close your PR if you decide you don't want to follow through with it (e.g. if it's too hard or you lose interest). Nothing ever locks you in to being *required* to finish something. But please do link the issue as soon as you open the PR. For this PR, we're already close to being done anyway.

python/sedonadb/tests/functions/test_functions.py
LakshmiSowmya04: can I ask how you figured out we have to use the polygon M and polygon ZM cases for this? I cannot see it in the documentation.

petern48: When it comes to geospatial knowledge. You just learn overtime that there's four possible dimensions. X and Y (required), and Z and M (both optional). These cases always existed, but they were less interesting cases for your previous functions which only returned counts of rings. Whether an M or Z dimension existed or not is unlikely to fail for those "counting" functions. We could totally add them, but they were less interesting or unlikely to break things.

For this ST_ExteriorRing, it doesn't just count rings, it **returns them**. Specifically, the returned geometry should also have Z and M dimensions if their inputs did. It's certainly possible for underlying implementations to forget to implement Z and M support. In that case, the function might right `LINESTRING` except missing the Z and M dimensions. That's why Z and M tests are more interesting for this function, hence we want to test them and make sure it doesn't test the wrong input.

In contrast, if ST_NumRings completely disregarded Z and M dimensions, it probably would still return the right count, since it doesn't really care about the exact contents of the rings, just the number of them. This intuition isn't quite something contributors would be expected to reason about, so don't worry too much if you can't follow the logic.

PR #413, opened 2025-12-05
chore(dev/release) Add publish scripts to crate.io and update rust dependency

zhangfengcdt:


dev/release/publish-crates.sh
paleolimbot, line 1: Can you add a license header and move this to `dev/release`?

PR #428, opened 2025-12-08
chore(docs): Add a validation script for SQL reference

yutannihilation:
Related to #180.

It seems it's a bit hard to tackle #180 directly, so this pull request adds a simple validation script to check if the document reflects the implementation correctly. This checks, (1) if the functions listed in the reference matches with the actual implementation, and (2) if the functions are sorted in alphabetical order.

I need to check if each one is actual problem or just a false-positive, but this is the current result:

```console
❯ uv run ./docs/scripts/validate_sql.py

Functions only in implementation:
  - st_aswkb
  - st_aswkt
  - st_distance
  - st_distancesphere
  - st_distancespheroid
  - st_frechetdistance
  - st_geogfromtext
  - st_geometryfromtext
  - st_geomfromtext
  - st_hausdorffdistance
  - st_line_interpolate_point
  - st_line_locate_point
  - st_numinteriorrings

Functions only in document:
  - st_closestpoint
  - st_linelocatepoint


Traceback (most recent call last):
  File "/path/to/sedona-db/./docs/scripts/validate_sql.py", line 51, in <module>
    raise RuntimeError(
RuntimeError: There are some mismatch between the SQL reference and the actual implementation!
```

TODO:
- [ ] Handle aliases
- [ ] Check the order of the functions section by section
- [ ] Insert missing functions to `sql.md`

### DIFF
--- a/docs/scripts/validate_sql.py
+++ b/docs/scripts/validate_sql.py
@@ -17,9 +17,11 @@
 
 import sedonadb
 import difflib
+from pathlib import Path
 
+HERE = Path(__file__).parent
 
-with open("docs/reference/sql.md", "r") as f:
+with open(HERE.parent / "reference" / "sql.md", "r") as f:
     lines = f.readlines()
     # Headers with `##` are the function names.
     funs_in_doc = [line[3:-1] for line in lines if line.startswith("## ")]


docs/scripts/validate_sql.py
```diff
@@ -0,0 +1,65 @@
...
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+import sedonadb
+import difflib
+
+
+with open("docs/reference/sql.md", "r") as f:
```
paleolimbot:
```suggestion
import sedonadb
import difflib
from pathlib import Path

HERE = Path(__file__).parent

with open(HERE.parent / "reference" / "sql.md", "r") as f:
```

yutannihilation:
Thanks! Yeah, it's always good to have a validation. What I'm still wondering is when we should enforce it... Before making this check mandatory, I hope this script can have some functionality to insert the new section to `sql.md` from the description and examples if there's any missing function name.

docs/scripts/validate_sql.py
```diff
@@ -0,0 +1,75 @@
...
+df = sd.sql(r"""
+SELECT DISTINCT
+  routine_name,
+  function_type,
+  CASE substr(routine_name, 1, 2)
+    WHEN 'st' THEN 'vector'
+    WHEN 'rs' THEN 'raster'
+    ELSE 'unknown'
+  END AS data_type,
+  count(*) OVER (PARTITION BY description) > 1 as has_alias
```
yutannihilation, line 42: It seems the information about aliases cannot be retrieved by SQL. I added a tweak to guess it from `description`, but I'm not sure if this is reliable enough.

I might rewrite the validation script by using Rust just like DataFusion does to generate the documentation.

https://github.com/apache/datafusion/blob/5a01e68643a198a1aaa7124524d7be5be7df24ec/datafusion/core/src/bin/print_functions_docs.rs#L174-L179

yutannihilation:
Closing this for now as I get lost (especially about handling aliases). I hope I can retry this someday...

PR #431, opened 2025-12-09
feat(c/sedona-geos): ST_ConcaveHull Implementation

Abeeujah:
This has been deeper than I expected, particularly the WKB generated after the `Concavity` has been done, It tends to produce geometrically identical `Polygon` as PostGIS, but the Polygon generated starts at a different point from the Postgis version.

First I tried with `rust-geo` It worked perfectly on some geometries, but was a bit rotated on other geometries, also, it tend to be more aggressive than Postgis; i.e truncating some points in the hull generated.

Secondly I tried `geos` using `geo-sys` it worked for other geometries, but provided Geometrically Identical WKBs but Geometries starting at different points.


Cargo.toml
```diff
@@ -97,6 +97,7 @@ libmimalloc-sys = { version = "0.1", default-features = false }
 once_cell = "1.20"
 
 geos = { version = "10.0.0", features = ["geo", "v3_10_0"] }
+geos-sys = { version = "2.0.6", features = ["v3_11_0"] }
```
paleolimbot: See my general note here...I think we can avoid this!

rust/sedona-geo/src/st_concavehull.rs
```diff
@@ -0,0 +1,231 @@
...
+    hull: geo_types::Polygon,
+) -> Result<()> {
+    write_wkb_polygon_header(writer, dims, hull.num_rings())
+        .map_err(|e| DataFusionError::Execution(e.to_string()))?;
+    write_concave_ring(writer, hull.exterior())?;
+    for interior in hull.interiors() {
+        write_concave_ring(writer, interior)?
+    }
+    Ok(())
+}
```
paleolimbot: I think you may be able to just use `wkb::writer::write_wkb()` here? I don't think you need to reverse the order of the output ring.

rust/sedona-geo/src/st_concavehull.rs
```diff
@@ -0,0 +1,231 @@
...
+        );
+        assert_eq!(tester.return_type().unwrap(), WKB_GEOMETRY);
+
+        let result = tester
+            .invoke_scalar_scalar(
+                "GEOMETRYCOLLECTION (MULTIPOINT((1 1), (3 3)), POINT(5 6), LINESTRING(4 5, 5 6))",
+                0.1,
+            )
+            .unwrap();
+        tester.assert_scalar_result_equals(result, "POLYGON ((3 3, 1 1, 4 5, 5 6, 3 3))");
```
paleolimbot: I think what you want here is to compare the result "topologically" (i.e., a rust version of ST_Equals(actual, expected):

https://github.com/apache/sedona-db/blob/9115c194b013d30bb21ade974642229387b9a5b9/rust/sedona-testing/src/compare.rs#L256-L288

c/sedona-geos/src/st_concavehull.rs
```diff
@@ -0,0 +1,237 @@
...
+
+/// ST_ConcaveHull() implementation using the geos crate
+pub fn st_concave_hull_impl() -> ScalarKernelRef {
+    Arc::new(STConcaveHull {})
+}
+
+#[derive(Debug)]
+struct STConcaveHull {}
+
+impl SedonaScalarKernel for STConcaveHull {
```
paleolimbot, line 73: For the purposes of this PR I think just adding one kernel makes sense (whichever one is faster!)

rust/sedona-geo/src/st_concavehull.rs
Abeeujah: Using this gets it right, thank you for recommending!

rust/sedona-geo/src/st_concavehull.rs
Abeeujah: This has been helpful, but for MultiPolygons, PostGis returns a MultiPolygon after ST_ConcaveHull, the `geo` crate here returns a Polygon.
```sh
actual WKB scalar:
  POLYGON((6 1,8 1,8 3,6 3,6 3,5 5,2 5,2 2,5 2,6 1))
expected WKB scalar:
  MULTIPOLYGON(((5 2,6 3,8 3,8 1,6 1,5 2)),((2 5,5 5,5 2,2 2,2 5)))
  ```
  They tend to be topologically identical, but I'm guessing being different geometries is the reason why it's failing.
  
  I tried processing the Polygons in the Multipolygon individually, I get this
  ```sh
  actual WKB scalar:
  MULTIPOLYGON(((5 2,5 5,2 5,2 2,5 2)),((8 1,8 3,6 3,6 1,8 1)))
expected WKB scalar:
  MULTIPOLYGON(((5 2,6 3,8 3,8 1,6 1,5 2)),((2 5,5 5,5 2,2 2,2 5)))
  ```

File: Cargo.toml (hunk 1 above)
Comment (paleolimbot) [reply Cargo.toml]: Ok I see...the concave hull feature was added in GEOS 3.11. The version of GEOS in Ubuntu 24.04 is 3.12 and so this shouldn't be a problem. If it is we could add some opt-in feature for various geos versions.

File: c/sedona-geos/src/st_concavehull.rs (hunk 4 above)
Comment (paleolimbot) [reply c/sedona-geos/src/st_concavehull.rs:73]: Ok I take this back...feel free to add both of these as long as they're tested, which they mostly are! (I'll leave a suggestion below about how to avoid the allow(dead_code).)

File: c/sedona-geos/src/st_concavehull.rs [hunk 5]
```diff
@@ -0,0 +1,237 @@
...
+}
+
+fn concave_hull_via_geos_sys(input_wkb: &[u8], ratio: f64, allow_holes: bool) -> Result<Vec<u8>> {
+    unsafe {
+        let ctx = geos_sys::GEOS_init_r();
+        if ctx.is_null() {
+            return Err(DataFusionError::Execution(
+                "Failed to initialize GEOS context".to_string(),
+            ));
+        }
```
paleolimbot: It looks like this has been added upstream:

https://github.com/georust/geos/blob/47afbad2483e489911ddb456417808340e9342c3/src/geometry.rs#L2443-L2454

I asked on the GeoRust discord about the possibility of a release...I think it's likely we can help with one before our next release. In the meantime, you could update our dependency in Cargo.toml to a Git dependency and use that function directly.

rust/sedona-geo/src/st_concavehull.rs
```diff
@@ -0,0 +1,270 @@
...
+        matcher.match_args(args)
+    }
+
+    fn invoke_batch(
+        &self,
+        arg_types: &[SedonaType],
+        args: &[ColumnarValue],
+    ) -> Result<ColumnarValue> {
+        invoke_batch_impl(arg_types, args)
+    }
```
paleolimbot: I'm not sure we need this wrapper function here (Maybe just inline `invoke_batch_impl()` here?)

rust/sedona-geo/src/st_concavehull.rs
```diff
@@ -0,0 +1,270 @@
...
+        );
+        assert_eq!(tester.return_type().unwrap(), WKB_GEOMETRY);
+
+        let result = tester
+            .invoke_scalar_scalar("LINESTRING (50 50, 150 150, 150 50)", 0.1)
+            .unwrap();
+        assert_scalar_equal_wkb_geometry_topologically(
+            &result,
+            Some("POLYGON ((50 50, 150 150, 150 50, 50 50))"),
+        );
```
paleolimbot: If we'd like to keep this implementation in the PR (which I think is great!), we need one test for every branch you have in the implementation (which here would be one for every geometry type plus the empty point). Testing all the empties is probably a good thing to do, too.

rust/sedona-geo/src/lib.rs
```diff
@@ -18,6 +18,7 @@ pub mod centroid;
 pub mod register;
 mod st_area;
 mod st_centroid;
+mod st_concavehull;
```
paleolimbot:
```suggestion
pub mod st_concavehull;
```

(then you can remove the allow dead code in that module and interested parties can manually register this implementation if they want to try it)

rust/sedona-geo/src/st_concavehull.rs
Abeeujah: I was working on two kernels, one to accept the optional boolean parameter to allow holes and the other that uses the default `false` value for the allow holes parameter, that's why I had the `invoke_batch_impl` there.

But this particular library wasn't actually supporting allow holes, so I'd just go ahead and inline here.

rust/sedona-geo/src/st_concavehull.rs
Abeeujah: Would do that!

rust/sedona-geo/src/lib.rs
Abeeujah: Yeah, didn't deep this at the time, I only unregistered the function as it was taking precedence over the other `geos` implementation, would remove the clippy rule, and fix this.

c/sedona-geos/src/st_concavehull.rs
Abeeujah: I'll update `geos` and remove the extra `geo-sys` dependency I was using.

Abeeujah: Upgrading to this version, and lots of errors, about 40 errors in 10 files, I could fix them all, but I don't think that would be good for this PR, would probably convert it to a two stage job, update to feature `3.11.0` and then implement `ST_ConcaveHull` using geos.

Would finish the `geo` version instead in this PR, unless you would be cool with reviewing all the changes in this one PR.

paleolimbot: I'm happy to review all the changes in this PR or in two PRs, whichever is easier for you!

paleolimbot (approved):
Thank you!

I see you removed the geo-based implementation...happy to review if you're ever in the mood to re-add!

Abeeujah:
> If we'd like to keep this implementation in the PR (which I think is great!), we need one test for every branch you have in the implementation (which here would be one for every geometry type plus the empty point). Testing all the empties is probably a good thing to do, too.

It was panic-ing for Empty Points, Empty Points in Multi-Points and a GeometryCollection, I didn't exactly have a fix for it, and didn't want it getting in the way. 

> I see you removed the geo-based implementation...happy to review if you're ever in the mood to re-add!

Would open up another PR for the geo-based implementation.

Issue #436, opened 2025-12-11
Proposal: Spatial-Partitioned Out-of-Core Spatial Join


Kontinuation:
SedonaDB's `SpatialJoinExec` currently buffers the entire build side, builds a single in-memory R-tree, and streams the probe side. This approach works well when the build side and its index fit in memory but fails with memory reservation failures or OOM errors for large datasets.

We propose introducing a **spatial-partitioned out-of-core execution path** that keeps memory usage bounded by:

1.  Partitioning the build side into disjoint partitions that each fit in memory.
2.  Building one in-memory R-tree per partition and probing it before moving to the next partition.

The design intentionally mirrors hybrid/grace hash join but adapts it to SedonaDB’s spatial join requirements.

paleolimbot:
Closing as this is now implemented (!!!)

PR #443, opened 2025-12-12
feat(rust/sedona-spatial-join): Add spatial partitioners for partitioning spatially joined data by partitioning grids

Kontinuation:
This is part of #436 . We defined interface for spatial partitioners and implement several spatial partitioners for different purposes:

1. KDB: The main partitioner for partitioning the build side
2. R-tree: partitioning dataset by a given set of partitioning grids
3. Similar to R-tree, but does not use spatial indexes. This is faster when the number of partitioning grid is small.

### DIFF
--- a/rust/sedona-spatial-join/src/partitioning/kdb.rs
+++ b/rust/sedona-spatial-join/src/partitioning/kdb.rs
@@ -123,6 +123,9 @@ impl KDBTree {
         max_levels: usize,
         extent: BoundingBox,
     ) -> Result<Self> {
+        if max_items_per_node == 0 {
+            return sedona_internal_err!("max_items_per_node must be greater than 0");
+        }
         let Some(extent_rect) = bbox_to_geo_rect(&extent)? else {
             return sedona_internal_err!("KDBTree extent cannot be empty");
         };


rust/sedona-spatial-join/src/partitioning/util.rs
```diff
@@ -0,0 +1,186 @@
...
+    if bbox.x().is_wraparound() {
+        return Err(DataFusionError::Execution(
+            "BoundingBox has wraparound coordinates, which is not supported yet".to_string(),
+        ));
+    }
+
+    // Convert lo bounds - these are inclusive, so we want to round down
+    // to ensure we don't exclude any points
+    let min_x = bbox.x().lo() as f32;
+    let min_y = bbox.y().lo() as f32;
```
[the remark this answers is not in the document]
Kontinuation: Reimplement this using geo-index's `f64_box_to_f32`.

rust/sedona-spatial-join/src/partitioning/flat.rs
```diff
@@ -0,0 +1,177 @@
...
+
+/// Spatial partitioner that linearly scans partition boundaries.
+pub struct FlatPartitioner {
+    boundaries: Vec<Rect<f32>>,
+    num_partitions: usize,
+}
+
+impl FlatPartitioner {
+    /// Create a new flat partitioner from explicit partition boundaries.
+    pub fn new(boundaries: Vec<BoundingBox>) -> Result<Self> {
```
paleolimbot: Are we passing around a sufficient number of these that it would be faster/use less memory to have a `BoundingBox2d`?

rust/sedona-spatial-join/src/partitioning/flat.rs
```diff
@@ -0,0 +1,177 @@
...
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+//! Flat (linear scan) spatial partitioner.
+//!
+//! This module provides a minimal partitioner that shares the same
+//! intersection semantics as [`crate::partitioning::rtree::RTreePartitioner`]
+//! but avoids the RTree indexing overhead. It stores partition boundaries
```
paleolimbot, line 22: If you happen to know how many partition boundaries it takes to make the RTree worth it, it might be nice to add it to this comment

rust/sedona-spatial-join/src/partitioning/flat.rs
```diff
@@ -0,0 +1,177 @@
...
+    }
+
+    #[test]
+    fn test_flat_partitioner_regular() {
+        let partitioner = FlatPartitioner::new(sample_partitions()).unwrap();
+        let bbox = BoundingBox::xy((10.0, 20.0), (10.0, 20.0));
+        match partitioner.partition(&bbox).unwrap() {
+            SpatialPartition::Regular(id) => assert_eq!(id, 0),
+            _ => panic!("expected Regular partition"),
+        }
```
paleolimbot: Is there a reason this won't work? (Here and below)

```suggestion
        assert_eq!(partitioner.partition(&bbox).unwrap(), SpatialPartition::Regular(0));
```

rust/sedona-spatial-join/src/partitioning/kdb.rs
```diff
@@ -0,0 +1,843 @@
...
+        });
+
+        Ok(sp.unwrap_or(SpatialPartition::None))
+    }
+
+    fn partition_no_multi(&self, bbox: &BoundingBox) -> Result<SpatialPartition> {
+        let rect = bbox_to_geo_rect(bbox)?;
+
+        let mut best_leaf_id: Option<u32> = None;
+        let mut max_area = 0.0_f32;
```
paleolimbot: Above you use `let mut max_area = -1.0_f32` and just compare `area > max_area`. Is there a reason it's different here?

rust/sedona-spatial-join/src/partitioning/kdb.rs
```diff
@@ -0,0 +1,843 @@
...
+        self.items.clear();
+        if let Some(ref mut children) = self.children {
+            for child in children.iter_mut() {
+                child.drop_elements();
+            }
+        }
+    }
+
+    /// Split this node along the specified axis.
+    /// Returns true if the split was successful, false otherwise.
```
paleolimbot, line 297: It might be worth briefly explaining here why this split would fail (it seems like it is mostly if you have 100% identical rects)

rust/sedona-spatial-join/src/partitioning/kdb.rs
```diff
@@ -0,0 +1,843 @@
...
+            } else {
+                item.min().y <= split_coord
+            };
+
+            if belongs_to_left {
+                left_child.insert_rect(item);
+            } else {
+                right_child.insert_rect(item);
+            }
+        }
```
paleolimbot: Not sure if this is better or not:

```suggestion
        if split_x {
            for item in self.items.drain(..) {
                if item.min().x <= split_coord {
                    left_child.insert_rect(item);
                } else {
                    right_child.insert_rect(item);
                }
            }
        } else {
            for item in self.items.drain(..) {
                if item.min().y <= split_coord {
                    left_child.insert_rect(item);
                } else {
                    right_child.insert_rect(item);
            }
        }
```

rust/sedona-spatial-join/src/partitioning/kdb.rs
```diff
@@ -0,0 +1,843 @@
...
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn test_kdb_insert_and_leaf_assignment() -> Result<()> {
```
paleolimbot: Is there a reason these tests return Result instead of using `unwrap()`? (I tend to like the failures by `unwrap()` but DataFusion seems to like returning results)

rust/sedona-spatial-join/src/partitioning/kdb.rs
Kontinuation: Switched to proposed code for splitting leaf nodes.

rust/sedona-spatial-join/src/partitioning/kdb.rs
Kontinuation: Switched to `unwrap()`. I initially wanted to avoid unwrapping things in test code so returning `Result<()>` would make the test code less verbose.

rust/sedona-spatial-join/src/partitioning/kdb.rs
Kontinuation: Added explanations, and documented how insert_rect handles failures.

rust/sedona-spatial-join/src/partitioning/kdb.rs
Kontinuation: We can use the same pattern here. Updated.

rust/sedona-spatial-join/src/partitioning/flat.rs
Kontinuation: I have removed all panics in the test and assert the resulting spatial partitions in a better way.

rust/sedona-spatial-join/src/partitioning/flat.rs
Kontinuation: I have updated the comment to provide a lower bound of 16.

rust/sedona-spatial-join/src/partitioning/flat.rs
Kontinuation: There's going to be just a few of them so it is not a big deal here.

Another place to consider is bounding box sampling (https://github.com/apache/sedona-db/pull/442). We'll save several megabytes of memory per partition. The total size of sampled bounding boxes won't be too large so we won't risk consuming too much additional memory by not using `BoundingBox2d`, so I decided to leave it as is for now.

Kontinuation:
I just realized that `BoundingBox` can be empty and we should handle empty bounding boxes correctly. The latest commit introduced quite some changes for addressing this problem.

paleolimbot (approved):
Thank you!

PR #444, opened 2025-12-12
feat(rust/sedona-raster-functions): Add RS_Envelope

jesspav:
Add the function `RS_Envelope`.

Also changed the test raster generation function to use a negative scale for y (and made skews smaller), since this is standard and made it far easier to think through the correct/expected results.

Example:
```
> select rs_envelope(rs_Example());
┌─────────────────────────────────────────────────────────────────────────────────────────────────────┐
│                                      rs_envelope(rs_example())                                      │
│                                               geometry                                              │
╞═════════════════════════════════════════════════════════════════════════════════════════════════════╡
│ POLYGON((43.08 79.07,171.07999999999998 143.07,203.07999999999998 207.07,75.08 143.07,43.08 79.07)) │
└─────────────────────────────────────────────────────────────────────────────────────────────────────┘
```

Perf:
```
Benchmarking native-raster-rs_envelope-Array(Raster(64, 64)): Collecting 100 samples in estimated 5.4152 s (5
native-raster-rs_envelope-Array(Raster(64, 64))
                        time:   [10.920 ms 10.952 ms 10.985 ms]
                        change: [+1.9087% +2.5539% +3.1910%] (p = 0.00 < 0.05)
                        Performance has regressed.
Found 3 outliers among 100 measurements (3.00%)
  1 (1.00%) low mild
  2 (2.00%) high mild
```

### DIFF
--- a/rust/sedona-raster-functions/src/rs_envelope.rs
+++ b/rust/sedona-raster-functions/src/rs_envelope.rs
@@ -18,12 +18,13 @@ use std::sync::Arc;
 
 use crate::executor::RasterExecutor;
 use arrow_array::builder::BinaryBuilder;
+use datafusion_common::DataFusionError;
 use datafusion_common::Result;
 use datafusion_expr::{
     scalar_doc_sections::DOC_SECTION_OTHER, ColumnarValue, Documentation, Volatility,
 };
 use sedona_expr::scalar_udf::{SedonaScalarKernel, SedonaScalarUDF};
-use sedona_geometry::wkb_factory::wkb_polygon;
+use sedona_geometry::wkb_factory::write_wkb_polygon;
 use sedona_raster::affine_transformation::to_world_coordinate;
 use sedona_raster::traits::RasterRef;
 use sedona_schema::datatypes::Edges;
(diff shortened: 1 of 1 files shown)


rust/sedona-raster-functions/src/rs_envelope.rs
```diff
@@ -0,0 +1,172 @@
...
+            Ok(())
+        })?;
+
+        executor.finish(Arc::new(builder.finish()))
+    }
+}
+
+/// Create WKB for a polygon for the raster
+fn create_envelope_wkb(raster: &dyn RasterRef) -> Result<Vec<u8>> {
+    // Compute the four corners of the raster in world coordinates.
```
paleolimbot, line 99: This should probably be written directly into the output to avoid the intermediate allocation (i.e., `create_envelope_wkb(raster: _, builder: &mut impl Write) -> Result<()>`). 

https://github.com/apache/sedona-db/blob/f93b9540976dcc0e6d3b5dd65888370b866cc42e/rust/sedona-functions/src/st_envelope.rs#L140-L147

paleolimbot (approved):
Thank you!

I think it's worth fixing the output writing to stream into the builder instead of an intermediate allocation (but also in the context raster iteration overhead this might not matter as much!)

rust/sedona-raster-functions/src/rs_envelope.rs
jesspav: Good catch!

jesspav: thanks... done!

PR #446, opened 2025-12-12
perf(rust/sedona-schema): Improve performance of `deserialize_crs()`

jesspav:
In flamegraphs for PR #430 revealed that the CRS serialization had some opportunities for improvement that would be essential for raster/geo per item CRS functions.  This was going to be especially important for CRS comparisons in joins.

Added a simple benchmarks for lnglat and auth code equality that was run repeatedly with different performance optimizations.  At the start it took 61ms for auth codes and 37ms for lnglat.  By the end auth codes have improved by over 20x and lnglat by over 10x, with both running in 2.7ms.

There are basically two changes: an LRU cache and avoiding unnecessary string allocations.  At the end, checked to see if we no longer needed the cache, but that brought the time back to 12ms.

Breakdown of performance changes:
```
Starting point:


     Running benches/crs_benchmarks.rs (target/release/deps/crs_benchmarks-c62a15af9667b709)
equality_lnglat_crs     time:   [37.442 ms 37.712 ms 37.960 ms]
                        change: [+2.5665% +4.0155% +5.4781%] (p = 0.00 < 0.05)
                        Performance has regressed.
Found 15 outliers among 100 measurements (15.00%)
  6 (6.00%) low severe
  9 (9.00%) low mild

Benchmarking equality_different_crs: Warming up for 3.0000 s
Warning: Unable to complete 100 samples in 5.0s. You may wish to increase target time to 6.3s, or reduce sample count to 70.
equality_different_crs  time:   [61.892 ms 62.119 ms 62.362 ms]
                        change: [+2.1679% +2.6956% +3.2470%] (p = 0.00 < 0.05)
                        Performance has regressed.
Found 7 outliers among 100 measurements (7.00%)
  1 (1.00%) low mild
  1 (1.00%) high mild
  5 (5.00%) high severe


After cache:
equality_lnglat_crs     time:   [28.265 ms 28.361 ms 28.461 ms]
                        change: [−25.338% −24.795% −24.189%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 2 outliers among 100 measurements (2.00%)
  2 (2.00%) high mild

equality_different_crs  time:   [27.473 ms 27.530 ms 27.586 ms]
                        change: [−55.878% −55.683% −55.494%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 2 outliers among 100 measurements (2.00%)
  2 (2.00%) high mild



Time after cache + switch value to str +  + colon rather than split

equality_lnglat_crs     time:   [11.324 ms 11.354 ms 11.385 ms]
                        change: [−60.148% −59.967% −59.798%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 2 outliers among 100 measurements (2.00%)
  1 (1.00%) low mild
  1 (1.00%) high mild

equality_different_crs  time:   [11.655 ms 11.701 ms 11.746 ms]
                        change: [−57.674% −57.497% −57.308%] (p = 0.00 < 0.05)
                        Performance has improved.



Time after cache + switch value to str + colon rather than split + pre-alloc instead of format for eq:
Benchmarking equality_lnglat_crs: Collecting 100 samples in estimated 5.4946 s (800 iterat
equality_lnglat_crs     time:   [6.6983 ms 6.7258 ms 6.7557 ms]
                        change: [+0.7120% +1.2645% +1.8552%] (p = 0.00 < 0.05)
                        Change within noise threshold.
  1 (1.00%) high mild
  4 (4.00%) high severe

Benchmarking equality_different_crs: Collecting 100 samples in estimated 5.4518 s (800 ite
equality_different_crs  time:   [6.7116 ms 6.7429 ms 6.7758 ms]
                        change: [+0.7410% +1.5266% +2.2864%] (p = 0.00 < 0.05)
                        Change within noise threshold.
Found 4 outliers among 100 measurements (4.00%)
  4 (4.00%) high mild


All of the above +  not splitting up the auth code:
equality_lnglat_crs     time:   [3.3993 ms 3.4098 ms 3.4200 ms]
                        change: [−49.633% −49.425% −49.229%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 3 outliers among 100 measurements (3.00%)
  3 (3.00%) low mild

equality_different_crs  time:   [3.4080 ms 3.4170 ms 3.4258 ms]
                        change: [−49.183% −48.996% −48.819%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 1 outliers among 100 measurements (1.00%)
  1 (1.00%) low mild

with thread local cache:
     Running benches/crs_benchmarks.rs (target/release/deps/crs_benchmarks-c62a15af9667b709)
equality_lnglat_crs     time:   [2.6705 ms 2.6748 ms 2.6790 ms]
                        change: [−1.5862% −1.2966% −1.0092%] (p = 0.00 < 0.05)
                        Performance has improved.

equality_different_crs  time:   [2.6570 ms 2.6619 ms 2.6668 ms]
                        change: [−2.7208% −1.8683% −1.2686%] (p = 0.00 < 0.05)
                        Performance has improved.

```

### DIFF
--- a/rust/sedona-schema/src/crs.rs
+++ b/rust/sedona-schema/src/crs.rs
@@ -61,7 +61,7 @@ pub fn deserialize_crs(crs_str: &str) -> Result<Crs> {
         Some(Arc::new(projjson) as Arc<dyn CoordinateReferenceSystem + Send + Sync>)
     };
 
-    // Cache the result
+    // Cache result
     CRS_CACHE.with(|cache| {
         cache.borrow_mut().put(crs_str.to_string(), crs.clone());
     });


c/sedona-proj/src/sd_order_lnglat.rs
```diff
@@ -190,8 +189,7 @@ mod test {
         assert_eq!(&result, &expected);
 
         // Check the "not already lnglat" case
-        let crs =
-            sedona_schema::crs::deserialize_crs(&Value::String("EPSG:3857".to_string())).unwrap();
```
jesspav, line 194: @paleolimbot These string allocations were clearly in the flame graph, and all tests continued to pass without them.  What do you think about this change?

rust/sedona-schema/src/crs.rs
```diff
@@ -15,41 +15,61 @@
...
 
+/// LRU cache for CRS deserialization
+static CRS_CACHE: OnceLock<Mutex<LruCache<String, Crs>>> = OnceLock::new();
+
+fn get_crs_cache() -> &'static Mutex<LruCache<String, Crs>> {
+    CRS_CACHE.get_or_init(|| {
+        // Cache up to 256 CRS strings
+        Mutex::new(LruCache::new(NonZeroUsize::new(256).unwrap()))
+    })
+}
```
paleolimbot: Can/should this be thread local to avoid contention when CRS computations are happening on many partitions at once?

rust/sedona-schema/src/datatypes.rs
```diff
@@ -312,7 +312,13 @@ fn deserialize_edges_and_crs(value: &Option<String>) -> Result<(Edges, Crs)> {
             };
 
             let crs = match json_value.get("crs") {
-                Some(crs_value) => deserialize_crs(crs_value)?,
+                Some(crs_value) => {
+                    if let Some(s) = crs_value.as_str() {
+                        deserialize_crs(s)?
+                    } else {
+                        deserialize_crs(&crs_value.to_string())?
```
paleolimbot: The original reason for keeping the `Value` here was to avoid the re-serialization. We don't benchmark query planning but this code path gets called every time we take a `FieldRef` from DataFusion and convert it to a `SedonaType` (at least once per batch when calling our scalar functions).

Maybe:

```rust
match &crs_value {
    Value::Object(obj) => // convert using the old path,
    Value::String(s) => deserialize_crs(s)?,
    Value::Null => None
}
```

...would improve the current state of this PR (which is FieldRef metadata -> serde_json Value -> string -> serde_json Value for PROJJSON crses).

jesspav: makes sense.  I'll switch it back

rust/sedona-schema/src/crs.rs
jesspav: Yes!!! That should have an impact on the perf of a single thread, too, since it avoids the lock.

any opinion on the magic number for size of the cache?  this could get very large if all the keys are json!!

jesspav: I made it 50 for now.  I think this should be configurable given the perf implications.  I'll can add an issue so we don't forget.

rust/sedona-schema/src/datatypes.rs
```diff
@@ -312,8 +312,13 @@ fn deserialize_edges_and_crs(value: &Option<String>) -> Result<(Edges, Crs)> {
             };
 
             let crs = match json_value.get("crs") {
-                Some(crs_value) => deserialize_crs(crs_value)?,
-                None => Crs::None,
+                Some(crs_value) => match &crs_value {
+                    Value::Object(_obj) => deserialize_crs_from_obj(crs_value)?,
+                    Value::String(s) => deserialize_crs(s)?,
+                    Value::Number(s) => deserialize_crs(&s.to_string())?,
+                    _ => None,
```
paleolimbot, line 319: This case should probably error (I think it would be the case where CRS is a JSON list)

paleolimbot (commented):
Thank you! One nit but other than that it looks great! I think the CI failure is just download issues.

PR #462, opened 2025-12-16
feat(rust/sedona-geo): ST_ConcaveHull Geo Implementation

Abeeujah:


rust/sedona-geo/src/st_concavehull.rs
```diff
@@ -0,0 +1,364 @@
...
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+#![allow(dead_code)]
+use std::sync::Arc;
```
paleolimbot:
```suggestion
use std::sync::Arc;
```

rust/sedona-geo/src/st_concavehull.rs
```diff
@@ -0,0 +1,364 @@
...
+use sedona_geometry::is_empty;
+use sedona_geometry::wkb_factory::WKB_MIN_PROBABLE_BYTES;
+use sedona_schema::datatypes::SedonaType;
+use sedona_schema::{datatypes::WKB_GEOMETRY, matchers::ArgMatcher};
+use wkb::reader::Wkb;
+use wkb::writer::{write_geometry, WriteOptions};
+
+use crate::to_geo::item_to_geometry;
+
+/// ST_ConcaveHull implementation using [ConcaveHull]
```
paleolimbot, line 38: This would be a good place to put a note about the difference between this implementation and the GEOS one (notably, that this one only returns a Polygon whereas the GEOS implementation sometimes returns a MultiPolygon).

rust/sedona-geo/src/st_concavehull.rs
```diff
@@ -0,0 +1,364 @@
...
+fn write_opts() -> WriteOptions {
+    WriteOptions {
+        endianness: wkb::Endianness::LittleEndian,
+    }
+}
+
+fn normalize_geometry(geom: &Wkb) -> Result<Geometry> {
+    match geom.as_type() {
+        geo_traits::GeometryType::GeometryCollection(gc) => {
+            let filtered: Vec<Geometry> = gc
```
paleolimbot, line 116: I see what you mean now...this is a nice strategy (filter out unsupported geometries since they don't affect the result).

rust/sedona-geo/src/st_concavehull.rs
```diff
@@ -0,0 +1,364 @@
...
+        );
+        assert_eq!(tester.return_type().unwrap(), WKB_GEOMETRY);
+
+        let result = tester
+            .invoke_scalar_scalar(ScalarValue::Null, ScalarValue::Null)
+            .unwrap();
+        assert!(result.is_null());
+
+        let result = tester.invoke_scalar_scalar("POINT EMPTY", 0.1).unwrap();
+        assert_scalar_equal_wkb_geometry_topologically(&result, Some("POLYGON EMPTY"));
```
paleolimbot: It might be nice to check all the EMPTYs in one place to make sure you have them all (I think this is missing POLYGON EMPTY).

Similarly, it would be easier to read these tests if there was a group with differing geometry types (but identical values for other parameters), and then a group with differing parameter values (using an identical geometry). I'm being slightly pickier about these tests because we can't rely on the PostGIS comparison for correctness.

Abeeujah: I wanted going the array route for grouping the tests but `assert_array_equal` doesn't have an arm that checks if two Geometries are Topologically equal.

rust/sedona-geo/src/st_concavehull.rs
Abeeujah: Yeah, exactly how it is!

rust/sedona-geo/src/st_concavehull.rs
Abeeujah: Noted, will fix!

rust/sedona-geo/src/st_concavehull.rs
paleolimbot: No problem! You can group the scalar calls in this way too and it will still be a little easier to read. The empties should work with the usual `assert_array_equal()` if you wanted to do it that way (but even just putting those test together and giving them a comment to separate them visually is great too).

rust/sedona-geo/src/st_concavehull.rs
```diff
@@ -0,0 +1,419 @@
...
+    #[case::collection("GEOMETRYCOLLECTION EMPTY")]
+    fn test_empty_geometries(
+        #[values(WKB_GEOMETRY, WKB_VIEW_GEOMETRY)] sedona_type: SedonaType,
+        #[case] input_wkt: &str,
+    ) {
+        let tester = create_tester(sedona_type);
+        // Empty geometries should result in an empty Polygon
+        let result = tester.invoke_scalar_scalar(input_wkt, 0.1).unwrap();
+        assert_scalar_equal_wkb_geometry_topologically(&result, Some("POLYGON EMPTY"));
+    }
```
paleolimbot, line 369: In other functions we test this as an array of input -> array of output. Is there a reason we use a parameterized version here?

We also probably don't need to test the output topologically for the empties (although it's not hurting to do so!)

rust/sedona-geo/src/st_concavehull.rs
```diff
@@ -0,0 +1,419 @@
...
+
+    #[rstest]
+    // =========================================================================
+    // Differing Geometry Types, Identical Percentage (0.1)
+    // =========================================================================
+
+    // Point
+    #[case("POINT (2.5 3.1)", 0.1, "POINT (2.5 3.1)")]
```
paleolimbot: We haven't tested most functions in this type of parameterized way so far, and I am not sure we should start doing so unless there's a reason I haven't considered. The reason that I don't like this is that it's harder in the IDE to debug a specific case when there are more than a few cases. In Python the IDE makes this a bit easier.

Also,

```rust
let result = tester.invoke_scalar_scalar("LINESTRING (50 50, 150 150, 150 50)", 0.1).unwrap();
assert_scalar_equal_wkb_geometry_topologically(result, "POLYGON ((50 50, 150 150, 150 50, 50 50))");
```

...is not meaningfully more verbose than the case statements you have here.

paleolimbot (approved):
Thank you!

PR #465, opened 2025-12-17
feat(rust/sedona-spatial-join): Add GPU-accelerated spatial join support

zhangfengcdt:
This commit introduces GPU-accelerated spatial join capabilities to SedonaDB, enabling significant performance improvements for large-scale spatial join operations.

Key changes:
- Add new `sedona-spatial-join-gpu` crate that provides GPU-accelerated spatial join execution using CUDA via the `sedona-libgpuspatial` library.
- Implement `GpuSpatialJoinExec` execution plan with build/probe phases that efficiently handles partitioned data by sharing build-side data across probes.
- Add GPU backend abstraction (`GpuBackend`) for geometry data transfer and spatial predicate evaluation on GPU.
- Extend the spatial join optimizer to automatically select GPU execution when available and beneficial, with configurable thresholds and fallback to CPU.
- Add configuration options in `SedonaOptions` for GPU spatial join settings including enable/disable, row thresholds, and CPU fallback behavior.
- Include comprehensive benchmarks and functional tests for GPU spatial join correctness validation against CPU reference implementations.

### DIFF
--- a/c/sedona-libgpuspatial/build.rs
+++ b/c/sedona-libgpuspatial/build.rs
@@ -126,22 +126,11 @@ fn main() {
             "Release"
         };
 
-        let mut config = cmake::Config::new("./libgpuspatial");
-        config
-            .define("CMAKE_CUDA_ARCHITECTURES", cuda_architectures) // or your variable
-            .define("CMAKE_POLICY_VERSION_MINIMUM", "3.5")
-            .define("LIBGPUSPATIAL_LOGGING_LEVEL", "WARN");
-        let profile = env::var("PROFILE").unwrap();
-        if profile == "debug" {
-            println!("cargo:warning=Building with AddressSanitizer (ASan) enabled.");
-            config
-                .define(
-                    "CMAKE_CXX_FLAGS",
-                    "-fsanitize=address -fno-omit-frame-pointer",
-                )
-                .define("CMAKE_SHARED_LINKER_FLAGS", "-fsanitize=address");
-        }
-        let dst = config.build();
+        let dst = cmake::Config::new("./libgpuspatial")
+            .define("CMAKE_CUDA_ARCHITECTURES", cuda_architectures)
+            .define("CMAKE_POLICY_VERSION_MINIMUM", "3.5") // Allow older CMake versions
+            .define("LIBGPUSPATIAL_LOGGING_LEVEL", "INFO") // Set logging level
+            .build();
         let include_path = dst.join("include");
         println!(
             "cargo:rustc-link-search=native={}",
--- a/c/sedona-libgpuspatial/libgpuspatial/include/gpuspatial/gpuspatial_c.h
+++ b/c/sedona-libgpuspatial/libgpuspatial/include/gpuspatial/gpuspatial_c.h
@@ -25,33 +25,34 @@ struct ArrowSchema;
 struct ArrowArray;
 
 // Interfaces for ray-tracing engine (OptiX)
-struct GpuSpatialRTEngineConfig {
+struct GpuSpatialRuntimeConfig {
   /** Path to PTX files */
   const char* ptx_root;
   /** Device ID to use, 0 is the first GPU */
   int device_id;
+  /** Ratio of initial memory pool size to total GPU memory, between 0.0 and 1.0; zero is
+   * effectively disable async memory allocation and using cudaMalloc */
+  float cuda_init_memory_pool_ratio;
 };
 
-struct GpuSpatialRTEngine {
-  /** Initialize the ray-tracing engine (OptiX) with the given configuration
+struct GpuSpatialRuntime {
+  /** Initialize the runtime (OptiX) with the given configuration
    * @return 0 on success, non-zero on failure
    */
-  int (*init)(struct GpuSpatialRTEngine* self, struct GpuSpatialRTEngineConfig* config);
-  void (*release)(struct GpuSpatialRTEngine* self);
-  const char* (*get_last_error)(struct GpuSpatialRTEngine* self);
+  int (*init)(struct GpuSpatialRuntime* self, struct GpuSpatialRuntimeConfig* config);
+  void (*release)(struct GpuSpatialRuntime* self);
+  const char* (*get_last_error)(struct GpuSpatialRuntime* self);
   void* private_data;
 };
 
-/** Create an instance of GpuSpatialRTEngine */
-int GpuSpatialRTEngineCreate(struct GpuSpatialRTEngine* instance);
+/** Create an instance of GpuSpatialRuntime */
+void GpuSpatialRuntimeCreate(struct GpuSpatialRuntime* runtime);
 
 struct GpuSpatialIndexConfig {
-  /** Pointer to an initialized GpuSpatialRTEngine struct */
-  struct GpuSpatialRTEngine* rt_engine;
+  /** Pointer to an initialized GpuSpatialRuntime struct */
+  struct GpuSpatialRuntime* runtime;
   /** How many threads will concurrently call Probe method */
   uint32_t concurrency;
-  /** Device ID to use, 0 is the first GPU */
-  int device_id;
 };
 
 // An opaque context for concurrent probing
@@ -114,12 +115,10 @@ int GpuSpatialIndexFloat2DCreate(struct SedonaFloatIndex2D* index,
                                  const struct GpuSpatialIndexConfig* config);
 
 struct GpuSpatialRefinerConfig {
-  /** Pointer to an initialized GpuSpatialRTEngine struct */
-  struct GpuSpatialRTEngine* rt_engine;
+  /** Pointer to an initialized GpuSpatialRuntime struct */
+  struct GpuSpatialRuntime* runtime;
   /** How many threads will concurrently call Probe method */
   uint32_t concurrency;
-  /** Device ID to use, 0 is the first GPU */
-  int device_id;
   /** Whether to compress the BVH structures to save memory */
   bool compress_bvh;
   /** Number of batches to pipeline for parsing and refinement; setting to 1 disables
--- a/c/sedona-libgpuspatial/libgpuspatial/src/gpuspatial_c.cc
+++ b/c/sedona-libgpuspatial/libgpuspatial/src/gpuspatial_c.cc
@@ -22,6 +22,10 @@
 #include "gpuspatial/rt/rt_engine.hpp"
 #include "gpuspatial/utils/exception.h"
 
+#include "rmm/mr/device/cuda_async_memory_resource.hpp"
+#include "rmm/mr/device/per_device_resource.hpp"
+#include "rmm/mr/device/pool_memory_resource.hpp"
+
 #include <threads.h>
 #include <algorithm>
 #include <cstring>
@@ -58,56 +62,81 @@ int SafeExecute(GpuSpatialWrapper<T>* wrapper, Func&& func) {
 // IMPLEMENTATION
 // -----------------------------------------------------------------------------
 
-struct GpuSpatialRTEngineExporter {
-  using private_data_t = GpuSpatialWrapper<std::shared_ptr<gpuspatial::RTEngine>>;
-  static void Export(private_data_t* private_data, struct GpuSpatialRTEngine* out) {
+struct GpuSpatialRuntimeExporter {
+  struct Payload {
+    std::shared_ptr<gpuspatial::RTEngine> rt_engine;
+    std::unique_ptr<rmm::mr::cuda_async_memory_resource> upstream_mr;
+    std::unique_ptr<rmm::mr::pool_memory_resource<rmm::mr::cuda_async_memory_resource>>
+        pool_mr;
+    int device_id;
+  };
+
+  using private_data_t = GpuSpatialWrapper<Payload>;
+  static void Export(struct GpuSpatialRuntime* out) {
+    private_data_t* private_data =
+        new private_data_t{Payload{std::make_shared<gpuspatial::RTEngine>()}, ""};
     out->init = CInit;
     out->release = CRelease;
     out->get_last_error = CGetLastError;
     out->private_data = private_data;
   }
 
-  static int CInit(GpuSpatialRTEngine* self, GpuSpatialRTEngineConfig* config) {
+  static int CInit(GpuSpatialRuntime* self, GpuSpatialRuntimeConfig* config) {
     return SafeExecute(static_cast<private_data_t*>(self->private_data), [&] {
       std::string ptx_root(config->ptx_root);
       auto rt_config = gpuspatial::get_default_rt_config(ptx_root);
 
+      GPUSPATIAL_LOG_INFO("Initializing GpuSpatialRuntime on device %d, PTX root %s",
+                          config->device_id, config->ptx_root);
+
       CUDA_CHECK(cudaSetDevice(config->device_id));
-      static_cast<private_data_t*>(self->private_data)->payload->Init(rt_config);
+
+      float mem_pool_ratio = config->cuda_init_memory_pool_ratio;
+
+      if (mem_pool_ratio < 0 || mem_pool_ratio > 1) {
+        throw std::invalid_argument(
+            "cuda_init_memory_pool_ratio must be between 0 and 1");
+      }
+
+      if (mem_pool_ratio > 0) {
+        auto async_mr = std::make_unique<rmm::mr::cuda_async_memory_resource>();
+        auto pool_size = rmm::percent_of_free_device_memory(mem_pool_ratio);
+
+        GPUSPATIAL_LOG_INFO("Creating RMM pool memory resource with size %zu MB",
+                            pool_size / 1024 / 1024);
+
+        auto pool_mr = std::make_unique<
+            rmm::mr::pool_memory_resource<rmm::mr::cuda_async_memory_resource>>(
+            async_mr.get(), pool_size);
+
+        rmm::mr::set_current_device_resource(pool_mr.get());
+        static_cast<private_data_t*>(self->private_data)->payload.upstream_mr =
+            std::move(async_mr);
+        static_cast<private_data_t*>(self->private_data)->payload.pool_mr =
+            std::move(pool_mr);
+      }
+
+      static_cast<private_data_t*>(self->private_data)
+          ->payload.rt_engine->Init(rt_config);
     });
   }
 
-  static void CRelease(GpuSpatialRTEngine* self) {
+  static void CRelease(GpuSpatialRuntime* self) {
     delete static_cast<private_data_t*>(self->private_data);
     self->private_data = nullptr;
   }
 
-  static const char* CGetLastError(GpuSpatialRTEngine* self) {
+  static const char* CGetLastError(GpuSpatialRuntime* self) {
     auto* private_data = static_cast<private_data_t*>(self->private_data);
     return private_data->last_error.c_str();
   }
 };
 
-int GpuSpatialRTEngineCreate(struct GpuSpatialRTEngine* instance) {
-  try {
-    auto rt_engine = std::make_shared<gpuspatial::RTEngine>();
-    GpuSpatialRTEngineExporter::Export(
-        new GpuSpatialWrapper<std::shared_ptr<gpuspatial::RTEngine>>{rt_engine},
-        instance);
-  } catch (std::exception& e) {
-    GpuSpatialRTEngineExporter::Export(
-        new GpuSpatialWrapper<std::shared_ptr<gpuspatial::RTEngine>>{nullptr, e.what()},
-        instance);
-    return EINVAL;
-  } catch (...) {
-    GpuSpatialRTEngineExporter::Export(
-        new GpuSpatialWrapper<std::shared_ptr<gpuspatial::RTEngine>>{nullptr,
-                                                                     "Unknown error"},
-        instance);
-    return EINVAL;
-  }
-  return 0;
+void GpuSpatialRuntimeCreate(struct GpuSpatialRuntime* runtime) {
+  GpuSpatialRuntimeExporter::Export(runtime);
 }
+
+using runtime_data_t = GpuSpatialRuntimeExporter::private_data_t;
 
 struct GpuSpatialIndexFloat2DExporter {
   using scalar_t = float;
@@ -117,7 +146,7 @@ struct GpuSpatialIndexFloat2DExporter {
 
   struct Payload {
     std::unique_ptr<spatial_index_t> index;
-    int device_id;
+    runtime_data_t* rdata;
   };
 
   struct ResultBuffer {
@@ -135,8 +164,20 @@ struct GpuSpatialIndexFloat2DExporter {
   using private_data_t = GpuSpatialWrapper<Payload>;
   using context_t = GpuSpatialWrapper<ResultBuffer>;
 
-  static void Export(std::unique_ptr<spatial_index_t> index, int device_id,
-                     const std::string& last_error, struct SedonaFloatIndex2D* out) {
+  static void Export(const struct GpuSpatialIndexConfig* config,
+                     struct SedonaFloatIndex2D* out) {
+    auto* rdata = static_cast<runtime_data_t*>(config->runtime->private_data);
+
+    gpuspatial::RTSpatialIndexConfig index_config;
+
+    index_config.rt_engine = rdata->payload.rt_engine;
+    index_config.concurrency = config->concurrency;
+
+    // Create SpatialIndex may involve GPU operations, set device here
+    CUDA_CHECK(cudaSetDevice(rdata->payload.device_id));
+
+    auto uniq_index = gpuspatial::CreateRTSpatialIndex<float, 2>(index_config);
+
     out->clear = &CClear;
     out->create_context = &CCreateContext;
     out->destroy_context = &CDestroyContext;
@@ -148,8 +189,7 @@ struct GpuSpatialIndexFloat2DExporter {
     out->get_last_error = &CGetLastError;
     out->context_get_last_error = &CContextGetLastError;
     out->release = &CRelease;
-    out->private_data =
-        new private_data_t{Payload{std::move(index), device_id}, last_error};
+    out->private_data = new private_data_t{Payload{std::move(uniq_index), rdata}, ""};
   }
 
   static void CCreateContext(struct SedonaSpatialIndexContext* context) {
@@ -220,35 +260,19 @@ struct GpuSpatialIndexFloat2DExporter {
 
   static spatial_index_t& use_index(self_t* self) {
     auto* private_data = static_cast<private_data_t*>(self->private_data);
+    auto* r_data = private_data->payload.rdata;
 
-    CUDA_CHECK(cudaSetDevice(private_data->payload.device_id));
-    if (private_data->payload.index == nullptr) {
-      throw std::runtime_error("SpatialIndex is not initialized");
-    }
+    CUDA_CHECK(cudaSetDevice(r_data->payload.device_id));
     return *(private_data->payload.index);
   }
 };
 
 int GpuSpatialIndexFloat2DCreate(struct SedonaFloatIndex2D* index,
                                  const struct GpuSpatialIndexConfig* config) {
-  gpuspatial::RTSpatialIndexConfig rt_index_config;
-  auto rt_engine = static_cast<GpuSpatialWrapper<std::shared_ptr<gpuspatial::RTEngine>>*>(
-                       config->rt_engine->private_data)
-                       ->payload;
-  rt_index_config.rt_engine = rt_engine;
-  rt_index_config.concurrency = config->concurrency;
   try {
-    if (rt_index_config.rt_engine == nullptr) {
-      throw std::runtime_error("RTEngine is not initialized");
-    }
-    // Create SpatialIndex may involve GPU operations, set device here
-    CUDA_CHECK(cudaSetDevice(config->device_id));
-
-    auto uniq_index = gpuspatial::CreateRTSpatialIndex<float, 2>(rt_index_config);
-    GpuSpatialIndexFloat2DExporter::Export(std::move(uniq_index), config->device_id, "",
-                                           index);
+    GpuSpatialIndexFloat2DExporter::Export(config, index);
   } catch (std::exception& e) {
-    GpuSpatialIndexFloat2DExporter::Export(nullptr, config->device_id, e.what(), index);
+    GPUSPATIAL_LOG_ERROR("Failed to create GpuSpatialIndexFloat2D: %s", e.what());
     return EINVAL;
   }
   return 0;
@@ -257,12 +281,26 @@ int GpuSpatialIndexFloat2DCreate(struct SedonaFloatIndex2D* index,
 struct GpuSpatialRefinerExporter {
   struct Payload {
     std::unique_ptr<gpuspatial::SpatialRefiner> refiner;
-    int device_id;
+    runtime_data_t* rdata;
   };
   using private_data_t = GpuSpatialWrapper<Payload>;
 
-  static void Export(std::unique_ptr<gpuspatial::SpatialRefiner> refiner, int device_id,
-                     const std::string& last_error, struct SedonaSpatialRefiner* out) {
+  static void Export(const GpuSpatialRefinerConfig* config,
+                     struct SedonaSpatialRefiner* out) {
+    auto* rdata = static_cast<runtime_data_t*>(config->runtime->private_data);
+
+    gpuspatial::RTSpatialRefinerConfig refiner_config;
+
+    refiner_config.rt_engine = rdata->payload.rt_engine;
+    refiner_config.concurrency = config->concurrency;
+    refiner_config.compact = config->compress_bvh;
+    refiner_config.pipeline_batches = config->pipeline_batches;
+
+    // Create Refinner may involve GPU operations, set device here
+    CUDA_CHECK(cudaSetDevice(rdata->payload.device_id));
+
+    auto refiner = gpuspatial::CreateRTSpatialRefiner(refiner_config);
+
     out->clear = &CClear;
     out->push_build = &CPushBuild;
     out->finish_building = &CFinishBuilding;
@@ -270,8 +308,7 @@ struct GpuSpatialRefinerExporter {
     out->refine = &CRefine;
     out->get_last_error = &CGetLastError;
     out->release = &CRelease;
-    out->private_data =
-        new private_data_t{Payload{std::move(refiner), device_id}, last_error};
+    out->private_data = new private_data_t{Payload{std::move(refiner), rdata}, ""};
   }
 
   static int CClear(SedonaSpatialRefiner* self) {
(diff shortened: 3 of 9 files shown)


c/sedona-libgpuspatial/build.rs
```diff
@@ -157,6 +158,7 @@ fn main() {
         println!("cargo:rustc-link-lib=static=gpuspatial");
         println!("cargo:rustc-link-lib=static=rmm");
         println!("cargo:rustc-link-lib=static=rapids_logger");
+        println!("cargo:rustc-link-lib=static=spdlog");
```
pwrliang: Should we remove this ".define("CMAKE_BUILD_TYPE", "Release")"
and let Cargo to determine the compiling mode?

Under Debug mode, the product of spdlog is named "spdlogd" while in the release mode, it is "spdlog" The following code can automatically determine linking to which target according to the building model
```rust
        let profile_mode = if cfg!(debug_assertions) {
            "debug"
        } else {
            "release"
        };
        // Use the 'd' suffix for the debug build of spdlog (libspdlogd.a)
        let spdlog_lib_name = if profile_mode == "debug" {
            "spdlogd"
        } else {
            "spdlog"
        };

        println!(
            "cargo:warning=Linking spdlog in {} mode: lib{}.a",
            profile_mode, spdlog_lib_name
        );
        println!("cargo:rustc-link-lib=static={}", spdlog_lib_name);
```

c/sedona-libgpuspatial/CMakeLists.txt
pwrliang, line 1: This file is not used and can be removed

zhangfengcdt:
Will merge the patch to fix the failing example build once this is merged - https://github.com/apache/sedona-db/pull/486

rust/sedona-spatial-join-gpu/benches/gpu_spatial_join.rs
```diff
@@ -0,0 +1,360 @@
...
+use sedona_testing::create::create_array_storage;
+use std::sync::Arc;
+use tokio::runtime::Runtime;
+
+// Helper execution plan that returns a single pre-loaded batch
+struct SingleBatchExec {
+    schema: Arc<Schema>,
+    batch: RecordBatch,
+    props: datafusion::physical_plan::PlanProperties,
+}
```
paleolimbot: This seems very similar to `SessionContext::register_batch()` and is a lot of lines of code. Do we need this?

rust/sedona-spatial-join-gpu/benches/gpu_spatial_join.rs
```diff
@@ -0,0 +1,360 @@
...
+    use rand::Rng;
+    let mut rng = rand::thread_rng();
+    (0..count)
+        .map(|_| {
+            let x: f64 = rng.gen_range(-180.0..180.0);
+            let y: f64 = rng.gen_range(-90.0..90.0);
+            format!("POINT ({} {})", x, y)
+        })
+        .collect()
+}
```
paleolimbot: We have a random geometry generator in sedona-testing (that is used in the non-GPU join tests and elsewhere) that I think we should be using here!

rust/sedona-spatial-join-gpu/src/config.rs
```diff
@@ -0,0 +1,89 @@
...
+            },
+            right_geom_column: GeometryColumnInfo {
+                name: "geometry".to_string(),
+                index: 0,
+            },
+            predicate: GpuSpatialPredicate::Relation(
+                sedona_libgpuspatial::SpatialPredicate::Intersects,
+            ),
+            device_id: 0,
+            batch_size: 8192,
```
paleolimbot: Should this be `Option<usize>` so that it can default to the `datafusion.batch_size` setting?

rust/sedona-spatial-join-gpu/src/exec.rs
```diff
@@ -0,0 +1,294 @@
...
+        let schema = Arc::new(join_schema);
+
+        // Create execution properties
+        // Output partitioning matches right side to enable parallelism
+        let eq_props = EquivalenceProperties::new(schema.clone());
+        let partitioning = right.output_partitioning().clone();
+        let properties = PlanProperties::new(
+            eq_props,
+            partitioning,
+            EmissionType::Final, // GPU join produces all results at once
```
paleolimbot: Just checking that this is correct (I thought that because one side is streaming the output might be incremental?)

rust/sedona-spatial-join-gpu/src/gpu_backend.rs
```diff
@@ -0,0 +1,294 @@
...
+
+/// GPU backend for spatial operations
+#[allow(dead_code)]
+pub struct GpuBackend {
+    device_id: i32,
+    gpu_context: Option<GpuSpatialContext>,
+}
+
+#[allow(dead_code)]
+impl GpuBackend {
```
paleolimbot: Can these dead code markers be removed?

rust/sedona-spatial-join-gpu/src/gpu_backend.rs
```diff
@@ -0,0 +1,294 @@
...
+
+    /// Convert BinaryView array to Binary array for GPU processing
+    /// OPTIMIZATION: Use Arrow's optimized cast instead of manual iteration
+    fn ensure_binary_array(array: &ArrayRef) -> Result<ArrayRef> {
+        match array.data_type() {
+            DataType::BinaryView => {
+                // OPTIMIZATION: Use Arrow's cast which is much faster than manual iteration
+                use arrow::compute::cast;
+                cast(array.as_ref(), &DataType::Binary).map_err(crate::Error::Arrow)
+            }
```
paleolimbot: This feels like it should be fixed on the C side in a follow-up (nanoarrow/geoarrow support view types and should be able to ingest them).

rust/sedona-spatial-join-gpu/src/utils/once_fut.rs
```diff
@@ -0,0 +1,165 @@
...
+            state: OnceFutState::Pending(
+                fut.map(|res| res.map(Arc::new).map_err(Arc::new))
+                    .boxed()
+                    .shared(),
+            ),
+        }
+    }
+
+    /// Get the result of the computation if it is ready, without consuming it
+    #[allow(unused)]
```
paleolimbot: Can we remove this if it is not used?

rust/sedona-spatial-join-gpu/tests/gpu_functional_test.rs
```diff
@@ -0,0 +1,588 @@
...
+use std::sync::Arc;
+
+/// Helper to create test geometry data
+#[allow(dead_code)]
+fn create_point_wkb(x: f64, y: f64) -> Vec<u8> {
+    let mut wkb = vec![0x01, 0x01, 0x00, 0x00, 0x00]; // Little endian point type
+    wkb.extend_from_slice(&x.to_le_bytes());
+    wkb.extend_from_slice(&y.to_le_bytes());
+    wkb
+}
```
paleolimbot: We have a helper for this in sedona-testing (and one specifically for points in sedona-geometry)

rust/sedona-spatial-join-gpu/tests/gpu_functional_test.rs
```diff
@@ -0,0 +1,588 @@
...
+    use sedona_libgpuspatial::GpuSpatialContext;
+
+    match GpuSpatialContext::new() {
+        Ok(mut ctx) => ctx.init().is_ok(),
+        Err(_) => false,
+    }
+}
+
+/// Mock execution plan that produces geometry data
+#[allow(dead_code)]
```
paleolimbot: Would something like:

```rust
#[cfg(feature = "gpu"))
mod gpu_functional_test;
```

enable you to remove all these unused/dead code bits?

rust/sedona-spatial-join-gpu/tests/gpu_functional_test.rs
```diff
@@ -0,0 +1,588 @@
...
+
+#[allow(dead_code)]
+impl GeometryDataExec {
+    fn new(ids: Vec<i32>, geometries: Vec<Vec<u8>>) -> Self {
+        let schema = Arc::new(Schema::new(vec![
+            Field::new("id", DataType::Int32, false),
+            Field::new("geometry", DataType::Binary, false),
+        ]));
+
+        let id_array = Int32Array::from(ids);
```
paleolimbot: It feels like this is duplicated from the single batch exec also in this PR (and either way I think this could be removed in favour of `SessionContext::register_batch()`).

rust/sedona-spatial-join-gpu/tests/gpu_functional_test.rs
```diff
@@ -0,0 +1,588 @@
...
+        "GPU spatial join completed successfully with {} result rows",
+        total_rows
+    );
+}
+/// Helper execution plan that returns a single pre-loaded batch
+struct SingleBatchExec {
+    schema: Arc<Schema>,
+    batch: RecordBatch,
+    props: datafusion::physical_plan::PlanProperties,
+}
```
paleolimbot: Duplicate of the other single batch exec?

rust/sedona-spatial-join-gpu/tests/gpu_functional_test.rs
```diff
@@ -0,0 +1,588 @@
...
+    .unwrap();
+
+    let point_batch =
+        RecordBatch::try_new(point_schema.clone(), vec![Arc::new(point_ids), points]).unwrap();
+
+    // Pre-create CPU testers for all predicates (shared across all tests)
+    let kernels = scalar_kernels();
+    let sedona_type = SedonaType::Wkb(Edges::Planar, lnglat());
+
+    let _cpu_testers: std::collections::HashMap<&str, ScalarUdfTester> = [
```
paleolimbot: Is there a reason this variable is not used / can we do this using a `for` loop to avoid this indirection?

rust/sedona-spatial-join/src/exec.rs
```diff
@@ -1151,26 +1156,195 @@ mod tests {
...
             }
             Ok(TreeNodeRecursion::Continue)
         })?;
-        Ok(spatial_join_execs)
+        Ok(count)
+    }
+
+    #[cfg(feature = "gpu")]
+    #[tokio::test]
+    #[ignore] // Requires GPU hardware
```
paleolimbot: We need to figure out a way to not ignore tests in this repo (in this case I think these tests shouldn't exist if the gpu feature isn't enabled so we shouldn't need the ignore it?)

rust/sedona-spatial-join/src/optimizer.rs
```diff
@@ -1054,6 +1080,282 @@ fn is_spatial_predicate_supported(
...
+        match predicate {
+            SpatialPredicate::Relation(rel) => {
+                let gpu_pred = match rel.relation_type {
+                    SpatialRelationType::Intersects => LibGpuPred::Intersects,
+                    SpatialRelationType::Contains => LibGpuPred::Contains,
+                    SpatialRelationType::Covers => LibGpuPred::Covers,
+                    SpatialRelationType::Within => LibGpuPred::Within,
+                    SpatialRelationType::CoveredBy => LibGpuPred::CoveredBy,
+                    SpatialRelationType::Touches => LibGpuPred::Touches,
+                    SpatialRelationType::Equals => LibGpuPred::Equals,
```
paleolimbot: Can we move `SpatialRelationType` to `sedona-geometry` or `sedona-common` to avoid two copies?

c/sedona-s2geography/s2geography
paleolimbot, line 1: `git submodule update --recursive` should remove this diff

python/sedonadb/Cargo.toml
```diff
@@ -29,6 +29,7 @@ crate-type = ["cdylib"]
 default = ["mimalloc"]
 mimalloc = ["dep:mimalloc", "dep:libmimalloc-sys"]
 s2geography = ["sedona/s2geography"]
+gpu = ["sedona/gpu"]
```
paleolimbot, line 32: Because we don't have any tests in Python for this feature I suggest leaving this out for now (a follow-up PR could add Python support + a test)

rust/sedona-spatial-join-gpu/benches/gpu_spatial_join.rs
pwrliang: @zhangfengcdt Hi Feng, do we still need this benchmark program? It compares to a brute-force CPU implementation.

rust/sedona-spatial-join-gpu/benches/gpu_spatial_join.rs
pwrliang: This file has been removed

rust/sedona-spatial-join-gpu/src/config.rs
pwrliang: batch_size does not exist anymore. To fully exploit the parallelism of the GPU, users need to manually set datafusion.execution.batch_size to a very large number. I want to overwrite this value when the GPU feature is enabled, but haven't figured out the right place to insert this logic.

rust/sedona-spatial-join-gpu/src/exec.rs
pwrliang: This file has been completely rewritten.

rust/sedona-spatial-join-gpu/src/gpu_backend.rs
pwrliang: This file has been removed.

rust/sedona-spatial-join-gpu/src/gpu_backend.rs
pwrliang: I will change the C interface to let it accept the BinaryView, so the cast can be prevented. The cast operation is very fast, though.

rust/sedona-spatial-join-gpu/tests/gpu_functional_test.rs
pwrliang: This function is not used and will be removed

rust/sedona-spatial-join-gpu/tests/gpu_functional_test.rs
pwrliang: I found this struct is unused and has been removed

rust/sedona-spatial-join-gpu/tests/gpu_functional_test.rs
pwrliang: The previous one, now, has been removed

rust/sedona-spatial-join-gpu/tests/gpu_functional_test.rs
pwrliang: This test was incomplete, but it has now been completed.

rust/sedona-spatial-join/src/optimizer.rs
pwrliang: I have extracted SpatialRelationType into a separate file under sedona-geometry. I involves some changes in sedona-spatial-join package, I'd like to ask @Kontinuation whether this change is appropriate

pwrliang:
Hi @zhangfengcdt @paleolimbot, the GPU-based spatial join has been completely rewritten.

The initial version wrapped both the filtering and refinement stages into the same interface (C library). However, that approach suffered from build-side parsing overhead because it could not be shared across partitions.

I have redesigned it to match the CPU-based join, where the two stages are separated. This version runs faster and takes less memory. It also supports WHERE clauses (which the initial version did not). The implementation is based on @Kontinuation 's spatial-join, so please give me some advice if you can.

I realize this PR is large, but the structural changes were necessary. Please review it when you have time. Thanks!

Kontinuation:
The modifications made to the existing CPU spatial join code is pretty trivial. If I understand it correctly, it has only moved `pub enum SpatialRelationType` to another module, and generate GPU join physical plans when the GPU feature is enabled. This part looks good to me.

I still need to take some time looking into sedona-spatial-join-gpu. It is enormous amount of code.

python/sedonadb/Cargo.toml
```diff
@@ -54,3 +55,4 @@ thiserror = { workspace = true }
 tokio = { workspace = true }
 mimalloc = { workspace = true, optional = true }
 libmimalloc-sys = { workspace = true, optional = true }
+env_logger = { workspace = true }
```
Kontinuation, line 58: I favored it a lot. I had it when testing my local out-of-core spatial join branch.

rust/sedona-spatial-join-gpu/src/index/spatial_index_builder.rs
```diff
@@ -0,0 +1,207 @@
...
+        let concat_timer = self.metrics.concat_time.timer();
+        let all_record_batches: Vec<&RecordBatch> =
+            indexed_batches.iter().map(|batch| &batch.batch).collect();
+
+        if all_record_batches.is_empty() {
+            return Err(DataFusionError::Internal(
+                "Build side has no batches".into(),
+            ));
+        }
+
```
Kontinuation: Is GPU join capable of handling empty build side gracefully even if `add_partitions` returns an error here?

c/sedona-libgpuspatial/src/lib.rs
```diff
@@ -77,197 +140,411 @@ impl GpuSpatialContext {
...
+            unsafe {
+                index.probe(&mut ctx, rects.as_ptr() as *const f32, rects.len() as u32)?;
+            }
 
-            Ok((build_indices, stream_indices))
+            // Get results
+            let build_indices = index.get_build_indices_buffer(&mut ctx).to_vec();
+            let probe_indices = index.get_probe_indices_buffer(&mut ctx).to_vec();
+            index.destroy_context(&mut ctx);
+            Ok((build_indices, probe_indices))
```
Kontinuation: ctx will leak when `index.probe` fails.

rust/sedona-spatial-join-gpu/src/index/spatial_index_builder.rs
```diff
@@ -0,0 +1,207 @@
...
+        // 2. Extract the schema from the first batch
+        let schema = all_record_batches[0].schema();
+
+        // 3. Pass the slice of references (&[&RecordBatch])
+        self.build_batch.batch = arrow::compute::concat_batches(&schema, all_record_batches)
+            .map_err(|e| {
+                DataFusionError::Execution(format!("Failed to concatenate left batches: {}", e))
+            })?;
```
Kontinuation: I wonder if it is necessary to concatenate all the batches and `gs.push_build` only once. It would be easier for us to unify the CPU and GPU based spatial join if we support `push_build` multiple times.

There might be some performance implications switching to multiple `gc.push_build` calls. If it is not feasible, we need to be aware that concatenating batches doubles the memory requirement, so we need to reserve more memory for building GPU indexes.

Kontinuation (commented):
The GPU spatial join code is mostly the same as the CPU code. The difference is mainly in the spatial index.

Notably, GPU index performs filters and refinements in batches, while the current spatial index in CPU spatial join works with one probe geometry at a time. I have already refactored the CPU spaital index to work with batches for better performance: https://github.com/Kontinuation/sedona-db/blob/485d45fb95fd278b47253bfc2cb1f3fd93798075/rust/sedona-spatial-join/src/index/spatial_index.rs#L437-L460. I believe that it is quite feasible to unify the code for CPU and GPU based spatial join.

I am fine with duplicating code for GPU spatial join, we can do some refactoring to unify them later on.

rust/sedona-spatial-join-gpu/src/index/spatial_index_builder.rs
pwrliang: It is feasible to call push_build many times, but it may slightly increase the build time.  I think it is worth making this change to unify the code.

pwrliang:
Hi @Kontinuation, I have merged the GPU code path into sedona-spatial-join, so there's no duplicated code issue. Could you review the changes to the spatial-join package?

rust/sedona-spatial-join-gpu/src/index/spatial_index_builder.rs
pwrliang: I have added an option that enables/disables the concatenation operations.

c/sedona-libgpuspatial/src/lib.rs
pwrliang: Fixed.

c/sedona-s2geography/s2geography
pwrliang: Fixed

rust/sedona-spatial-join/src/optimizer.rs
pwrliang: We have a unified spatial-join package. This piece of code has been removed

rust/sedona-spatial-join/src/exec.rs
pwrliang: Fixed.

rust/sedona-spatial-join/src/index/spatial_index.rs
```diff
@@ -441,1543 +138,78 @@ impl SpatialIndex {
...
-
-        let indexed_batch = EvaluatedBatch {
-            batch,
-            geom_array: EvaluatedGeometryArray::try_new(geom_batch, &WKB_GEOMETRY).unwrap(),
-        };
-        builder.add_batch(indexed_batch).unwrap();
+pub(crate) trait SpatialIndexFull: SpatialIndex + SpatialIndexInternal {}
 
-        let index = builder.finish().unwrap();
+impl<T> SpatialIndexFull for T where T: SpatialIndex + SpatialIndexInternal {}
```
Kontinuation: I prefer defining the SpatialIndex trait as one rather than splitting it into external and internal interfaces. At least for now SpatialIndex is only intended to be used by spatial join so we don't need to define external interfaces for it.

rust/sedona-spatial-join/src/exec.rs
```diff
@@ -1338,13 +1408,16 @@ mod tests {
...
+        let use_gpu = cfg!(feature = "gpu");
+
         let mark_exec = SpatialJoinExec::try_new(
             original_exec.left.clone(),
             original_exec.right.clone(),
             original_exec.on.clone(),
             original_exec.filter.clone(),
             &join_type,
             None,
+            use_gpu,
```
Kontinuation: Do we still run CPU spatial join tests when GPU feature is enabled?

rust/sedona-spatial-join/src/index/gpu_spatial_index_builder.rs
```diff
@@ -0,0 +1,248 @@
...
+            spatial_predicate,
+            options,
+            join_type,
+            probe_threads_count,
+            metrics,
+            indexed_batches: vec![],
+            reservation,
+        }
+    }
+    /// Build visited bitmaps for tracking left-side indices in outer joins.
```
Kontinuation, line 79: We should add empty lines in between methods. I'm not sure if there's a cargo fmt option for automatically doing this.

Kontinuation (commented):
The overall approach looks good to me.

rust/sedona-spatial-join/src/exec.rs
pwrliang: I initially thought we would run tests twice with and without `--features gpu`. This can eliminate duplicated code. However, I think it is more natural to run both CPU and GPU tests when using `--features gpu`. I have updated the tests accordingly.

rust/sedona-spatial-join/src/index/spatial_index.rs
pwrliang: I have fixed this. Now, it only has a unified interface.

python/sedonadb/Cargo.toml
pwrliang: I have removed this feature.

pwrliang:
Here's the evaluation result from [a *variant* of Sedona-SpatialBench](https://github.com/pwrliang/sedona-bench)

Each query was executed 5 times with one round of warmup to cache datasets in buff/cache, and the average times are reported. Using GPU join is faster for queries with heavy joins, such as Q10 and Q11.

Commit: https://github.com/apache/sedona-db/pull/465/commits/b690b127a626daf51b350deaf99f3234f22851f5
Machine: AWS g5.2xlarge (NVIDIA A10)

| Mode | Scale Factor | Query | Time (s) |
| :--- | :--- | :--- | :--- |
| gpu | sf_1 | q2 | 4.721 |
| gpu | sf_1 | q4 | 1.293 |
| gpu | sf_1 | q8 | 0.413 |
| gpu | sf_1 | q9 | 0.041 |
| gpu | sf_1 | q10 | 4.396 |
| gpu | sf_1 | q11 | 6.773 |
| gpu | sf_10 | q2 | 8.036 |
| gpu | sf_10 | q4 | 2.01 |
| gpu | sf_10 | q8 | 1.78 |
| gpu | sf_10 | q9 | 0.144 |
| gpu | sf_10 | q10 | 9.092 |
| gpu | sf_10 | q11 | 17.378 |

Commit: https://github.com/apache/sedona-db/commit/e106ce65b6ded198298c36d87776dae878c88282
Machine: m7i.2xlarge
| Mode | Scale Factor | Query | Time (s) |
| :--- | :--- | :--- | :--- |
| cpu | sf_1 | q2 | 4.732 |
| cpu | sf_1 | q4 | 0.742 |
| cpu | sf_1 | q8 | 0.489 |
| cpu | sf_1 | q9 | 0.021 |
| cpu | sf_1 | q10 | 5.126 |
| cpu | sf_1 | q11 | 8.625 |
| cpu | sf_10 | q2 | 31.165 |
| cpu | sf_10 | q4 | 1.249 |
| cpu | sf_10 | q8 | 4.53 |
| cpu | sf_10 | q9 | 0.062 |
| cpu | sf_10 | q10 | 28.368 |
| cpu | sf_10 | q11 | 44.737 |

rust/sedona-spatial-join/src/index/gpu_spatial_index_builder.rs
pwrliang: Fixed

jiayuasu:
@pwrliang Great work!

c/sedona-libgpuspatial/libgpuspatial/include/gpuspatial/gpuspatial_c.h
```diff
@@ -20,19 +20,109 @@
...
-struct GpuSpatialJoinerContext {
+struct GpuSpatialIndexContext {
   const char* last_error;  // Pointer to std::string to store last error message
-  void* private_data;      // GPUSpatial context
   void* build_indices;     // Pointer to std::vector<uint32_t> to store results
-  void* stream_indices;
+  void* probe_indices;
+};
+
+struct GpuSpatialIndexFloat2D {
```
paleolimbot: Can we call this `SedonaFloatIndex2D`?

c/sedona-libgpuspatial/libgpuspatial/include/gpuspatial/gpuspatial_c.h
```diff
@@ -20,19 +20,109 @@
...
+  /** Initialize the spatial index with the given configuration
+   *
+   * @return 0 on success, non-zero on failure
+   */
+  int (*init)(struct GpuSpatialIndexFloat2D* self, struct GpuSpatialIndexConfig* config);
```
paleolimbot: If we move the `struct GpuSpatialIndexConfig` parameter to `GpuSpatialIndexFloat2DCreate()` (which I think is possible), this structure will no longer be GPU-specific (which will make it easier to test by implementing on the CPU or implement with Apple Metal later)

c/sedona-libgpuspatial/libgpuspatial/include/gpuspatial/gpuspatial_c.h
```diff
@@ -20,19 +20,109 @@
...
+  /** Pointer to an initialized GpuSpatialRTEngine struct */
+  struct GpuSpatialRTEngine* rt_engine;
+  /** How many threads will concurrently call Probe method */
+  uint32_t concurrency;
+  /** Device ID to use, 0 is the first GPU */
+  int device_id;
 };
 
-struct GpuSpatialJoinerContext {
+struct GpuSpatialIndexContext {
```
paleolimbot: It would be more future-proof to keep this as `last_error` and `private_data`. (Or maybe just `private_data`, with the last error accessible via a callback on the GpuSpatialIndexFloat2D). It seems like the intention is this object is opaque to the user of this API except to pass to methods that support concurrent access.

c/sedona-libgpuspatial/libgpuspatial/include/gpuspatial/gpuspatial_c.h
```diff
@@ -20,19 +20,109 @@
...
+   */
+  int (*probe)(struct GpuSpatialIndexFloat2D* self,
+               struct GpuSpatialIndexContext* context, const float* buf,
+               uint32_t n_rects);
+  /** Get the build indices buffer from the context
+   *
+   * @return A pointer to the buffer and its length
+   */
+  void (*get_build_indices_buffer)(struct GpuSpatialIndexContext* context,
+                                   void** build_indices, uint32_t* build_indices_length);
```
paleolimbot: Can the `build_indices` be any more strongly typed here? (Or can the docstring be more specific about what exactly it contains?)

c/sedona-libgpuspatial/libgpuspatial/include/gpuspatial/gpuspatial_c.h
```diff
@@ -20,19 +20,109 @@
...
+  /** Get the probe indices buffer from the context
+   *
+   * @return A pointer to the buffer and its length
+   */
+  void (*get_probe_indices_buffer)(struct GpuSpatialIndexContext* context,
+                                   void** probe_indices, uint32_t* probe_indices_length);
```
paleolimbot: Can the `probe_indices` be any more strongly typed here? (Or can the docstring be more specific about what is expected?)

c/sedona-libgpuspatial/libgpuspatial/include/gpuspatial/gpuspatial_c.h
```diff
@@ -20,19 +20,109 @@
...
+  /** Release the spatial index and free all resources */
+  void (*release)(struct GpuSpatialIndexFloat2D* self);
+  void* private_data;
+  const char* last_error;
```
paleolimbot: I know it is annoying but it is safer to have this be a callback. If you leave it as a bare public member of a struct you need some very specific documented rules about when it is safe to access.

c/sedona-libgpuspatial/libgpuspatial/include/gpuspatial/gpuspatial_c.h
```diff
@@ -43,31 +133,35 @@ enum GpuSpatialPredicate {
...
-                     int64_t offset, int64_t length, enum GpuSpatialPredicate predicate,
-                     int32_t array_index_offset);
-  void (*get_build_indices_buffer)(struct GpuSpatialJoinerContext* context,
-                                   void** build_indices, uint32_t* build_indices_length);
-  void (*get_stream_indices_buffer)(struct GpuSpatialJoinerContext* context,
-                                    void** stream_indices,
-                                    uint32_t* stream_indices_length);
-  void (*release)(struct GpuSpatialJoiner* self);
+struct GpuSpatialRefiner {
+  int (*init)(struct GpuSpatialRefiner* self, struct GpuSpatialRefinerConfig* config);
```
paleolimbot: As above, can this be a `SedonaSpatialRefiner` with the `GpuSpatialRefinerConfig` moved to `GpuSpatialRefinerCreate()`? We can (and should!) reuse this struct for things that are not just NVIDIA GPUs.

c/sedona-libgpuspatial/libgpuspatial/include/gpuspatial/gpuspatial_c.h
```diff
@@ -43,31 +133,35 @@ enum GpuSpatialPredicate {
...
+
+  int (*clear)(struct GpuSpatialRefiner* self);
+
+  int (*push_build)(struct GpuSpatialRefiner* self,
+                    const struct ArrowSchema* build_schema,
```
paleolimbot, line 144: Is it possible to pass the `build_schema` and the `probe_schema` separately (maybe as arguments to `init`?) and remove them from the other methods? (Totally fine if this is not possible, just might help reduce the number of arguments to the other callbacks)

c/sedona-libgpuspatial/libgpuspatial/include/gpuspatial/gpuspatial_c.h
```diff
@@ -43,31 +133,35 @@ enum GpuSpatialPredicate {
...
+                       enum GpuSpatialRelationPredicate predicate,
+                       uint32_t* build_indices, uint32_t* probe_indices,
+                       uint32_t indices_size, uint32_t* new_indices_size);
+
+  int (*refine)(struct GpuSpatialRefiner* self, const struct ArrowSchema* schema1,
+                const struct ArrowArray* array1, const struct ArrowSchema* schema2,
+                const struct ArrowArray* array2,
+                enum GpuSpatialRelationPredicate predicate, uint32_t* indices1,
+                uint32_t* indices2, uint32_t indices_size, uint32_t* new_indices_size);
+  void (*release)(struct GpuSpatialRefiner* self);
```
paleolimbot: This `last_error` should probably be a callback (as above)

c/sedona-libgpuspatial/libgpuspatial/src/gpuspatial_c.cc
paleolimbot, line 1: I see some `.cc` files and some `.cpp` files...if it's easy to rename these I usually see projects choose one and rename files to match (except for vendored files).

c/sedona-libgpuspatial/libgpuspatial/src/rt/rt_engine.cpp
```diff
@@ -14,7 +14,7 @@
 // KIND, either express or implied.  See the License for the
 // specific language governing permissions and limitations
 // under the License.
-#include "gpuspatial/index/detail/rt_engine.hpp"
+#include "gpuspatial/rt/rt_engine.hpp"
```
paleolimbot, line 17: Similarly, if `.h` and `.hpp` are both used it usually means that `.h` files are intended for C usage. If it's easy to rename the `.h` files intended for C++ usage it may look slightly cleaner.

paleolimbot (commented):
Thank you for working on this! This is a substantial improvement on the initial work (which was also a substantial improvement!). In particular I am grateful for your work to ensure that the integration with the existing sedona-spatial-join code.

I think this would work best as at least two PRs:

- Update the C++ library portion with the new approach (I left review comments on the C++ portion), since this is nicely contained
- Add a Rust wrapper around `gpuspatial_c.h` with a few "is it plugged in" tests
- Integrate with sedona-spatial-join

In addition to being easier for both the community and LLMs to review, @Kontinuation has rather heroically split a very large change implementing larger-than-memory spatial joins into small PRs and splitting out the C library change and the wrapper first may give him a few extra days to merge or align conflicting pieces.

This C API is great and is much better than what we initially came up with when we started this project! Now that it has been more thought out, I left some comments to perhaps help polish it a bit. It would help to ensure every member has docstrings now that you've nicely showed that it works!

I'm not a CUDA expert and I'll have to rely on copilot to review that bit when it's in a dedicated PR. It seems like you have great test coverage and were kind to ensure the licensing and TODOs were all handled.

pwrliang:
We may close this PR now because of https://github.com/apache/sedona-db/pull/722

PR #467, opened 2025-12-21
fix(rust/sedona-functions): Let `ST_Collect_Agg()` on a Geography return a Geography

### LINKED ISSUE #280: ST_Collect() on a Geography returns a Geometry
```python
sd.sql("SELECT ST_Collect(ST_GeogPoint(0, 1))").schema
#> SedonaSchema with 1 field:
#>   st_collect(st_geogpoint(Int64(0),Int64(1))): geometry<Wkb>
```

yutannihilation:
Fix #280

I guess there are more functions that need similar fix, but this only addresses `ST_Collect_Agg()`, which looks straightforward.

```
> SELECT ST_Collect_Agg(ST_GeogPoint(0, 1));
┌─────────────────────────────────────────────────┐
│ st_collect_agg(st_geogpoint(Int64(0),Int64(1))) │
│                    geography                    │
╞═════════════════════════════════════════════════╡
│ MULTIPOINT((0 1))                               │
└─────────────────────────────────────────────────┘
```

### DIFF
--- a/rust/sedona-functions/src/st_collect_agg.rs
+++ b/rust/sedona-functions/src/st_collect_agg.rs
@@ -48,7 +48,12 @@ use sedona_schema::{
 pub fn st_collect_agg_udf() -> SedonaAggregateUDF {
     SedonaAggregateUDF::new(
         "st_collect_agg",
-        vec![Arc::new(STCollectAggr {})],
+        vec![
+            Arc::new(STCollectAggr {
+                is_geography: false,
+            }),
+            Arc::new(STCollectAggr { is_geography: true }),
+        ],
         Volatility::Immutable,
         Some(st_collect_agg_doc()),
     )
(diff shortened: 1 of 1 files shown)


rust/sedona-functions/src/st_collect_agg.rs
```diff
@@ -70,8 +70,14 @@ struct STCollectAggr {}
...
-        let matcher = ArgMatcher::new(vec![ArgMatcher::is_geometry_or_geography()], WKB_GEOMETRY);
-        matcher.match_args(args)
+        let geom_matcher = ArgMatcher::new(vec![ArgMatcher::is_geometry()], WKB_GEOMETRY);
+        let geog_matcher = ArgMatcher::new(vec![ArgMatcher::is_geography()], WKB_GEOGRAPHY);
+        for matcher in [geom_matcher, geog_matcher] {
+            if let result @ Ok(Some(_)) = matcher.match_args(args) {
+                return result;
+            }
+        }
+        Ok(None)
```
paleolimbot: This is very similar to what the `SedonaAggregateUDF` does when resolving the `SedonaAccumulator` based on the input arguments.

yutannihilation:
Thanks for the hint, I'll implement so! (Probably later this week or next year)

PR #468, opened 2025-12-28
feat(r/sedonadb): Translate R expressions to DataFusion logical Expr

paleolimbot:
This PR adds constructors for DataFusion/SedonaDB (logical) `Expr` and a framework for constructing the using an R expression. I also added a few function translations to jog my memory on the corner cases with respect to constructing them. This is to support #474 (i.e., a stepping stone to help implement dplyr verbs).

``` r
library(sedonadb)

schema <- nanoarrow::na_struct(list(x = nanoarrow::na_int32()))
ctx <- sedonadb:::sd_expr_ctx(schema)
sedonadb:::sd_eval_expr(quote(sum(x + 5)), expr_ctx = ctx)
#> <SedonaDBExpr>
#> sum(x + Float64(5)) RESPECT NULLS
```

This implementation comes from the work I did in arrow ( https://github.com/apache/arrow/blob/main/r/R/dplyr-funcs.R ) and substrait-r ( https://github.com/voltrondata/substrait-r/blob/e3045d21b1765c15d2b774409f4ca2b1ebb01796/R/compiler.R#L448-L466 ) and the suggestions in #474 with respect to how this is organized in duckdb-r/duckplyr.

In arrow and substrait-r we used a slightly more straightforward approach than this PR, which was basically to load an environment with "override" functions and field references. When those packages evaluate an expression, we arrange the environments such that column reference expressions and our "translated" functions (that accept expressions and return expressions) are resolved first. This mostly works but:

- It's tricky to determine the output type of an expression. For example, `/`, which in R is never floor division, we need to check if both arguments are integers (and if so, cast one of them to a float).
- It's tricky to scope translations. For example, we mostly want things scoped to a SessionContext and not globally (although currently we only have a global session context in R).
- Aggregations and window functions need some extra context that isn't available when translations are global and stateless.

For this reason, I introduced a `sd_expr_ctx()` (expression context) here that is faithfully passed down when evaluating translated expressions. This gives each translation access to things like the input schema (so that the output type of the expression can be computed), session options when we have them, and information like grouping expressions for aggregate functions.

This means that evaluation is a bit more "manual", but I think the extra control is worth it (although we can revisit as we implement more translations and I remember why various things are hard). We could accomplish some of this with a global variable (like the global `substrait_compiler()` in substrait-r) and less complicated evaluation as well.

There is a lot to do after this PR, like add support for more functions, more types of literals, and wire up actual dplyr verbs like `filter()`.

### DIFF
--- a/r/sedonadb/R/expression.R
+++ b/r/sedonadb/R/expression.R
@@ -67,9 +67,7 @@ sd_expr_negative <- function(expr, factory = sd_expr_factory()) {
 #' @export
 sd_expr_scalar_function <- function(function_name, args, factory = sd_expr_factory()) {
   args_as_expr <- lapply(args, as_sd_expr, factory = factory)
-  # Not sure why we need this exactly (something about savvy)
-  args_as_expr_ptr <- lapply(args_as_expr, "[[", ".ptr")
-  factory$scalar_function(function_name, args_as_expr_ptr)
+  factory$scalar_function(function_name, args_as_expr)
 }
 
 #' @rdname sd_expr_column
@@ -77,9 +75,7 @@ sd_expr_scalar_function <- function(function_name, args, factory = sd_expr_facto
 sd_expr_aggregate_function <- function(function_name, args, ...,
                                        na.rm = FALSE, distinct = FALSE, factory = sd_expr_factory()) {
   args_as_expr <- lapply(args, as_sd_expr, factory = factory)
-  # Not sure why we need this exactly (something about savvy)
-  args_as_expr_ptr <- lapply(args_as_expr, "[[", ".ptr")
-  factory$aggregate_function(function_name, args_as_expr_ptr, na_rm = na.rm, distinct = distinct)
+  factory$aggregate_function(function_name, args_as_expr, na_rm = na.rm, distinct = distinct)
 }
 
 #' @rdname sd_expr_column
--- a/r/sedonadb/src/rust/src/expression.rs
+++ b/r/sedonadb/src/rust/src/expression.rs
@@ -22,7 +22,7 @@ use datafusion_expr::{
     expr::{AggregateFunction, FieldMetadata, NullTreatment, ScalarFunction},
     BinaryExpr, Cast, Expr, Operator,
 };
-use savvy::{savvy, savvy_err};
+use savvy::{savvy, savvy_err, EnvironmentSexp};
 use sedona::context::SedonaContext;
 
 use crate::{
@@ -179,11 +179,21 @@ impl SedonaDBExprFactory {
         savvy::ListSexp::try_from(exprs_sexp)?
             .iter()
             .map(|(_, item)| -> savvy::Result<Expr> {
-                // This seems to require $.ptr from the list() input (can't just
-                // use list of R SedonaDBExpr objects)
-                let expr_wrapper: &SedonaDBExpr = item.try_into()?;
+                // item here is the Environment wrapper around the external pointer
+                let expr_wrapper: &SedonaDBExpr = EnvironmentSexp::try_from(item)?.try_into()?;
                 Ok(expr_wrapper.inner.clone())
             })
             .collect()
+    }
+}
+
+impl TryFrom<EnvironmentSexp> for &SedonaDBExpr {
+    type Error = savvy::Error;
+
+    fn try_from(env: EnvironmentSexp) -> Result<Self, Self::Error> {
+        env.get(".ptr")?
+            .map(<&SedonaDBExpr>::try_from)
+            .transpose()?
+            .ok_or(savvy_err!("Invalid SedonaDBExpr object."))
     }
 }


r/sedonadb/tools/savvy-update.sh
```diff
@@ -0,0 +1,94 @@
...
+
+    # Add license header to api.h
+    echo "${LICENSE_C}" > "${api_h}"
+    cat "${api_h}.tmp" >> "${api_h}"
+
+    # Add license header, put includes on their own lines, and fix a typo in init.c
+    echo "${LICENSE_C}" > "${init_c}"
+    sed 's/#include/\n#include/g' "${init_c}.tmp" | \
+        sed '1s/^\n//' | \
+        sed 's/initialzation/initialization/g' >> "${init_c}"
```
paleolimbot, line 45: @yutannihilation I will open an issue and/or try to fix this in savvy when I have a moment, but while I'm thinking about it:

- Includes have to go on their own lines, or else clang-format will reorder them in a way such that they won't compile
- There's a typo in one of the comments that causes our pre-commit to fail ("initialzation")

...the other stuff (license files, clang-format) is specific to our setup and I'm not sure that's savvy's issue.

r/sedonadb/R/expression.R
```diff
@@ -0,0 +1,287 @@
...
+#' @examples
+#' sd_expr_column("foofy")
+#' sd_expr_literal(1L)
+#' sd_expr_scalar_function("abs", list(1L))
+#' sd_expr_cast(1L, nanoarrow::na_int64())
+#' sd_expr_alias(1L, "foofy")
+#'
+sd_expr_column <- function(column_name, qualifier = NULL, factory = sd_expr_factory()) {
+  factory$column(column_name, qualifier)
+}
```
paleolimbot, line 46: @e-kotov I tried to mirror the DuckDB expression constructors you linked! The names are slightly different because I used DataFusion naming conventions.

r/sedonadb/src/rust/src/expression.rs
```diff
@@ -0,0 +1,190 @@
...
+
+impl SedonaDBExprFactory {
+    fn exprs(exprs_sexp: savvy::Sexp) -> savvy::Result<Vec<Expr>> {
+        savvy::ListSexp::try_from(exprs_sexp)?
+            .iter()
+            .map(|(_, item)| -> savvy::Result<Expr> {
+                // This seems to require $.ptr from the list() input (can't just
+                // use list of R SedonaDBExpr objects)
+                let expr_wrapper: &SedonaDBExpr = item.try_into()?;
+                Ok(expr_wrapper.inner.clone())
```
paleolimbot: @yutannihilation Is there a trick here that I should be using to get a list of a `#[savvy] struct XXX` back into Rust land?

r/sedonadb/tools/savvy-update.sh
yutannihilation: Thanks for catching!

yutannihilation: Regarding license files, I found `bindgen` has option to insert arbitrary strings at the top of the generated file. Will it be useful if I implement this in savvy-cli?

```console
bindgen input.h -o bindings.rs \
    --raw-line "/*" \
    --raw-line " * Copyright (c) 2026 Your Name" \
    --raw-line " * Licensed under Apache-2.0" \
    --raw-line " */"
```

r/sedonadb/src/rust/src/expression.rs
yutannihilation: I tried to remember if there's anything, but I think there's no trick...

Regarding `$.ptr`, it's posible to return a bare EXTPTRSXP without a wrapper, but it requres you to add several wrapper R functions (e.g. add S3 class, and print it nicely). So, I'm not sure which is easier.

cf. https://yutannihilation.github.io/savvy/guide/struct.html#external-pointer

r/sedonadb/R/expression.R
e-kotov: @paleolimbot note taken. Sorry, I will get back to https://github.com/apache/sedona-db/pull/475 in a few days.

r/sedonadb/src/rust/src/expression.rs
eitsupi: I think r-polars does that sort of thing.

https://github.com/pola-rs/r-polars/blob/a6c028b6341f6910616a1cffa246a1b218b0ba32/src/rust/src/expr/mod.rs#L32-L41
https://github.com/pola-rs/r-polars/blob/a6c028b6341f6910616a1cffa246a1b218b0ba32/src/rust/src/conversion/mod.rs#L214-L234
https://github.com/pola-rs/r-polars/blob/a6c028b6341f6910616a1cffa246a1b218b0ba32/src/rust/src/lazyframe/general.rs#L91-L95

r/sedonadb/R/expression.R
paleolimbot: No worries!

r/sedonadb/tools/savvy-update.sh
paleolimbot: I think we'll probably need a shell helper script even if you do, so I wouldn't worry about it. Apache projects are just full of weird constraints.

r/sedonadb/src/rust/src/expression.rs
paleolimbot: Thanks! I'll add that helper here, too.

yutannihilation: Thanks for sharing, @eitsupi!

PR #469, opened 2025-12-28
feat(rust/sedona-geo): Add ST_AsGeoJSON

prantogg:
Implements the `ST_AsGeoJSON` function for converting geometry objects to GeoJSON format.

### DIFF
--- a/python/sedonadb/tests/functions/test_functions.py
+++ b/python/sedonadb/tests/functions/test_functions.py
@@ -123,6 +123,9 @@ def test_st_astext(eng, geom):
 @pytest.mark.parametrize(
     ("geom", "expected"),
     [
+        # Note: Using coordinates with decimal values instead of integers
+        # because PostGIS returns integer coordinates in GeoJSON when the geometry has integer
+        # coordinates, while SedonaDB always returns floats. See issue #472.
         (None, None),
         ("POINT EMPTY", '{"type":"Point","coordinates":[]}'),
         ("LINESTRING EMPTY", '{"type":"LineString","coordinates":[]}'),
@@ -131,30 +134,30 @@ def test_st_astext(eng, geom):
         ("MULTILINESTRING EMPTY", '{"type":"MultiLineString","coordinates":[]}'),
         ("MULTIPOLYGON EMPTY", '{"type":"MultiPolygon","coordinates":[]}'),
         ("GEOMETRYCOLLECTION EMPTY", '{"type":"GeometryCollection","geometries":[]}'),
-        ("POINT (1 2)", '{"type":"Point","coordinates":[1.0,2.0]}'),
+        ("POINT (1.5 2.5)", '{"type":"Point","coordinates":[1.5,2.5]}'),
         (
-            "LINESTRING (0 0, 1 1)",
-            '{"type":"LineString","coordinates":[[0.0,0.0],[1.0,1.0]]}',
+            "LINESTRING (0.5 0.5, 1.5 1.5)",
+            '{"type":"LineString","coordinates":[[0.5,0.5],[1.5,1.5]]}',
         ),
         (
-            "POLYGON ((0 0, 1 0, 1 1, 0 1, 0 0))",
-            '{"type":"Polygon","coordinates":[[[0.0,0.0],[1.0,0.0],[1.0,1.0],[0.0,1.0],[0.0,0.0]]]}',
+            "POLYGON ((0.5 0.5, 1.5 0.5, 1.5 1.5, 0.5 1.5, 0.5 0.5))",
+            '{"type":"Polygon","coordinates":[[[0.5,0.5],[1.5,0.5],[1.5,1.5],[0.5,1.5],[0.5,0.5]]]}',
         ),
         (
-            "MULTIPOINT ((0 0), (1 1))",
-            '{"type":"MultiPoint","coordinates":[[0.0,0.0],[1.0,1.0]]}',
+            "MULTIPOINT ((0.5 0.5), (1.5 1.5))",
+            '{"type":"MultiPoint","coordinates":[[0.5,0.5],[1.5,1.5]]}',
         ),
         (
-            "MULTILINESTRING ((0 0, 1 1), (2 2, 3 3))",
-            '{"type":"MultiLineString","coordinates":[[[0.0,0.0],[1.0,1.0]],[[2.0,2.0],[3.0,3.0]]]}',
+            "MULTILINESTRING ((0.5 0.5, 1.5 1.5), (2.5 2.5, 3.5 3.5))",
+            '{"type":"MultiLineString","coordinates":[[[0.5,0.5],[1.5,1.5]],[[2.5,2.5],[3.5,3.5]]]}',
         ),
         (
-            "MULTIPOLYGON (((0 0, 1 0, 1 1, 0 1, 0 0)), ((2 2, 3 2, 3 3, 2 3, 2 2)))",
-            '{"type":"MultiPolygon","coordinates":[[[[0.0,0.0],[1.0,0.0],[1.0,1.0],[0.0,1.0],[0.0,0.0]]],[[[2.0,2.0],[3.0,2.0],[3.0,3.0],[2.0,3.0],[2.0,2.0]]]]}',
+            "MULTIPOLYGON (((0.5 0.5, 1.5 0.5, 1.5 1.5, 0.5 1.5, 0.5 0.5)), ((2.5 2.5, 3.5 2.5, 3.5 3.5, 2.5 3.5, 2.5 2.5)))",
+            '{"type":"MultiPolygon","coordinates":[[[[0.5,0.5],[1.5,0.5],[1.5,1.5],[0.5,1.5],[0.5,0.5]]],[[[2.5,2.5],[3.5,2.5],[3.5,3.5],[2.5,3.5],[2.5,2.5]]]]}',
         ),
         (
-            "GEOMETRYCOLLECTION (POINT (0 0), LINESTRING (1 1, 2 2))",
-            '{"type":"GeometryCollection","geometries":[{"type":"Point","coordinates":[0.0,0.0]},{"type":"LineString","coordinates":[[1.0,1.0],[2.0,2.0]]}]}',
+            "GEOMETRYCOLLECTION (POINT (0.5 0.5), LINESTRING (1.5 1.5, 2.5 2.5))",
+            '{"type":"GeometryCollection","geometries":[{"type":"Point","coordinates":[0.5,0.5]},{"type":"LineString","coordinates":[[1.5,1.5],[2.5,2.5]]}]}',
         ),
     ],
 )


rust/sedona-functions/src/st_asgeojson.rs
```diff
@@ -0,0 +1,456 @@
...
+        convert_to_geojson(&arg_types[..1], &args[..1], geojson_type)
+    }
+}
+
+fn convert_to_geojson(
+    arg_types: &[SedonaType],
+    args: &[ColumnarValue],
+    geojson_type: GeoJsonType,
+) -> Result<ColumnarValue> {
+    let executor = WkbExecutor::new(arg_types, args);
```
paleolimbot: A good reason to put this in the `sedona-geo` crate is that there we have the `GeoTypesExecutor` (which will let you remove the wkb -> geo_types conversion)

rust/sedona-functions/src/st_asgeojson.rs
```diff
@@ -0,0 +1,456 @@
...
+        // Test with a simple point
+        let result = tester.invoke_wkb_scalar(Some("POINT (1 2)")).unwrap();
+        if let ScalarValue::Utf8(Some(json_str)) = result {
+            // Verify it's valid JSON and contains expected structure
+            let parsed: serde_json::Value = serde_json::from_str(json_str.as_str()).unwrap();
+            assert_eq!(parsed["type"], "Point");
+            assert!(parsed["coordinates"].is_array());
+        } else {
+            panic!("Expected Utf8 scalar value");
+        }
```
paleolimbot: Our tester helper is probably the best way to test these results:

```suggestion
        tester.assert_scalar_result_equal(result, #"{"key": "value"}"#);
```

rust/sedona-functions/src/st_asgeojson.rs
```diff
@@ -0,0 +1,456 @@
...
+            &ScalarValue::Utf8(None),
+        );
+
+        // Test with array
+        let result_array = tester
+            .invoke_wkb_array(vec![Some("POINT(1 2)"), None, Some("POINT(3 5)")])
+            .unwrap();
+
+        // Verify the array has the expected number of elements
+        assert_eq!((*result_array).len(), 3);
```
paleolimbot: We probably do want to create an explicit array here and check for equality. I believe we do this for other functions that return a string (like `ST_GeometryType()`).

rust/sedona-functions/src/st_asgeojson.rs
```diff
@@ -0,0 +1,456 @@
...
+    .build()
+}
+
+#[derive(Debug)]
+struct STAsGeoJSON {}
+
+impl SedonaScalarKernel for STAsGeoJSON {
+    fn return_type(&self, args: &[SedonaType]) -> Result<Option<SedonaType>> {
+        let matcher = ArgMatcher::new(
+            vec![ArgMatcher::is_geometry_or_geography()],
```
paleolimbot:
```suggestion
            vec![ArgMatcher::is_geometry()],
```

(GeoJSON has explicit planar edges, unlike geographies)

rust/sedona-functions/src/st_asgeojson.rs
```diff
@@ -0,0 +1,456 @@
...
+}
+
+#[derive(Debug)]
+struct STAsGeoJSONWithType {}
+
+impl SedonaScalarKernel for STAsGeoJSONWithType {
+    fn return_type(&self, args: &[SedonaType]) -> Result<Option<SedonaType>> {
+        let matcher = ArgMatcher::new(
+            vec![
+                ArgMatcher::is_geometry_or_geography(),
```
paleolimbot:
```suggestion
                ArgMatcher::is_geometry(),
```

rust/sedona-functions/src/st_asgeojson.rs
```diff
@@ -0,0 +1,456 @@
...
+            ColumnarValue::Scalar(datafusion_common::ScalarValue::Utf8(Some(type_str))) => {
+                GeoJsonType::from_str(type_str.as_str())?
+            }
+            ColumnarValue::Scalar(datafusion_common::ScalarValue::Utf8(None)) => {
+                GeoJsonType::Simple // Default to Simple if NULL
+            }
+            _ => {
+                return Err(DataFusionError::Execution(
+                    "Second argument to ST_AsGeoJSON must be a string literal".to_string(),
+                ));
```
paleolimbot:
```suggestion
                return exec_err!("Second argument to ST_AsGeoJSON must be a string literal");
```

rust/sedona-functions/src/st_asgeojson.rs
```diff
@@ -0,0 +1,456 @@
...
+
+                match geo_geometry {
+                    Some(geom) => {
+                        // Convert geo_types::Geometry to geojson::Geometry
+                        let geojson_geom: geojson::Geometry = (&geom).try_into().map_err(|e| {
+                            DataFusionError::Execution(format!(
+                                "Failed to convert to GeoJSON: {:?}",
+                                e
+                            ))
+                        })?;
```
paleolimbot: We can do this in a follow-up PR and not this one; however, this chain of WKB bytes -> Wkb object -> GeoTypes geometry -> GeoJSON geometry is probably very slow. If we can use:

https://github.com/geoarrow/geoarrow-rs/blob/e17e9e48a6e1b357d5c71839dd33c7cb72d853c0/rust/geoarrow-geojson/src/encoder/geometry.rs#L31-L48

(which should operate directly on a `Wkb` object), then I think we might be able to make this quite a bit faster (and support Z values).

rust/sedona-functions/src/st_asgeojson.rs
```diff
@@ -0,0 +1,456 @@
...
+
+    let geometries = geometries?;
+
+    Ok(Some(Geometry::GeometryCollection(
+        geo_types::GeometryCollection(geometries),
+    )))
+}
+
+#[cfg(test)]
+mod tests {
```
petern48, line 50: In addition to rust tests, we need integration tests in [test_functions.py](https://github.com/apache/sedona-db/blob/e0e1d109480727faaf7be25923b57b4686144438/python/sedonadb/tests/functions/test_functions.py#L4). Instructions for testing it locally and iterating are [here](https://github.com/apache/sedona-db/blob/e0e1d109480727faaf7be25923b57b4686144438/docs/contributors-guide.md#python).

rust/sedona-functions/src/st_asgeojson.rs
```diff
@@ -0,0 +1,456 @@
...
+        "ST_AsGeoJSON (A: Geometry [, type: String])",
+    )
+    .with_argument("geom", "geometry: Input geometry or geography")
+    .with_argument(
+        "type",
+        "string (optional): Output type - 'Simple' (default), 'Feature', or 'FeatureCollection'",
+    )
+    .with_sql_example("SELECT ST_AsGeoJSON(ST_Point(1.0, 2.0))")
+    .with_sql_example("SELECT ST_AsGeoJSON(ST_Point(1.0, 2.0), 'Feature')")
+    .with_sql_example("SELECT ST_AsGeoJSON(ST_Point(1.0, 2.0), 'FeatureCollection')")
```
petern48: I'm not sure if you already discussed this or not, but I'll point out that this follows [the Sedona API](https://sedona.apache.org/latest/api/sql/Function/#st_asgeojson). The [PostGIS one](https://postgis.net/docs/ST_AsGeoJSON.html) is completely different.

```
text ST_AsGeoJSON(record feature, text geom_column="", integer maxdecimaldigits=9, boolean pretty_bool=false, text id_column='');

text ST_AsGeoJSON(geometry geom, integer maxdecimaldigits=9, integer options=8);

text ST_AsGeoJSON(geography geog, integer maxdecimaldigits=9, integer options=0);
```

We have generally followed the PostGIS behavior for all previous functions that I know of.

rust/sedona-functions/src/st_asgeojson.rs
petern48: +1 to moving to this from `rust/sedona-functions/` to `rust/sedona-geo`. Note you need to do this *in order to* access these helpful methods/structs.

`GeoTypesExecutor` uses a [item_to_geometry()](https://github.com/apache/sedona-db/blob/e0e1d109480727faaf7be25923b57b4686144438/rust/sedona-geo/src/to_geo.rs#L58) function that implements exactly what you did in `wkb_to_geometry()` below (hence we can remove it). If you find `GeoTypesExecutor` isn't flexible enough for this function ([like I did](https://github.com/apache/sedona-db/pull/233#discussion_r2443552552) for a different function), you could consider using `item_to_geometry()` directly instead.

rust/sedona-functions/src/st_asgeojson.rs
petern48: Interesting, there's also a note on the Sedona page that says

"**This method is not recommended**. Please use Sedona GeoJSON data source to write GeoJSON files." (bold is mine)

... haha. Maybe it's because the api is different 🤷

rust/sedona-functions/src/st_asgeojson.rs
prantogg: Thanks! `GeoTypesExecutor` is exactly what i needed!

rust/sedona-functions/src/st_asgeojson.rs
```diff
@@ -0,0 +1,67 @@
...
+        "ST_AsGeoJSON (A: Geometry [, type: String])",
+    )
+    .with_argument("geom", "geometry: Input geometry")
+    .with_argument(
+        "type",
+        "string (optional): Output type - 'Simple' (default), 'Feature', or 'FeatureCollection'",
+    )
+    .with_sql_example("SELECT ST_AsGeoJSON(ST_Point(1.0, 2.0))")
+    .with_sql_example("SELECT ST_AsGeoJSON(ST_Point(1.0, 2.0), 'Feature')")
+    .with_sql_example("SELECT ST_AsGeoJSON(ST_Point(1.0, 2.0), 'FeatureCollection')")
```
petern48:
```suggestion
```

prantogg:
Thanks for the feedback! I've moved the code to `rust/sedona-geo` and am now using `GeoTypesExecutor` as suggested. I've also removed the output type parameter for now - we can address that in a later PR.

However, I'm running into some [CI python test failures](https://github.com/apache/sedona-db/actions/runs/20584413572/job/59118209352?pr=469#step:14:3679) when comparing GeoJSON output between SedonaDB and PostGIS.
The differences arise from both the geojson crate's serialization and how it converts from `geo_types::Geometry`. Specifically:
- Coordinate formatting (integers vs floats): The `geojson` crate always serializes coordinates as floats, even for whole numbers. PostGIS optimizes by omitting .0 for whole numbers.
- Empty polygon representation: The geojson crate's conversion from `geo_types::Polygon` to `geojson::Geometry` represents `POLYGON EMPTY` as `{"type":"Polygon","coordinates":[[]]}` (one empty ring), while PostGIS returns `{"type":"Polygon","coordinates":[]}` (no rings).

What's your preference for handling these format differences? Should we:
- Normalize these in the tests (converting PostGIS output to match SedonaDB's format)?
- Look into customizing the GeoJSON serialization to match PostGIS exactly?

rust/sedona-functions/src/st_asgeojson.rs
prantogg: Thanks! The current implementation still does the full chain (via `GeoTypesExecutor` -> `geo_types::Geometry` -> `geojson::Geometry` -> JSON string). 
I'm happy to address the optimization using geoarrow-geojson's encoder in a follow-up PR if that's fine by you?

rust/sedona-functions/src/st_asgeojson.rs
```diff
@@ -0,0 +1,67 @@
...
+fn st_asgeojson_doc() -> Documentation {
+    Documentation::builder(
+        DOC_SECTION_OTHER,
+        "Return the GeoJSON representation of a geometry",
+        "ST_AsGeoJSON (A: Geometry [, type: String])",
+    )
+    .with_argument("geom", "geometry: Input geometry")
+    .with_argument(
+        "type",
+        "string (optional): Output type - 'Simple' (default), 'Feature', or 'FeatureCollection'",
```
petern48:
```suggestion
```
(based on you removing it from the actual kernel implementation)

rust/sedona-functions/src/st_asgeojson.rs
petern48:
```suggestion
```

rust/sedona-geo/src/st_asgeojson.rs
```diff
@@ -0,0 +1,149 @@
...
+    fn invoke_batch(
+        &self,
+        arg_types: &[SedonaType],
+        args: &[ColumnarValue],
+    ) -> Result<ColumnarValue> {
+        let executor = GeoTypesExecutor::new(arg_types, args);
+
+        // Estimate the minimum probable memory requirement of the output.
+        // GeoJSON is typically longer than WKT due to JSON formatting.
+        let min_probable_geojson_size = executor.num_iterations() * 50;
```
petern48:
```suggestion
        // Minimal GeoJSON: {"type":"Point","coordinates":[]}
        let min_probable_geojson_size = executor.num_iterations() * 33;
```
how about this?

petern48 (commented):
> The geojson crate's conversion from geo_types::Polygon to geojson::Geometry represents POLYGON EMPTY as {"type":"Polygon","coordinates":[[]]} (one empty ring), while PostGIS returns {"type":"Polygon","coordinates":[]} (no rings).

I do think this is technically a bug upstream (see the toggle for why), and there does seem to be a maintainer responding to new isssues/PRs in the [geojson rust repo](https://github.com/georust/geojson). I'd say if you're eager to fix it, then go for it, but I think it's fine to kick this down the road since this is such a specific edge case. At least it's valid traditional JSON. We can just file an issue and comment the test out.

<details>
<summary>Proof from the [GeoJSON spec](https://datatracker.ietf.org/doc/html/rfc7946) (I asked an LLM and verified it 🤠)</summary>

`A linear ring is a closed LineString with four or more positions.`

`For type "Polygon", the "coordinates" member MUST be an array of linear ring coordinate arrays.`

</details>

> Coordinate formatting (integers vs floats): The geojson crate always serializes coordinates as floats, even for whole numbers. PostGIS optimizes by omitting .0 for whole numbers.

I think fixing this would be convenient to do when we decide to add support for the `integer maxdecimaldigits` argument for `ST_AsGeoJSON`, which would likely involve modifying similar code upstream. Speaking of, it seems like there's an old PR that might be fixing this for us: https://github.com/georust/geojson/pull/234, though I didn't look too deeply into it.

For both cases, I'm in favor of tweaking the tests a bit to pass and keeping existing behavior for convenience. We'll see what Dewey thinks.

rust/sedona-functions/src/st_asgeojson.rs
petern48: filed a ticket so we don't forget: https://github.com/apache/sedona-db/issues/472

rust/sedona-geo/src/st_asgeojson.rs
prantogg: Sure sounds good!

python/sedonadb/tests/functions/test_functions.py
```diff
@@ -119,6 +119,49 @@ def test_st_astext(eng, geom):
...
+        ("POLYGON EMPTY", '{"type":"Polygon","coordinates":[[]]}'),
+        ("MULTIPOINT EMPTY", '{"type":"MultiPoint","coordinates":[]}'),
+        ("MULTILINESTRING EMPTY", '{"type":"MultiLineString","coordinates":[]}'),
+        ("MULTIPOLYGON EMPTY", '{"type":"MultiPolygon","coordinates":[]}'),
+        ("GEOMETRYCOLLECTION EMPTY", '{"type":"GeometryCollection","geometries":[]}'),
+        ("POINT (1 2)", '{"type":"Point","coordinates":[1.0,2.0]}'),
+        (
+            "LINESTRING (0 0, 1 1)",
+            '{"type":"LineString","coordinates":[[0.0,0.0],[1.0,1.0]]}',
+        ),
```
paleolimbot: For the failing tests, how about something like:

```python
if eng is PostGIS:
   expected = expected.replace(".0", "")
```

Slightly less hacky but might not work (not sure if Python dictionary equality works exactly in the way we need it to for this):

```python
result = eng.execute_and_collect()
df = eng.result_to_pandas(result)
assert df.shape == (1, 1)
assert json.loads(df.iloc[0, 0]) == json.loads(expected)
```

Either works for me! I don't particularly mind the difference (in the process of making it faster we might be able to fix the incompatibility as well).

python/sedonadb/tests/functions/test_functions.py
```diff
@@ -119,6 +119,49 @@ def test_st_astext(eng, geom):
     eng.assert_query_result(f"SELECT ST_AsText({geom_or_null(geom)})", expected)
 
 
+@pytest.mark.parametrize("eng", [SedonaDB, PostGIS])
+@pytest.mark.parametrize(
+    ("geom", "expected"),
+    [
+        (None, None),
+        ("LINESTRING EMPTY", '{"type":"LineString","coordinates":[]}'),
```
paleolimbot, line 131:
```suggestion
        (None, None),
        ("POINT EMPTY", '{"type":"Point","coordinates":[]}'),
        ("LINESTRING EMPTY", '{"type":"LineString","coordinates":[]}'),
```

rust/sedona-geo/src/st_asgeojson.rs
```diff
@@ -0,0 +1,148 @@
...
+
+        // Initialize an output builder of the appropriate type
+        let mut builder =
+            StringBuilder::with_capacity(executor.num_iterations(), min_probable_geojson_size);
+
+        executor.execute_wkb_void(|maybe_geom| {
+            match maybe_geom {
+                Some(geom) => {
+                    // Convert geo_types::Geometry to geojson::Geometry
+                    let geojson_geom: geojson::Geometry = (&geom).into();
```
paleolimbot: You can solve the lack of empty point support and empty polygon incompatiblity here by special casing them. Maybe:

```rust
match geom.as_type() {
  geo_traits::GeometryType::Point(pt) => {
    if pt.coord().is_none() { /* special case the empty point output and return */ }
  }
  geo_traits::GeometryType::Polygon(pl) => {
    if pl.exterior().is_none() { /* special case the empty polygon output and return */ }
  }

  let geo_geom = item_to_geometry(geom)?;
  /* current implementation */
}
```

This means you'd have to go back to the `WkbExecutor` because `geo_types` can't represent an empty point. Most of the functions implemented in this crate have to do that (the ones that don't aren't used by default at the moment).

python/sedonadb/tests/functions/test_functions.py
paleolimbot: Alternatively, just add `0.5` to all of your coordinates (maybe add a comment with a link tp https://github.com/apache/sedona-db/issues/472 or a link to this PR and a short explanation of why you aren't using any integer coordinates in your test data given that all the other tests do).

rust/sedona-geo/src/st_asgeojson.rs
petern48: Opened a quick fix for POLYGON EMPTY in geojson: https://github.com/georust/geojson/pull/262. Though we should just move forward with the workaround.

paleolimbot (approved):
Thank you!

PR #471, opened 2025-12-30
feat(rust/sedona-pointcloud) Initial LAZ format support

b4l:
This provides basic listing table provider functionality for LAZ files, featuring partitioned/parallel read capabilities with low memory footprint.


Cargo.toml
```diff
@@ -83,12 +85,14 @@ datafusion = { version = "50.2.0", default-features = false }
 datafusion-catalog = { version = "50.2.0" }
 datafusion-common = { version = "50.2.0", default-features = false }
 datafusion-common-runtime = { version = "50.2.0", default-features = false }
+datafusion-datasource = { version = "50.2.0", default-features = false }
```
paleolimbot: Apologies that I updated to DataFusion 51 in a recent PR 😬  (there are a few datasource related changes that may force some minor changes here)

examples/sedonadb-rust-pointcloud/src/main.rs
```diff
@@ -0,0 +1,35 @@
...
+use datafusion::common::Result;
+use sedona::context::{SedonaContext, SedonaDataFrame};
+
+#[tokio::main]
+async fn main() -> Result<()> {
+    let ctx = SedonaContext::new_local_interactive().await?;
+    let url = "https://basisdata.nl/hwh-ahn/ahn4/01_LAZ/C_69AZ1.LAZ";
+
+    // let _ = ctx.sql(&format!("SET laz.point_encoding = 'plain'")).await.unwrap();
+    // let df = ctx.sql(&format!("SELECT x, y, z FROM \"{url}\"")).await.unwrap();
```
paleolimbot:
```suggestion
```

Can these be removed?

rust/sedona-pointcloud/src/laz/builder.rs
```diff
@@ -0,0 +1,569 @@
...
+                    let x = unsafe { x.value_unchecked(i) };
+                    let y = unsafe { y.value_unchecked(i) };
+                    let z = unsafe { z.value_unchecked(i) };
+
+                    wkb_bytes[5..13].copy_from_slice(x.to_le_bytes().as_slice());
+                    wkb_bytes[13..21].copy_from_slice(y.to_le_bytes().as_slice());
+                    wkb_bytes[21..29].copy_from_slice(z.to_le_bytes().as_slice());
+
+                    unsafe { builder.push_wkb_unchecked(Some(&wkb_bytes)) };
+                }
```
paleolimbot, line 198: This can probably use an `arrow_array` `BinaryBuilder` since all we're doing is pushing raw bytes into an array

rust/sedona-pointcloud/src/laz/format.rs
```diff
@@ -0,0 +1,210 @@
...
+
+#[async_trait::async_trait]
+impl FileFormat for LazFormat {
+    fn as_any(&self) -> &dyn Any {
+        self
+    }
+
+    fn get_ext(&self) -> String {
+        LazFormatFactory::new().get_ext()
+        // DEFAULT_LAZ_EXTENSION[1..].to_string()
```
paleolimbot:
```suggestion
        LazFormatFactory::new().get_ext()
```

rust/sedona-pointcloud/src/laz/opener.rs
```diff
@@ -0,0 +1,130 @@
...
+            if let Some(pruning_predicate) = &pruning_predicate {
+                if let Some(statistics) = file.statistics {
+                    let prunable_statistics = PrunableStatistics::new(vec![statistics], schema);
+                    if let Ok(filter) = pruning_predicate.prune(&prunable_statistics) {
+                        if !filter[0] {
+                            return Ok(futures::stream::empty().boxed());
+                        }
+                    }
+                }
+            }
```
paleolimbot, line 101: Does this handle file pruning based on `st_intersects()` and friends? Our helper for that is:

https://github.com/apache/sedona-db/blob/e80fb37b2a15a9708420fef0cc575b9e5274972e/rust/sedona-geoparquet/src/file_opener.rs#L128

rust/sedona-pointcloud/src/laz/opener.rs
```diff
@@ -0,0 +1,130 @@
...
+                for chunk_meta in chunk_table.into_iter() {
+                    // limit
+                    if let Some(limit) = limit {
+                        if row_count >= limit {
+                            break;
+                        }
+                    }
+
+                    // fetch batch
+                    let record_batch = laz_reader.get_batch(&chunk_meta).await?;
```
paleolimbot, line 128: Does this fetch respect the requested batch size? (I think it's in the partitioned file/meta, which merged in DataFusion 51)

rust/sedona-pointcloud/src/laz/opener.rs
```diff
@@ -0,0 +1,130 @@
...
+                    row_count += record_batch.num_rows();
+
+                    // project
+                    let record_batch = record_batch
+                        .project(&projection)
```
paleolimbot, line 134: Follow-up territory, but this could probably save some decoding effort if it pushed the projection into the builder.

rust/sedona-pointcloud/src/laz/opener.rs
```diff
@@ -0,0 +1,130 @@
...
+                        .map_err(|e| DataFusionError::ArrowError(Box::new(e), None))?;
+
+                    yield record_batch
+                }
+
+            };
+
+            // let stream = futures::stream::iter(batches);
+
```
paleolimbot:
```suggestion
```

rust/sedona-pointcloud/src/laz/options.rs
```diff
@@ -0,0 +1,147 @@
...
+    extensions_options,
+};
+
+/// Geometry representation
+#[derive(Clone, Copy, Default, PartialEq, Eq, Debug)]
+pub enum LasPointEncoding {
+    #[default]
+    Plain,
+    Wkb,
+    Nativ,
```
paleolimbot:
```suggestion
    Native,
```

rust/sedona-pointcloud/src/laz/options.rs
```diff
@@ -0,0 +1,147 @@
...
+
+use datafusion_common::{
+    config::{ConfigExtension, ConfigField, Visit},
+    error::DataFusionError,
+    extensions_options,
+};
+
+/// Geometry representation
+#[derive(Clone, Copy, Default, PartialEq, Eq, Debug)]
+pub enum LasPointEncoding {
```
paleolimbot: This would benefit from some docstrings to explain how this option affects the output schema. For the purposes of SedonaDB, it might make sense to have the default be `Wkb` so that it works with `ST_Intersects()` pruning when doing `SELECT * FROM 'foofy.laz' WHERE ST_Intersects(...)`. This can also be something we apply at a higher level (e.g., as a default in the Python bindings).

rust/sedona-pointcloud/src/laz/schema.rs
```diff
@@ -0,0 +1,147 @@
...
+            Field::new("z", DataType::Float64, false),
+        ],
+        LasPointEncoding::Wkb => {
+            let point_type = WkbType::new(Arc::new(Metadata::new(crs, None)));
+            vec![Field::new("geometry", DataType::Binary, false).with_extension_type(point_type)]
+        }
+        LasPointEncoding::Nativ => {
+            let point_type = PointType::new(Dimension::XYZ, Arc::new(Metadata::new(crs, None)))
+                .with_coord_type(CoordType::Separated);
+            vec![point_type.to_field("geometry", false)]
```
paleolimbot: No need to change this, but I'll have to add `SedonaType::GeoArrowNative` or `SedonaType::GenericExtension` for this to work with SedonaDB (or else it will give an unsupported type error).

rust/sedona-pointcloud/src/laz/schema.rs
```diff
@@ -0,0 +1,147 @@
...
+// Arrow schema for LAS points
+pub fn schema_from_header(
+    header: &Header,
+    point_encoding: LasPointEncoding,
+    extra_bytes: LasExtraBytes,
+) -> Schema {
+    let crs = header
+        .get_wkt_crs_bytes()
+        .and_then(|b| String::from_utf8(b.to_vec()).ok().map(Crs::from_wkt2_2019))
+        .unwrap_or_default();
```
paleolimbot: I'll have to add WKT2 CRS support to SedonaDB's CRS helper (I think right now it will error).

rust/sedona-pointcloud/src/laz/metadata.rs
```diff
@@ -0,0 +1,403 @@
...
+}
+
+impl FileMetadata for LazMetadata {
+    fn as_any(&self) -> &dyn Any {
+        self
+    }
+
+    fn memory_size(&self) -> usize {
+        // TODO: proper size
+        std::mem::size_of_val(self)
```
paleolimbot: + `chunk_table` capacity in bytes + `extra_attributes` capacity in bytes?

rust/sedona-pointcloud/src/laz/metadata.rs
```diff
@@ -0,0 +1,403 @@
...
+                metadata
+                    .chunk_table
+                    .iter()
+                    .map(|meta| meta.byte_range.end - meta.byte_range.start)
+                    .sum::<u64>() as usize,
+            ));
+
+        let bounds = metadata.header.bounds();
+        for field in table_schema.fields() {
+            let cs = match field.name().as_str() {
```
paleolimbot, line 190: For `"geometry"` we haven't currently fudged `ColumnStatistics` to store a bbox, but in an early version of the GeoParquet reader I serialized `sedona_expr::GeoStatistics` as JSON as the "sum". Perhaps it's just a separate method here to return `GeoStatistics`, which is what you'll need to prune using `SpatialFilter`.

rust/sedona-pointcloud/src/laz/metadata.rs
```diff
@@ -0,0 +1,403 @@
...
+                3 => DataType::UInt16,
+                4 => DataType::Int16,
+                5 => DataType::UInt32,
+                6 => DataType::Int32,
+                7 => DataType::UInt64,
+                8 => DataType::Int64,
+                9 => DataType::Float32,
+                10 => DataType::Float64,
+                11..=30 => panic!("deprecated extra bytes data type"),
+                31..=255 => panic!("reserved extra butes data type"),
```
paleolimbot: These panics (and the unwraps elsewhere in this file) should probably return errors instead of panics (they could occur in normal usage when reading a fuzzed/corrupted file).

rust/sedona-pointcloud/src/laz/opener.rs
b4l: Addressed in https://github.com/apache/sedona-db/pull/471/commits/aa669539894e24acbee941d216cf544da4f29e3a

rust/sedona-pointcloud/src/laz/builder.rs
b4l: Done in https://github.com/apache/sedona-db/pull/471/commits/a3211f8fd2e4d4ac662cb1adec5d83f15284b5ec

examples/sedonadb-rust-pointcloud/src/main.rs
b4l: removed

Cargo.toml
b4l: No worries, rebased it.

rust/sedona-pointcloud/src/laz/opener.rs
b4l: This fetches a whole laz chunk (commonly 50k points) and does not respect the requested batch size. Not sure if it is required or has any implications. A proper stream reader would be nice at some point, but it's a little work.

rust/sedona-pointcloud/src/laz/opener.rs
b4l: Possibly. I haven't looked into the lower-level libraries much yet. Pasture did something like this: https://github.com/igd-geo/pasture

rust/sedona-pointcloud/src/laz/options.rs
b4l: thanks!

rust/sedona-pointcloud/src/laz/opener.rs
b4l: removed

rust/sedona-pointcloud/src/laz/options.rs
b4l: Yeah, it is set to `wkb` in the Sedona context, so pruning with 'ST_` works.

rust/sedona-pointcloud/src/laz/schema.rs
b4l: Changed it to use a helper library to extract the EPSG code in https://github.com/apache/sedona-db/pull/471/commits/cd2b8ba4d2b1bbf64452c02472f911dd311d4bf7

rust/sedona-pointcloud/src/laz/metadata.rs
b4l: Improved in https://github.com/apache/sedona-db/pull/471/commits/cd2b8ba4d2b1bbf64452c02472f911dd311d4bf7

Maybe also need to calculate the header, which can blow up with arbitrarily sized (E)VLRs

rust/sedona-pointcloud/src/laz/metadata.rs
b4l: Creating the bbox from the header is sufficient to prune the files for now. Maybe we can address more elaborate statistics extraction and handling in a follow-up?

rust/sedona-pointcloud/src/laz/metadata.rs
b4l: Converted to errors

rust/sedona-pointcloud/src/laz/opener.rs
paleolimbot: This reader really should output batches according to `batch_size`, as it has implications for for all the operators that come after (too high and the batch no longer fits in L2 cache and may have more limited parallelization; too low and batch overhead dominates).

I think there is a helper for this (we should probably do this on our RecordBatchReaderProvider):

https://github.com/kosiew/datafusion/blob/2e8b9c956dddae2d639633c70002e7d9a3856c50/datafusion/physical-plan/src/stream.rs#L528-L658

Details: https://github.com/apache/datafusion/issues/16717

rust/sedona-pointcloud/src/laz/opener.rs
paleolimbot: My guess is that most queries will select a small handful of columns and that decoding them all is sort of waste most of the time (but definitely a future concern as that requires reworking the approach here).

rust/sedona-pointcloud/src/laz/reader.rs
```diff
@@ -0,0 +1,232 @@
...
+    use las::{point::Format, Builder, Point, Writer};
+    use object_store::{memory::InMemory, path::Path, ObjectStore, PutPayload};
+
+    use crate::laz::reader::LazFileReaderFactory;
+
+    #[allow(static_mut_refs)]
+    #[tokio::test]
+    async fn reader_basic_e2e() {
+        // create laz file
+        static mut LAZ: Vec<u8> = Vec::new();
```
paleolimbot: Thanks for adding this!

Does this really need to be static mut? For some other tests we use a temporary directory, which would also let you check a multi file read.

https://github.com/apache/sedona-db/blob/90658675a6b849356599e2ae86477c0f8e60545c/rust/sedona/src/context.rs#L622

rust/sedona-pointcloud/src/laz/metadata.rs
```diff
@@ -0,0 +1,451 @@
...
+    use las::{point::Format, Builder, Reader, Writer};
+    use object_store::{memory::InMemory, path::Path, ObjectStore, PutPayload};
+
+    use crate::laz::metadata::LazMetadataReader;
+
+    #[allow(static_mut_refs)]
+    #[tokio::test]
+    async fn header_basic_e2e() {
+        // create laz file
+        static mut LAZ: Vec<u8> = Vec::new();
```
paleolimbot: Maybe also just a tempfile unless this really needs to be static/mutable?

rust/sedona-pointcloud/src/laz/reader.rs
```diff
@@ -0,0 +1,232 @@
...
+            )
+            .unwrap();
+        let metadata = laz_file_reader.get_metadata().await.unwrap();
+
+        let batch = laz_file_reader
+            .get_batch(&metadata.chunk_table[0])
+            .await
+            .unwrap();
+
+        assert_eq!(batch.num_rows(), 1);
```
paleolimbot, line 230: This level of granularity is OK if all the lower-level pieces are tested, although in this case they are mostly not.

rust/sedona-pointcloud/src/laz/builder.rs
```diff
@@ -0,0 +1,571 @@
...
+
+        dt => {
+            return Err(ArrowError::ExternalError(
+                format!("Unsupported data type for extra bytes: `{dt}`").into(),
+            ))
+        }
+    }
+
+    Ok(width)
+}
```
paleolimbot, line 501: Some suggestions for tests that should live in this file:

- Create a builder with zero rows, check all of the output options to ensure they give you the schema (or at least number of columns) you are expecting
- The building of attributes (tests should cover each branch here). One of the nice parts about refactoring this to use GATs if you can would be that there are fewer branches to test (although probably easier to use an `rstest` parameterized test like `#[values(DataType::Int8, DataType::Int16, ...)]`).

I don't think you'll need a test file for any of that (but you might need to create a mock header and mock attributes with/without offset, scale, and nodata).

rust/sedona-pointcloud/src/laz/metadata.rs
```diff
@@ -0,0 +1,451 @@
...
+pub struct ExtraAttribute {
+    pub data_type: DataType,
+    pub no_data: Option<[u8; 8]>,
+    pub scale: Option<f64>,
+    pub offset: Option<f64>,
+}
+
+pub(crate) fn extra_bytes_attributes(
+    header: &Header,
+) -> Result<Vec<ExtraAttribute>, Box<dyn Error + Send + Sync>> {
```
paleolimbot, line 278: Maybe?

```suggestion
) -> Result<Vec<ExtraAttribute>, DataFusionError> {
```

rust/sedona-pointcloud/src/laz/metadata.rs
```diff
@@ -0,0 +1,451 @@
...
+    }
+
+    Ok(attributes)
+}
+
+pub(crate) async fn chunk_table(
+    store: &(impl ObjectStore + ?Sized),
+    object_meta: &ObjectMeta,
+    header: &Header,
+) -> Result<Vec<ChunkMeta>, Box<dyn Error + Send + Sync>> {
```
paleolimbot, line 343:
```suggestion
) -> Result<Vec<ChunkMeta>, DataFusionError> {
```

Probably use `plan_err!()` for these?

rust/sedona-pointcloud/src/laz/metadata.rs
```diff
@@ -0,0 +1,451 @@
...
+
+        // read with las `Reader`
+        let read = unsafe { Cursor::new(&mut LAZ) };
+        let reader = Reader::new(read).unwrap();
+
+        assert_eq!(
+            reader.header(),
+            &metadata_reader.fetch_header().await.unwrap()
+        );
+    }
```
paleolimbot, line 451: With a dummy header (perhaps you can inline the bytes of a known test file, or perhaps you can create one using the Builder that exercises the code path for the full matrix of data types by offset/scale/nodata, or create a function that accepts `DataType`, `offset`, `scale`, `nodata`, and outputs a file with exactly one extra attribute that we can use in a parameterized test to check that it roundtrips.

rust/sedona-pointcloud/src/laz/opener.rs
b4l: Hmm not sure this applies here. This is already a part of a partitioned stream, so all cores should be working. And I am very suspicious about data remaining in l2 across operations in datafusion in general, even when the size is right given there is no explicit fusion and execution is driven by tokio. I might be wrong though here, it's just a hunch. 

Regarding the bytes range partitioning, I however made the experience that for pointclouds, randomly splitting the points across as many files as you have cores gives a certain performance boost as all partions always contribute the same amount to the result set. Otherwise round robin is superior to byte ranges as there is almost always a certain order (even if it is just scanlines) and thus the workload and result set is more evenly distributed among partitions for selective queries that can end up in a single byte range.

paleolimbot: Downstream operators are expecting batches of ~8k rows so we really do need to respect that value. It's very possible that increasing that value to 50k makes no difference for your files, but if you don't implement that option we can't benchmark the effect of that. (Mostly the effect that has is not on your code here but on downstream operators that, say, buffer the input and create a much larger batch size where the difference between 8k and 50k really does matter).

rust/sedona-pointcloud/src/laz/reader.rs
b4l: converted to using temptir

rust/sedona-pointcloud/src/laz/metadata.rs
b4l: done

rust/sedona-pointcloud/src/laz/reader.rs
b4l: There are now tests for any point format and single value extra attributes, each plane, scaled, and with nodata. Does this suffice?

rust/sedona-pointcloud/src/laz/builder.rs
b4l: Addressed in https://github.com/apache/sedona-db/pull/471/commits/7b5bd3684be69b2409cd0959ea01d8e7fe82a4b3

Unfortunately, the Rust library does not support reading/writing extra attributes, so I had to use laspy to generate a file. Not sure where this should live.

rust/sedona-pointcloud/src/laz/metadata.rs
b4l: I would keep it as is if you don't mind.

rust/sedona-pointcloud/src/laz/metadata.rs
b4l: I don't feel like planning fits here.

rust/sedona-pointcloud/src/laz/metadata.rs
b4l: Addressed in https://github.com/apache/sedona-db/pull/471/commits/7b5bd3684be69b2409cd0959ea01d8e7fe82a4b3

rust/sedona-pointcloud/src/laz/opener.rs
b4l: Addressed in https://github.com/apache/sedona-db/pull/471/commits/90b56bd620dccda73b3a533e29ac6bc87a6a58ed

b4l:
@paleolimbot Thank you for the review! Sorry for being slow to address it.

I was wondering where the test data/generation should live. Is it in `sedona-testing`?

c/sedona-s2geography/s2geography
paleolimbot, line 1: Does this submodule need to be reverted or updated?

examples/sedonadb-rust-pointcloud/src/main.rs
```diff
@@ -0,0 +1,33 @@
...
+// Because a number of methods only return Err() for not implemented,
+// the compiler doesn't know how to guess which impl RecordBatchReader
+// will be returned. When we implement the methods, we can remove this.
+
+use std::error::Error;
+
+use sedona::context::{SedonaContext, SedonaDataFrame};
+
+#[tokio::main]
+async fn main() -> Result<(), Box<dyn Error>> {
```
paleolimbot: Optional nit (we use `datafusion_common::Result` pretty much everywhere else)

```suggestion
async fn main() -> Result<()> {
```

examples/sedonadb-rust-pointcloud/src/main.rs
paleolimbot, line 1: Thanks!

We can add this here to run in CI:

https://github.com/apache/sedona-db/blob/d17d7f08dd65fe42d9bf64126ae2a383666c72f6/.github/workflows/examples.yml#L76-L90

rust/sedona-pointcloud/src/laz/options.rs
```diff
@@ -0,0 +1,150 @@
...
+    /// Resolves the coordinates to a fields `geometry` with separated GeoArrow encoding.
+    Native,
+}
+
+impl Display for LasPointEncoding {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        match self {
+            LasPointEncoding::Plain => f.write_str("plain"),
+            LasPointEncoding::Wkb => f.write_str("wkb"),
+            LasPointEncoding::Native => f.write_str("nativ"),
```
paleolimbot: Optional nit:

```suggestion
            LasPointEncoding::Native => f.write_str("native"),
```

rust/sedona-pointcloud/src/laz/options.rs
```diff
@@ -0,0 +1,150 @@
...
+    pub struct LazTableOptions {
+        pub point_encoding: LasPointEncoding, default = LasPointEncoding::default()
+        pub extra_bytes: LasExtraBytes, default = LasExtraBytes::default()
+    }
+
+}
+
+impl ConfigExtension for LazTableOptions {
+    const PREFIX: &'static str = "laz";
+}
```
paleolimbot: Should this be `pointcloud`? (e.g., will these apply to las files as well as laz files?)

rust/sedona-pointcloud/src/laz/builder.rs
```diff
@@ -0,0 +1,926 @@
...
+                3 | 5 | 7 => assert_eq!(batch.num_columns(), 21),
+                8 | 10 => assert_eq!(batch.num_columns(), 22),
+                _ => unreachable!(),
+            }
+        }
+    }
+
+    #[tokio::test]
+    async fn extra_attributes() {
+        // create extra file
```
paleolimbot: Can you add a quick description of what this file contains?

Probably you can just include this as a file and perhaps some other files in the rust/sedona-pointcloud directory with the Python script that generates it (and we can follow up with adding them to sedona-testing without blocking this PR). I believe cargo test sets the working directory such that this is reasonable to do.

rust/sedona-pointcloud/src/laz/builder.rs
```diff
@@ -0,0 +1,926 @@
...
+            0, 0, 0, 0, 0, 0, 0, 1, 0, 0, 0, 1, 0, 0, 0, 0, 0, 0, 0, 1, 0, 0, 0, 82, 237, 50, 0, 0,
+        ];
+
+        let tmpdir = tempfile::tempdir().unwrap();
+        let tmp_path = tmpdir.path().join("extra.laz");
+        std::fs::write(&tmp_path, extra).unwrap();
+
+        // TODO: use sedona-testing or similar
+        // let path = "../../submodules/sedona-testing/data/pointclouds/extra.laz";
+
```
paleolimbot: (This is a good idea but for now I think the test bytes work)

```suggestion
```

rust/sedona-pointcloud/src/laz/format.rs
```diff
@@ -0,0 +1,209 @@
...
+        _state: &dyn Session,
+        conf: FileScanConfig,
+    ) -> Result<Arc<dyn ExecutionPlan>, DataFusionError> {
+        Ok(DataSourceExec::from_data_source(conf))
+    }
+
+    fn file_source(&self) -> Arc<dyn FileSource> {
+        Arc::new(LazSource::default().with_options(self.options.clone()))
+    }
+}
```
paleolimbot, line 212: In `sedona-geoparquet/src/format.rs` you'll see a few tests for the Parquet format that verify a few basic properties of the format in the spirit of "is it plugged in". I think this file would benefit from having those (both to make sure everything here is plugged in and to ensure future contributors to this file have a place to test bug fixes).

rust/sedona-pointcloud/src/laz/builder.rs
paleolimbot: I left a suggestion inline...anywhere in rust/sedona-pointcloud that makes sense for the purposes of this PR!

paleolimbot (commented):

Mostly nits...I think `rust/sedona-pointcloud/testing` would work for any example files/scripts to generate the for now to unblock this PR!

I have one more suggestion for testing the final bit of untested code here, which is the file format...I think you can copy any relevant tests from `rust/sedona-geoparquet/src` except they should be easier since you can use `assert_batches_eq!()` (I couldn't there because I had geometry column output and your output should just be xyz)`. Among other things, this makes sure that filters, projections, and options work and are all piped through all the steps.

c/sedona-s2geography/s2geography
b4l: No sure how this ended up here so probably it needs to be reverted.

examples/sedonadb-rust-pointcloud/src/main.rs
b4l: done

rust/sedona-pointcloud/src/laz/options.rs
b4l: changed

examples/sedonadb-rust-pointcloud/src/main.rs
b4l: this would entail adding `datafusion_common` as a dependency, so i will leaf it as is

rust/sedona-pointcloud/src/laz/options.rs
b4l: I refactored the config in https://github.com/apache/sedona-db/pull/471/commits/43803c3a79bd30414f8a1823330aad31527b75f3 to reflect the potential scope of future usage.

rust/sedona-pointcloud/src/laz/builder.rs
b4l: Added more descriptive comment, also the generating source is now checked in for reference.

rust/sedona-pointcloud/src/laz/builder.rs
b4l: Added the file and generating script to the crate in `/tests/data`.

rust/sedona-pointcloud/src/laz/format.rs
b4l: Ported some tests.

PR #476, opened 2026-01-02
perf: Implement method to directly write GeosGeometry into buffer and use on st_unaryunion + st_buffer

### LINKED ISSUE #461: perf: Directly write `GeosGeometry` into binary builder buffer
We should probably also look into optimizing `GeosGeometry` writing to `BinaryBuilder`...`to_wkb()` and then `builder.write()` is probably not as fast as peeking into all the geometries and writing them to the output in place. (I think this will still be faster though and is easier since it's all rust!)

_Originally posted by @paleolimbot in https://github.com/apache/sedona-db/pull/233#pullrequestreview-3578780925_

petern48:
closes #461

This PR implements the method and applies the change to two functions (st_unaryunion and st_buffer) only. Changes to other functions will follow later.

Results after PR merged.
ST_UnaryUnion
image
ST_Buffer
image

### DIFF
--- a/c/sedona-geos/src/geos_to_wkb.rs
+++ b/c/sedona-geos/src/geos_to_wkb.rs
@@ -17,10 +17,14 @@
 use std::io::Write;
 
 use byteorder::{LittleEndian, WriteBytesExt};
-use datafusion_common::{error::Result, exec_err, DataFusionError};
+use datafusion_common::{error::Result, DataFusionError};
+use geo_traits::Dimensions;
 use geos::{Geom, Geometry, GeometryTypes};
-
-const ENDIANNESS: u8 = 1;
+use sedona_geometry::wkb_factory::{
+    write_wkb_geometrycollection_header, write_wkb_linestring_header,
+    write_wkb_multilinestring_header, write_wkb_multipoint_header, write_wkb_multipolygon_header,
+    write_wkb_point_header, write_wkb_polygon_header,
+};
 
 /// Write a GEOS geometry to WKB format.
 ///
@@ -35,24 +39,6 @@ fn write_geometry(geom: &impl Geom, writer: &mut impl Write) -> Result<()> {
         .geometry_type()
         .map_err(|e| DataFusionError::Execution(format!("Failed to get geometry type: {e}")))?;
 
-    match geom_type {
-        GeometryTypes::Point => write_point(geom, writer),
-        GeometryTypes::LineString => write_line_string(geom, writer),
-        GeometryTypes::Polygon => write_polygon(geom, writer),
-        GeometryTypes::MultiPoint => write_multi_point(geom, writer),
-        GeometryTypes::MultiLineString => write_multi_line_string(geom, writer),
-        GeometryTypes::MultiPolygon => write_multi_polygon(geom, writer),
-        GeometryTypes::GeometryCollection => write_geometry_collection(geom, writer),
-        _ => Err(DataFusionError::Execution(format!(
-            "Unsupported geometry type: {geom_type:?}"
-        ))),
-    }
-}
-
-fn write_point(geom: &impl Geom, writer: &mut impl Write) -> Result<()> {
-    // Write byte order
-    writer.write_u8(ENDIANNESS)?;
-
     let has_z = geom
         .has_z()
         .map_err(|e| DataFusionError::Execution(format!("Failed to check has_z: {e}")))?;
@@ -60,14 +46,29 @@ fn write_point(geom: &impl Geom, writer: &mut impl Write) -> Result<()> {
         .has_m()
         .map_err(|e| DataFusionError::Execution(format!("Failed to check has_m: {e}")))?;
 
-    let wkb_type = match (has_z, has_m) {
-        (false, false) => 1,   // Point
-        (true, false) => 1001, // Point Z
-        (false, true) => 2001, // Point M
-        (true, true) => 3001,  // Point ZM
+    let dim = match (has_z, has_m) {
+        (false, false) => Dimensions::Xy,
+        (true, false) => Dimensions::Xyz,
+        (false, true) => Dimensions::Xym,
+        (true, true) => Dimensions::Xyzm,
     };
+    match geom_type {
+        GeometryTypes::Point => write_point(geom, dim, writer),
+        GeometryTypes::LineString => write_line_string(geom, dim, writer),
+        GeometryTypes::Polygon => write_polygon(geom, dim, writer),
+        GeometryTypes::MultiPoint => write_multi_point(geom, dim, writer),
+        GeometryTypes::MultiLineString => write_multi_line_string(geom, dim, writer),
+        GeometryTypes::MultiPolygon => write_multi_polygon(geom, dim, writer),
+        GeometryTypes::GeometryCollection => write_geometry_collection(geom, dim, writer),
+        _ => Err(DataFusionError::Execution(format!(
+            "Unsupported geometry type: {geom_type:?}"
+        ))),
+    }
+}
 
-    writer.write_u32::<LittleEndian>(wkb_type)?;
+fn write_point(geom: &impl Geom, dim: Dimensions, writer: &mut impl Write) -> Result<()> {
+    write_wkb_point_header(writer, dim)
+        .map_err(|e| DataFusionError::Execution(format!("Failed to write point header: {e}")))?;
 
     let is_empty = geom
         .is_empty()
@@ -77,10 +78,10 @@ fn write_point(geom: &impl Geom, writer: &mut impl Write) -> Result<()> {
         // Write NaN coordinates for empty point
         writer.write_f64::<LittleEndian>(f64::NAN)?; // x
         writer.write_f64::<LittleEndian>(f64::NAN)?; // y
-        if has_z {
+        if matches!(dim, Dimensions::Xyz | Dimensions::Xyzm) {
             writer.write_f64::<LittleEndian>(f64::NAN)?; // z
         }
-        if has_m {
+        if matches!(dim, Dimensions::Xym | Dimensions::Xyzm) {
             writer.write_f64::<LittleEndian>(f64::NAN)?; // m
         }
     } else {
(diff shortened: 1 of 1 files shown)


c/sedona-geos/src/st_unaryunion.rs
```diff
@@ -77,7 +78,30 @@ fn invoke_scalar(geos_geom: &geos::Geometry) -> Result<Vec<u8>> {
...
+
+    // wkb_writer method maintains proper z and m values
+    use geos::WKBWriter;
+    let mut wkb_writer = WKBWriter::new().unwrap();
+    // use geos::OutputDimension;
+    // wkb_writer.set_output_dimension(if has_z && has_m { OutputDimension::FourD } else if has_z { OutputDimension::ThreeD } else if has_m { OutputDimension::ThreeD } else { OutputDimension::TwoD });
+    let wkb = wkb_writer.write_wkb(&geometry).unwrap();
+    let tmp_geom = geos::Geometry::new_from_wkb(&wkb).unwrap();
+    let (has_z, has_m) = (tmp_geom.has_z().unwrap(), tmp_geom.has_m().unwrap());
+    println!("after write_wkb(): has_z: {:?}, has_m: {:?}", has_z, has_m);  // proper z and m values
```
petern48: Side note: I sketched out some code here that led to some interesting behavior (I'm immediately deleting it, but I'll comment here to save my thoughts somewhere. I found that `.to_wkb()` drops the Z and M dimensions.

After looking into it, I see `.to_wkb()` is [implemented using GEOSGeomToWKB_buf_r](https://github.com/georust/geos/blob/47afbad2483e489911ddb456417808340e9342c3/src/geometry.rs#L2229), which [is deprecated and suggestes using GEOSWKBWriter_write instead](https://github.com/libgeos/geos/blob/158c588566e08fd2f31f8c9e7ae069518e8264c5/capi/geos_c.h.in#L6685). In GeoRust, that cooresponds to using [WKBWriter.write_wkb()](https://github.com/georust/geos/blob/47afbad2483e489911ddb456417808340e9342c3/src/wkb_writer.rs#L71) instead. I tried it out, and it seems to work. tldr; `WKBWriter.write_wkb` should be used instead of `.to_wkb()`. 

This code used on `POINT ZM (1 2 3 4)` results in the following.
```
before to_wkb: has_z: true, has_m: true
after to_wkb(): has_z: false, has_m: false
after write_wkb(): has_z: true, has_m: true
```

Though in reality, I think this new method `write_geos_geometry` should replace any need for `write_wkb()`.

c/sedona-geos/src/geos_to_wkb.rs
```diff
@@ -0,0 +1,503 @@
...
+
+    let coords = coord_seq
+        .as_buffer(Some(dims))
+        .map_err(|e| DataFusionError::Execution(format!("Failed to get coord seq buffer: {e}")))?;
+
+    // TODO: I guess we could compute the exact size buffer we expect here and check it. Should we do it or just assume geos correct?
+    // TODO: do we need to consider endianness??
+
+    // Cast Vec<f64> to &[u8] so we can write the bytes directly to the writer buffer
+    let byte_slice: &[u8] = bytemuck::cast_slice(&coords);
```
petern48, line 273: I originally went with the approach of using the .get_x, get_y, etc methods to parse out the coords (code is still commented a few lines down atm), but I found we can just use this `as_buffer()` method and copy the buffer directly.

1. You think this is safe enough or would you prefer to go the getter route? (assuming tests pass)
2. WDYT about the questions in the TODO comments here? Should we check size? Do we need to support writing big endian? If I do, maybe geos handles that already?
3. `bytemuck` (a new external library) was the only way I could find to cast a `Vec<f64>` to a `&[u8]` so we can call `writer.write_all()` on the result of `as_buffer()`. Any better idea?

c/sedona-geos/src/geos_to_wkb.rs
```diff
@@ -0,0 +1,503 @@
...
+            .map_err(|e| DataFusionError::Execution(format!("Failed to get geometry {i}: {e}")))?;
+
+        write_geometry(&sub_geom, writer)?;
+    }
+
+    Ok(())
+}
+
+// TBH. um not sure this is needed for geos
+fn get_geomcol_dimension(geomcol: &impl Geom) -> Result<u32> {
```
petern48: reminder for myself to check on if this is really a problem, it's what I faced in wkb_header PR. an LLM told me this was the problem and i implemented it without verifying that was indeed a problem. the root of the problem i cared for turned out to be something else, and i hadn't revisited this yet

Cargo.toml
```diff
@@ -100,12 +101,11 @@ mimalloc = { version = "0.1", default-features = false }
 libmimalloc-sys = { version = "0.1", default-features = false }
 once_cell = "1.20"
 
-geos = { git="https://github.com/georust/geos.git", rev="47afbad2483e489911ddb456417808340e9342c3", features = ["geo", "v3_11_0"] }
+geos = { git="https://github.com/georust/geos.git", rev="47afbad2483e489911ddb456417808340e9342c3", features = ["geo", "v3_12_0"] }
```
petern48, line 104: need 3_12_0 so i can access `has_m()`. `get_m()` is not implemented yet, but i found a nice way around that by copying the whole buffer

c/sedona-geos/src/geos_to_wkb.rs
```diff
@@ -0,0 +1,503 @@
...
+use geos::{Geom, Geometry, GeometryTypes};
+
+// TODO:
+const NATIVE_ENDIANNESS: u8 = 1;
+
+/// Write a GEOS geometry to WKB format.
+///
+/// This is a fast, custom implementation that directly extracts coordinates
+/// from GEOS geometries and writes them in WKB format into a buffer.
+pub fn write_geos_geometry(geom: &Geometry, writer: &mut impl Write) -> Result<()> {
```
petern48, line 33: I just realized I kinda just picked ISO WKB and it worked. Should I keep this, support EWKB instead or support both? i don't quite have the strongest understanding of how SRIDs fit into this picture. ik it's stored in metadata, so i'm guessing that just handled completely separate from the geos geometries and there's no need to worry about that here?

python/sedonadb/tests/functions/test_functions.py
```diff
@@ -874,17 +827,70 @@ def test_st_convexhull(eng, geom, expected):
             "GEOMETRYCOLLECTION (POLYGON ((0 0, 1 0, 1 1, 0 1, 0 0)), POLYGON ((1 0, 2 0, 2 1, 1 1, 1 0)))",
             "POLYGON ((0 0, 0 1, 1 1, 2 1, 2 0, 1 0, 0 0))",
         ),
+        # Z and M support
+        ("POINT Z (0 0 0)", "POINT Z(0 0 0)"),
+        ("POINT M (1 2 3)", "POINT M(1 2 3)"),
+        ("POINT ZM (1 2 3 4)", "POINT ZM(1 2 3 4)"),
+        ("LINESTRING Z (0 0 0, 1 1 1)", "LINESTRING Z(0 0 0,1 1 1)"),
```
petern48: I mainly added these Z/M tests for st_unaryunion as a more convenient way to test `geos_to_wkb.rs`. I need to add rust tests. We can keep them or delete them, but Z/M support is new here since `.to_wkb()` apparently doesn't support them.

c/sedona-geos/src/geos_to_wkb.rs
paleolimbot: (Native or little endian, whichever is easier) ISO WKB is the only thing we need to support here...our SRID support is handled outside of GEOS/Wkb (as a struct with the crs in a separate column).

c/sedona-geos/src/geos_to_wkb.rs
paleolimbot: If we had control over all the parts, the best thing we could possibly do is to call `GEOSCoordSeq_copyToBuffer()` on the sequence such that it copies the bytes directly into the output buffer ( https://github.com/libgeos/geos/blob/158c588566e08fd2f31f8c9e7ae069518e8264c5/capi/geos_c.h.in#L2459-L2469 ). The second best thing we could do would be to keep a mutable `Vec<f64>` on hand that we re-use every time to copy the sequence into, then write the bytes into the output. Given the constraints of the current `as_buffer()` I think your solution is the best we can do for now!

As a side note, there's a possibly a bug in `as_buffer()` (the logic won't work for XYM geometries): https://github.com/georust/geos/blob/47afbad2483e489911ddb456417808340e9342c3/src/coord_seq.rs#L521-L540 .

python/sedonadb/tests/functions/test_functions.py
paleolimbot: That's a good catch! We should definitely keep these tests.

python/sedonadb/tests/functions/test_functions.py
```diff
@@ -827,17 +827,70 @@ def test_st_convexhull(eng, geom, expected):
...
     ],
 )
 def test_st_unaryunion(eng, geom, expected):
+    is_postgis = eng == PostGIS
     eng = eng.create_or_skip()
 
     if expected is None:
         eng.assert_query_result(f"SELECT ST_UnaryUnion({geom_or_null(geom)})", expected)
+    # PostGIS doesn't support M dimensions so we skip the test
+    elif is_postgis and (" M" in expected or "ZM" in expected):
```
paleolimbot: Instead of the `if` statements here, can this be a separate test?

python/sedonadb/tests/functions/test_functions.py
```diff
@@ -827,17 +827,70 @@ def test_st_convexhull(eng, geom, expected):
...
+    elif is_postgis and "Z" in expected:  # and "MULTIPOLYGON" not in expected:
+        pytest.skip(
+            "Temporarily skip PostGIS z tests for now since WKT is spaced differently, and ST_Equals() doesn't work"
+        )
     elif "EMPTY" in expected.upper():
         eng.assert_query_result(
             f"SELECT ST_IsEmpty(ST_UnaryUnion({geom_or_null(geom)}))", True
         )
+    # ST_Equals() (in the else case) doesn't test for Z and M dimensions so we test for string equality for 3-4D
+    elif "Z" in expected.upper() or "M " in expected.upper():
```
paleolimbot: I hope string equality works here...the alternative would probably be to check for 2D equality and then dump the coordinates and check for equal z/m values (maybe by sorting them and checking Python list equality?)

c/sedona-geos/src/geos_to_wkb.rs
paleolimbot: I suppose removing this and seeing if your tests still pass would do it? If not this workaround seems fine to me (it's theoretically slower, but only slower for geometry collections and that's not something at least I'm super worried about the performance implications of)

Cargo.toml
paleolimbot: GEOS 3.12 is in Ubuntu 24.04 and so I don't think will be a problem for CI or contributors!

paleolimbot (commented):
This is huge! I'm also curious about ST_Buffer() benchmarks since that function typically returns very complicated geometries. I know you're still working here but in general this is great.

c/sedona-geos/src/geos_to_wkb.rs
petern48: Yes, I think you're right. I found my exact conditional logic was allowing `XYM` cases to slip to the `ST_Equals()` assert to bypass the M dimension check, hence why the tests were passing. Seems like a breaking change is necessary to fix the API, considering the [dims argument is an int](https://github.com/georust/geos/blob/47afbad2483e489911ddb456417808340e9342c3/src/coord_seq.rs#L522).

I also see that the `CoordSeq::new()` is buggy too, [it's implementation](https://github.com/georust/geos/blob/47afbad2483e489911ddb456417808340e9342c3/src/coord_seq.rs#L63-L65) uses the legacy `GEOSCoordSeq_create_r` function. Looking at a [git blame](https://github.com/libgeos/geos/blame/158c588566e08fd2f31f8c9e7ae069518e8264c5/capi/geos_c.h.in#L501-L511) at these lines in geos, I see `GEOSCoordSeq_create_r` is 18 years old and it takes a simple `int dims` param. while the `GEOSCoordSeq_createWithDimensions_r` function is only 9 months old, and it takes a separate `hasz` and `hasm` param, as desired. Would also need some breaking fixes to multiple enums in [enums.rs](https://github.com/georust/geos/blob/47afbad2483e489911ddb456417808340e9342c3/src/enums.rs#L6). Looks like none of geos-rust supports XYM and a lot of breaking changes would need to be made 😬

petern48:
> I'm also curious about ST_Buffer() benchmarks since that function typically returns very complicated geometries

I updated the PR description with results for st_buffer. Only 12-13% speedup. A little disappointing, but it makes sense. st_buffer is such an expensive algorithm that the optimizing the wkb writing is a smaller portion of the total time. `geo` should still be faster. This does at least speed up the buffer params kernel, which still uses geos. I want to leave the rest of the functions for a separate PR(s) to avoid cramming this one with so many benchmark results.

PR is still WIP as I need to look into fixing XYM.

python/sedonadb/tests/functions/test_functions.py
petern48: I moved the ZM tests to a separate test, this exact if condition though still remains in that new test since I still want to test postgis on the Z cases. personally like it this way rather than pulling it out into a separate Z and M test. though let me know if you feel strongly otherwise

c/sedona-geos/src/geos_to_wkb.rs
petern48: removed this in https://github.com/apache/sedona-db/pull/476/commits/beb0733e1756934b5200e4598bae4c9a23953cb9 after i saw the rust and python tests still pass

petern48:
Alright, I've added rust tests, refactored the python tests a bit, and did a bit of cleaning up.

For the python tests, I went the route of using exact string matching, but removing all white spaces to avoid formatting differences. It may be slightly odd, but I think it's a safe testing approach. Looks like this for example: `LINESTRINGZM(1234,5678)`.

After further thought, I don't really want to spend much time trying to fix XYM upstream. I would expect some friction and discussion, and i don't think it's top priority. Not having `XYM` would not be a regression, so I personally think it would be okay to merge this without addressing `XYM`. I filed a ticket (https://github.com/apache/sedona-db/issues/481) to track it.

c/sedona-geos/src/geos_to_wkb.rs
```diff
@@ -0,0 +1,646 @@
...
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+use std::io::Write;
+
+use byteorder::{LittleEndian, WriteBytesExt};
+use datafusion_common::{error::Result, exec_err, DataFusionError};
+use geos::{Geom, Geometry, GeometryTypes};
+
+const ENDIANNESS: u8 = 1;
```
paleolimbot:
```suggestion
/// WKB byte for little endian
const ENDIANNESS: u8 = 1;
```

c/sedona-geos/src/geos_to_wkb.rs
```diff
@@ -0,0 +1,646 @@
...
+        GeometryTypes::GeometryCollection => write_geometry_collection(geom, writer),
+        _ => Err(DataFusionError::Execution(format!(
+            "Unsupported geometry type: {geom_type:?}"
+        ))),
+    }
+}
+
+fn write_point(geom: &impl Geom, writer: &mut impl Write) -> Result<()> {
+    // Write byte order
+    writer.write_u8(ENDIANNESS)?;
```
paleolimbot: Is there any way we can use the helpers here (maybe at least for the header writing?):

https://github.com/apache/sedona-db/blob/c4ad8c9386b10e4b785d0ea99d36dc4ff56937e3/rust/sedona-geometry/src/wkb_factory.rs#L69-L83

c/sedona-geos/src/geos_to_wkb.rs
petern48: this var was removed

c/sedona-geos/src/geos_to_wkb.rs
petern48: yes! very clean now.

PR #478, opened 2026-01-04
chore(rust/sedona-functions): Make ST_Transform not accept geography

yutannihilation:
PostGIS's `ST_Transform()` doesn't take GEOGRAPHY. So, probably `ST_Transform` on GEOGRAPHY doesn't make sense...?

https://postgis.net/docs/en/ST_Transform.html

(Honestly, I don't understand how GEOGRAPHY's CRS actually matters, so I might be wrong)

### DIFF
--- a/c/sedona-proj/src/st_transform.rs
+++ b/c/sedona-proj/src/st_transform.rs
@@ -159,7 +159,7 @@ impl SedonaScalarKernel for STTransform {
     ) -> Result<Option<SedonaType>> {
         let matcher = ArgMatcher::new(
             vec![
-                ArgMatcher::is_geometry_or_geography(),
+                ArgMatcher::is_geometry(),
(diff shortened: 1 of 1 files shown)


yutannihilation:
> The CRS of a geography can be any lon/lat CRS

Thanks for clarifying. I mean, it definitely can be in theory, but I was wondering if the difference of CRSs is actually handled correctly by the underlying libraries. For example, it seems PostGIS's GEOGRAPHY assumes WGS84, and probably WGS84 is the only CRS that it can handle.

> In order to load geometry data into a geography table, the geometry first needs to be projected into EPSG:4326 (longitude/latitude), then it needs to be changed into geography.
https://postgis.net/workshops/en/postgis-intro/geography.html

Anyway, probably this is a broader topic, so I don't want to discuss the details here. I'll study more about how other DBs handle GEOGRAPHY.

yutannihilation:
Thanks for sharing! Looks very useful.

PR #479, opened 2026-01-04
feat(rust/sedona): Add Azure Blob Storage support for GeoParquet

isaaccorley:
Adds support for reading GeoParquet files from Azure Blob Storage.

## Changes

- New `azure` feature flag in `rust/sedona/Cargo.toml` enabling `object_store/azure`
- `AzureOptions` struct supporting common auth methods:
  - `account_name`, `sas_token`, `access_key`
  - `bearer_token`, `client_id`, `client_secret`, `tenant_id`, `authority_id`
- URL scheme support for `az://`, `abfs://`, `abfss://`
- Fixed GeoParquet metadata parsing for files missing `geometry_types` field (e.g., MS Building Footprints)
- Fixed URL extension detection to strip query params before checking file type

## Motivation

Wanted to query Microsoft Planetary Computer datasets (MS Building Footprints, etc.) directly from SedonaDB. These are hosted on Azure Blob Storage and use SAS token auth.

## Usage

```python
import sedonadb

sd = sedonadb.connect()
df = sd.read_parquet(
    "abfss://container@account.dfs.core.windows.net/path/",
    options={
        'azure.account_name': 'blobstorage',
        'azure.sas_token': 'sv=2023-01-03&st=...'
    }
)
```

## Testing

- Ran `pre-commit run --all-files` 
- `cargo clippy --workspace --all-targets --all-features -- -Dwarnings`
- `cargo test -p sedona -p sedona-geoparquet --all-features` (86 tests pass)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -262,9 +262,9 @@ checksum = "7c02d123df017efcdfbd739ef81735b36c5ba83ec3c59c80a9d7ecc718f92e50"
 
 [[package]]
 name = "arrow"
-version = "57.0.0"
+version = "57.1.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "4df8bb5b0bd64c0b9bc61317fcc480bad0f00e56d3bc32c69a4c8dada4786bae"
+checksum = "cb372a7cbcac02a35d3fb7b3fc1f969ec078e871f9bb899bf00a2e1809bec8a3"
 dependencies = [
  "arrow-arith",
  "arrow-array",
@@ -283,9 +283,9 @@ dependencies = [
 
 [[package]]
 name = "arrow-arith"
-version = "57.0.0"
+version = "57.1.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "a1a640186d3bd30a24cb42264c2dafb30e236a6f50d510e56d40b708c9582491"
+checksum = "0f377dcd19e440174596d83deb49cd724886d91060c07fec4f67014ef9d54049"
 dependencies = [
  "arrow-array",
  "arrow-buffer",
@@ -297,9 +297,9 @@ dependencies = [
 
 [[package]]
 name = "arrow-array"
-version = "57.0.0"
+version = "57.1.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "219fe420e6800979744c8393b687afb0252b3f8a89b91027d27887b72aa36d31"
+checksum = "a23eaff85a44e9fa914660fb0d0bb00b79c4a3d888b5334adb3ea4330c84f002"
 dependencies = [
  "ahash",
  "arrow-buffer",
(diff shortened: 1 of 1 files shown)


rust/sedona-geoparquet/src/metadata.rs
```diff
@@ -329,6 +329,7 @@ pub struct GeoParquetColumnMetadata {
     /// and multipolygons, it is not sufficient to specify `["MultiPolygon"]`, but it is expected
     /// to specify `["Polygon", "MultiPolygon"]`. Or if having 3D points, it is not sufficient to
     /// specify `["Point"]`, but it is expected to list `["Point Z"]`.
+    #[serde(default)]
```
paleolimbot, line 336: I'm guessing the planetary computer files are missing the `geometry_types` key? (Probably add a comment here explaining that, because this is technically out-of-spec)

rust/sedona-geoparquet/src/provider.rs
```diff
@@ -139,6 +136,49 @@ impl GeoParquetReadOptions<'_> {
...
+                            "Unknown Azure option '{}'. Did you mean: {}?",
+                            key,
+                            close_matches.join(", ")
+                        ));
+                    } else {
+                        return Err(format!(
+                            "Unknown Azure option '{}'. Valid options are: {}",
+                            key,
+                            common_azure_options.join(", ")
+                        ));
```
paleolimbot, line 179: Can you add a test for this like this one?

https://github.com/apache/sedona-db/blob/a43c9fc62365aab10d3c639dc72c59df01802c8f/python/sedonadb/tests/test_context.py#L116-L133

rust/sedona/src/object_storage.rs
```diff
@@ -503,6 +585,121 @@ impl ConfigExtension for GcpOptions {
     const PREFIX: &'static str = "gcp";
 }
 
+/// This struct encapsulates Azure options one uses when setting up object storage.
+#[cfg(feature = "azure")]
+#[derive(Default, Debug, Clone)]
+pub struct AzureOptions {
```
paleolimbot, line 591: Can you add a test for azure storage mirroring the ones for gcp and aws?

https://github.com/apache/sedona-db/blob/a43c9fc62365aab10d3c639dc72c59df01802c8f/rust/sedona/src/object_storage.rs#L632-L680

paleolimbot (commented):
Thank you! I have a few suggestions for tests but this is exciting!

rust/sedona-geoparquet/src/metadata.rs
isaaccorley: Correct they are missing the column which doesn't really give us a performance gain here but we should at least be able to read it so a user doesn't need to switch to a different library, if that makes sense

paleolimbot: I think it's pretty unequivocal that if the GeoParquet JSON metadata is missing the `geometry_types` key that we have no information about the geometry types (I never understood why we put that in the spec in the first place 🙂 )

isaaccorley:
Addressed all review feedback in c70e419:

1. **metadata.rs**: Added docstring explaining `#[serde(default)]` for out-of-spec GeoParquet files missing `geometry_types`
2. **test_context.py**: Added `test_read_parquet_invalid_azure_option()` mirroring the AWS test
3. **object_storage.rs**: Added `azure_object_store_builder` test mirroring GCP/AWS tests

All Rust tests pass locally. Python venv was broken so couldn't run Python tests locally, but the test follows the exact same pattern as the existing AWS test.

Issue #487, opened 2026-01-05
rust: Add item-level CRS wrappers for scalar functions


paleolimbot:
In #410 we implemented item-level CRS support, including a wrapper kernel that could turn any function into one that supported various combinations of item-level CRS. We can now use this to wrap existing scalar functions where this makes sense (most of them!). The `st_xxx_udf()` functions in each file implementing a function is probably the best place to do this (and each function should have a basic test to ensure the wrapper is doing what it says it will!).

PR #492, opened 2026-01-06
feat(rust/sedona-functions,rust/sedona-geo): Add item CRS implementations for existing scalar kernels that can be trivially wrapped

paleolimbot:
This PR is the first half of #487 (the second half will apply this to the scalar kernels in `c/`). This PR applies the `ItemCrsKernel` wrapper to all functions to which it can be trivially applied. "Trivial" here means that the function can be applied to the raw geometry/geography values as long as all CRSes match (and that if there are ItemCrs inputs, geometry outputs are also ItemCrs).

```python
import pandas as pd
import sedona.db

sd = sedona.db.connect()

df = pd.DataFrame({"wkt": ["POINT (0 1)", "POINT (2 3)"], "srid": [4326, 3857]})
sd.create_data_frame(df).to_view("foofy")
sd.sql("SELECT ST_Translate(ST_SetSRID(ST_GeomFromWKT(wkt), srid), 5, 10) as g FROM foofy").show()
#> ┌─────────────────────────────────────┐
#> │                  g                  │
#> │                struct               │
#> ╞═════════════════════════════════════╡
#> │ {item: POINT(5 11), crs: OGC:CRS84} │
#> ├╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌┤
#> │ {item: POINT(7 13), crs: EPSG:3857} │
#> └─────────────────────────────────────┘
```

### DIFF
--- a/rust/sedona-expr/src/item_crs.rs
+++ b/rust/sedona-expr/src/item_crs.rs
@@ -75,7 +75,7 @@ impl ItemCrsKernel {
     pub fn wrap_impl(inner: impl IntoScalarKernelRefs) -> Vec<ScalarKernelRef> {
         let kernels = inner.into_scalar_kernel_refs();
 
-        let mut out = Vec::new();
+        let mut out = Vec::with_capacity(kernels.len() * 2);
 
         // Add ItemCrsKernels first (so they will be resolved last)
         for inner_kernel in &kernels {


rust/sedona-expr/src/item_crs.rs
```diff
@@ -58,9 +58,36 @@ pub struct ItemCrsKernel {
...
+
+    /// Wrap a vector of kernels by appending all ItemCrs versions followed by
+    /// the contents of inner
+    ///
+    /// This is the recommended way to add kernels when all of them should support
+    /// ItemCrs inputs.
+    pub fn wrap_impl(inner: impl IntoScalarKernelRefs) -> Vec<ScalarKernelRef> {
+        let kernels = inner.into_scalar_kernel_refs();
+
+        let mut out = Vec::new();
```
Kontinuation:
```suggestion
        let mut out = Vec::with_capacity(kernels.len() * 2);
```

Kontinuation (approved):

PR #493, opened 2026-01-06
perf: Write wkb directly from geos for st_boundary, centroid, concavehull, convexhull, simplify, snap

petern48:
This is a follow up to https://github.com/apache/sedona-db/pull/476, which added a new `write_geos_geometry()` method that's faster than calling `.to_wkb()` and `write_all()`. This PR replaces more instances in the following functions and presents there performance improvements. This is about half of the rest of the functions. I'll do one more PR for the other half because it would be a lot to put all the benchmarks in one description.

Perf benchmarks:
ST_Boundary: (-42% and -72% reduction)
image

ST_Snap (-13.6% and -7.9% reduction)
image

ST_Simplify (-11.6% and -3% reduction)
image

ST_ConvexHull (-7.6% reduction, other one was taking too long)
image

ST_ConcaveHull (-5.5% reduction, other took too long)
image

ST_Centroid (-3% reduction). Improvement is small because it's always writing points which is cheap
image

### DIFF
--- a/python/sedonadb/tests/functions/test_functions.py
+++ b/python/sedonadb/tests/functions/test_functions.py
@@ -3119,6 +3119,9 @@ def test_st_simplifypreservetopology(eng, geom, tolerance, expected):
     ],
 )
 def test_st_snap(eng, input, reference, tolerance, expected):
+    # PostGIS incorrectly returns LINESTRING EMPTY here, so we skip this case for PostGIS.
+    if eng == PostGIS and input == "POLYGON EMPTY":
+        pytest.skip("PostGIS's result for POLYGON EMPTY is incorrect")
     eng = eng.create_or_skip()
     eng.assert_query_result(
         f"SELECT ST_Snap({geom_or_null(input)}, {geom_or_null(reference)}, {val_or_null(tolerance)})",


petern48:
Seems like there's an upstream bug, in `geos` where `.simplify()` adds a Z dimension when processing `POINT EMPTY` or `LINESTRING EMPTY`. I filed a ticket (https://github.com/libgeos/geos/issues/1356) in the original geos repo. This was causing [rust](https://github.com/apache/sedona-db/actions/runs/20763714948/job/59624864044?pr=493) and [Python tests](https://github.com/apache/sedona-db/actions/runs/20763714895/job/59624864115?pr=493) to fail now that write_geos_geometry() actually writes the Z dimension, unlike before.

<details>
<summary>
Debug output:
</summary>

```
input geometry: "LINESTRING EMPTY"
initial_type LineString. post_geometry_type LineString
raw geometry after simplify(): "LINESTRING Z EMPTY"
geometry after post-processing: LINESTRING Z EMPTY
```

```
input geometry: "POINT EMPTY"
initial_type Point. post_geometry_type Point
raw geometry after simplify(): "POINT Z EMPTY"
geometry after post-processing: POINT Z EMPTY
```
</details>

There was also a [test failure](https://github.com/apache/sedona-db/actions/runs/20763714948/job/59624864044?pr=493) in the Rust ST_Snap tests. I see it has the same problem of adding Z dimensions to `POINT EMPTY` and `LINESTRING EMPTY`.

For both functions, I've added extra logic to handle empty geometries separately and return the input value instead of using the simplify() or snap() results.

Ran benchmarks again, and see no noticeable performance degradations. So the benches in the PR description still stand.

PR #495, opened 2026-01-06
chore(r): Add air format + lintr checks to R package

paleolimbot:
There are a few forthcoming PRs that will involve a higher volume of (possibly generated) R code, so I'd like to set up some linters and formatters to help.

### DIFF
--- a/dev/release/rat_exclude_files.txt
+++ b/dev/release/rat_exclude_files.txt
@@ -21,7 +21,6 @@ c/sedona-tg/src/tg/*
 Cargo.lock
 ci/scripts/windows/Cargo.lock
 dev/release/rat_exclude_files.txt
-r/sedonadb/.lintr
 r/sedonadb/.Rbuildignore
 r/sedonadb/DESCRIPTION
 r/sedonadb/NAMESPACE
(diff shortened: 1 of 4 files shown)


r/sedonadb/R/000-wrappers.R
eitsupi, line 1: I think this file should be excluded.
(IIUC, air exclude this file by default, but lintr does not.)

eitsupi: This file may useful.
https://github.com/pola-rs/r-polars/blob/154d8d28877a6492241e59e0afe32f1d55813935/.lintr.R

I recommend to use `.lintr.R` instead of `.lintr`

eitsupi: Ah, now I understand why air didn't exclude this file.
Air sees the "Generated by" comment but assumes it's probably on the first line.

paleolimbot: Thanks! I changed this to use .lintr.R instead for the day we need more complex rules 🚀 

We have to modify the R wrappers anyway (because of licenses and because of the two blank lines at the end of the file), and we need a shell script anyway (because of the license files and pre-commit), so it's just as easy to keep it in for now. It's a good tip and we can circle back if this gets in the way (I think I can move the generated-by comment to the first line without the license checker getting mad in a pinch).

Issue #490, opened 2026-01-05
rust/sedona-functions: Implement ST_AsEWKB and ST_GeomFromEWKB


paleolimbot:
In #410 we implemented ItmCrs types that can handle the SRID provided by the EWKB; however, we didn't implement an EWKB parser that can actually persist this value (or export the values in such a way that the EWKB can be read by something else like PostGIS.

PR #498, opened 2026-01-07
feat(rust/sedona-functions): Implement ST_GeomFromEWKT

yutannihilation:
Now that we have item-level CRS, I think `ST_GeomFromEWKT` can be implemented. Unlike `ST_GeomFromEWKB` (https://github.com/apache/sedona-db/issues/490), if I understand correctly, EWKT doesn't need a dedicated parser except for spliting `SRID=...;` and pass the rest to the existing WKT parser.

It seems to work. I think I still don't understand how the item-level CRS should be handled, so this pull request takes some more time to finish.

```
> select st_geomfromewkt('SRID=3857;POINT (1 1)');
┌────────────────────────────────────────────────┐
│ st_geomfromewkt(Utf8("SRID=3857;POINT (1 1)")) │
│                     struct                     │
╞════════════════════════════════════════════════╡
│ {item: POINT(1 1), crs: EPSG:3857}             │
└────────────────────────────────────────────────┘
```

```
> select st_geomfromewkt('POINT (1 1)');
┌──────────────────────────────────────┐
│ st_geomfromewkt(Utf8("POINT (1 1)")) │
│                struct                │
╞══════════════════════════════════════╡
│ {item: POINT(1 1), crs: }            │
└──────────────────────────────────────┘
```

### DIFF
--- a/python/sedonadb/tests/functions/test_functions.py
+++ b/python/sedonadb/tests/functions/test_functions.py
@@ -1301,9 +1301,9 @@ def test_st_geomfromtext_with_srid(eng, wkt, srid, expected):
     ("ewkt", "expected", "expected_srid"),
     [
         (None, None, None),
-        ("SRID=4326;POINT (0 0)", "POINT (0 0)", 4326),
-        ("SRID=3857;POINT (0 0)", "POINT (0 0)", 3857),
-        ("SRID=0;POINT (0 0)", "POINT (0 0)", 0),
+        ("SRID=4326;POINT (0 0)", "POINT(0 0)", 4326),
+        ("SRID=3857;POINT (0 0)", "POINT(0 0)", 3857),
+        ("SRID=0;POINT (0 0)", "POINT(0 0)", 0),
     ],
 )
 def test_st_geomfromewkt(eng, ewkt, expected, expected_srid):


rust/sedona-functions/src/st_geomfromwkt.rs
```diff
@@ -147,8 +149,139 @@ fn invoke_scalar(wkt_bytes: &str, builder: &mut BinaryBuilder) -> Result<()> {
...
+        vec![Arc::new(STGeoFromEWKT {})],
+        Volatility::Immutable,
+        Some(doc),
+    )
+}
+
+#[derive(Debug)]
+struct STGeoFromEWKT {
+    // TODO
+    // engine: Option<Arc<dyn CrsEngine + Send + Sync>>,
```
paleolimbot: Feel free to leave this out for now...the CRS engine variants of some of the other functions aren't currently used with an engine and we probably need a more reliable way to inject support for validation.

rust/sedona-functions/src/st_geomfromwkt.rs
```diff
@@ -147,8 +149,139 @@ fn invoke_scalar(wkt_bytes: &str, builder: &mut BinaryBuilder) -> Result<()> {
...
+
+fn parse_maybe_srid(maybe_srid: &str) -> Option<String> {
+    if !maybe_srid.starts_with("SRID=") {
+        return None;
+    }
+    let srid_str = &maybe_srid[5..];
+    let srid: u16 = match srid_str.parse() {
+        Ok(srid) => srid,
+        Err(_) => return None,
+    };
```
paleolimbot: Can this be a U32?

rust/sedona-functions/src/st_geomfromwkt.rs
```diff
@@ -233,6 +366,64 @@ mod tests {
...
+            DataType::Utf8View => arrow_array::create_array!(
+                Utf8View,
+                [
+                    Some("SRID=4326;POINT (1 2)"),
+                    Some("SRID=3857;POINT (1 2)"),
+                    None,
+                    Some("POINT (3 4)"),
+                    Some("SRID=0;POINT (5 6)")
+                ]
+            ) as ArrayRef,
```
paleolimbot: I don't think you need the repeated case here (I'm pretty sure `tester.invoke_array()` will do that cast for you).

paleolimbot (commented):
Only minor nits from me...this looks great!

> I think I still don't understand how the item-level CRS should be handled

Me too! Supporting both type-level and item-level CRS is new territory and feel free to flag anything that doesn't make sense.

rust/sedona-functions/src/st_geomfromwkt.rs
yutannihilation: Oh, thanks. Then, I remove this. I noticed engine is currently unused, but wondered if this is somehow useful for testing.

rust/sedona-functions/src/st_geomfromwkt.rs
```diff
@@ -147,8 +149,132 @@ fn invoke_scalar(wkt_bytes: &str, builder: &mut BinaryBuilder) -> Result<()> {
...
+    let srid_str = &maybe_srid[5..];
+    let auth_code = match srid_str.parse::<u32>() {
+        Ok(0) => return None,
+        Ok(4326) => "OGC:CRS84".to_string(),
+        Ok(srid) => format!("EPSG:{srid}"),
+        Err(_) => return None,
+    };
+
+    // TODO: the CRS should be validated
+    // validate_crs(&auth_code, maybe_engine)?;
```
paleolimbot:
```suggestion
    // CRS could be validated here
    // https://github.com/apache/sedona-db/issues/501
```

yutannihilation: I see. Thanks!

yutannihilation:
I kept this pull request draft as I was wondering if I should add Python tests. However, since the output is different from PostGIS, probably it's not worth...? If I should add some, I can do.

yutannihilation:
Thanks, I added some tests, but I found `ST_AsText()` doesn't work on per-item-CRS geometry yet. So, I think testing only the SRID is enough for now.

`> SELECT ST_AsText(ST_GeomFromEWKT('POINT(0 0)'));`
`This feature is not implemented: st_astext([Arrow(Struct([Field { data_type: Binary, nullable: true, metadata: {"ARROW:extension:name": "geoarrow.wkb", "ARROW:extension:metadata": "{}"} }, Field { name: "crs", data_type: Utf8View, nullable: true }]))]): No kernel matching arguments`

python/sedonadb/tests/functions/test_functions.py
```diff
@@ -1296,6 +1296,30 @@ def test_st_geomfromtext_with_srid(eng, wkt, srid, expected):
     )
 
 
+@pytest.mark.parametrize("eng", [SedonaDB, PostGIS])
+@pytest.mark.parametrize(
+    ("ewkt", "expected", "expected_srid"),
+    [
+        ("SRID=4326;POINT (0 0)", "POINT (0 0)", 4326),
```
```suggestion
        (None, None, None),
        ("SRID=4326;POINT (0 0)", "POINT (0 0)", 4326),
```

I think this should work (nulls propagate)

python/sedonadb/tests/functions/test_functions.py
```diff
@@ -1296,6 +1296,30 @@ def test_st_geomfromtext_with_srid(eng, wkt, srid, expected):
...
+)
+def test_st_geomfromewkt(eng, ewkt, expected, expected_srid):
+    if ewkt is not None:
+        ewkt = f"'{ewkt}'"
+    eng = eng.create_or_skip()
+    # TODO: currently, a per-item-CRS geometry cannot be cast to string
+    # eng.assert_query_result(
+    #     f"SELECT ST_GeomFromEWKT({val_or_null(ewkt)})",
+    #     expected,
+    # )
```
paleolimbot: If you rebase or merge in the main branch I think this will work:

```suggestion
    eng.assert_query_result(
        f"SELECT ST_AsWKT(ST_GeomFromEWKT({val_or_null(ewkt)}))",
        expected,
    )
```

https://github.com/apache/sedona-db/blob/af5e94799418470fec391657666ae0daad68ff44/rust/sedona-functions/src/st_astext.rs#L38

Also feel free to remove this bit and I can follow up when I do ST_GeomFromEKWB() (FWIW I prefer not to merge TODOs that don't have an tracker link associated with them, since TODOs in the code base are much harder to prioritize than issues.)

yutannihilation: Fair enough! Then, I remove this for now as `ST_AsWKT()` doesn't work either.

yutannihilation:
Sorry, I just didn't notice `ST_AsText()` needs https://github.com/apache/sedona-db/pull/492 to work. Added the test again.

PR #503, opened 2026-01-10
feat(rust/sedona-geos): Implement ST_LineMerge()

yutannihilation:
This pull request implements `ST_LineMerge`.

https://postgis.net/docs/en/ST_LineMerge.html

This is mostly straightforward using GEOS's LineMerge via the geos crate. One thing to note is that, the direction-aware version (`.line_merge_directed()`) requires GEOS >= v3.11, which was [released 3.5 years ago](https://github.com/libgeos/geos/releases/tag/3.11.0). I don't think this is a difficult requirement, but I'm not really sure.

https://github.com/georust/geos/blob/47afbad2483e489911ddb456417808340e9342c3/src/geometry.rs#L2796-L2802

```
> SELECT ST_LineMerge(ST_GeomFromWKT('MULTILINESTRING ((0 0, 1 0), (1 0, 1 1))'));
┌────────────────────────────────────────────────────────────────────────────────┐
│ st_linemerge(st_geomfromwkt(Utf8("MULTILINESTRING ((0 0, 1 0), (1 0, 1 1))"))) │
│                                    geometry                                    │
╞════════════════════════════════════════════════════════════════════════════════╡
│ LINESTRING(0 0,1 0,1 1)                                                        │
└────────────────────────────────────────────────────────────────────────────────┘
```

### DIFF
--- a/python/sedonadb/tests/functions/test_functions.py
+++ b/python/sedonadb/tests/functions/test_functions.py
@@ -1814,7 +1814,7 @@ def test_st_isring_non_linestring_error(eng, geom):
             "MULTILINESTRING ((0 0, 1 0), (8 8, 9 9))",
         ),
         # empty cases
-        # ("POINT EMPTY", "POINT EMPTY"), # PostGIS returns POINT (nan, nan)
+        ("POINT EMPTY", "POINT (nan nan)"),
         ("LINESTRING EMPTY", "LINESTRING EMPTY"),
         ("POLYGON EMPTY", "POLYGON EMPTY"),
         ("MULTIPOINT EMPTY", "MULTIPOINT EMPTY"),


c/sedona-geos/src/st_line_merge.rs
```diff
@@ -0,0 +1,177 @@
...
+    let geom =
+        result.map_err(|e| DataFusionError::Execution(format!("Failed to merge lines: {e}")))?;
+
+    let wkb = geom
+        .to_wkb()
+        .map_err(|e| DataFusionError::Execution(format!("Failed to convert result to WKB: {e}")))?;
+
+    writer
+        .write_all(wkb.as_ref())
+        .map_err(|e| DataFusionError::Execution(format!("Failed to write result WKB: {e}")))?;
```
petern48:
```suggestion
    write_geos_geometry(&geom, writer)?;
```

https://github.com/apache/sedona-db/pull/476 recently introduced this helper method that is faster than doing these two calls

c/sedona-geos/src/st_line_merge.rs
```diff
@@ -0,0 +1,177 @@
...
+
+use arrow_array::builder::BinaryBuilder;
+use datafusion_common::{error::Result, DataFusionError, ScalarValue};
+use datafusion_expr::ColumnarValue;
+use geos::Geom;
+use sedona_expr::scalar_udf::{ScalarKernelRef, SedonaScalarKernel};
+use sedona_geometry::wkb_factory::WKB_MIN_PROBABLE_BYTES;
+use sedona_schema::{datatypes::WKB_GEOMETRY, matchers::ArgMatcher};
+
+use crate::executor::GeosExecutor;
```
petern48, line 28:
```suggestion
use crate::executor::GeosExecutor;
use crate::geos_to_wkb::write_geos_geometry;
```
Here's the import for my above suggestion.

python/sedonadb/tests/functions/test_functions.py
```diff
@@ -1797,6 +1797,58 @@ def test_st_isring_non_linestring_error(eng, geom):
...
+        ),  # opposite direction
+        (
+            "MULTILINESTRING ((0 0, 1 0), (8 8, 9 9))",
+            "MULTILINESTRING ((0 0, 1 0), (8 8, 9 9))",
+        ),
+        # Note that the behaviour on non-multilinestring geometry is not documented.
+        # But, we test such cases here as well to detect if there's any difference.
+        ("POINT (0 0)", "GEOMETRYCOLLECTION EMPTY"),
+        ("LINESTRING (0 0, 1 0)", "LINESTRING (0 0, 1 0)"),
+        ("POLYGON ((0 0, 0 1, 1 0, 0 0))", "LINESTRING (0 0, 0 1, 1 0, 0 0)"),
```
petern48: interesting. The [docs](https://postgis.net/docs/ST_LineMerge.html) say "Other geometry types return an empty GeometryCollection" in the note. Though that doesn't seem to be the case for POLYGON or LINESTRING input 🤷.

Could we test some empty geometries, for both of these tests? Those can often catch weird edge cases.

c/sedona-geos/src/st_line_merge.rs
yutannihilation: Oh, thanks! I missed it.

python/sedonadb/tests/functions/test_functions.py
yutannihilation: Thanks for catching. I simply didn't notice the note...

I'll add empty cases.

yutannihilation:
> Did you want to add a Python integration test for this here?

Do you mean benchmark tests? I just forgot to do so, but now I don't immediately see how to add tests as it probably requires specific inputs of touching line segments.

python/sedonadb/tests/functions/test_functions.py
```diff
@@ -1803,14 +1803,24 @@ def test_st_isring_non_linestring_error(eng, geom):
...
             "LINESTRING (0 0, 1 0, 1 1)",
-        ),  # opposite direction
+        ),
+        # non-touching
         (
             "MULTILINESTRING ((0 0, 1 0), (8 8, 9 9))",
             "MULTILINESTRING ((0 0, 1 0), (8 8, 9 9))",
         ),
+        # empty cases
+        # ("POINT EMPTY", "POINT EMPTY"), # PostGIS returns POINT (nan, nan)
```
yutannihilation: While I actually get this error,

```
E                     [('POINT EMPTY',)]
E                   Got:
E                     [('POINT (nan nan)',)]
```

it seems this is not PostGIS's fault. Something seems wrong in conversion between PostGIS and Python.

```
postgres=# SELECT ST_AsText(ST_LineMerge(ST_GeomFromText('POINT EMPTY')));
  st_astext  
-------------
 POINT EMPTY
(1 row)
```

yutannihilation: I found `ST_AsText()` is needed here!

petern48: This is known. This is just an issue of how geoarrow renders the WKB as a string since `POINT EMPTY` and `POINT (nan nan)` are represented the same as WKB. In most test functions, we just specify expected as `POINT (nan nan)` instead. If you `ctrl + f` for `POINT (nan nan)`, you'll see a lot. e.g. here:

https://github.com/apache/sedona-db/blob/edcb909413328ebc028703b3cc7fe72e4907ec02/python/sedonadb/tests/functions/test_functions.py#L493-L497

I'm okay with your approach too, though I personally just use `POINT (nan nan)`, so I don't have to deal with the annoyances of manually fixing the different spacing when doing string comparison.

petern48 (approved):
(I will leave this for @paleolimbot to clarify what he meant and answer this: https://github.com/apache/sedona-db/pull/503#issuecomment-3733650122, given there were python integration tests already).

python/sedonadb/tests/functions/test_functions.py
yutannihilation: Ah, thanks for the details. I think I saw some discussion about this somewhere, but didn't realize it relates here. Removed `ST_AsText()` now.

yutannihilation:
I see. Thanks for clarifying!

Issue #47, opened 2025-09-10
Enhancement: ST_TRANSFORM to support z


jesspav:
Some transforms change the z coordinate, for example with a conversion from feet into meters.

See example in: https://proj.org/en/stable/usage/transformation.html

We should add support for this in sedona-db.

paleolimbot:
Closed in #544.

PR #504, opened 2026-01-11
feat(rust/sedona-functions): Implement ST_Affine(), ST_Rotate(), and ST_Scale()

yutannihilation:
This pull request implements `ST_Affine()`, `ST_Rotate` and `ST_Scale` using [the glam crate](https://github.com/bitshifter/glam-rs). I chose glam because

- while affine transformation is probably not very hard to implement, I might make some silly mistakes
- glam is widely used
- glam doesn't require any additional dependency by default

but I don't have strong opinion. If it is desirable to implement from scratch, I can do so.

```
> SELECT ST_Affine(ST_GeomFromText('POLYGON ZM ((1 0 1 1, 1 1 1 1, 2 2 2 1, 1 0 1 1))'), 10, 0, 0, 0, 10, 0, 0, 0, 10, 0, 0, 0);
┌──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┐
│ st_affine(st_geomfromtext(Utf8("POLYGON ZM ((1 0 1 1, 1 1 1 1, 2 2 2 1, 1 0 1 1))")),Int64(10),Int64(0),Int64(0),Int64(0),Int64(10),Int64(0),Int64(0),Int64(0),Int64(10),Int64(0),Int64(0),Int64(0)) │
│                                                                                               geometry                                                                                               │
╞══════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════╡
│ POLYGON ZM((10 0 10 1,10 10 10 1,20 20 20 1,10 0 10 1))                                                                                                                                              │
└──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┘
```

- [x] `ST_Affine()`
- [x] `ST_Rotate()`
  - [x] angle
  - [ ] ~~angle and origin~~ (I'll address in another pull request)
- [x] `ST_Scale()`
  - [x] scale
  - [ ] ~~scale and origin~~ (I'll address in another pull request)

### DIFF
--- a/rust/sedona-functions/src/st_rotate.rs
+++ b/rust/sedona-functions/src/st_rotate.rs
@@ -148,6 +148,8 @@ impl SedonaScalarKernel for STRotate {
 
 #[cfg(test)]
 mod tests {
+    use std::f64;
+
     use arrow_array::Array;
     use datafusion_expr::ScalarUDF;
     use rstest::rstest;
@@ -165,21 +167,31 @@ mod tests {
         assert!(st_rotate_udf.documentation().is_some());
 
         let st_rotate_x_udf: ScalarUDF = st_rotate_x_udf().into();
-        assert_eq!(st_rotate_x_udf.name(), "st_rotate_x");
+        assert_eq!(st_rotate_x_udf.name(), "st_rotatex");
         assert!(st_rotate_x_udf.documentation().is_some());
 
         let st_rotate_y_udf: ScalarUDF = st_rotate_y_udf().into();
-        assert_eq!(st_rotate_y_udf.name(), "st_rotate_y");
+        assert_eq!(st_rotate_y_udf.name(), "st_rotatey");
         assert!(st_rotate_y_udf.documentation().is_some());
     }
 
     #[rstest]
     fn udf(#[values(WKB_GEOMETRY, WKB_VIEW_GEOMETRY)] sedona_type: SedonaType) {
-        let tester = ScalarUdfTester::new(
+        let tester_z = ScalarUdfTester::new(
             st_rotate_udf().into(),
             vec![sedona_type.clone(), SedonaType::Arrow(DataType::Float64)],
         );
-        tester.assert_return_type(WKB_GEOMETRY);
+        let tester_x = ScalarUdfTester::new(
+            st_rotate_x_udf().into(),
+            vec![sedona_type.clone(), SedonaType::Arrow(DataType::Float64)],
+        );
+        let tester_y = ScalarUdfTester::new(
+            st_rotate_y_udf().into(),
+            vec![sedona_type.clone(), SedonaType::Arrow(DataType::Float64)],
+        );
+        tester_z.assert_return_type(WKB_GEOMETRY);
+        tester_x.assert_return_type(WKB_GEOMETRY);
+        tester_y.assert_return_type(WKB_GEOMETRY);
 
         let points = create_array(
             &[
@@ -188,6 +200,7 @@ mod tests {
                 Some("POINT M EMPTY"),
                 Some("POINT (1 2)"),
                 Some("POINT M (1 2 3)"),
+                Some("POINT Z (1 2 3)"),
             ],
             &sedona_type,
         );
@@ -198,14 +211,25 @@ mod tests {
                 Some("POINT M EMPTY"),
                 Some("POINT (1 2)"),
                 Some("POINT M (1 2 3)"),
+                Some("POINT Z (1 2 3)"),
             ],
             &WKB_GEOMETRY,
         );
 
-        let result_identity = tester
-            .invoke_arrays(prepare_args(points.clone(), &[Some(0.0_f64.to_radians())]))
+        let result_identity_z = tester_z
+            .invoke_arrays(prepare_args(points.clone(), &[Some(0.0_f64)]))
             .unwrap();
-        assert_array_equal(&result_identity, &expected_identity);
+        assert_array_equal(&result_identity_z, &expected_identity);
+
+        let result_identity_x = tester_x
+            .invoke_arrays(prepare_args(points.clone(), &[Some(0.0_f64)]))
+            .unwrap();
+        assert_array_equal(&result_identity_x, &expected_identity);
+
+        let result_identity_y = tester_y
+            .invoke_arrays(prepare_args(points.clone(), &[Some(0.0_f64)]))
+            .unwrap();
+        assert_array_equal(&result_identity_y, &expected_identity);
 
         // Don't test the rotated results here since it's hard to match with the exact number.
     }


yutannihilation:
I was hoping to add the variants with `origin` of rotation or scaling, but, since this pull request is already a bit lengthy, I think it's probably better to address in a separate pull request.

rust/sedona-functions/src/st_affine.rs
```diff
@@ -0,0 +1,450 @@
...
+
+    fn invoke_batch(
+        &self,
+        arg_types: &[SedonaType],
+        args: &[ColumnarValue],
+    ) -> Result<ColumnarValue> {
+        let executor = WkbExecutor::new(arg_types, args);
+        let mut builder = BinaryBuilder::with_capacity(
+            executor.num_iterations(),
+            WKB_MIN_PROBABLE_BYTES * executor.num_iterations(),
```
paleolimbot, line 125: No need to do this here because I think we need a helper for it to be compact, but in this particular case we know the output will have the exact number of bytes as the input)

rust/sedona-functions/src/st_affine_helpers.rs
```diff
@@ -0,0 +1,531 @@
...
+            if pt.coord().is_some() {
+                write_wkb_point_header(writer, dims)
+                    .map_err(|e| DataFusionError::Execution(e.to_string()))?;
+                write_transformed_coord(writer, pt.coord().unwrap(), mat, dim)?;
+            } else {
+                write_wkb_empty_point(writer, dims)
+                    .map_err(|e| DataFusionError::Execution(e.to_string()))?;
+            }
+        }
+
```
paleolimbot: If this can be written as a `CrsTransform` we might be able re-use some existing infrastructure:

https://github.com/apache/sedona-db/blob/9bf02f474d475b61a444e561ec74df449413f46b/rust/sedona-functions/src/st_translate.rs#L123-L135

https://github.com/apache/sedona-db/blob/9bf02f474d475b61a444e561ec74df449413f46b/rust/sedona-functions/src/st_translate.rs#L106-L108

https://github.com/apache/sedona-db/blob/main/rust/sedona-geometry/src/transform.rs#L259-L263

There's a chance what you have here is faster, though (no `dyn` like our current `transform()`), and we haven't implemented non-XY support for the `CrsTransform` yet ( https://github.com/apache/sedona-db/issues/47 ).

No need to take on any of that here (I'll just add a note to that issue that we can update this section when we do get there).

rust/sedona-functions/src/st_affine_helpers.rs
```diff
@@ -0,0 +1,531 @@
...
+        .into_iter()
+        .try_for_each(|coord| write_transformed_coord(writer, coord, affine, dim))
+}
+
+fn write_transformed_coord<C>(
+    writer: &mut impl Write,
+    coord: C,
+    affine: &DAffine,
+    dim: &geo_traits::Dimensions,
+) -> Result<()>
```
paleolimbot: (Optional) It may be unfounded, but my intuition would be that `match (dim)` should be in `write_transformed_coords()` to avoid a `match()` for every coordinate. Perhaps the compiler does this anyway or the transform operation is sufficiently expensive that it doesn't matter 🙂

rust/sedona-functions/src/st_affine_helpers.rs
```diff
@@ -0,0 +1,531 @@
...
+    d: &'a PrimitiveArray<Float64Type>,
+    e: &'a PrimitiveArray<Float64Type>,
+    x_offset: &'a PrimitiveArray<Float64Type>,
+    y_offset: &'a PrimitiveArray<Float64Type>,
+}
+
+impl<'a> DAffine2Iterator<'a> {
+    pub(crate) fn new(array_args: &'a [Arc<dyn Array>]) -> Result<Self> {
+        if array_args.len() != 6 {
+            return internal_err!("Invalid number of arguments are passed");
```
paleolimbot:
```suggestion
            return sedona_internal_err!("Invalid number of arguments are passed");
```

rust/sedona-functions/src/st_affine_helpers.rs
```diff
@@ -0,0 +1,531 @@
...
+    i: &'a PrimitiveArray<Float64Type>,
+    x_offset: &'a PrimitiveArray<Float64Type>,
+    y_offset: &'a PrimitiveArray<Float64Type>,
+    z_offset: &'a PrimitiveArray<Float64Type>,
+}
+
+impl<'a> DAffine3Iterator<'a> {
+    pub(crate) fn new(array_args: &'a [Arc<dyn Array>]) -> Result<Self> {
+        if array_args.len() != 12 {
+            return internal_err!("Invalid number of arguments are passed");
```
paleolimbot:
```suggestion
            return sedona_internal_err!("Invalid number of arguments are passed");
```

rust/sedona-functions/src/st_affine_helpers.rs
```diff
@@ -0,0 +1,531 @@
...
+pub(crate) struct DAffine2ScaleIterator<'a> {
+    index: usize,
+    x_scale: &'a PrimitiveArray<Float64Type>,
+    y_scale: &'a PrimitiveArray<Float64Type>,
+}
+
+impl<'a> DAffine2ScaleIterator<'a> {
+    pub(crate) fn new(array_args: &'a [Arc<dyn Array>]) -> Result<Self> {
+        if array_args.len() != 2 {
+            return internal_err!("Invalid number of arguments are passed");
```
paleolimbot:
```suggestion
            return sedona_internal_err!("Invalid number of arguments are passed");
```

rust/sedona-functions/src/st_affine_helpers.rs
```diff
@@ -0,0 +1,531 @@
...
+
+pub(crate) struct DAffine2Iterator<'a> {
+    index: usize,
+    a: &'a PrimitiveArray<Float64Type>,
+    b: &'a PrimitiveArray<Float64Type>,
+    d: &'a PrimitiveArray<Float64Type>,
+    e: &'a PrimitiveArray<Float64Type>,
+    x_offset: &'a PrimitiveArray<Float64Type>,
+    y_offset: &'a PrimitiveArray<Float64Type>,
+}
```
paleolimbot, line 35: Do we need multiple versions of this (e.g., can the core logic be distilled to `struct NumericArgs { args: Vec<&'a PrimitiveArray<Float64Type>> }`? Or maybe the lifetimes make this difficult?).

My reading of these is that they don't handle nulls in any of the arguments. I am not sure if it's optimizing this but you could check ahead of time if the `null_count()` is >0 for any of the inputs and use the slower iterator if this is the case.

rust/sedona-functions/src/st_affine_helpers.rs
```diff
@@ -0,0 +1,531 @@
...
+            }
+            DAffineIterator::DAffine3Scale(daffine3_scale_iterator) => {
+                daffine3_scale_iterator.next().map(DAffine::DAffine3)
+            }
+            DAffineIterator::DAffineRotate(daffine_rotate_iterator) => {
+                daffine_rotate_iterator.next().map(DAffine::DAffine3)
+            }
+        }
+    }
+}
```
paleolimbot, line 532: This module would benefit from tests (e.g., for null item handling since that is non-trivial across 12 input arguments). One benefit of using the existing `transform()` would be that you don't have to write tests for it, and a benefit of consolidating iteration over `n` numeric arguments into a single struct is that you don't have to test it four times.

In particular, I'm not sure there is test coverage at a higher level for null propagation or the full matrix of geometry types by XY XYM XYZ XYZM including empties, which I think is a concern of the transformed writer here.

rust/sedona-functions/src/st_affine.rs
```diff
@@ -0,0 +1,450 @@
...
+        );
+
+        let array_args = args[1..]
+            .iter()
+            .map(|arg| {
+                arg.cast_to(&DataType::Float64, None)?
+                    .to_array(executor.num_iterations())
+            })
+            .collect::<Result<Vec<Arc<dyn Array>>>>()?;
```
paleolimbot, line 134: No need to do this now, but a future optimization would be to handle the case where all 9/12 input arguments are scalars. In general I haven't been able to notice a speed improvement for the case where one or two scalar numerics have to get expanded into arrays but I'm not sure that's the case with 12 input arguments 🙂

python/sedonadb/tests/functions/test_functions.py
```diff
@@ -189,6 +189,177 @@ def test_st_azimuth(eng, geom1, geom2, expected):
...
+        f"{val_or_null(d)}, {val_or_null(e)}, {val_or_null(f)}, "
+        f"{val_or_null(g)}, {val_or_null(h)}, {val_or_null(i)}, "
+        f"{val_or_null(xoff)}, {val_or_null(yoff)}, {val_or_null(zoff)})"
+    )
+    try:
+        eng.assert_query_result(query, expected)
+    except Exception as exc:
+        if isinstance(eng, PostGIS):
+            pytest.skip(f"PostGIS may not support 3D ST_Affine: {exc}")
+        raise
```
paleolimbot: Is this an issue with the version of PostGIS in our docker compose setup?

python/sedonadb/tests/functions/test_functions.py
```diff
@@ -189,6 +189,177 @@ def test_st_azimuth(eng, geom1, geom2, expected):
...
+        ("POINT (1 2)", 1.5707963267948966, -2.0, 1.0),
+        ("POINT (1 2)", 3.141592653589793, -1.0, -2.0),
+    ],
+)
+def test_st_rotate(eng, geom, angle, expected_x, expected_y):
+    eng = eng.create_or_skip()
+    x_query = f"SELECT ST_X(ST_Rotate({geom_or_null(geom)}, {val_or_null(angle)}))"
+    y_query = f"SELECT ST_Y(ST_Rotate({geom_or_null(geom)}, {val_or_null(angle)}))"
+    eng.assert_query_result(x_query, expected_x, numeric_epsilon=1e-12)
+    eng.assert_query_result(y_query, expected_y, numeric_epsilon=1e-12)
```
paleolimbot: There is also `wkt_precision` if you just want to check the result of `ST_Rotate()`:

https://github.com/apache/sedona-db/blob/9bf02f474d475b61a444e561ec74df449413f46b/python/sedonadb/python/sedonadb/testing.py#L186

paleolimbot (commented):
Thank you! This has to be the most arguments I have ever seen in a SQL function and managing all of them is a pain that you've nicely abstracted here.

> using [the glam crate](https://github.com/bitshifter/glam-rs).

I think this is OK. We sort of have this for 2D already but I'm not sure that will simplify this PR given the 3D support. If we do inline this we should do it in sedona-geometry and add the appropriate tests.

> I was hoping to add the variants with origin of rotation or scaling, but, since this pull request is already a bit lengthy, I think it's probably better to address in a separate pull request.

Great!

python/sedonadb/tests/functions/test_functions.py
yutannihilation: Ah, sorry, these test code is generated by Codex, and it added this kind of skips for unknown reason. I removed it, but it seems I forgot here...

rust/sedona-functions/src/st_affine.rs
yutannihilation: Ah, true. Thanks for catching.

rust/sedona-functions/src/st_affine.rs
yutannihilation: This part, I just followed what `ST_Translate` does. Do you mean this should be handled differently when all the input arguments are scalars? If all are scalars, I think I can create a `DAffine2` / `DAffine3` here immediately without expanding.

https://github.com/apache/sedona-db/blob/9bf02f474d475b61a444e561ec74df449413f46b/rust/sedona-functions/src/st_translate.rs#L92-L99

rust/sedona-functions/src/st_affine_helpers.rs
yutannihilation: Sounds good to me. I'll try.

(As translation can also be handled by affine transformation, I was wondering if ST_Transform can also be implemented using glam. But, I guess there will be not much difference)

rust/sedona-geometry/src/transform.rs
```diff
@@ -58,13 +58,27 @@ pub trait CrsEngine: Debug {
...
 pub trait CrsTransform: std::fmt::Debug {
     fn transform_coord(&self, coord: &mut (f64, f64)) -> Result<(), SedonaGeometryError>;
+
+    // CrsTransform can optionally handle 3D coordinates. If this method is not implemented,
+    // the Z coordinate is simply ignored.
+    fn transform_coord_3d(&self, coord: &mut (f64, f64, f64)) -> Result<(), SedonaGeometryError> {
+        let mut coord_2d = (coord.0, coord.1);
+        self.transform_coord(&mut coord_2d)?;
+        coord.0 = coord_2d.0;
+        coord.1 = coord_2d.1;
```
yutannihilation, line 68: I'm wondering if I should change the signature to this so that we don't need to copy the values here. But, as it would affect several other functions, I think I should not do this in this pull request anyway.

```rs
fn transform_coord(&self, x: &mut f64, y: &mut f64) -> Result<(), SedonaGeometryError>;
fn transform_coord_3d(&self, x: &mut f64, y: &mut f64, z: &mut f64) -> Result<(), SedonaGeometryError>;
```

yutannihilation:
I think I addressed most of the comments. Thanks for the advices!

rust/sedona-geometry/src/transform.rs
```diff
@@ -58,13 +58,27 @@ pub trait CrsEngine: Debug {
...
 
 /// A boxed trait object for dynamic dispatch of CRS transformations.
 impl CrsTransform for Box<dyn CrsTransform> {
     fn transform_coord(&self, coord: &mut (f64, f64)) -> Result<(), SedonaGeometryError> {
         self.as_ref().transform_coord(coord)
     }
+
+    fn transform_coord_3d(&self, coord: &mut (f64, f64, f64)) -> Result<(), SedonaGeometryError> {
+        self.as_ref().transform_coord_3d(coord)
+    }
```
paleolimbot, line 81: Thank you for this!

To ensure coverage in this file I think this could use a quick test of the non-default 3D transform case:

```rust
#[derive(Debug)]
    struct Mock3DTransform {}
    impl CrsTransform for Mock3DTransform {
        fn transform_coord_3d(&self, coord: &mut (f64, f64, f64)) -> Result<(), SedonaGeometryError> {
            coord.0 += 10.0;
            coord.1 += 20.0;
            coord.2 += 30.0;
            Ok(())
        }
    }

#[test]
    fn test_transform_dimensions_3d() {
         // like test_transform_dimensions but when there's a non-default 3D transform
    }
```

rust/sedona-geometry/src/transform.rs
paleolimbot: I think this is OK...copying small arguments is usually not faster and there are possibly lower hanging fruit to optimize (optimized transform of an array of coordinates, or possibly just non-aligned bytes -> bytes).

rust/sedona-functions/src/st_rotate.rs
```diff
@@ -0,0 +1,241 @@
...
+            ColumnarValue::Scalar(geom),
+            ColumnarValue::Scalar(ScalarValue::Float64(Some(0.0))),
+        ];
+
+        let result = tester.invoke(args).unwrap();
+        if let ColumnarValue::Scalar(scalar) = result {
+            tester.assert_scalar_result_equals(scalar, "POINT (1 2)");
+        } else {
+            panic!("Expected scalar result from item CRS rotate invoke");
+        }
```
paleolimbot:
```suggestion
        let result = tester.invoke_scalar_scalar("POINT (1 2)", 0.0);
        tester.assert_scalar_result_equals("POINT (1 2)");
```

rust/sedona-functions/src/st_rotate.rs
```diff
@@ -0,0 +1,241 @@
...
+        assert_eq!(st_rotate_udf.name(), "st_rotate");
+        assert!(st_rotate_udf.documentation().is_some());
+    }
+
+    #[rstest]
+    fn udf(#[values(WKB_GEOMETRY, WKB_VIEW_GEOMETRY)] sedona_type: SedonaType) {
+        let tester = ScalarUdfTester::new(
+            st_rotate_udf().into(),
+            vec![sedona_type.clone(), SedonaType::Arrow(DataType::Float64)],
+        );
```
paleolimbot, line 183: Maybe a quick check that `st_rotate_y()` and `st_rotate_x()` is also plugged in? It can be very short (just enough to check for crossed wires).

rust/sedona-functions/src/st_affine.rs
paleolimbot: My thinking was that all our existing functions (like ST_Translate()) have a small number of arguments but 12 arguments might be enough that we'd notice the copies. That's a very much future concern...implementing it like you did here I think is the right thing to do.

rust/sedona-functions/src/st_scale.rs
```diff
@@ -0,0 +1,359 @@
...
+            ColumnarValue::Scalar(ScalarValue::Float64(Some(2.0))),
+            ColumnarValue::Scalar(ScalarValue::Float64(Some(3.0))),
+        ];
+
+        let result = tester.invoke(args).unwrap();
+        if let ColumnarValue::Scalar(scalar) = result {
+            tester.assert_scalar_result_equals(scalar, "POINT (2 6)");
+        } else {
+            panic!("Expected scalar result from item CRS scale invoke");
+        }
```
paleolimbot:
```suggestion
        let result = tester.invoke_scalar_scalar_scalar("POINT (1 2)", 2, 3).unwrap();
        tester.assert_scalar_result_equals("POINT (2 6)"); 
```

rust/sedona-functions/src/st_affine_helpers.rs
paleolimbot: I think we can have fewer tests now that you've implemented the transform (thanks!), but I think we do need *something* here (or more tests in the Rust functions). (For example, you check every null by name individually and an LLM may come through and obliterate one of them in a refactor without a test failing that I can see).

paleolimbot (commented):
I tried to get through all of this before the end of the day but I may have missed some things. Thank you for these edits...I think it is very close!

rust/sedona-geometry/src/transform.rs
yutannihilation: Thanks, I added some tests here.

rust/sedona-functions/src/st_rotate.rs
yutannihilation: Thanks for catching, I added tests here, and some in Python tests as well.

rust/sedona-functions/src/st_affine.rs
yutannihilation: I see, thanks for the details. Then, I'll leave this for now.

paleolimbot (approved):
Thank you!

yutannihilation:
Thanks so much for reviewing!

PR #508, opened 2026-01-12
fix(rust/sedona-spatial-join): Fix several bugs related to KNN join

Kontinuation:
This patch fixes 3 bugs related to KNN join:

1. Panic when the number of partitions of joined relations are different.
2. Incorrect join result when the number of projected columns from the left and right side is different.
3. Incorrect K-nearest neighbor search result when query object is not a point and `include_tie_breakers` is enabled.

Rust tests were also added to make testing KNN joins easier.

### DIFF
--- a/rust/sedona-spatial-join/src/exec.rs
+++ b/rust/sedona-spatial-join/src/exec.rs
@@ -614,15 +614,15 @@ mod tests {
         prelude::{SessionConfig, SessionContext},
     };
     use datafusion_common::tree_node::{TreeNode, TreeNodeRecursion};
+    use datafusion_expr::ColumnarValue;
     use geo::{Distance, Euclidean};
     use geo_types::{Coord, Rect};
     use rstest::rstest;
-    use sedona_geo::to_geo::item_to_geometry;
+    use sedona_geo::to_geo::GeoTypesExecutor;
     use sedona_geometry::types::GeometryTypeId;
     use sedona_schema::datatypes::{SedonaType, WKB_GEOGRAPHY, WKB_GEOMETRY};
     use sedona_testing::datagen::RandomPartitionedDataBuilder;
     use tokio::sync::OnceCell;
-    use wkb::reader::read_wkb;
 
     use crate::register_spatial_join_optimizer;
     use sedona_common::{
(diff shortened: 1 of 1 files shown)


rust/sedona-spatial-join/src/exec.rs
```diff
@@ -1173,4 +1202,157 @@ mod tests {
...
+                let geoms_binary_view = geoms_col
+                    .as_any()
+                    .downcast_ref::<arrow_array::BinaryViewArray>();
+
+                if geoms_binary.is_none() && geoms_binary_view.is_none() {
+                    panic!(
+                        "Column 'geometry' should be Binary or BinaryView. Schema: {:?}",
+                        batch.schema()
+                    );
+                }
```
paleolimbot: I wonder if you could simplify this part with an executor:

```rust
let executor = GeoTypesExecutor::new(...);
let idx = geom_idx.iter();
executor.execute_wkb_void(|maybe_geom| { result.push((idx.next().unwrap().unwrap(), maybe_geom.unwrap())); })
```

rust/sedona-spatial-join/src/stream.rs
```diff
@@ -244,10 +244,10 @@ impl SpatialJoinStream {
         // Extract the necessary data first to avoid borrowing conflicts
         let (batch_opt, is_complete) = match &mut self.state {
             SpatialJoinStreamState::ProcessProbeBatch(iterator) => {
-                // For KNN joins, we swapped build/probe sides, so build_side should be Right
-                // For regular joins, build_side is Left
+                // For KNN joins, we may have swapped build/probe sides, so build_side might be Right;
+                // For regular joins, build_side is always Left.
                 let build_side = match &self.spatial_predicate {
-                    SpatialPredicate::KNearestNeighbors(_) => JoinSide::Right,
+                    SpatialPredicate::KNearestNeighbors(knn) => knn.probe_side.negate(),
```
paleolimbot, line 250: When debugging my partition out of range issue with `sd_random_geometry()`, copilot sent me to this line and asked me to debug print the output of build_side (I should have listened!)

rust/sedona-spatial-join/src/exec.rs
Kontinuation: Refactored this code to use an executor.

PR #510, opened 2026-01-13
feat(rust/sedona-expr): Add GroupsAccumulator to framework and implementation for ST_Envelope_Agg

### LINKED ISSUE #167: rust/sedona-expr: Aggregate function framework should expose GroupsAccumulator
The framework for writing an aggregate function implementation currently doesn't expose the `GroupsAccumulator`:

https://github.com/apache/sedona-db/blob/a93b322f0f60e5927cb21e33a48afb130ac42c61/rust/sedona-expr/src/aggregate_udf.rs#L162-L181

...which is an important optimization for low cardinality groups:

https://github.com/apache/datafusion/blob/5cc0be51f9b60224c0667b89dd0318a591e733d9/datafusion/expr/src/udaf.rs#L581-L605

There are a few potentially interesting things we could exploit in our aggregate functions but this is probably the most important one (or perhaps we need to separate the dispatch mechanism so that we don't have to repeat concepts when certain functions might benefit from a deeper integration with DataFusion).

paleolimbot:
This PR allows aggregate functions implemented using the `SedonaAccumulator` to implement their own `GroupsAccumulator` (which they should usually do for performance on tiny groups). This also implements one for `ST_Envelope_Agg()` and the appropriate extension to the `AggregateUdfTester` (since I couldn't seem to trigger all the features of the `GroupsAccumulator` reliably with simple queries).

This would also benefit other aggregate functions (but they would require that implementation to exist and it might take a bit).

Closes #167.

### DIFF
--- a/rust/sedona-functions/src/st_envelope_agg.rs
+++ b/rust/sedona-functions/src/st_envelope_agg.rs
@@ -337,7 +337,7 @@ impl GroupsAccumulator for BoundsGroupsAccumulator2D {
     }
 
     fn size(&self) -> usize {
-        size_of::<BoundsAccumulator2D>()
+        size_of::<BoundsGroupsAccumulator2D>()
             + self.xs.capacity() * size_of::<Interval>()
             + self.ys.capacity() * size_of::<Interval>()
     }


jiayuasu:
Did you see any performance gain from this?

jiayuasu:
Great work!

PR #513, opened 2026-01-14
fix(rust/sedona-spatial-join): Fix KDB partitioner to produce no less than specified number of partitions

Kontinuation:
The initial KDB tree holds bounding boxes in both internal nodes and leaf nodes. This will produce slightly less number of partitions than specified. We'll risk overshooting the memory budget in such cases.

This patch makes KDB tree storing all the bounding boxes in leaf nodes. The KDB tree will produce slightly more number of partitions, which is fine and gives us some buffer to compensate for the inaccuracy of memory usage estimation. We've also added a method for dumping the partitioning bounding box in human readable format for easier debugging.

### DIFF
--- a/rust/sedona-spatial-join/src/partitioning/kdb.rs
+++ b/rust/sedona-spatial-join/src/partitioning/kdb.rs
@@ -562,6 +562,32 @@ mod tests {
 
     use super::*;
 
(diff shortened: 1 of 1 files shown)


Kontinuation:
I have added a regression test in https://github.com/apache/sedona-db/pull/513/commits/191d7dcdb8a01714e965c26e2b6dea2dae26785a.

PR #515, opened 2026-01-14
chore(rust/sedona-spatial-join): More accurate batch in-memory size estimation

Kontinuation:
This patch improves the accuracy of memory usage estimation by implementing our own functions for estimating the in-memory sizes of record batches and arrow arrays.

The rationale is similar to https://github.com/apache/datafusion/pull/13377. If we don't roll our own memory usage estimation function but call `RecordBatch::get_array_memory_size` instead, we'll get insanely inaccurate numbers for spilled batches read using `arrow::ipc::reader::StreamReader`.

Future work: use the memory pool API of arrow-rs for more accurate memory usage accounting. See https://github.com/apache/arrow-rs/issues/8137.

### DIFF
--- a/rust/sedona-spatial-join/src/utils/arrow_utils.rs
+++ b/rust/sedona-spatial-join/src/utils/arrow_utils.rs
@@ -265,8 +265,8 @@ mod tests {
 
         // Slice: first struct row only
         let slice0 = struct_ref.slice(0, 1);
-        let expected_bv_slice0 = 1 * OFFSET_BYTES + 2 * VIEW_BYTES + long1.len();
-        let expected_i32_slice0 = 1 * OFFSET_BYTES + 3 * I32_BYTES;
+        let expected_bv_slice0 = OFFSET_BYTES + 2 * VIEW_BYTES + long1.len();
+        let expected_i32_slice0 = OFFSET_BYTES + 3 * I32_BYTES;
         assert_eq!(
             get_array_memory_size(&slice0).unwrap(),
             expected_bv_slice0 + expected_i32_slice0
@@ -274,8 +274,8 @@ mod tests {
 
         // Slice: second struct row only
         let slice1 = struct_ref.slice(1, 1);
-        let expected_bv_slice1 = 1 * OFFSET_BYTES + 1 * VIEW_BYTES + long2.len();
-        let expected_i32_slice1 = 1 * OFFSET_BYTES + 1 * I32_BYTES;
+        let expected_bv_slice1 = OFFSET_BYTES + VIEW_BYTES + long2.len();
+        let expected_i32_slice1 = OFFSET_BYTES + I32_BYTES;
         assert_eq!(
             get_array_memory_size(&slice1).unwrap(),
             expected_bv_slice1 + expected_i32_slice1


rust/sedona-spatial-join/src/utils/arrow_utils.rs
```diff
@@ -0,0 +1,166 @@
...
+    Ok(total_size)
+}
+
+/// Estimate the in-memory size of a given Arrow array. This function estimates the
+/// size as if the underlying buffers were copied to somewhere else and not shared.
+pub(crate) fn get_array_memory_size(array: &ArrayRef) -> Result<usize> {
+    let array_data = array.to_data();
+    let size = count_array_data_memory_size(&array_data)?;
+    Ok(size)
+}
```
paleolimbot, line 44: Is https://docs.rs/arrow/latest/arrow/array/struct.ArrayData.html#method.get_slice_memory_size helpful here?

rust/sedona-spatial-join/src/utils/arrow_utils.rs
```diff
@@ -0,0 +1,166 @@
...
+/// Count the memory usage of `array_data` and its children recursively.
+fn count_array_data_memory_size(array_data: &ArrayData) -> core::result::Result<usize, ArrowError> {
+    Ok(get_binary_view_value_size(array_data)? + array_data.get_slice_memory_size()?)
+}
+
+fn get_binary_view_value_size(array_data: &ArrayData) -> Result<usize, ArrowError> {
+    let mut result: usize = 0;
+    let array_data_type = array_data.data_type();
+
+    if matches!(array_data_type, DataType::BinaryView | DataType::Utf8View) {
```
paleolimbot, line 66: This seems like it is missing a branch for non BinaryView types. I imagine we aren't trying to ignore the memory occupied by non-view types?

rust/sedona-spatial-join/src/utils/arrow_utils.rs
Kontinuation: It is already using `get_slice_memory_size` under the hood. However, `get_slice_memory_size` does not take buffers of binary/utf8 views into account, `count_array_data_memory_size` is meant to fix that.

rust/sedona-spatial-join/src/utils/arrow_utils.rs
Kontinuation: Non-view types were already taken care by `get_slice_memory_size`.

paleolimbot: Got it! To prevent readers/llm readers from making this mistake again, how about one function:

```rust
let mut result = 0;
let array_data_type = array_data.data_type();

match array_data_type {
  DataType::BinaryView | DataType::Utf8View => {
    result += array_data.get_slice_memory_size()?;
    // Also take into account view buffer memory
    
  }
  // If you're being super complete you might have to take into account
  // dictionaries, run end encoded, and/or list views but those are newer/rarer
  _ => {
    result += array_data.get_slice_memory_size()?;
  }
}
```

rust/sedona-spatial-join/src/utils/arrow_utils.rs
```diff
@@ -0,0 +1,166 @@
...
+        // * Short strings, length <= 12
+        // | Bytes 0-3  | Bytes 4-15                            |
+        // |------------|---------------------------------------|
+        // | length     | data (padded with 0)                  |
+        //
+        // * Long strings, length > 12
+        // | Bytes 0-3  | Bytes 4-7  | Bytes 8-11 | Bytes 12-15 |
+        // |------------|------------|------------|-------------|
+        // | length     | prefix     | buf. index | offset      |
+        let views = &array_data.buffer::<u128>(0)[..array_data.len()];
```
paleolimbot, line 79: Should this take into account the offset?

```suggestion
        let views = &array_data.buffer::<u128>(0)[array_data.offset()..array_data.len()];
```

rust/sedona-spatial-join/src/utils/arrow_utils.rs
```diff
@@ -0,0 +1,166 @@
...
+                } else {
+                    0
+                }
+            })
+            .sum();
+    }
+
+    for child in array_data.child_data() {
+        result += get_binary_view_value_size(child)?;
+    }
```
paleolimbot: Does this apply to a Utf8View/BinaryView array?

rust/sedona-spatial-join/src/utils/arrow_utils.rs
```diff
@@ -0,0 +1,166 @@
...
+    for array in batch.columns() {
+        let array_data = array.to_data();
+        total_size += count_array_data_memory_size(&array_data)?;
+    }
+
+    Ok(total_size)
+}
+
+/// Estimate the in-memory size of a given Arrow array. This function estimates the
+/// size as if the underlying buffers were copied to somewhere else and not shared.
```
paleolimbot:
```suggestion
/// Estimate the in-memory size of a given Arrow array. This function estimates the
/// size as if the underlying buffers were copied to somewhere else and not shared,
/// including the sizes of each BinaryView item (which is otherwise not counted by
/// `array_data.get_slice_memory_size()`).
```

rust/sedona-spatial-join/src/utils/arrow_utils.rs
paleolimbot, line 97: Possibly worth a comment since it wasn't clear to me initially what was going on:

```suggestion
    // If this was not a BinaryView/Utf8View array, count the bytes of any BinaryView/Utf8View
    // children, taking into account the slice of this array that applies to the child.
    for child in array_data.child_data() {
        result += get_binary_view_value_size(child)?;
    }
```

...or change the comment to acknowledge that we aren't taking into account the slice of the child we are dealing with, which is super annoying math and is very tricky to get right with little payoff (probably not that many sliced list arrays are coming through this function).

rust/sedona-spatial-join/src/utils/arrow_utils.rs
```diff
@@ -0,0 +1,166 @@
...
+    }
+
+    #[test]
+    fn test_sliced_view_array_memory_size() {
+        let array = StringViewArray::from(vec![
+            "short",
+            "Long string that is definitely longer than 12 bytes",
+            "Another long string to make buffer larger",
+        ]);
+        let sliced = array.slice(0, 2);
```
paleolimbot, line 175: We can also test `array.slice(2, 1)` to test the non-zero offset case

rust/sedona-spatial-join/src/utils/arrow_utils.rs
```diff
@@ -0,0 +1,166 @@
...
+            (
+                Arc::new(Field::new("b", DataType::Boolean, false)),
+                Arc::new(boolean_array) as ArrayRef,
+            ),
+        ]);
+
+        let array_ref: ArrayRef = Arc::new(struct_array);
+        let size = get_array_memory_size(&array_ref).unwrap();
+        // 83 (StringView) + 1 (Boolean values) = 84
+        assert_eq!(size, 84);
```
paleolimbot, line 157: I think `array_ref.slice(0, 1)` will give the wrong answer with the current code (the slice is not projected into the children)

rust/sedona-spatial-join/src/utils/arrow_utils.rs
paleolimbot: Got it! I suggested a comment so that future me also doesn't think that 🙂

rust/sedona-spatial-join/src/utils/arrow_utils.rs
Kontinuation: This does not look correct to me. `ArrayData::buffer` has already taken offset into consideration: https://github.com/apache/arrow-rs/blob/57.2.0/arrow-data/src/data.rs#L580-L588

I'll add the test cases for buffers not starting at 0 anyway.

Kontinuation:
I have added the requested tests and also renamed the private function `count_array_data_memory_size` to `get_array_data_memory_size` since the original name looked awful.

rust/sedona-spatial-join/src/utils/arrow_utils.rs
paleolimbot: What do you think about this test case? (Also OK to punt with a comment in the appropriate location that we arenot calculating correct memory usage of sliced lists/structs?)

Kontinuation: I have added `array_ref.slice(0, 1)` test below and it gave correct result.

rust/sedona-spatial-join/src/utils/arrow_utils.rs
Kontinuation: `array.slice(2, 1)` is invalid since the array contains only 2 rows. I have tested `array.slice(1, 1)` and `array.slice(2, 0)` instead.

rust/sedona-spatial-join/src/utils/arrow_utils.rs
Kontinuation: Comment was added to `get_array_data_memory_size`.

rust/sedona-spatial-join/src/utils/arrow_utils.rs
paleolimbot: Ah, arrow-rs propagates the slice for a StructArray on `slice()`: https://docs.rs/arrow-array/57.2.0/src/arrow_array/array/struct_array.rs.html#337

It doesn't do this for list arrays: https://docs.rs/arrow-array/57.2.0/src/arrow_array/array/list_array.rs.html#396

...and possibly not for some other types like dictionaries, list views, or run-end-encoded. I think it's OK to punt on all of that but perhaps open an issue and link to it here in case this comes up.

paleolimbot (approved):
Apologies for the nits on offset/length (I still have nightmares from implementing nested offsets in nanoarrow). I think this is still not right for some cases but feel free to file an issue and follow up later.

rust/sedona-spatial-join/src/utils/arrow_utils.rs
Kontinuation: I'd like to leave `get_array_data_memory_size` as is and log a ticket. https://github.com/apache/sedona-db/issues/519

Kontinuation: And also added a XFAIL test case for list array slices.

PR #526, opened 2026-01-19
doc: Add `validate` arg to `ST_GeomFromWKB()` SQL reference

2010YOUY01:
Skipping validation is supported in `st_geomfromwkb()` function

```sh
# In sedona-cli
> SELECT ST_AsText(
    ST_GeomFromWKB(decode('0101000000000000000000F03F0000000000000040', 'hex'), false)
);
┌──────────────────────────────────────────────────────────────────────────────────────────────────────────────────┐
│ st_astext(st_geomfromwkb(decode(Utf8("0101000000000000000000F03F0000000000000040"),Utf8("hex")),Boolean(false))) │
│                                                       utf8                                                       │
╞══════════════════════════════════════════════════════════════════════════════════════════════════════════════════╡
│ POINT(1 2)                                                                                                       │
└──────────────────────────────────────────────────────────────────────────────────────────────────────────────────┘
```

This PR includes this optional arg to the SQL reference

### DIFF
--- a/docs/reference/sql.md
+++ b/docs/reference/sql.md
@@ -585,7 +585,8 @@ SELECT ST_GeometryN(ST_GeomFromText('MULTIPOINT((1 2), (3 4), (5 6), (8 9))'), 1
 
 #### Description
 
-Construct a Geometry from WKB.
+Construct a Geometry from WKB. By default, input is validated and invalid WKB raises an error.
+Pass `false` as the `validate` argument to skip validation for efficiency; invalid WKB then results in undefined behavior.
 
 Since: v0.1.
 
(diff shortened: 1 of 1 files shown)


2010YOUY01:
> Thank you for opening!
> 
> I think we need to actually expose this still (either via the argument or via a separate function like `ST_GeomFromWKBUnchecked()` or `sd_wrap_wkb()`).
> 
> ```python
> import sedona.db
> 
> sd = sedona.db.connect()
> 
> sd.sql("SELECT ST_GeomFromWkb(ST_AsBinary(ST_GeomFromText('POINT (0 1)')), false) as s").show()
> #> SedonaError: st_geomfromwkb([Arrow(Binary), Arrow(Boolean)]): No kernel matching arguments
> ```

Sorry for the confusion. I mistakenly thought it was already exposed due to some local AI changes 🤦🏼

I think using an additional API would be better than adding an argument:
- it makes the unsafe operation more explicit;
- it also aligns better with PostGIS as a reference system, and they don’t have such an argument: https://postgis.net/docs/ST_GeomFromWKB.html

I’ll open another PR for this.

PR #532, opened 2026-01-20
fix(sedona-raster-functions): Handle array inputs in coordinate conversion functions

Kontinuation:
Previously, RS_RasterToWorldCoord and RS_WorldToRasterCoord only handled scalar inputs and would error with 'Expected scalar ... argument' when array inputs were passed.

This commit fixes both functions to properly handle array inputs by:
- Using into_array() to expand scalar inputs to arrays
- Processing all values in a loop with proper null handling

Added test cases for array input handling.

### DIFF
--- a/rust/sedona-raster-functions/src/rs_rastercoordinate.rs
+++ b/rust/sedona-raster-functions/src/rs_rastercoordinate.rs
@@ -18,10 +18,8 @@ use std::{sync::Arc, vec};
 
 use crate::executor::RasterExecutor;
 use arrow_array::builder::{BinaryBuilder, Int64Builder};
-use arrow_array::cast::AsArray;
-use arrow_array::types::Float64Type;
-use arrow_array::Array;
 use arrow_schema::DataType;
+use datafusion_common::cast::as_float64_array;
 use datafusion_common::error::Result;
 use datafusion_expr::{
     scalar_doc_sections::DOC_SECTION_OTHER, ColumnarValue, Documentation, Volatility,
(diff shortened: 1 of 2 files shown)


rust/sedona-raster-functions/src/rs_rastercoordinate.rs
```diff
@@ -139,17 +141,35 @@ impl SedonaScalarKernel for RsCoordinateMapper {
...
         let mut builder = Int64Builder::with_capacity(executor.num_iterations());
 
-        let coord_opt = extract_scalar_coord(&args[1], &args[2])?;
-        executor.execute_raster_void(|_i, raster_opt| {
-            match (raster_opt, coord_opt) {
-                (Some(raster), (Some(x), Some(y))) => {
+        // Expand world x and y coordinate parameters to arrays and cast to Float64
+        let world_x_array = args[1].clone().cast_to(&DataType::Float64, None)?;
+        let world_x_array = world_x_array.into_array(executor.num_iterations())?;
+        let world_x_array = world_x_array.as_primitive::<Float64Type>();
```
paleolimbot: Totally optional, but I have mostly been using DataFusion's helpers for these so far (`as_float64_array()`), which return errors in debug mode.

rust/sedona-raster-functions/src/rs_rastercoordinate.rs
```diff
@@ -139,17 +141,35 @@ impl SedonaScalarKernel for RsCoordinateMapper {
...
+            let x_opt = if world_x_array.is_null(i) {
+                None
+            } else {
+                Some(world_x_array.value(i))
+            };
+            let y_opt = if world_y_array.is_null(i) {
+                None
+            } else {
+                Some(world_y_array.value(i))
+            };
```
paleolimbot: Also optional, but so far we have mostly implemented this by resolving the iterator before the loop and calling `x_opt.next().unwrap()`. In theory the iterator could be faster than calling `is_null()` and `value()` in a loop but probably isn't in practice.

rust/sedona-raster-functions/src/rs_rastercoordinate.rs
Kontinuation: Use as_float64_array to replace `as_primitive`.

rust/sedona-raster-functions/src/rs_rastercoordinate.rs
Kontinuation: Switched to `.next().unwrap()`.

PR #533, opened 2026-01-21
feat(rust/sedona-functions): Implement `ST_GeomFromWKBUnchecked()`

2010YOUY01:
# Rationale

`ST_GeomFromWKBUnchecked()` is equivalent to `ST_GeomFromWKB()`, but it skips validation during conversion.

This API is more efficient but potentially unsafe, so it is exposed as a separate function to make the lack of safety explicit.

An alternative would be to control validation via an additional argument, but this was not chosen for the above reason.
```sql
select ST_GeomFromWKB('0x01....', false); -- Skip validation via `false` arg
```

# Implementation
The function kernel for `ST_GeomFromWKBUnchecked()` is already implemented, this PR only exposed such functionality through a new function, and add tests.

### DIFF
--- a/python/sedonadb/tests/functions/test_functions.py
+++ b/python/sedonadb/tests/functions/test_functions.py
@@ -16,6 +16,7 @@
 # under the License.
 import math
 
+import pyarrow
 import pytest
 import shapely
 from sedonadb.testing import PostGIS, SedonaDB, geom_or_null, val_or_null
@@ -1707,9 +1708,9 @@ def test_st_geomfromwkbunchecked_invalid_wkb(eng):
         "SELECT ST_AsBinary(ST_GeomFromWKBUnchecked(0x01))", b"\x01"
     )
 
-    # TODO: Test error case after error message check is supported in `assert_query_result`
     # Using invalid WKB elsewhere may result in undefined behavior.
-    # eng.assert_query_result("SELECT ST_AsText(ST_GeomFromWKBUnchecked(0x01))", None)
+    with pytest.raises(pyarrow.lib.ArrowInvalid, match="failed to fill whole buffer"):
+        eng.execute_and_collect("SELECT ST_AsText(ST_GeomFromWKBUnchecked(0x01))")
 
 
 @pytest.mark.parametrize("eng", [SedonaDB, PostGIS])


python/sedonadb/tests/functions/test_functions.py
```diff
@@ -1668,6 +1669,49 @@ def test_st_geomfromwkb(eng, geom):
...
+    eng = eng.create_or_skip()
+
+    # Invalid WKB payload can still convert to geometry column
+    eng.assert_query_result(
+        "SELECT ST_AsBinary(ST_GeomFromWKBUnchecked(0x01))", b"\x01"
+    )
+
+    # TODO: Test error case after error message check is supported in `assert_query_result`
+    # Using invalid WKB elsewhere may result in undefined behavior.
+    # eng.assert_query_result("SELECT ST_AsText(ST_GeomFromWKBUnchecked(0x01))", None)
```
2010YOUY01: Follow-up PR perhaps: I found it's not possible to assert error message if the query fails.

rust/sedona-functions/src/st_geomfromwkb.rs
```diff
@@ -77,6 +92,29 @@ fn doc(name: &str, out_type_name: &str) -> Documentation {
     .build()
 }
 
+/// Documentation for `ST_GeomFromWKBUnchecked()`.
+///
+/// Parameterized for reuse if `ST_GeogFromWKBUnchecked()` is implemented in the future.
+fn doc_unchecked(name: &str, out_type_name: &str) -> Documentation {
```
2010YOUY01, line 98: Question: Is it the case that SQL reference (e.g. https://sedona.apache.org/sedonadb/latest/reference/sql/#st_geomfromwkb) can be auto-generated through these `doc()` traits, but we still have to update them manually?

python/sedonadb/tests/functions/test_functions.py
paleolimbot: I think sometimes we do:

```python
with pytest.raises(lib.SedonaError, match="foofy"):
    eng.execute_and_collect("SELECT ...")
```

Would that work here? If not I think it is OK to skip this check for now.

rust/sedona-functions/src/st_geomfromwkb.rs
paleolimbot: Yes, it's currently being updated manually. It is also in a state of flux right now where we're trying to separate the pages from each other in a way that each page can have a bit more detail, but still generate the main page. For now having the doc reference like you did here is perfect.

2010YOUY01:
Thank you for the review.

> I think sometimes we do:
> 
> ```python
> with pytest.raises(lib.SedonaError, match="foofy"):
>     eng.execute_and_collect("SELECT ...")
> ```
> 
> Would that work here? If not I think it is OK to skip this check for now.

Yes it works, applied in [3524859](https://github.com/apache/sedona-db/pull/533/commits/35248596225206e2bae4ebaa25d53e1ee2e81f75)

I was thinking wrapping such raises syntax with something like
```
eng.assert_query_error("select foo()", Exception, "Function did not exist")
```
To make error assertion slightly more ergonomic, should we do this as follow up?

2010YOUY01:
CI python test failed in https://github.com/apache/sedona-db/actions/runs/21236733252/job/61106200223?pr=533

Looks like it's not related to the change, and I can't reproduce it locally with repeated test runs. Let me re-trigger CI to see if it is reproducible.

paleolimbot (approved):
Thank you!

I believe the CI failure is due to pandas 3.0 (released today).

PR #535, opened 2026-01-21
feat(rust/sedona-functions): Implement ST_AsEWKB with item CRS support

paleolimbot:
This PR implements an EWKB writer and uses it to implement ST_AsEWKB.

```python
import sedona.db

sd = sedona.db.connect()

sd.sql("SELECT ST_AsEWKB(ST_GeomFromEWKT('SRID=26920;POINT (0 1)')) as ewkb").show()
#> ┌────────────────────────────────────────────────────┐
#> │                        ewkb                        │
#> │                       binary                       │
#> ╞════════════════════════════════════════════════════╡
#> │ 0101000020286900000000000000000000000000000000f03f │
#> └────────────────────────────────────────────────────┘

# Regular binary is shorter because it doesn't have the embedded SRID
sd.sql("SELECT ST_AsBinary(ST_GeomFromEWKT('SRID=26920;POINT (0 1)')) as ewkb").show()
#> ┌────────────────────────────────────────────┐
#> │                    ewkb                    │
#> │                   binary                   │
#> ╞════════════════════════════════════════════╡
#> │ 01010000000000000000000000000000000000f03f │
#> └────────────────────────────────────────────┘
```


rust/sedona-geometry/src/ewkb_factory.rs
```diff
@@ -0,0 +1,349 @@
...
+            wkt::to_wkt::write_geometry(
+                &mut wkt_roundtrip_with_srid,
+                &wkb_geom_roundtrip_with_srid,
+            )
+            .unwrap();
+            assert_eq!(wkt_roundtrip_with_srid, wkt_str);
+        }
+    }
+
+    const ROUNDTRIP_CASES: [&str; 56] = [
```
zhangfengcdt: Can we add a test case that have very large SRID values and nested GeometryCollection?

paleolimbot: Good call! A few questionable things were happening for CRSes that couldn't be cleanly exported as an SRID

rust/sedona-functions/src/executor.rs
```diff
@@ -383,6 +383,20 @@ impl ScalarGeo for ScalarValue {
             | ScalarValue::BinaryView(maybe_item)
             | ScalarValue::LargeBinary(maybe_item) => Ok(maybe_item.as_deref()),
             ScalarValue::Null => Ok(None),
+            ScalarValue::Struct(s)
+                if s.fields().len() == 2
+                    && s.fields()[0].name() == "item"
+                    && s.fields()[1].name() == "crs" =>
```
Kontinuation, line 389: If I remember it correctly, this pattern for decomposing fields in struct type representing item with crs has appeared many times in the code base. Maybe we can define a helper method for doing this.

PR #542, opened 2026-01-23
chore(rust/sedona-spatial-join): Integrate bounding box sampler and spilling support into build side collector

Kontinuation:
This is an intermediate step that integrates multiple pieces together. The overall behavior is unchanged when memory limit of DataFusion is not set (which is the default case). The collected bounding box samples are unused for now, the performance overhead of sampling boxes is negligible according to our tests. Committing this won't drive the project into an unstable or unreleasable state.

The next step will be adding a partitioned spatial index provider and integrate spatial partitioner into the main spatial join workflow, but will effectively only work on one single partition for now. This will also be a non-breaking change.

### DIFF
--- a/rust/sedona-spatial-join/src/exec.rs
+++ b/rust/sedona-spatial-join/src/exec.rs
@@ -33,6 +33,7 @@ use datafusion_physical_plan::{
     PlanProperties,
 };
 use parking_lot::Mutex;
+use sedona_common::SpatialJoinOptions;
 
 use crate::{
     build_index::build_index,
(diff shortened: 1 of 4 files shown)


rust/sedona-spatial-join/src/exec.rs
```diff
@@ -178,6 +180,7 @@ impl SpatialJoinExec {
             filter.as_ref(),
             converted_from_hash_join,
         )?;
+        let seed = fastrand::u64(0..0xFFFF);
```
paleolimbot: No need to do this here, but should this be configurable in some way or logged so we can reproduce failures?

rust/sedona-spatial-join/Cargo.toml
```diff
@@ -66,6 +66,7 @@ geo-index = { workspace = true }
 geos = { workspace = true }
 float_next_after = { workspace = true }
 fastrand = { workspace = true }
+log = { workspace = true }
```
paleolimbot, line 69: Are `log` and `env_logger` doing the same thing or do we need both of them?

rust/sedona-spatial-join/src/exec.rs
Kontinuation: I have added an option `sedona.spatial_join.debug.random_seed` for setting the random seed; also the seed is printed in debug log.

rust/sedona-spatial-join/Cargo.toml
Kontinuation: We need both. `log` is the logging facade, `env_logger` is the logging implementation we actually use.

Reference: https://docs.rs/log/latest/log/#available-logging-implementations

PR #544, opened 2026-01-23
feat(rust/sedona-proj): Let ST_Transform() handle 3D coordinates

yutannihilation:
As `Proj::transform()` already transforms 4D, the implementation is fairly simple.

I guess the hard part is to find proper test cases...

```
> SELECT ST_Transform(ST_GeomFromWKT('POINT Z (-124 45 100)'), 4979, 4978);
┌─────────────────────────────────────────────────────────────────────────────────────┐
│ st_transform(st_geomfromwkt(Utf8("POINT Z (-124 45 100)")),Int64(4979),Int64(4978)) │
│                                       geometry                                      │
╞═════════════════════════════════════════════════════════════════════════════════════╡
│ POINT Z(-2526244.3011458996 -3745311.197806657 4487419.119544039)                   │
└─────────────────────────────────────────────────────────────────────────────────────┘
```

### DIFF
--- a/python/sedonadb/tests/functions/test_transforms.py
+++ b/python/sedonadb/tests/functions/test_transforms.py
@@ -35,7 +35,7 @@ def test_st_transform(eng):
 def test_st_transform_3d(eng):
     eng = eng.create_or_skip()
     eng.assert_query_result(
-        "SELECT ST_Transform(ST_GeomFromText('POINT Z (1 1 1)'), 'EPSG:4326', 'EPSG:4978')",
+        "SELECT ST_Transform(ST_GeomFromText('POINT Z (1 1 1)'), 'EPSG:4979', 'EPSG:4978')",
         "POINT Z (6376201.805927448 111297.016517882 110568.792276973)",
         wkt_precision=9,
     )


python/sedonadb/tests/functions/test_transforms.py
```diff
@@ -31,6 +31,16 @@ def test_st_transform(eng):
...
 
 
+@pytest.mark.parametrize("eng", [SedonaDB, PostGIS])
+def test_st_transform_3d(eng):
+    eng = eng.create_or_skip()
+    eng.assert_query_result(
+        "SELECT ST_Transform(ST_GeomFromText('POINT Z (1 1 1)'), 'EPSG:4326', 'EPSG:4978')",
+        "POINT Z (6376201.805927448 111297.016517882 110568.792276973)",
+        wkt_precision=9,
+    )
```
paleolimbot, line 41: This particular transformation is a great test case and an interesting one because it is 2D -> 3D. Does 2D input (e.g., POINT (-64 45)) work? I believe you can query the number of axes of a CRS to handle this:

https://github.com/apache/sedona-db/blob/main/c/sedona-proj/src/proj_dyn_bindgen.rs#L127-L128

...although we can do that later because I think 3D->3D is what the current code is set up to handle.

Another good test would be 3D -> 2D (the inverse of this transform would be good) and 3D -> 3D like `EPSG:4326+EPSG:5701` (lon/lat with Z in meters) to `EPSG:4326+EPSG:8050` (lon/lat with Z in feet). You might have to get pyproj to render those to PROJJSON to get them to work in ST_Transform because our internal `deserialize_crs()` doesn't know how to parse compound CRSes yet.

yutannihilation:
> handle changes in dimensions

Do you mean conversions between 2D CRS and 3D CRS should change the dimension of the geometry? In PostGIS, I don't see such changes.

```
postgres=# SELECT ST_AsText(ST_Transform(ST_GeomFromText('POINT (1 1 1)'), 'EPSG:4979', 'EPSG:4326'));
    st_astext
-----------------
 POINT Z (1 1 1)
(1 row)
```

yutannihilation:
Thanks for clarifying. I'm not immediately sure what's the right behavior, but at least I agree with this part:

> There is significantly less precedent around 3D/4D/spatiotemporal CRS handling and I don't think this is one of the things we should perfectly match PostGIS if we can do a better job.

yutannihilation:
Thanks for merging! I'll comment there once I've formed an opinion on this.

PR #449, opened 2025-12-14
feat(rust/sedona-functions): Add SRID to ST_GeomFromWKT

yutannihilation:
Part of https://github.com/apache/sedona-db/issues/126

This pull request applies the `SRIDifiedKernel` wrapper implemented in https://github.com/apache/sedona-db/pull/275 to `ST_GeomFromWKT` ~~and `ST_GeogFromWKT`~~ .

```sql
> SELECT ST_SRID(ST_GeomFromText('POINT(40.7128 -74.0060)', 4326));
┌──────────────────────────────────────────────────────────────────────┐
│ st_srid(st_geomfromwkt(Utf8("POINT(40.7128 -74.0060)"),Int64(4326))) │
│                                uint32                                │
╞══════════════════════════════════════════════════════════════════════╡
│                                                                 4326 │
└──────────────────────────────────────────────────────────────────────┘
```

PR #545, opened 2026-01-23
docs: Remove unnecessary ST_SRID in example queries

yutannihilation:
After https://github.com/apache/sedona-db/pull/275 and https://github.com/apache/sedona-db/pull/449, `ST_GeomFromText` and `ST_Point` accepts SRID as the last argument. So, in theory, we can replace the following `ST_SetSrid()` usages.

- `ST_SetSrid(ST_GeomFromText('wkt'), srid)` -> `ST_GeomFromText('wkt', srid)`
- `ST_SetSrid(ST_Point(x, y), srid)` -> `ST_Point(x, y, srid)`

### DIFF
--- a/r/sedonadb/tests/testthat/test-context.R
+++ b/r/sedonadb/tests/testthat/test-context.R
@@ -41,11 +41,11 @@ test_that("views can be created and dropped", {
 })
 
 test_that("scalar udfs can be registered", {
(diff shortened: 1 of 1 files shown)


yutannihilation:
Oh, thanks so much for investigating and fixing! I was stuck as I had no idea what was happening here.

PR #555, opened 2026-01-27
feat(rust/sedona-spatial-join) Add partitioned index provider

Kontinuation:
This patch adds a index provider for coordinating the creation of spatial index for specified partitions. It is also integrated into `SpatialJoinExec` so we use it to create the spatial index even when there's only one spatial partition (the degenerate case). The handling for multiple spatial partitions will be added in a subsequent PR.

The memory reservations growed in the build side collection phase will be held by `PartitionedIndexProvider`. Spatial indexes created by the provider does not need to hold memory reservations.

The next step is to support partitioned probe side by adding a `PartitionedProbeStreamProvider`, and modify the state machine of `SpatialJoinStream` to process multiple spatial partitions sequentially.

### DIFF
--- a/rust/sedona-spatial-join/src/index/partitioned_index_provider.rs
+++ b/rust/sedona-spatial-join/src/index/partitioned_index_provider.rs
@@ -305,7 +305,7 @@ impl PartitionedIndexProvider {
         // Spawn tasks to load indexed batches from spilled files concurrently
         let (spill_files, geo_statistics, _) = spilled_partition.into_inner();
         let mut join_set: JoinSet<Result<(), DataFusionError>> = JoinSet::new();
-        let (tx, mut rx) = mpsc::channel(self.probe_threads_count);
+        let (tx, mut rx) = mpsc::channel(spill_files.len() * 2 + 1);
         for spill_file in spill_files {
             let tx = tx.clone();
             join_set.spawn(async move {
@@ -364,15 +364,18 @@ async fn get_index_from_cell(
 #[cfg(test)]
 mod tests {
     use super::*;
-    use crate::evaluated_batch::{
-        evaluated_batch_stream::{
-            in_mem::InMemoryEvaluatedBatchStream, SendableEvaluatedBatchStream,
-        },
-        EvaluatedBatch,
-    };
     use crate::operand_evaluator::EvaluatedGeometryArray;
     use crate::partitioning::partition_slots::PartitionSlots;
     use crate::utils::bbox_sampler::BoundingBoxSamples;
+    use crate::{
+        evaluated_batch::{
+            evaluated_batch_stream::{
+                in_mem::InMemoryEvaluatedBatchStream, SendableEvaluatedBatchStream,
+            },
+            EvaluatedBatch,
+        },
+        index::CollectBuildSideMetrics,
+    };
     use arrow_array::{ArrayRef, BinaryArray, Int32Array, RecordBatch};
     use arrow_schema::{DataType, Field, Schema, SchemaRef};
     use datafusion::config::SpillCompression;
@@ -471,6 +474,7 @@ mod tests {
             bbox_samples: BoundingBoxSamples::empty(),
             estimated_spatial_index_memory_usage: estimated_usage,
             reservation: new_reservation(memory_pool),
+            metrics: CollectBuildSideMetrics::new(0, &ExecutionPlanMetricsSet::new()),
         })
     }
 
--- a/rust/sedona-spatial-join/src/prepare.rs
+++ b/rust/sedona-spatial-join/src/prepare.rs
@@ -33,9 +33,10 @@ use sedona_geometry::bounding_box::BoundingBox;
 
 use crate::{
     index::{
-        memory_plan::{compute_memory_plan, PartitionMemorySummary},
+        memory_plan::{compute_memory_plan, MemoryPlan, PartitionMemorySummary},
         partitioned_index_provider::PartitionedIndexProvider,
-        BuildSideBatchesCollector, CollectBuildSideMetrics, SpatialJoinBuildMetrics,
+        BuildPartition, BuildSideBatchesCollector, CollectBuildSideMetrics,
+        SpatialJoinBuildMetrics,
     },
     partitioning::{
         kdb::KDBPartitioner,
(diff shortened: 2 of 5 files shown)


rust/sedona-spatial-join/src/stream.rs
```diff
@@ -213,16 +232,97 @@ impl SpatialJoinStream {
...
+            let spatial_join_components =
+                ready!(self.once_fut_spatial_join_components.get_shared(cx))?;
+            let provider = Arc::clone(&spatial_join_components.partitioned_index_provider);
+            self.num_regular_partitions = Some(provider.num_regular_partitions() as u32);
+            self.index_provider = Some(provider);
+        }
+
+        let num_partitions = self
+            .num_regular_partitions
+            .expect("num_regular_partitions should be available");
```
paleolimbot, line 249: Should this return a `sedona_internal_err!()`?

rust/sedona-spatial-join/src/stream.rs
```diff
@@ -213,16 +232,97 @@ impl SpatialJoinStream {
...
         Poll::Ready(Ok(StatefulStreamResult::Continue))
     }
 
+    fn wait_build_index(
+        &mut self,
+        cx: &mut std::task::Context<'_>,
+    ) -> Poll<Result<StatefulStreamResult<Option<RecordBatch>>>> {
+        let num_partitions = self
+            .num_regular_partitions
+            .expect("num_regular_partitions should be available");
```
paleolimbot, line 272: Should this return a `sedona_internal_err!()`?

rust/sedona-spatial-join/src/stream.rs
```diff
@@ -213,16 +232,97 @@ impl SpatialJoinStream {
...
+        if partition_id >= num_partitions {
+            self.state = SpatialJoinStreamState::Completed;
+            return Poll::Ready(Ok(StatefulStreamResult::Continue));
+        }
+
+        if self.pending_index_future.is_none() {
+            let provider = Arc::clone(
+                self.index_provider
+                    .as_ref()
+                    .expect("Partitioned index provider should be available"),
```
paleolimbot, line 283: Should this return a `sedona_internal_err!()`?

rust/sedona-spatial-join/src/stream.rs
```diff
@@ -213,16 +232,97 @@ impl SpatialJoinStream {
...
+                );
+                async move { provider.build_or_wait_for_index(partition_id).await }.boxed()
+            };
+            self.pending_index_future = Some(future);
+        }
+
+        let future = self
+            .pending_index_future
+            .as_mut()
+            .expect("pending future must exist");
```
paleolimbot, line 299: Should this return a `sedona_internal_err!()`?

rust/sedona-spatial-join/src/prepare.rs
```diff
@@ -0,0 +1,416 @@
...
+    seed: u64,
+) -> Result<SpatialJoinComponents> {
+    let session_config = context.session_config();
+    let target_batch_size = session_config.batch_size();
+    let sedona_options = session_config
+        .options()
+        .extensions
+        .get::<SedonaOptions>()
+        .cloned()
+        .unwrap_or_default();
```
paleolimbot: If it is easy to avoid it may be a good idea to not clone all of the SedonaOptions (lest some day it contain something large).

rust/sedona-spatial-join/src/prepare.rs
```diff
@@ -0,0 +1,416 @@
...
+
+            let extent = geo_stats
+                .bbox()
+                .cloned()
+                .unwrap_or(BoundingBox::xy((0, 0), (0, 0)));
+            let extent = if extent.is_empty() {
+                BoundingBox::xy((0, 0), (0, 0))
+            } else {
+                extent
+            };
```
paleolimbot: Do we have to assign a finite but meaningless value here? Can the KDBPartitioner accept an empty extent so that the implications of assigning `(0, 0), (0, 0)` arbitrarily are more clear?

rust/sedona-spatial-join/src/prepare.rs
```diff
@@ -0,0 +1,416 @@
...
+            let kdb_partitioner =
+                KDBPartitioner::build(samples.into_iter(), max_items_per_node, max_levels, extent)?;
+            log::debug!(
+                "Built KDB spatial partitioner with {} partitions",
+                num_partitions
+            );
+            let mut kdb_dbg_str = String::new();
+            if kdb_partitioner.debug_print(&mut kdb_dbg_str).is_ok() {
+                log::debug!("KDB partitioner debug info:\n{}", kdb_dbg_str);
+            }
```
paleolimbot: Should this/is this going to compute a debug string even in release mode?

rust/sedona-spatial-join/src/prepare.rs
```diff
@@ -0,0 +1,416 @@
...
+pub(crate) async fn prepare_spatial_join_components(
+    context: Arc<TaskContext>,
+    build_schema: SchemaRef,
+    build_streams: Vec<SendableRecordBatchStream>,
+    spatial_predicate: SpatialPredicate,
+    join_type: JoinType,
+    probe_threads_count: usize,
+    metrics: ExecutionPlanMetricsSet,
+    seed: u64,
+) -> Result<SpatialJoinComponents> {
```
paleolimbot: You don't have to do this if it is too much work, but this is a very long function that does quite a lot that has a lot of arguments. Would it be cleaner to do something like:

```rust
struct SpatialJoinComponentsBuilder {
    pub context: Arc<TaskContext>,
    pub build_schema: SchemaRef,
    pub build_streams: Vec<SendableRecordBatchStream>,
    pub spatial_predicate: SpatialPredicate,
    pub join_type: JoinType,
    pub probe_threads_count: usize,
    pub metrics: ExecutionPlanMetricsSet,
    pub seed: u64,
}

impl SpatialJoinComponentsBuilder {
    pub fn build(self) -> Result<SpatialJoinComponents> {
        let sedona_options = self.collect_options();
        if self.build_streams.is_empty() {
            return self.empty_components();
        }

        let (memory, stuff) = self.configure_reservations();
        let build_partitions = self.collect_build_partitions().await?;
        if num_partitions == 1 {
            return self.single_partition_components();
        }

        let build_partitioner = self.make_partitioner();
        let mut partitioned_spill_files_vec = self.collect_build_side_in_parallel().await?;
        self.merge_spilled_partitions(&partitioned_spill_files_vec)?;
        self.sanity_check_partitions();
        self.multi_partition_components()
    }
}
```

Totally understandable if there's too much intermediate state between those steps that needs to be shared to split that up...what is being done here is a very tricky thing to accomplish and I get that at some point it just has to be done.

In the absence of function names to help, perhaps at least some comments around what this is doing, which seems like a critical high-level piece of the out of core implementation!

rust/sedona-spatial-join/src/prepare.rs
```diff
@@ -0,0 +1,416 @@
...
+                partitioned_spill_files
+            });
+            reservations.push(reservation);
+        }
+
+        let results = join_set.join_all().await;
+        let mut partitioned_spill_files_vec = Vec::with_capacity(results.len());
+        for result in results {
+            partitioned_spill_files_vec.push(result?);
+        }
```
paleolimbot: Not sure if it's possible/cleaner, but maybe:


```suggestion
        let mut partitioned_spill_files_vec = results.into_iter().collect::<Result<Vec<_>>>();
```

rust/sedona-spatial-join/src/utils/disposable_async_cell.rs
```diff
@@ -0,0 +1,194 @@
...
+    /// The cell already has a value.
+    AlreadySet,
+}
+
+/// An asynchronous cell that can be set at most once before either being
+/// disposed or read by any number of waiters.
+///
+/// Awaiters calling [`DisposableAsyncCell::get`] will park until a value is set
+/// or the cell is disposed. Once disposed, `get` returns `None` and `set`
+/// returns [`CellSetError::Disposed`].
```
paleolimbot, line 48: I trust that you need this, but perhaps add to this comment what this is used for?

rust/sedona-spatial-join/src/index/memory_plan.rs
```diff
@@ -0,0 +1,217 @@
...
+    /// The memory budget for intermittent usage, such as buffering data during repartitioning.
+    pub memory_for_intermittent_usage: usize,
+    /// The number of spatial partitions to split the build side into.
+    pub num_partitions: usize,
+}
+
+impl MemoryPlan {
+    /// Write debug info for this memory plan
+    pub fn debug_print(&self, f: &mut impl std::fmt::Write) -> std::fmt::Result {
+        writeln!(f, "Memory Plan:")?;
```
paleolimbot: Is this better than `#[derive(Debug)]` + `format!("self:#?")`

rust/sedona-spatial-join/src/index/partitioned_index_provider.rs
```diff
@@ -0,0 +1,598 @@
...
+                Some(
+                    self.build_index_for_spilled_partition(spilled_partition)
+                        .await,
+                )
+            }
+        }
+    }
+
+    #[allow(unused)]
+    pub async fn wait_for_index(&self, partition_id: u32) -> Option<Result<Arc<SpatialIndex>>> {
```
paleolimbot: Should this be `#[cfg(test)]` or can it be removed?

rust/sedona-spatial-join/src/index/partitioned_index_provider.rs
```diff
@@ -0,0 +1,598 @@
...
+            self.join_type,
+            self.probe_threads_count,
+            self.metrics.clone(),
+        )?;
+
+        // Spawn tasks to load indexed batches from spilled files concurrently
+        let (spill_files, geo_statistics, _) = spilled_partition.into_inner();
+        let mut join_set: JoinSet<Result<(), DataFusionError>> = JoinSet::new();
+        let (tx, mut rx) = mpsc::channel(spill_files.len() * 2 + 1);
+        for spill_file in spill_files {
```
paleolimbot, line 309: Is there any value to limiting the concurrency? (i.e., will there are 2000 spill files and if so will we get/do we want to spawn 2000 threads?)

rust/sedona-spatial-join/src/prepare.rs
```diff
@@ -0,0 +1,416 @@
...
+
+        let num_partitions = build_partitioner.num_regular_partitions();
+        log::debug!("Actual number of spatial partitions: {}", num_partitions);
+
+        // Spawn each task for each build partition to repartition the data using the spatial partitioner for
+        // the build/indexed side
+        let mut join_set = JoinSet::new();
+        let buffer_bytes_threshold = memory_for_intermittent_usage / build_partitions.len();
+        let mut reservations = Vec::with_capacity(build_partitions.len());
+        for (k, partition) in build_partitions.into_iter().enumerate() {
```
paleolimbot: Does this JoinSet limit the concurrency in any way? (i.e., if there are 2000 partitions will we get/do we want 2000 threads to spawn?)

paleolimbot (approved):
🚀 !

rust/sedona-spatial-join/src/stream.rs
Kontinuation: I don't think so. `num_regular_partitions` must exist at this point and this can be easily proved. Something must be going very very wrong if `expect` panics. The same applies to the remaining .expect calls.

rust/sedona-spatial-join/src/prepare.rs
Kontinuation: We can use an Arc if that's the case. We pass SedonaOptions into spawned async tasks so it is nearly impossible to avoid moving owned values.

rust/sedona-spatial-join/src/prepare.rs
Kontinuation: We can add another method `debug_str` to avoid evaluating it when debug log is not enabled.

rust/sedona-spatial-join/src/index/partitioned_index_provider.rs
Kontinuation: We won't spawn 2000 threads. We'll spawn 2000 async tasks to be scheduled by the async runtime on fixed number of workers. I also don't see the number of threads increasing when there are lots of spill files in practice.

The capacity of the channel may be a problem. It should not be proportional to the number of spill files. I'll use `self.probe_threads_count` instead.

rust/sedona-spatial-join/src/prepare.rs
Kontinuation: Spawning lots of async tasks does not show up as a problem. I had some workloads that handled 500+ partitions, and I didn't observe heavy scheduling overhead.

rust/sedona-spatial-join/src/prepare.rs
Kontinuation: We can make `KDBPartitioner` accept empty extent, but we still need to store a sentinel rect inside KDB tree. I think this will be better, since it makes the KDBPartitioner exposing a cleaner interface, and the selection of sentinel rect is the implementation detail of KDBPartitioner.

rust/sedona-spatial-join/src/prepare.rs
paleolimbot: Got it! This is not spawning threads, it is spawning tasks and there is a scheduler limiting the concurrency.

rust/sedona-spatial-join/src/index/memory_plan.rs
Kontinuation: `{:#?}` is pretty readable so I'll remove the custom debug_print function.

rust/sedona-spatial-join/src/index/partitioned_index_provider.rs
Kontinuation: It will be used by subsequent patches. I'll change it to `#[cfg(test)]` for now.

rust/sedona-spatial-join/src/prepare.rs
Kontinuation: I have refactored this method by splitting it into smaller functions. There are some other related changes made to `BuildPartition` to make the flow of data simpler.

rust/sedona-spatial-join/src/prepare.rs
Kontinuation: Switched to use collect.

rust/sedona-spatial-join/src/utils/disposable_async_cell.rs
Kontinuation: Added comment for how it is used by PartitionedIndexProvider and its role in the partitioned spatial join workflow.

rust/sedona-spatial-join/src/index/partitioned_index_provider.rs
Kontinuation: I made a mistake in the previous comment. The capacity of the channel is not a problem. We need to load everything in that spatial partition from spill files into memory anyway, there's no point reducing the capacity of the channel.

PR #556, opened 2026-01-27
feat(c/sedona-sedona-libgpuspatial): Refactoring GPU Spatial Join Library

pwrliang:
This PR introduces a major refactoring of the GPU-accelerated library. Compared to the previous design, this version decouples the spatial join into distinct filtering and refinement stages, making it easier to integrate it into rust/sedona-spatial-join in an upcoming PR. Additionally, this update includes performance optimizations and minor structural improvements:

- ParallelWkbLoader: Enhances parsing performance by balancing workloads across threads based on byte counts.
- RelateEngine: Improves Point-in-Polygon (PIP) performance by allowing a single AABB to bound multiple line segments (conceptually similar to TG's Natural index).
- MemoryManager: Enables the memory pool to reduce frequent allocation and deallocation overheads.
- Header Renaming: Standardizes naming conventions by using .cuh for CUDA-exclusive headers and .hpp for mixed CUDA/C++ code.


petern48:
As you wait for a proper review, I'll suggest that you consider breaking this into multiple PRs. It looks like you described 4 separate changes (the bullets) that would tend nicely to at least 4 isolated PRs. I know it's more work for you, but it reduces the review burden significantly, as reviewers don't need to figure out which of the 4 bullets a particular code change applies to.

You can always base branches off of each other to reuse work from your other branches. Something like below, or however you see fit.

```
Header Renaming -> MemoryManager -> RelateEngine
       \
        \---> ParallelWkbLoader

(this is a random example, i didn't actually look into it that much)
```

Doing a separate PR for Header Renaming and straightforward changes would cut down the 88-file diff significantly, and could even be reviewed by people with less context about GPU join (like me).

Separate PRs could help your changes land faster, and help with future debugging / understanding when someone tries to figure out what happened. WDYT? Would breaking this up be reasonable?

petern48:
> (or now, if Peter or other members of the community are blocked from participating because of this PR's size)

To be clear, I'm not dying to participate 😅. Definitely too busy for that. Just figured I'd encourage making it more modular, for both the purposes of reviewing and for cleaner traceability and git history. Though yeah, the fact that this is isolated to `c/sedona-libgpuspatial` makes those less important.

> kindly split this one out from the larger parent PR at my request

I guess my point is that it *looks* like it could be broken up even more (primarily based on the PR description). But if you're all good with reviewing it as it is, go for it! I don't mean to stand in the way.

pwrliang:
Thanks for your suggestions. For the future PRs, I will make incremental changes to make them easy to review.

c/sedona-libgpuspatial/libgpuspatial/include/gpuspatial/loader/parallel_wkb_loader.hpp
```diff
@@ -756,102 +799,99 @@ class ParallelWkbLoader {
...
+        uint32_t local_seen_mask = 0;
 
         for (uint32_t work_offset = thread_work_start; work_offset < thread_work_end;
              work_offset++) {
-          auto arrow_offset = work_offset + offset;
-          // handle null value
-          if (ArrowArrayViewIsNull(&array_view_, arrow_offset)) {
+          auto arrow_offset = begin[work_offset];
+
+          if (ArrowArrayViewIsNull(array_view_.get(), arrow_offset)) {
```
paleolimbot: Since it looks like you're trying to avoid branching here, I believe it's faster to decompose this into three branchless loops:

- Assert no nulls
- Assert all lengths > 5
- Accumulate the bitmask
- Fall back on an implementation with branches inside the loop if any of the assertions fails

c/sedona-libgpuspatial/libgpuspatial/include/gpuspatial/loader/parallel_wkb_loader.hpp
```diff
@@ -875,21 +915,107 @@ class ParallelWkbLoader {
...
+      prefix_sum.push_back(total_bytes);
+    }
+
+    // 3. Calculate balanced split points
+    std::vector<uint32_t> split_points;
+    split_points.reserve(num_threads + 1);
+    split_points.push_back(0);  // The start index for the first thread
+
+    // Avoid division by zero
+    if (num_threads > 0) {
```
paleolimbot: Should something like this fail with an assertion or exception?

c/sedona-libgpuspatial/libgpuspatial/include/gpuspatial/loader/parallel_wkb_loader.hpp
```diff
@@ -875,21 +915,107 @@ class ParallelWkbLoader {
...
     size_t total_bytes = 0;
-    for (int64_t i = 0; i < length; i++) {
-      if (!ArrowArrayViewIsNull(&array_view_, offset + i)) {
-        auto item = ArrowArrayViewGetBytesUnsafe(&array_view_, offset + i);
+    for (auto it = begin; it != end; ++it) {
+      auto offset = *it;
+      if (!ArrowArrayViewIsNull(array_view_.get(), offset)) {
+        auto item = ArrowArrayViewGetBytesUnsafe(array_view_.get(), offset);
         total_bytes += item.size_bytes - 1      // byte order
                        - 2 * sizeof(uint32_t);  // type + size
```
paleolimbot: Later you just use `item.size_bytes`. Are the extra five bytes per item important here and not elsewhere?

c/sedona-libgpuspatial/libgpuspatial/include/gpuspatial/loader/parallel_wkb_loader.hpp
```diff
@@ -875,21 +915,107 @@ class ParallelWkbLoader {
...
       }
     }
     return total_bytes;
   }
+
+  template <typename OFFSET_IT>
+  std::vector<uint32_t> assignBalancedWorks(OFFSET_IT begin, OFFSET_IT end,
+                                            uint32_t num_threads) const {
+    size_t total_bytes = 0;
```
paleolimbot, line 938: To my reading these are only iterating over sequences representable by `offset` and `length`. If this is the case you can optmize some things (e.g., calling `ArrowArrayViewIsNull()` in a loop is slow and can often be skipped by checking for a zero null count and proceeding with a branchless loop).

Also, if this can be reduced to an offset/length you can compute the value you're looking for without peeking into the values at all depending on the array type (binary/large binary this is "just" reading two values from the offsets buffer).

c/sedona-libgpuspatial/libgpuspatial/include/gpuspatial/refine/spatial_refiner.hpp
```diff
@@ -0,0 +1,44 @@
...
+  virtual ~SpatialRefiner() = default;
+
+  virtual void Clear() = 0;
+
+  virtual void PushBuild(const ArrowSchema* build_schema,
+                         const ArrowArray* build_array) = 0;
+
+  virtual void FinishBuilding() = 0;
+
+  virtual uint32_t Refine(const ArrowSchema* probe_schema, const ArrowArray* probe_array,
```
paleolimbot: For the purposes of the internal C++ class for this, can you simplify these signatures by passing a `const ArrowArrayView*`?

c/sedona-libgpuspatial/libgpuspatial/include/gpuspatial/utils/markers.hpp
```diff
@@ -0,0 +1,145 @@
...
+  static constexpr uint32_t IntervalWorkitems = 2;
+};
+
+// Colors in ARGB format (Alpha, Red, Green, Blue)
+struct Color {
+  static constexpr uint32_t Red = 0xFF880000;
+  static constexpr uint32_t Green = 0xFF008800;
+  static constexpr uint32_t Blue = 0xFF000088;
+  static constexpr uint32_t Yellow = 0xFFFFFF00;
+  static constexpr uint32_t Default = 0;
```
paleolimbot, line 42: Do we need these?

c/sedona-libgpuspatial/libgpuspatial/include/gpuspatial/refine/rt_spatial_refiner.cuh
```diff
@@ -0,0 +1,124 @@
...
+
+#include "geoarrow/geoarrow_type.h"
+#include "nanoarrow/nanoarrow.h"
+
+#include "rmm/cuda_stream_pool.hpp"
+#include "rmm/cuda_stream_view.hpp"
+
+#include <thread>
+
+#define GPUSPATIAL_PROFILING
```
paleolimbot: Should this define be in some configuration header somewhere or is this intentional?

c/sedona-libgpuspatial/libgpuspatial/include/gpuspatial/refine/spatial_refiner.hpp
```diff
@@ -0,0 +1,44 @@
...
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+#pragma once
+#include "gpuspatial/relate/predicate.hpp"
+
+#include "nanoarrow/nanoarrow.h"
+
+namespace gpuspatial {
+class SpatialRefiner {
```
paleolimbot, line 30: This class and its methods would benefit from docstrings because it is one of the places where the general approach of this library is constrained. You can link to the docstrings here in the C interface documentation to avoid duplicating them.

c/sedona-libgpuspatial/libgpuspatial/include/gpuspatial/refine/rt_spatial_refiner.hpp
```diff
@@ -0,0 +1,52 @@
...
+#include "gpuspatial/refine/spatial_refiner.hpp"
+#include "gpuspatial/rt/rt_engine.hpp"
+
+#include <memory>
+
+namespace gpuspatial {
+
+struct RTSpatialRefinerConfig {
+  std::shared_ptr<RTEngine> rt_engine;
+  // Prefer fast build the BVH
```
paleolimbot, line 30: Can these be proper docstrings?

c/sedona-libgpuspatial/libgpuspatial/src/memory_manager.cc
```diff
@@ -0,0 +1,128 @@
...
+  }
+  return 0;
+
+#elif defined(__linux__)
+  // --- Linux (sysinfo) ---
+  struct sysinfo info;
+  if (sysinfo(&info) == 0) {
+    return (long long)info.freeram * (long long)info.mem_unit;
+  }
+  return 0;
```
paleolimbot, line 46: There is at least one other place where you query this information by calling `sysinfo()`. Should that code use this helper?

c/sedona-libgpuspatial/libgpuspatial/test/refiner_test.cu
```diff
@@ -0,0 +1,738 @@
...
+#include "gpuspatial/index/rt_spatial_index.cuh"
+#include "gpuspatial/refine/rt_spatial_refiner.cuh"
+
+namespace gpuspatial {
+// Function to read a single Parquet file and extract a column.
+static arrow::Status ReadParquetFromFile(
+    arrow::fs::FileSystem* fs,     // 1. Filesystem pointer (e.g., LocalFileSystem)
+    const std::string& file_path,  // 2. Single file path instead of a folder
+    int64_t batch_size, const char* column_name,
+    std::vector<std::shared_ptr<arrow::Array>>& out_arrays) {
```
paleolimbot: This sequence of tests is a little hard to follow:

- Do we need both a Parquet and IPC loader to load test files? I wonder if this would make it slightly easier to avoid indirection associated with Arrow C++/Array vs nanoarrow/UniqueArray/UniqueSchema. In the early version of this library we were reading Parquet from S3 as part of the end-to-end bechmarks; however, I think once we hook this up to Rust it is better to do those benchmarks there.
- It may help to split some of these longer functions into pieces with more descriptive names. Perhaps encapsulate the GEOS portions as a function like `GenerateReferenceResultUsingGEOS()` and `GenerateResultUsingSpatialJoiner()` that both return the same thing (a vector of index pairs?).

c/sedona-libgpuspatial/libgpuspatial/test/index_test.cu
```diff
@@ -0,0 +1,300 @@
...
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+#include "array_stream.hpp"
+#include "gpuspatial/index/rt_spatial_index.cuh"
+#include "test_common.hpp"
+
+#include <geos/geom/Envelope.h>
+#include <geos/index/ItemVisitor.h>
+#include <geos/index/strtree/STRtree.h>
```
paleolimbot, line 23: It may be helpful in your other test to also use the GEOS C++ API since you have it here. (Also your other test utilities, which here are nicely defined as functions with nice names!)

c/sedona-libgpuspatial/src/lib.rs
```diff
@@ -77,197 +152,480 @@ impl GpuSpatialContext {
...
+                    .unwrap();
+                if result == true.into() {
+                    ans_build_indices.push(poly_index as u32);
+                    ans_probe_indices.push(point_index as u32);
+                }
+            }
+        }
+
+        ans_build_indices.sort();
+        ans_probe_indices.sort();
```
paleolimbot: It might be nice to encapsulate this as a standalone function generating the output index pairs to avoid breaking up the flow of the test.

c/sedona-libgpuspatial/src/libgpuspatial.rs
```diff
@@ -331,179 +327,387 @@ impl GpuSpatialJoinerWrapper {
...
+pub enum GpuSpatialRelationPredicateWrapper {
+    Equals = 0,
+    Disjoint = 1,
+    Touches = 2,
+    Contains = 3,
+    Covers = 4,
+    Intersects = 5,
+    Within = 6,
+    CoveredBy = 7,
+}
```
paleolimbot: Is there a reason that we need two enums defined like this?

c/sedona-libgpuspatial/Cargo.toml
```diff
@@ -40,8 +40,11 @@ which = "8.0"
 arrow-array = { workspace = true, features = ["ffi"] }
 arrow-schema = { workspace = true }
 thiserror = { workspace = true }
+geo = { workspace = true }
+wkt = { workspace = true }
```
paleolimbot: Can these additions be `[dev-dependencies]` (i.e., only used in tests)

c/sedona-libgpuspatial/Cargo.toml
```diff
@@ -40,8 +40,11 @@ which = "8.0"
...
 log = "0.4"
 sedona-schema = { path = "../../rust/sedona-schema" }
+nvml-wrapper = "0.10.0"
```
paleolimbot: Can this be a test dependency as well or can it be removed? (I thought all of the nvidia-related code was behind the C interface?)

c/sedona-libgpuspatial/src/lib.rs
paleolimbot, line 1: For the purposes of this getting this particular PR merged, I think it would be better to remove all the code that no longer compiles from this crate (you could even remove everything except the bindgen bindings) and open another PR for those changes after this one merges.

I can also iterate on these changes with you here, but I think we will both be happier polishing up the C/C++ and tackling this next. A few high level comments if you'd like to go this route:

There is a lot of `#[cfg(gpu_available)]` and we need to find a way to not sprinkle that everywhere. Maybe:

```rust
#[cfg(gpu_available)]
mod gpu_stuff {
   // any code that needs gpu stuff
}

#[cfg(not(gpu_available))]
mod gpu_stuff {
   // dummy version of pub functions or structs from gpu_stuff that error
}
```

There are also a lot of undocumented `pub` functions. It's not clear to me what the intended interface is that should be used outside this crate and some documentation would help highlight that.

c/sedona-libgpuspatial/libgpuspatial/include/gpuspatial/gpuspatial_c.h
```diff
@@ -14,60 +14,162 @@
...
+  SedonaSpatialPredicateDisjoint,
+  SedonaSpatialPredicateTouches,
+  SedonaSpatialPredicateContains,
+  SedonaSpatialPredicateCovers,
+  SedonaSpatialPredicateIntersects,
+  SedonaSpatialPredicateWithin,
+  SedonaSpatialPredicateCoveredBy
+};
+
+struct SedonaSpatialRefiner {
```
paleolimbot, line 156: Can this (and its members) have docstrings?

c/sedona-libgpuspatial/libgpuspatial/include/gpuspatial/gpuspatial_c.h
```diff
@@ -14,60 +14,162 @@
...
+  int (*clear)(struct SedonaSpatialRefiner* self);
+
+  int (*push_build)(struct SedonaSpatialRefiner* self,
+                    const struct ArrowSchema* build_schema,
+                    const struct ArrowArray* build_array);
```
paleolimbot, line 170: Is it possible to add a callback `int (*init_schema)(struct ArrowSchema* build_schema, struct ArrowSchema* probe_schema)` and eliminate all the `const struct ArrowSchema*` arguments for the other callbacks?

I know this doesn't help you in Rust because arrow-rs likes to export the schema whether you want it or not. But this might be fixed at some point and from a C API standpoint it's much cleaner.

c/sedona-libgpuspatial/libgpuspatial/include/gpuspatial/gpuspatial_c.h
```diff
@@ -14,60 +14,162 @@
...
+
+  int (*finish_building)(struct SedonaSpatialRefiner* self);
+
+  int (*refine_loaded)(struct SedonaSpatialRefiner* self,
+                       const struct ArrowSchema* probe_schema,
+                       const struct ArrowArray* probe_array,
+                       enum SedonaSpatialRelationPredicate predicate,
+                       uint32_t* build_indices, uint32_t* probe_indices,
+                       uint32_t indices_size, uint32_t* new_indices_size);
```
paleolimbot: I noticed in a recent PR that @Kontinuation  used 64-bit indices for the build side but 32-bit indices for the probe side. I am guessing that here the extra 4 bytes per row probably makes a difference because you have to shuffle them to and from the GPU, but if it doesn't, using a `uint64` for the build side might better align with Kristin's join code.

c/sedona-libgpuspatial/libgpuspatial/include/gpuspatial/gpuspatial_c.h
```diff
@@ -14,60 +14,162 @@
...
+
+  int (*refine)(struct SedonaSpatialRefiner* self, const struct ArrowSchema* schema1,
+                const struct ArrowArray* array1, const struct ArrowSchema* schema2,
+                const struct ArrowArray* array2,
+                enum SedonaSpatialRelationPredicate predicate, uint32_t* indices1,
+                uint32_t* indices2, uint32_t indices_size, uint32_t* new_indices_size);
```
paleolimbot: Some documentation for these parameters would be helpful (are these schemas the same as probe/build?)

c/sedona-libgpuspatial/libgpuspatial/test/c_wrapper_test.cc
```diff
@@ -84,23 +182,138 @@ TEST_F(CWrapperTest, InitializeJoiner) {
...
+                                        (uint32_t*)probe_indices_ptr + new_len);
+
+    struct Payload {
+      GEOSContextHandle_t handle;
+      const GEOSGeometry* geom;
+      std::vector<uint32_t> build_indices;
+      std::vector<uint32_t> stream_indices;
+      SedonaSpatialRelationPredicate predicate;
+    };
+
```
paleolimbot: I believe a lot of this code also exists in another test? Is it worth testing the C wrapper in that test as well instead of duplicating the code that calculates the "correct" answer or consolidating the three places where that is currently done into some common code?

paleolimbot (commented):
Thank you for working on this, and apologies for taking so long to review it! I promise all of my comments are just details...there just happens to be a lot of code in this PR and so there are also a lot of details. Generally:

- This is great! I really appreciate the test coverage and the new C interface that aligns more readily with the way our existing spatial join works. I'm glad to know this approach is also faster
- There are a few things around iterating over WKB arrays that could be faster. It is OK if you don't want to explore ways to make this faster in this PR...I was just here and looking at your code and so I put a comment if I noticed something that could be faster. I will be looking at making it easier to do this sort of thing in geoarrow-c soon because I need it for Geography support (also a binding to a C++ library) and at least now I know that functions like memory size estimates or rechunking are worth building in when I'm there.
- I do think we should deduplicate the three tests where you compare results to the GEOS-calculated version. The one that uses GEOS C++ is the cleanest and perhaps the other two could use that approach as well. It is also OK if they are testing the same bit of code to have them be one test.
- A few minor comments on the C API
- Removing the Rust section of this PR (e.g., by just deleting everything from the Rust wrapper crate except the bindgen bindings) will make it easier to merge this PR and review the Rust code separately (see inline comment).

c/sedona-libgpuspatial/libgpuspatial/include/gpuspatial/loader/parallel_wkb_loader.hpp
pwrliang: Thanks for your suggestions. I have changed it to two cases (has nulls or no nulls). I think it is not worth the many lines of code to handle cases with fewer than 5 bytes, since I have measured that this function does not take much runtime.

c/sedona-libgpuspatial/libgpuspatial/include/gpuspatial/loader/parallel_wkb_loader.hpp
pwrliang: It does not matter. We just need to gauge the total number of bytes loaded into memory to avoid OOM with chunk parsing.

c/sedona-libgpuspatial/libgpuspatial/include/gpuspatial/loader/parallel_wkb_loader.hpp
pwrliang: I changed it to two branches with/without nulls.

c/sedona-libgpuspatial/libgpuspatial/include/gpuspatial/utils/markers.hpp
pwrliang: These are used with Nsight Systems for profling purposes

c/sedona-libgpuspatial/libgpuspatial/src/memory_manager.cc
pwrliang: parallel_wkb_loader.hpp used to call sysinfo and I have moved it to `memory_manger.cc`

c/sedona-libgpuspatial/src/libgpuspatial.rs
pwrliang: This one maps the C library enum to a Rust enum. Where's the second one?

c/sedona-libgpuspatial/Cargo.toml
pwrliang: wkt = { workspace = true } has been moved to `[dev-dependencies]`, but we need `nvml-wrapper` to check whether the GPU is available to use

c/sedona-libgpuspatial/libgpuspatial/include/gpuspatial/gpuspatial_c.h
pwrliang: I guess you refer to batch refinement. Kristin's changes are compatible with the current interface design. https://github.com/zhangfengcdt/sedona-db/blob/bba62f552317d0e8f61b5af3fcbfd1c916296628/rust/sedona-spatial-join/src/index/gpu_spatial_index.rs#L180

c/sedona-libgpuspatial/libgpuspatial/include/gpuspatial/gpuspatial_c.h
pwrliang: This interface is currently unused, so I have removed it.

c/sedona-libgpuspatial/libgpuspatial/include/gpuspatial/refine/rt_spatial_refiner.cuh
pwrliang: These have been removed. We can enable it in `build.rs` when profiling

c/sedona-libgpuspatial/libgpuspatial/include/gpuspatial/gpuspatial_c.h
pwrliang: I think passing a pair of <ArrowSchema, ArrowArray> could be more flexible as it allows loading arrays with different schemas, e.g., push a Binary array first and push another BinaryView array later.

c/sedona-libgpuspatial/src/lib.rs
pwrliang: I have removed the Rust code and will add them back in a later PR

c/sedona-libgpuspatial/libgpuspatial/include/gpuspatial/gpuspatial_c.h
paleolimbot: I see! At this point in refinement these build indices are local and not global.

c/sedona-libgpuspatial/libgpuspatial/include/gpuspatial/gpuspatial_c.h
paleolimbot: I would prefer to have that be an error...DataFusion should never be producing arrays with different schemas for the build and probe side and knowing the types up front allows better reuse of nanoarrow/geoarrow readers in the implementation (e.g., ArrowArrayView, GeoArrowArrayReader).

c/sedona-libgpuspatial/libgpuspatial/include/gpuspatial/utils/markers.hpp
paleolimbot: Can you add a comment to this effect so these are not removed by somebody else?

c/sedona-libgpuspatial/libgpuspatial/test/test_common.hpp
```diff
@@ -91,6 +102,139 @@ std::string GetCanonicalPath(const std::string& relative_path_str) {
...
+// GEOS Visitor Implementation
+class JoinVisitor : public geos::index::ItemVisitor {
+ public:
+  JoinVisitorContext* ctx;
+  explicit JoinVisitor(JoinVisitorContext* c) : ctx(c) {}
+
+  void visitItem(void* item) override {
+    const auto* build_geom = static_cast<const geos::geom::Geometry*>(item);
+
+    // Use the existing predicate evaluator from your TestUtils
```
paleolimbot:
```suggestion
    // Use the existing predicate evaluator from TestUtils
```

c/sedona-libgpuspatial/libgpuspatial/test/refiner_test.cu
```diff
@@ -0,0 +1,356 @@
...
+    rects.reserve(array->length);
+
+    for (int64_t i = 0; i < array->length; i++) {
+      // Parse WKB
+      ArrowStringView wkb_view = ArrowArrayViewGetStringUnsafe(array_view.get(), i);
+      // Copy the view to a buffer because WKBReader reads from istream or byte array
+      // We can cast directly if the underlying type allows
+      std::stringstream iss;
+      // WKBReader::readHEX parses hex strings, read parses raw bytes
+      // Arrow WKB is usually raw bytes (binary)
```
paleolimbot:
```suggestion
```

c/sedona-libgpuspatial/libgpuspatial/test/refiner_test.cu
```diff
@@ -0,0 +1,356 @@
...
+        << error.message;
+    ASSERT_EQ(ArrowArrayViewSetArray(probe_view.get(), probe_array, &error), NANOARROW_OK)
+        << error.message;
+
+    std::vector<box_t> queries;
+    std::vector<std::unique_ptr<geos::geom::Geometry>> probe_geoms;
+    probe_geoms.reserve(probe_array->length);
+
+    for (int64_t i = 0; i < probe_array->length; i++) {
+      ArrowStringView wkb_view = ArrowArrayViewGetStringUnsafe(probe_view.get(), i);
```
paleolimbot:
```suggestion
      ArrowBufferView wkb_view = ArrowArrayViewGetBytesUnsafe(probe_view.get(), i);
```

c/sedona-libgpuspatial/libgpuspatial/include/gpuspatial/gpuspatial_c.h
pwrliang: Okay, I have another reason to stick with the current design. When using the spatial index, you will see the build side first. You will not see the stream side when building the index ( https://github.com/apache/sedona-db/blob/d17d7f08dd65fe42d9bf64126ae2a383666c72f6/rust/sedona-spatial-join/src/build_index.rs#L41). This introduces another issue. When should I call `init_schema`? We may also need to consider thread-safety as there are many partitions on the stream side that may call `init_schema` at the same time, and we need to ensure the correctness under such a case, e.g., with init_once_future.

c/sedona-libgpuspatial/libgpuspatial/include/gpuspatial/refine/rt_spatial_refiner.hpp
```diff
@@ -0,0 +1,54 @@
...
+  uint32_t parsing_threads = std::thread::hardware_concurrency();
+  // How many threads are allowed to call PushStream concurrently
+  uint32_t concurrency = 1;
+  // Overlapping parsing and refinement by pipelining multiple batches; 1 means no
+  // pipelining
+  uint32_t pipeline_batches = 1;
+  // the host memory quota for WKB parser compared to the available memory
+  float wkb_parser_memory_quota = 0.8;
+  // the device memory quota for relate engine compared to the available memory
+  float relate_engine_memory_quota = 0.8;
```
paleolimbot, line 45: Optional, but these are probably more effective to tweak from SedonaDB as memory quotas in MB (since SedonaDB has its own memory accounting).

c/sedona-libgpuspatial/libgpuspatial/include/gpuspatial/refine/spatial_refiner.hpp
```diff
@@ -0,0 +1,45 @@
...
+
+  virtual void Clear() = 0;
+
+  virtual void PushBuild(const ArrowArrayView* build_array) = 0;
+
+  virtual void FinishBuilding() = 0;
+
+  virtual uint32_t Refine(const ArrowArrayView* probe_array, Predicate predicate,
+                          uint32_t* build_indices, uint32_t* probe_indices,
+                          uint32_t len) = 0;
```
paleolimbot, line 65: Can you also add docstrings to these functions since this is one of the main APIs? In particular that `build_indices` and `probe_indices` are output parameters and what `len` is referring to (is `probe_array->length` sufficient?).

I think one of my earlier comments was not correct...here, `build_indices` is global and not local, correct? This implies the build side can never be larger than 4 billion rows (which is probably fine given the current memory limitations of GPUs). All of this is fine, but these parameters do need to be documented.

c/sedona-libgpuspatial/libgpuspatial/include/gpuspatial/gpuspatial_c.h
```diff
@@ -14,60 +14,199 @@
...
+  int cuda_memory_pool_init_precent;
 };
 
-struct GpuSpatialJoinerContext {
-  const char* last_error;  // Pointer to std::string to store last error message
-  void* private_data;      // GPUSpatial context
-  void* build_indices;     // Pointer to std::vector<uint32_t> to store results
-  void* stream_indices;
+/** Opaque runtime for GPU spatial operations
+ * Each process should only has one instance of GpuSpatialRuntime exactly
```
paleolimbot:
```suggestion
 * Each process should have exactly one instance of GpuSpatialRuntime
```

c/sedona-libgpuspatial/libgpuspatial/include/gpuspatial/gpuspatial_c.h
```diff
@@ -14,60 +14,199 @@
...
+  void (*create_context)(struct SedonaSpatialIndexContext* context);
+  /** Destroy a previously created context */
+  void (*destroy_context)(struct SedonaSpatialIndexContext* context);
+  /** Push rectangles for building the spatial index, each rectangle is represented by 4
+   * floats: [min_x, min_y, max_x, max_y] Points can also be indexed by providing [x, y,
+   * x, y] but points and rectangles cannot be mixed
+   *
+   * @return 0 on success, non-zero on failure
+   */
+  int (*push_build)(struct SedonaFloatIndex2D* self, const float* buf, uint32_t n_rects);
```
paleolimbot, line 80: I am not sure I understand the comment about mixing points and rectangles. Are points specified by only two `float`s? Or are degenerate rectangles like (0, 1, 0, 1) not valid?

c/sedona-libgpuspatial/libgpuspatial/include/gpuspatial/gpuspatial_c.h
```diff
@@ -14,60 +14,199 @@
...
+  void (*get_build_indices_buffer)(struct SedonaSpatialIndexContext* context,
+                                   uint32_t** build_indices,
+                                   uint32_t* build_indices_length);
+  /** Get the probe indices buffer from the context
+   *
+   * @return A pointer to the buffer and its length
+   */
+  void (*get_probe_indices_buffer)(struct SedonaSpatialIndexContext* context,
+                                   uint32_t** probe_indices,
+                                   uint32_t* probe_indices_length);
```
paleolimbot, line 110: In the `refine` callback the build and probe indices are returned directly from the callback into caller-allocated memory. Is it possible to use the same pattern for both of these interfaces?

(Also worth clarifying the local vs. globalness of the build indices here)

c/sedona-libgpuspatial/libgpuspatial/include/gpuspatial/gpuspatial_c.h
```diff
@@ -14,60 +14,199 @@
...
+   * @param probe_indices The probe indices of candidate pairs
+   * @param indices_size The number of candidate pairs
+   * @param new_indices_size Output parameter to store the number of refined pairs
+   * @return 0 on success, non-zero on failure
+   */
+  int (*refine)(struct SedonaSpatialRefiner* self, const struct ArrowSchema* probe_schema,
+                const struct ArrowArray* probe_array,
+                enum SedonaSpatialRelationPredicate predicate, uint32_t* build_indices,
+                uint32_t* probe_indices, uint32_t indices_size,
+                uint32_t* new_indices_size);
```
paleolimbot, line 196: It is worth clarifying here that build indices are global and probe indices are local, that build_indices and probe_indices and new_indices_size are output arguments, and what happens when there are more indices than `indices_size`.

c/sedona-libgpuspatial/libgpuspatial/test/c_wrapper_test.cc
```diff
@@ -15,92 +15,304 @@
...
+              NANOARROW_OK)
+        << error.message;
+    ASSERT_EQ(ArrowArrayViewSetArray(probe_view.get(), probe_array.get(), &error),
+              NANOARROW_OK)
+        << error.message;
 
-    void* build_indices_ptr;
-    void* stream_indices_ptr;
+    for (int64_t j = 0; j < probe_array->length; j++) {
+      ArrowStringView wkb = ArrowArrayViewGetStringUnsafe(probe_view.get(), j);
```
paleolimbot:
```suggestion
      ArrowBufferView wkb = ArrowArrayViewGetBytesUnsafe(probe_view.get(), j);
```

c/sedona-libgpuspatial/libgpuspatial/include/gpuspatial/loader/parallel_wkb_loader.hpp
```diff
@@ -105,6 +66,7 @@ template <typename POINT_T, typename INDEX_T>
 struct HostParsedGeometries {
   constexpr static int n_dim = POINT_T::n_dim;
   using mbr_t = Box<Point<float, n_dim>>;
+  GeometryType type;  // A general type that can reprs
```
paleolimbot:
```suggestion
  GeometryType type;
```

c/sedona-libgpuspatial/libgpuspatial/include/gpuspatial/loader/parallel_wkb_loader.hpp
```diff
@@ -531,9 +548,7 @@ class ParallelWkbLoader {
 
  public:
   struct Config {
-    // How many rows of WKBs to process in one chunk
-    // This value affects the peak memory usage and overheads
-    int chunk_size = 16 * 1024;
+    float memory_quota = 0.8f;  // percentage of free memory to use
```
paleolimbot, line 551: Just making sure this value is updated from the input config somewhere (I think I just missed where this happens)

c/sedona-libgpuspatial/libgpuspatial/include/gpuspatial/loader/parallel_wkb_loader.hpp
```diff
@@ -543,82 +558,94 @@ class ParallelWkbLoader {
...
+    size_t num_offsets = std::distance(begin, end);
+    if (num_offsets == 0) return;
+
     auto parallelism = thread_pool_->num_threads();
-    auto est_bytes = estimateTotalBytes(array, offset, length);
-    auto free_memory = detail::get_free_physical_memory_linux();
+    uint64_t est_bytes = estimateTotalBytes(array_view, begin, end);
+
+    uint64_t free_memory = MemoryManager::get_available_host_memory();
+    uint64_t memory_quota = free_memory * config_.memory_quota;
```
paleolimbot, line 590: Not for this PR but I think eventually we'll have to specify a host memory quote in MB based on some accounting in the RuntimeEnv of DataFusion.

paleolimbot (commented):
Thank you for these updates and apologies for the delay...the deduplication of tests and cleaning up of the array size in bytes looks great!

Everything here is just details...the one thing I would love to see is a few clarifications on the API and ensuring we only need to pass the `ArrowSchema` once instead of per batch.

zhangfengcdt (approved):
The refactoring cleanly separates GPU spatial join into distinct filtering (index) and refinement stages. The new design introduces proper abstractions with clear C API boundaries. Two suggessions for the future improvement:

- add safe Rust wrappers around SedonaFloatIndex2D and SedonaSpatialRefiner with proper lifetime management and error handling and integrate them to the rust code in spatial join module
- would be beneficial to use pipeline_batches in SpatialRefinerConfig async pipelining in GPU so GPU and CPU codes can be more streamlined

c/sedona-libgpuspatial/libgpuspatial/include/gpuspatial/gpuspatial_c.h
pwrliang: It is not possible for the index because the result buffer size is unknown to the caller.

c/sedona-libgpuspatial/libgpuspatial/include/gpuspatial/loader/parallel_wkb_loader.hpp
pwrliang: The refiner passes this config to the WKB parser.

paleolimbot (approved):
Thank you!

c/sedona-libgpuspatial/src/lib.rs
pwrliang: https://github.com/apache/sedona-db/pull/586

PR #557, opened 2026-01-28
feat(rust/sedona-expr): Pass ConfigOptions into UDFs

### LINKED ISSUE #248: Pipe actual ConfigOptions to UDF calls instead of using defaults
In the DataFusion 50 upgrade, we added `config_options: Arc::new(ConfigOptions::default())` to all `ScalarFunctionArgs` and `ScalarFunctionExpr::new` calls throughout the codebase to satisfy the new API requirements.

While this works functionally, it means these UDF calls don't respect user-configured DataFusion options that might affect query behavior.

Proposed Improvement
Pass actual `ConfigOptions` from the session/context through the call stack instead of creating default instances.

Locations to Address

  - `rust/sedona/src/show.rs`**
  - `rust/sedona-geoparquet/src/writer.rs` - BBox UDF calls (lines with `ScalarFunctionExpr::new`)
  - `rust/sedona-geoparquet/src/file_opener.rs` - Any UDF invocations
  - `rust/sedona-spatial-join/src/optimizer.rs` - Already gets config from context, verify it's passed through
  - `rust/sedona-testing/src/testers.rs` - Test utilities (lower priority, defaults likely fine for tests)

Please also find other locations that might be relevant.

Kontinuation:
Closes #248

This will allow altering the behavior of UDFs. For instance, allowing the size of per-thread proj cache used by ST_Transform to be configured before invoking any CRS transformation functions.

### DIFF
--- a/rust/sedona-geoparquet/src/format.rs
+++ b/rust/sedona-geoparquet/src/format.rs
@@ -302,11 +302,18 @@ impl FileFormat for GeoParquetFormat {
     async fn create_writer_physical_plan(
         &self,
         input: Arc<dyn ExecutionPlan>,
-        _state: &dyn Session,
+        session: &dyn Session,
         conf: FileSinkConfig,
         order_requirements: Option<LexRequirement>,
     ) -> Result<Arc<dyn ExecutionPlan>> {
-        create_geoparquet_writer_physical_plan(input, conf, order_requirements, &self.options)
+        let session_config_options = session.config().options();
+        create_geoparquet_writer_physical_plan(
+            input,
+            conf,
+            order_requirements,
+            &self.options,
+            session_config_options,
+        )
     }
 
     fn file_source(&self) -> Arc<dyn FileSource> {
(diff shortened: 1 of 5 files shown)


c/sedona-extension/src/scalar_kernel.rs
```diff
@@ -560,6 +562,7 @@ impl ExportedScalarKernelImpl {
                     &args,
                     return_type,
                     num_rows as usize,
+                    None,
```
Kontinuation, line 565: We do not support passing ConfigOptions to FFI UDFs yet (this also won't be easy). I wonder should we simply pass in a default ConfigOptions here. This will simplify the type of `config_options` parameter to be `&ConfigOptions`.

paleolimbot: I think it's OK that functions using the FFI extension mechanism don't have access to options. There are probably only a few functions where it makes sense to use this.

paleolimbot: I like the `Option<>` but if there is no cost to constructing default options that is OK too. If a Sedona UDF is specifically making use of options it's probably better for it to error in the unlikely event it is FFIed by accident?

Kontinuation: I agree that functions relying on options should fail instead of silently using default options. Let's keep using `Option<>` then.

PR #558, opened 2026-01-28
fix(python/sedonadb): Ensure that Python UDFs executing with >1 batch do not cause deadlock

paleolimbot:
I had intended to post a reprex to GeoPandas regarding crashes when threading but was caught by this issue, where the way we collected things into Python caused a lot of attempts to acquire the GIL (or perhaps a lock related to tokio) which interfered with UDF execution.

Briefly, before this PR, the Python bindings always collected via a special `RecordBatchReader` that called `block_on()` + `allow_threads()`, waiting for the next batch in the output `SendableRecordBatchStream`. To ensure cancellation requests worked, we aquired the GIL every 1 second to check for signals. After this PR, we use one `block_on()` + `allow_threads()` and collect all the batches at once when we know this is what needs to happen anyway. Streaming output still works it just is now only invoked when required.

This constant `block_on()` + GIL acquisition caused a deadlock when Python UDFs were also trying to acquire the GIL (or perhaps a tokio lock of some kind).

The workaround here is not a full solution but covers the most common case, where a user wants to collect the entire result (e.g., `.to_pandas()`. This is simpler to orchestrate.

### DIFF
--- a/python/sedonadb/src/dataframe.rs
+++ b/python/sedonadb/src/dataframe.rs
@@ -163,18 +163,12 @@ impl InternalDataFrame {
         let batches = wait_for_future(py, &self.runtime, async move {
             let mut stream = df.execute_stream().await?;
             let schema = stream.schema();
-            let mut count = 0usize;
             let mut batches = Vec::new();
             while let Some(batch) = stream.try_next().await? {
-                count += batch.num_rows();
                 batches.push(batch);
             }
 
-            Ok::<_, DataFusionError>(Batches {
-                schema,
-                batches,
-                count,
-            })
+            Ok::<_, DataFusionError>(Batches { schema, batches })
         })??;
 
         Ok(batches)
@@ -314,15 +308,10 @@ impl InternalDataFrame {
 pub struct Batches {
     schema: SchemaRef,
     batches: Vec<RecordBatch>,
-    count: usize,
 }
 
 #[pymethods]
 impl Batches {
-    fn __len__(&self) -> usize {
-        self.count
-    }
-
     #[pyo3(signature = (requested_schema=None))]
     fn __arrow_c_stream__<'py>(
         &self,


python/sedonadb/src/dataframe.rs
```diff
@@ -289,3 +309,51 @@ impl InternalDataFrame {
...
+pub struct Batches {
+    schema: SchemaRef,
+    batches: Vec<RecordBatch>,
+    count: usize,
+}
+
+#[pymethods]
+impl Batches {
+    fn __len__(&self) -> usize {
+        self.count
```
zhangfengcdt: Is this the number of rows or number of batches? Looks like __len__ is expected to be the later.

python/sedonadb/python/sedonadb/dataframe.py
```diff
@@ -253,10 +253,9 @@ def to_arrow_table(self, schema: Any = None) -> "pyarrow.Table":
         import pyarrow as pa
         import geoarrow.pyarrow  # noqa: F401
 
-        if schema is None:
-            return pa.table(self)
-        else:
-            return pa.table(self, schema=pa.schema(schema))
+        # Collects all batches into an object that exposes __arrow_c_stream__()
+        batches = self._impl.to_batches(schema)
+        return pa.table(batches)
```
zhangfengcdt, line 258: I assume we do not need to pass schema to pa.table anymore.

zhangfengcdt (approved):
minor comments, otherwise, lgtm

python/sedonadb/src/dataframe.rs
paleolimbot: That's a great point...I removed it since it's not used for anything yet.

python/sedonadb/python/sedonadb/dataframe.py
paleolimbot: Yes, passing it via `to_batches()` ensures it makes it to Rust!

PR #559, opened 2026-01-29
fix(rust/sedona-raster-functions): RasterExecutor always call the closure only once for scalar raster input

Kontinuation:
## Summary
- Fix RasterExecutor::execute_raster_void() to broadcast scalar rasters across num_iterations() when other args are arrays.
- Add regression tests for scalar raster + array coordinate inputs in rs_worldcoordinate and rs_rastercoordinate.

## Why
DataFusion can invoke these scalar UDFs with a scalar raster and array coordinate arguments. Previously the executor only invoked the per-row closure once for scalar rasters, producing output arrays of length 1 instead of the expected batch length.

## Testing
- cargo test -p sedona-raster-functions


rust/sedona-raster-functions/src/executor.rs
```diff
@@ -59,7 +59,7 @@ impl<'a, 'b> RasterExecutor<'a, 'b> {
     /// 4. Calling the provided function with each raster
     pub fn execute_raster_void<F>(&self, mut func: F) -> Result<()>
     where
-        F: FnMut(usize, Option<RasterRefImpl<'_>>) -> Result<()>,
+        F: FnMut(usize, Option<&RasterRefImpl<'_>>) -> Result<()>,
```
Kontinuation, line 62: Passing reference instead of moving value into the closure to avoid repeatedly constructing `RasterRefImpl` from raster array or cloning it. We will not want to persist the `RasterRefImpl` value passed in for most of the time.

rust/sedona-raster-functions/src/rs_rastercoordinate.rs
```diff
@@ -368,4 +371,108 @@ mod tests {
...
+        // Use raster 1 (invertible) as scalar.
+        let rasters = generate_test_rasters(2, Some(0)).unwrap();
+        let raster_struct = rasters.as_any().downcast_ref::<StructArray>().unwrap();
+        let raster_struct_one = raster_struct
+            .slice(1, 1)
+            .as_any()
+            .downcast_ref::<StructArray>()
+            .unwrap()
+            .clone();
+        let scalar_raster = ScalarValue::Struct(Arc::new(raster_struct_one));
```
paleolimbot: Does `try_from_array()` work for this?

```suggestion
        let raster_struct = rasters.as_any().downcast_ref::<StructArray>().unwrap();
        let scalar_raster = ScalarValue::try_from_array(&raster_struct, 1).unwrap();
```

rust/sedona-raster-functions/src/rs_rastercoordinate.rs
```diff
@@ -368,4 +371,108 @@ mod tests {
...
+            .as_any()
+            .downcast_ref::<arrow_array::Int64Array>()
+            .expect("Expected Int64Array");
+        assert_eq!(int_array.len(), 3);
+
+        // At upper-left world coordinate, raster coords should be (0,0).
+        for i in 0..3 {
+            assert!(!int_array.is_null(i));
+            assert_eq!(int_array.value(i), 0_i64);
+        }
```
paleolimbot: Does this work too?

```suggestion
        let expected = create_array!(Int64, [Some(0), Some(0), Some(0)]) as ArrayRef;
        assert_eq!(&array, &expected);
```

rust/sedona-raster-functions/src/rs_rastercoordinate.rs
```diff
@@ -368,4 +371,108 @@ mod tests {
...
+        let raster_struct_one = raster_struct
+            .slice(1, 1)
+            .as_any()
+            .downcast_ref::<StructArray>()
+            .unwrap()
+            .clone();
+        let scalar_raster = ScalarValue::Struct(Arc::new(raster_struct_one));
+
+        let world_x = Arc::new(arrow_array::Float64Array::from(vec![2.0, 2.0, 2.0]));
+        let world_y = Arc::new(arrow_array::Float64Array::from(vec![3.0, 3.0, 3.0]));
```
paleolimbot: This particular test seems like it might benefit from having >1 unique value in the input.

rust/sedona-raster-functions/src/rs_rastercoordinate.rs
```diff
@@ -368,4 +371,108 @@ mod tests {
...
```
paleolimbot: Same comments as above (I think the scalar raster can be generated more compactly and probably more than one unique value in the output would be a better test).

rust/sedona-raster-functions/src/rs_rastercoordinate.rs
Kontinuation: Changed the array argument to have non-unique values.

rust/sedona-raster-functions/src/rs_rastercoordinate.rs
Kontinuation: Changed the array argument to have non-unique values.

rust/sedona-raster-functions/src/rs_rastercoordinate.rs
```diff
@@ -368,4 +372,112 @@ mod tests {
...
+                (Some(x), Some(y)) => {
+                    let (rx, ry) = to_raster_coordinate(&raster_ref, x, y).unwrap();
+                    Some(match coord {
+                        Coord::X => rx,
+                        Coord::Y => ry,
+                    })
+                }
+                _ => None,
+            })
+            .collect();
```
paleolimbot: It would probably be better (and more compact) to hard-code these values. This is a bit circular since it's essentially the same as the UDF implementation.

(Also for other tests)

rust/sedona-raster-functions/src/rs_rastercoordinate.rs
```diff
@@ -368,4 +372,112 @@ mod tests {
...
+                RASTER,
+                SedonaType::Arrow(DataType::Float64),
+                SedonaType::Arrow(DataType::Float64),
+            ],
+        );
+
+        // Use raster 1 (invertible) as scalar.
+        let rasters = generate_test_rasters(2, Some(0)).unwrap();
+        let raster_struct = rasters.as_any().downcast_ref::<StructArray>().unwrap();
+        let scalar_raster = ScalarValue::try_from_array(raster_struct, 1).unwrap();
```
paleolimbot:
```suggestion
        let scalar_raster = ScalarValue::try_from_array(&rasters, 1).unwrap();
```

I know this was my suggestion but I think it can be even more compact 😬 

(Also for the other tests)

rust/sedona-raster-functions/src/rs_rastercoordinate.rs
Kontinuation: Makes sense. My initial attempt is to test the RasterExecutor but not the RS function itself. Hardcoding expected values is the correct approach since this test lives in rs_rastercoordinate.rs so it should test the correctness of the RS function as its primary goal.

rust/sedona-raster-functions/src/rs_rastercoordinate.rs
Kontinuation: Removed redundant code generated by LLM

Issue #530, opened 2026-01-20
Python API to cast binary columns to WKB columns


2010YOUY01:
# Motivation

To convert legacy Parquet files that store geometry as a `BINARY` column whose payload is WKB into GeoParquet, the snippet below can be used. It explicitly converts the binary WKB payload into a geometry value (and sets the SRID), so that SedonaDB recognizes the column as geometry and `to_parquet()` can write GeoParquet metadata correctly.

```python
# geo_legacy.parquet schema
# - geo_bin: Binary (payload is WKB)
# - c1: Int32
# - c2: Int32

df = sd.read_parquet("/data/geo_legacy.parquet")

# Register a view name for SQL
df = df.to_view("t", overwrite=True)

df = sd.sql("""
  SELECT
    ST_SetSRID(ST_GeomFromWKB(geo_bin), 4326) AS geometry,
    * EXCLUDE (geo_bin)
  FROM t
""")

df.to_parquet("geo_geoparquet.parquet")
```

# Proposed new API

It would be helpful to have an easier API for this. Using a dedicated method (instead of fusing the cast into `read_parquet()` or `to_parquet()`) makes the conversion more flexible, especially when “logically geometry, physically WKB-in-binary” columns come from other sources or are produced mid-query.

```python
def with_geometry(...):
    """
    Convert one or more binary WKB columns into geometry columns.

    Args:
        columns: A column name or list of column names containing WKB payloads.
        crs: Optional CRS identifier (e.g., 4326 or "EPSG:4326").
        validate: If True, validate WKB payloads while converting.
        primary: Optional name to mark as the primary geometry column.

    Examples:
        >>> sd = sedona.db.connect()
        >>> df = sd.read_parquet("geo_legacy.parquet").with_geometry(
        ...     columns=["geo_bin"],
        ...     crs="EPSG:4326",
        ...     validate=True,
        ...     primary="geo_bin",
        ... )
    """
```

## Example usage

```python
# geo_legacy.parquet schema
# - geo_bin: Binary (payload is WKB)
# - c1: Int32
# - c2: Int32

df = sd.read_parquet("/data/geo_legacy.parquet")

df = df.with_geometry(
    columns="geo_bin",
    crs=4326,
    validate=True,
    primary="geo_bin",
)

df.to_parquet("geo_geoparquet.parquet")
```

## Implementation

Internally, it's simply add expression projection on geometry columns (with `ST_SetSRID(ST_GeomFromWKB(geo_bin), 4326)`)

[...]

PR #560, opened 2026-01-29
feat(rust/sedona-geoparquet): Support `geometry_columns` option in `read_parquet(..)` to mark additional geometry columns

2010YOUY01:
Closes https://github.com/apache/sedona-db/issues/530

## Motivation

Today, converting legacy Parquet files that store geometry as raw WKB payloads inside `BINARY` columns into GeoParquet requires a full SQL rewrite pipeline. Users must explicitly parse WKB, assign CRS, and reconstruct the geometry column before writing:

```python
# geo_legacy.parquet schema
# - geo_bin: Binary (payload is WKB)
# - c1: Int32
# - c2: Int32

df = sd.read_parquet("/data/geo_legacy.parquet")

df = df.to_view("t", overwrite=True)

df = sd.sql("""
  SELECT
    ST_SetSRID(ST_GeomFromWKB(geo_bin), 4326) AS geometry,
    * EXCLUDE (geo_bin)
  FROM t
""")

df.to_parquet("geo_geoparquet.parquet")
```

This works, but it would be have a easier to use python API:

> “Treat this binary column as a geometry column with encoding=WKB and CRS=EPSG:4326.”

This PR introduces a `geometry_columns` option on `read_parquet()` so legacy Parquet files can be interpreted as GeoParquet directly, without SQL rewriting.

---

### Demo
```python
df = sd.read_parquet(
    "/data/geo_legacy.parquet",
    geometry_columns={
        "geo_bin": {
            "encoding": "WKB",
            "crs": 4326,
        }
    },
)

df.to_parquet("geo_geoparquet.parquet")
```

### Specification
```
            geometry_columns: Optional mapping of column name to GeoParquet column
                metadata (e.g., {"geom": {"encoding": "WKB"}}). Use this to mark
                binary WKB columns as geometry columns. Supported keys:
                - encoding: "WKB" (required)
                - crs: string (e.g., "EPSG:4326") or integer SRID (e.g., 4326).
                  If not provided, the default CRS is OGC:CRS84
                  (https://www.opengis.net/def/crs/OGC/1.3/CRS84), which means
                  the data in this column must be stored in longitude/latitude
                  based on the WGS84 datum.
                - edges: "planar" (default) or "spherical"
                Useful for:
                - Legacy Parquet files with Binary columns containing WKB payloads.
                - Overriding GeoParquet metadata when fields like `crs` are missing.
                Precedence:
                - If a column appears in both GeoParquet metadata and this option,
                  the geometry_columns entry takes precedence.
                Example:
                - For `geo.parquet(geo1: geometry, geo2: geometry, geo3: binary)`,
                  `read_parquet("geo.parquet", geometry_columns={"geo2": {...}, "geo3": ...})`
                  will override `geo2` metadata and treat `geo3` as a geometry column.
                Safety:
                - Columns specified here are not validated for WKB correctness.
                  Invalid WKB payloads may cause undefined behavior.
```
The key points:
- The geometry columns specified in the option overrides what's already in the metadata, I think this can be useful if the metadata is missing some configurations like `crs`, and we can use this API to provide more details
- No validation for now, this can be done in a follow-on PR

---

## Key Changes

1. Parse python option fields into rust `GeoParquetColumnMetadata` struct
2. In the schema inference step, first infer the schema from `GeoParquet` metadata as before, next look at the options, to add/override additional geometry columns

### DIFF
--- a/python/sedonadb/python/sedonadb/context.py
+++ b/python/sedonadb/python/sedonadb/context.py
@@ -140,14 +140,14 @@ class SedonaContext:
                 GeoParquet column metadata (e.g.,
                 {"geom": {"encoding": "WKB"}}). Use this to mark binary WKB
                 columns as geometry columns or correct metadata such as the
-                column CRS. If a dict is provided, it is JSON-serialized before
-                calling the underlying implementation.
+                column CRS.
 
-                Supported keys (others in the spec are not implemented):
+                Supported keys:
                 - encoding: "WKB" (required)
                 - crs: (e.g., "EPSG:4326")
                 - edges: "planar" (default) or "spherical"
-                See spec for details: https://geoparquet.org/releases/v1.1.0/
+                - ...other supported keys
+                See the specification for details: https://geoparquet.org/releases/v1.1.0/
 
                 Useful for:
                 - Legacy Parquet files with Binary columns containing WKB payloads.
@@ -176,8 +176,10 @@ class SedonaContext:
 
 
                 Safety:
-                - Columns specified here are not validated for WKB correctness.
-                  Invalid WKB payloads may cause undefined behavior.
+                - Columns specified here are not validated against the provided options
+                  (e.g., WKB encoding checks); inconsistent data may cause undefined
+                  behavior.
+
 
         Examples:
 


2010YOUY01:
I would love to hear feedbacks on the design and specification(in PR writeup)

Once we reach agreement on that, I will:
- Add comprehensive tests
- Polish the implementation — so far I have only validated the high-level structure; I haven’t yet reviewed the details (e.g. metadata parsing) carefully

python/sedonadb/python/sedonadb/context.py
```diff
@@ -126,6 +126,7 @@ def read_parquet(
         self,
         table_paths: Union[str, Path, Iterable[str]],
         options: Optional[Dict[str, Any]] = None,
+        geometry_columns: Optional[Dict[str, Union[str, Dict[str, Any]]]] = None,
```
paleolimbot: I would probably make this just `Optional[Mapping[str, Any]]`. In Python a user can rather easily decode that from JSON if they need to.

python/sedonadb/python/sedonadb/context.py
```diff
@@ -134,13 +135,46 @@ def read_parquet(
...
+            geometry_columns: Optional mapping of column name to GeoParquet column
+                metadata (e.g., {"geom": {"encoding": "WKB"}}). Use this to mark
+                binary WKB columns as geometry columns. Supported keys:
+                - encoding: "WKB" (required)
+                - crs: string (e.g., "EPSG:4326") or integer SRID (e.g., 4326).
+                  If not provided, the default CRS is OGC:CRS84
+                  (https://www.opengis.net/def/crs/OGC/1.3/CRS84), which means
+                  the data in this column must be stored in longitude/latitude
+                  based on the WGS84 datum.
+                - edges: "planar" (default) or "spherical"
```
paleolimbot, line 148: Perhaps just "See the GeoParquet specification for the required fields for each column".

https://geoparquet.org/releases/v1.1.0/

python/sedonadb/python/sedonadb/context.py
```diff
@@ -134,13 +135,46 @@ def read_parquet(
                 files.
             options: Optional dictionary of options to pass to the Parquet reader.
                 For S3 access, use {"aws.skip_signature": True, "aws.region": "us-west-2"} for anonymous access to public buckets.
+            geometry_columns: Optional mapping of column name to GeoParquet column
+                metadata (e.g., {"geom": {"encoding": "WKB"}}). Use this to mark
+                binary WKB columns as geometry columns. Supported keys:
```
```suggestion
                metadata (e.g., {"geom": {"encoding": "WKB"}}). Use this to mark
                binary WKB columns as geometry columns or correct metadata such
                as the column CRS. Supported keys:
```

python/sedonadb/src/context.rs
```diff
@@ -30,6 +34,125 @@ use crate::{
     udf::PySedonaScalarUdf,
 };
 
+fn parse_geometry_columns<'py>(
+    py: Python<'py>,
+    geometry_columns: HashMap<String, PyObject>,
+) -> Result<HashMap<String, GeoParquetColumnMetadata>, PySedonaError> {
```
paleolimbot: I think this bit can be avoided by just passing a string at this point (i.e., in Python, use `json.dumps()` before passing to Rust).

python/sedonadb/src/context.rs
```diff
@@ -80,6 +203,7 @@ impl InternalContext {
         py: Python<'py>,
         table_paths: Vec<String>,
         options: HashMap<String, PyObject>,
+        geometry_columns: Option<HashMap<String, PyObject>>,
```
paleolimbot:
```suggestion
        geometry_columns: Option<String>,
```

..I think JSON is the right format for this particular step (reduces bindings code considerably!)

python/sedonadb/tests/test_context.py
```diff
@@ -100,6 +104,38 @@ def test_read_parquet_options_parameter(con, geoarrow_data):
...
 
+# Basic test for `geometry_columns` option for `read_parquet(..)`
+def test_read_parquet_geometry_columns_roundtrip(con, tmp_path):
+    # Write regular parquet table with wkb column in Binary type
+    geom = shapely.from_wkt("POINT (0 1)").wkb
+    table = pa.table({"id": [1], "geom": [geom]})
+    src = tmp_path / "plain.parquet"
+    pq.write_table(table, src)
+
+    # Check metadata: geoparquet meatadata should not be available
```
paleolimbot:
```suggestion
    # Check metadata: geoparquet metadata should not be available
```

python/sedonadb/tests/test_context.py
```diff
@@ -100,6 +104,38 @@ def test_read_parquet_options_parameter(con, geoarrow_data):
...
+
+    out = tmp_path / "geo.parquet"
+    df.to_parquet(out)
+    metadata = pq.read_metadata(out).metadata
+    assert metadata is not None
+    geo = metadata.get(b"geo")
+    assert geo is not None
+    geo_metadata = json.loads(geo.decode("utf-8"))
+    print(json.dumps(geo_metadata, indent=2, sort_keys=True))
+    assert geo_metadata["columns"]["geom"]["crs"] == "EPSG:4326"
```
paleolimbot: I think you can probably skip this bit of the test (verifying the geometry-ness and CRS of the input seems reasonable to me).

rust/sedona-geoparquet/src/provider.rs
```diff
@@ -81,6 +89,7 @@ pub async fn geoparquet_listing_table(
 pub struct GeoParquetReadOptions<'a> {
     inner: ParquetReadOptions<'a>,
     table_options: Option<HashMap<String, String>>,
+    geometry_columns: Option<HashMap<String, GeoParquetColumnMetadata>>,
```
paleolimbot, line 86:
```suggestion
    geometry_columns: Option<String>,
```

...just keeping this as a String will make this work for SQL, too (easy to import from the `HashMap<String, String>` that DataFusion gives us)

rust/sedona-schema/src/crs.rs
```diff
@@ -51,7 +51,7 @@ pub fn deserialize_crs(crs_str: &str) -> Result<Crs> {
     }
 
     // Handle JSON strings "OGC:CRS84", "EPSG:4326", "{AUTH}:{CODE}" and "0"
-    let crs = if LngLat::is_str_lnglat(crs_str) {
+    let crs = if crs_str == "OGC:CRS84" {
```
paleolimbot: These changes should be reverted (there is >1 string that can represent lon/lat)

rust/sedona-schema/src/crs.rs
```diff
@@ -88,6 +88,10 @@ pub fn deserialize_crs_from_obj(crs_value: &serde_json::Value) -> Result<Crs> {
         }
     }
 
+    if let Some(number) = crs_value.as_number() {
```
paleolimbot: This part is OK (but perhaps add a test)

2010YOUY01:
> Mostly I think you can avoid exposing `GeoParquetMetadata` via the options and just accept a string for that parameter. I believe `serde_json` can automatically deserialize that for you to avoid the parsing code here.
> 
> https://github.com/apache/sedona-db/blob/3f91e2657c2473dc06357999dec9282d5e45e35a/rust/sedona-geoparquet/src/metadata.rs#L292-L293
> 
> Exposing a `HashMap<GeoParquetColumnMetadata>` in the options is OK, too if you feel strongly about it (probably helpful if this is being used from Rust), but for our bulit-in frontends (Python, R, SQL) a String is easier to deal with.

It's a great idea to make the rust internal API easier to use for different frontend bindings, WDYT:
- For the builder of Rust Struct `GeoParquetReadOptions`, it takes string for Json options to make it easier to use
- Inside `GeoParquetReadOptions`, it keeps typed/parsed field for `geometry_columns`, to make the rust backend impl cleaner

Now the API looks like:
```rust
pub struct GeoParquetReadOptions<'a> {
    inner: ParquetReadOptions<'a>,
    table_options: Option<HashMap<String, String>>,
    // Keep it typed to make backend impl cleaner
    geometry_columns: Option<HashMap<String, GeoParquetColumnMetadata>>,
}

impl GeoParquetReadOptions {
    // ...
    pub fn with_geometry_columns(
        mut self,
        // Json config string like "{"geo": {"encoding": "wkb"}}"
        geometry_columns: String,
    ) -> Self {
        let parsed = parse_geometry_columns(geometry_columns);
        self.geometry_columns = Some(parsed);
        self
    }
}
```

2010YOUY01:
This PR is reworked, the TLDR for the option semantics is:

1. For a regular parquet file with binary column (but physically WKB-encoded), use this option to specify Binary column as geometry
```
sd.read_parquet(
    "geo_legacy.parquet",
    geometry_columns={
        "geometry": {"encoding": "WKB", "crs": "EPSG:4326", "edges": "planar"}
    },
)
```
2. If a column is already geometry (inferred from parquet metadata), this option can be used to provide optional but missing field; if one field is already inferred from metadata, and set again from the option, an error occur. This feels safer to me, but I'm open to other opinions.
```
# Inferred option from metadata:
#     {"encoding": "WKB"} # "crs" is missing

# Provided 'crs' option from `geometry_columns` is allowed
sd.read_parquet(
    "geo.parquet",
    geometry_columns={
        "geometry": {"crs": "EPSG:4326"}
    },
)
# Now 'geometry' column is a geometry column with crs=4326
```

```
# Inferred option from metadata:
#     {"encoding": "WKB", "crs": "EPSG:4326"}

# Not allowed to provide option that is already inferred from metadata
sd.read_parquet(
    "geo.parquet",
    geometry_columns={
        "geometry": {"crs": "EPSG:3857"}
    },
)
# Errors...
```

## Implementation/Key changes
```text
(existing)
geoparquet metadata --> (per col) GeoParquetColumnMetadata --> schema

(PR)
geoparquet metadata --> (per col) GeoParquetColumnMetadata ----+
                                                               | (combine)
                                                               |
user option geometry_columns --> GeoParquetColumnMetadata -----+--> schema
```
1. Parse option with `serde_json::from_str`, the same as parquet metadata, and store the column options inside `GeoParquetFormat -> TableGeoParquetOption`, since `FileFormat` trait is used to build schema. When `infer_schema()` is called, combine the `GeoParquetColumnMetadata` from both metadata and `geometry_columns` option.
2. Refactor the `GeoParquetColumnMetadata` to make its `encoding` field optional. Since this is a required field for GeoParquet spec, assertions are added to the existing deserializer to ensure it exist in the parquet metadata

python/sedonadb/python/sedonadb/context.py
```diff
@@ -151,7 +198,9 @@ def read_parquet(
 
         return DataFrame(
             self._impl,
-            self._impl.read_parquet([str(path) for path in table_paths], options),
+            self._impl.read_parquet(
+                [str(path) for path in table_paths], options, geometry_columns
+            ),
```
paleolimbot, line 204: Optional, but accepting a dictionary here would probably be an easy/nice quality of life improvement for the Python user.

```suggestion
            if geometry_columns is not None and not isinstance(geometry_columns, str):
                geometry_columns = json.dumps(geometry_columns)

            self._impl.read_parquet(
                [str(path) for path in table_paths], options, geometry_columns
            ),
```

python/sedonadb/tests/test_context.py
```diff
@@ -14,10 +14,37 @@
...
+
+def _parse_geo_metadata(geoparquet_path: Path) -> Mapping[str, Any]:
+    """Return the GeoParquet "geo" metadata map, asserting it exists."""
+    metadata = pq.read_metadata(geoparquet_path).metadata
+    assert metadata is not None
+
+    geo = metadata.get(b"geo")
+    assert geo is not None
+
+    return json.loads(geo.decode("utf-8"))
```
paleolimbot: Optional nit:

```suggestion
    return json.loads(geo.decode())
```

rust/sedona-geoparquet/src/provider.rs
```diff
@@ -213,23 +244,32 @@ impl ReadOptions<'_> for GeoParquetReadOptions<'_> {
...
             options.format = Arc::new(GeoParquetFormat::new(geoparquet_options));
             return options;
         }
 
         unreachable!("GeoParquetReadOptions with non-ParquetFormat ListingOptions");
     }
 
+    /// Infer schema from GeoParquet metadata, then apply the user option
+    /// `geometry_columns` from `read_parquet()` to override if provided. See the
+    /// Python DataFrame `read_parquet(..)` documentation for details.
```
paleolimbot:
```suggestion
```

I'm not sure this comment is the best place to document the rules for calculating a GeoParquet schema (which are about to change again with Parquet geometry/geography).

rust/sedona-geoparquet/src/provider.rs
```diff
@@ -185,13 +188,41 @@ impl GeoParquetReadOptions<'_> {
...
 
     /// Get the table options
     pub fn table_options(&self) -> Option<&HashMap<String, String>> {
         self.table_options.as_ref()
     }
+
+    /// Add geometry column metadata (JSON string) to apply during schema resolution
+    /// See python `read_parquet(..)` comments for details.
+    ///
+    /// Errors if invalid json configuration string is provided.
```
paleolimbot:
```suggestion
    ///
    /// Reads Parquet files as if GeoParquet metadata with the `"geometry_columns"`
    /// key were present. If GeoParquet metadata is already present, the values provided
    /// here will override any definitions provided in the original metadata.
    /// Errors if an invalid JSON configuration string is provided.
```

rust/sedona-geoparquet/src/lib.rs
```diff
@@ -20,3 +20,5 @@ mod metadata;
 pub mod options;
 pub mod provider;
 mod writer;
+
+pub use metadata::{GeoParquetColumnEncoding, GeoParquetColumnMetadata};
```
paleolimbot: I am not sure we want to expose these in this way...these are more like internal utilities at the moment (they could be exposed later if there is a good reason to do so and we're confident they won't change).

rust/sedona-geoparquet/src/format.rs
```diff
@@ -146,6 +150,114 @@ impl GeoParquetFormat {
...
+/// Merge geometry columns metadata.
+/// `overrides` columns may only provide additional data; for example,
+/// if `crs` is None (missing) in the `base` metadata, the combined
+/// metadata uses the `crs` field from `overrides`.
+///
+/// If conflicting field data is provided, returns a plan error.
+fn merge_geometry_columns(
+    base: &mut HashMap<String, GeoParquetColumnMetadata>,
+    overrides: &HashMap<String, GeoParquetColumnMetadata>,
+) -> Result<()> {
```
paleolimbot, line 158: I think we can make this quite a bit simpler by just overriding columns instead of attempting to merge them, which serves your specific use case and I think is closer to what people actually want to do (ensure that a column is interpreted in a particular way regardless of what is specified by the file).

rust/sedona-geoparquet/src/format.rs
```diff
@@ -167,6 +279,9 @@ impl FileFormat for GeoParquetFormat {
         self.inner().compression_type()
     }
 
+    /// Infer schema from GeoParquet metadata, then optionally override it using
+    /// user-provided geometry column options (see Python DataFrame API
+    /// `read_parquet(..)` `geometry_columns` for details).
```
paleolimbot:
```suggestion
```

The rules for this are about to change with Parquet geometry/geography support and in that PR I will try to find a good place to document the rules for schema inference based on the file schema, GeoParquet metadata, and overrides.

rust/sedona-geoparquet/src/format.rs
paleolimbot, line 231: Can we apply the column overrides here and eliminate the somewhat complicated logic below?

Note that in https://github.com/apache/sedona-db/pull/561 this is simplified to just use `try_from_parquet_metadata()`.

rust/sedona-geoparquet/src/format.rs
```diff
@@ -222,38 +339,66 @@ impl FileFormat for GeoParquetFormat {
...
-            Ok(Arc::new(Schema::new(new_fields?)))
-        } else {
-            Ok(inner_schema_without_metadata)
+        if !remaining.is_empty() {
+            let mut missing: Vec<_> = remaining.into_iter().collect();
+            missing.sort();
+            return plan_err!(
+                "Geometry columns not found in schema: {}",
+                missing.join(", ")
+            );
```
paleolimbot, line 292: It's great to fix other issues with the Parquet Schema/GeoParquetMetadata interactions like this one (but if we do it in this PR we should add a test). I can also do these as part of the Parquet geometry/geometry PR since that further complicates the interaction.

python/sedonadb/python/sedonadb/context.py
```diff
@@ -134,14 +135,60 @@ def read_parquet(
...
+                  and geometry_columns provides:
+                  - `geo: {"crs": 4326}`
+                  then the result is (only override provided keys):
+                  - `geo: {"encoding": "wkb", "crs": "EPSG:4326", "edges": "spherical"...}`
+                - If `geo` inferred from metadata has:
+                  - `geo: {"encoding": "wkb", "crs": "EPSG:4326"}`
+                  and geometry_columns provides:
+                  - `geo: {"crs": "EPSG:3857"}`
+                  an error is returned for a conflicting key. This option is only
+                  allowed to provide missing optional fields in geometry columns.
```
```suggestion
```

If constraining this option to only ever override the columns provided by the file we can make this interaction considerably simpler.

python/sedonadb/python/sedonadb/context.py
2010YOUY01: This is a good point, updated in [5113c2f](https://github.com/apache/sedona-db/pull/560/commits/5113c2fc8c5f1836f910fa6c22e750c199ebe65d)

python/sedonadb/python/sedonadb/context.py
2010YOUY01: addressed in [5113c2f](https://github.com/apache/sedona-db/pull/560/commits/5113c2fc8c5f1836f910fa6c22e750c199ebe65d)

rust/sedona-geoparquet/src/format.rs
2010YOUY01: Added a test case to cover this in [dd75822](https://github.com/apache/sedona-db/pull/560/commits/dd75822c1bf9e39c17e68ad3c2a70f8c19542acb)

rust/sedona-geoparquet/src/format.rs
2010YOUY01: I think this should not be applicable after we change to the fully overwrite semantics? This combining step is only merging `bbox` and geo types, and reject any other conflicts, and the overwriting step has to be done separately.

rust/sedona-geoparquet/src/format.rs
2010YOUY01: cleaned up in [dd75822](https://github.com/apache/sedona-db/pull/560/commits/dd75822c1bf9e39c17e68ad3c2a70f8c19542acb)

rust/sedona-geoparquet/src/lib.rs
2010YOUY01: cleaned up in [dd75822](https://github.com/apache/sedona-db/pull/560/commits/dd75822c1bf9e39c17e68ad3c2a70f8c19542acb)

rust/sedona-geoparquet/src/provider.rs
2010YOUY01: Updated in [dd75822](https://github.com/apache/sedona-db/pull/560/commits/dd75822c1bf9e39c17e68ad3c2a70f8c19542acb)

rust/sedona-geoparquet/src/provider.rs
2010YOUY01: Cleaned up in [dd75822](https://github.com/apache/sedona-db/pull/560/commits/dd75822c1bf9e39c17e68ad3c2a70f8c19542acb)

python/sedonadb/tests/test_context.py
2010YOUY01: [dd75822](https://github.com/apache/sedona-db/pull/560/commits/dd75822c1bf9e39c17e68ad3c2a70f8c19542acb)

2010YOUY01:
> Thank you!
> 
> This is great! My main remaining high-level comment is that I think we should just have the `geometry_columns` be a pure override and not attempt to merge anything. I think this is simpler but also potentially more useful (can be used as an escape hatch to read a wider variety of input that provides incomplete or incorrect information that is difficult to otherwise fix).
> 
> Note that there is some minor overlap with #561 ...I'd prefer to merge your PR first and I can handle whatever merge conflict may arise.

Thank you for the review — that makes sense. I’ve switched to pure override semantics. Just realized for geoparquets with incorrect/missing metadata, `sedonadb` is likely the tool to fix them, so we need full overwrites here.

For safety, adding validation support can be a follow-up.

rust/sedona-geoparquet/src/format.rs
paleolimbot: I can give this a try in the Parquet geometry/geography PR. I think we want to apply them before the call to `try_update()` so the overrides can be used to avoid an erroneous or difficult to work around conflict.

PR #562, opened 2026-01-30
fix(rust/sedona-spatial-join) Reimplement planner for spatial join

### LINKED ISSUE #282: Revise compute_properties of SpatialJoinExec
The [`compute_properties`](https://github.com/apache/sedona-db/blob/9cecb5b42e7ed2e5c234d61c24ea286ba3d58245/rust/sedona-spatial-join/src/exec.rs#L230-L248) and other output data properties such as [`maintains_input_order`](https://github.com/apache/sedona-db/blob/9cecb5b42e7ed2e5c234d61c24ea286ba3d58245/rust/sedona-spatial-join/src/exec.rs#L203-L223) of SpatialJoinExec may be incorrect.

* `maintains_input_order` says that the order of the right side is maintained when running an inner/right/right-anti/right-semi join, which is incorrect for KNN join, since the left side might be the probe side.
* `compute_properties` may not be correct when the SpatialJoinExec was converted from a HashJoin. It states that the [output partitioning of the left side could be maintained](https://github.com/apache/sedona-db/blob/9cecb5b42e7ed2e5c234d61c24ea286ba3d58245/rust/sedona-spatial-join/src/exec.rs#L274-L278), but this is not the case.

We can change the compute_properties and maintains_input_order to be more conservative, but that will break the data distribution and ordering requirements of the downstream physical operators and fail the [query plan sanity check](https://github.com/apache/sedona-db/blob/9cecb5b42e7ed2e5c234d61c24ea286ba3d58245/rust/sedona-spatial-join/src/optimizer.rs#L562), unless we run enforce distribution and enforce sorting rules again after swapping NestedLoopJoinExec or HashJoinExec.

A better way is to inject our custom physical planning rules and emit SpatialJoinExec directly for logical join operator in the planning phase. We chose to mess around with physical plan optimizations rather than extending the planner because DataFusion does not have a good API for plugging in custom planner. We'll take another look into this issue and see how can we do this cleanly.

Kontinuation:
Closes #282 

Emit `SpatialJoinExec` directly from logical plan instead of performing the error-prone physical -> physical plan translation (NestedLoopJoinExec/HashJoinExec -> SpatialJoinExec). We follow the standard practice of doing this in DataFusion, see https://github.com/apache/datafusion/blob/52.1.0/datafusion/core/tests/user_defined/user_defined_plan.rs

We have also revised the plan properties of SpatialJoinExec. The plan properties were all wrong for KNN join, this patch fixed that and make sure that we won't break the data distribution enforcement.

### DIFF
--- a/rust/sedona-spatial-join/src/planner/logical_plan_node.rs
+++ b/rust/sedona-spatial-join/src/planner/logical_plan_node.rs
@@ -19,9 +19,10 @@ use std::cmp::Ordering;
 use std::fmt;
 use std::sync::Arc;
 
-use datafusion_common::{plan_err, DFSchemaRef, NullEquality, Result};
+use datafusion_common::{DFSchemaRef, NullEquality, Result};
 use datafusion_expr::logical_plan::UserDefinedLogicalNodeCore;
 use datafusion_expr::{Expr, JoinConstraint, JoinType, LogicalPlan};
+use sedona_common::sedona_internal_err;
 
 /// Logical extension node used as a planning hook for spatial joins.
 ///
--- a/rust/sedona-spatial-join/src/planner/physical_planner.rs
+++ b/rust/sedona-spatial-join/src/planner/physical_planner.rs
@@ -43,14 +43,18 @@ use sedona_common::option::SedonaOptions;
 
 /// Registers a query planner that can produce [`SpatialJoinExec`] from a logical extension node.
 pub(crate) fn register_spatial_join_planner(builder: SessionStateBuilder) -> SessionStateBuilder {
-    builder.with_query_planner(Arc::new(SedonaSpatialQueryPlanner))
+    let extension_planners: Vec<Arc<dyn ExtensionPlanner + Send + Sync>> =
+        vec![Arc::new(SpatialJoinExtensionPlanner {})];
+    builder.with_query_planner(Arc::new(SedonaSpatialQueryPlanner { extension_planners }))
 }
 
 /// Query planner that enables Sedona's spatial join planning.
 ///
 /// Installs an [`ExtensionPlanner`] that recognizes `SpatialJoinPlanNode` and produces
 /// `SpatialJoinExec` when supported and enabled.
-pub struct SedonaSpatialQueryPlanner;
+struct SedonaSpatialQueryPlanner {
+    extension_planners: Vec<Arc<dyn ExtensionPlanner + Send + Sync>>,
+}
 
 impl fmt::Debug for SedonaSpatialQueryPlanner {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
@@ -65,9 +69,8 @@ impl QueryPlanner for SedonaSpatialQueryPlanner {
         logical_plan: &LogicalPlan,
         session_state: &SessionState,
     ) -> Result<Arc<dyn ExecutionPlan>> {
-        let physical_planner = DefaultPhysicalPlanner::with_extension_planners(vec![Arc::new(
-            SpatialJoinExtensionPlanner {},
-        )]);
+        let physical_planner =
+            DefaultPhysicalPlanner::with_extension_planners(self.extension_planners.clone());
         physical_planner
             .create_physical_plan(logical_plan, session_state)
             .await


rust/sedona-spatial-join/src/planner/logical_plan_node.rs
```diff
@@ -0,0 +1,129 @@
...
+            join_constraint: &self.join_constraint,
+            null_equality: &self.null_equality,
+        };
+        let comparable_other = ComparableJoin {
+            left: &other.left,
+            right: &other.right,
+            filter: &other.filter,
+            join_type: &other.join_type,
+            join_constraint: &other.join_constraint,
+            null_equality: &self.null_equality,
```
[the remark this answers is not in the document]
Kontinuation: Fixed

rust/sedona-spatial-join/src/planner/optimizer.rs
```diff
@@ -0,0 +1,231 @@
...
+/// Logical optimizer rule that enables spatial join planning.
+///
+/// This rule turns eligible `Filter(Join(filter=...))` nodes into a `Join(filter=...)` node,
+/// so that the spatial join can be rewritten later by [SpatialJoinLogicalRewrite].
+#[derive(Debug, Default)]
+struct MergeSpatialProjectionIntoJoin;
+
+impl OptimizerRule for MergeSpatialProjectionIntoJoin {
+    fn name(&self) -> &str {
+        "spatial_join_optimizer"
```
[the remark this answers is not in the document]
Kontinuation: Fixed

rust/sedona-spatial-join/src/utils/join_utils.rs
```diff
@@ -715,11 +778,154 @@ pub(crate) fn boundedness_from_children<'a>(
...
+    join_filter: Option<&JoinFilter>,
+    join_on: &SpatialPredicate,
+) -> Result<Option<JoinPushdownData>> {
+    let Some(projection_as_columns) = physical_to_column_exprs(projection.expr()) else {
+        return Ok(None);
+    };
+
+    // Mark joins produce a synthetic column that does not belong to either child. This synthetic
+    // `mark` column will make `new_join_children` fail, so we skip pushdown for such joins.
+    // This limitation if inherited from DataFusion's builtin `try_pushdown_through_join`.
```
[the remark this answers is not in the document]
Kontinuation: Fixed

rust/sedona-spatial-join/src/planner/physical_planner.rs
```diff
@@ -0,0 +1,263 @@
...
+        let Some(ext) = session_state
+            .config_options()
+            .extensions
+            .get::<SedonaOptions>()
+        else {
+            return sedona_internal_err!("SedonaOptions not found in session state extensions");
+        };
+
+        if !ext.spatial_join.enable {
+            return sedona_internal_err!("Spatial join is disabled in SedonaOptions");
```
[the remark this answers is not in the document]
Kontinuation: This is intentional. There must be some programming error somewhere when we get a SpatialJoinPlanNode with spatial join feature disabled.

rust/sedona-spatial-join/src/exec.rs
```diff
@@ -212,61 +180,89 @@ impl SpatialJoinExec {
...
-            false,
-            matches!(
-                join_type,
-                JoinType::Inner | JoinType::Right | JoinType::RightAnti | JoinType::RightSemi
-            ),
-        ]
+    /// This function should be called BEFORE inserting any repartitioning
+    /// operators on the join's children. Check [`super::HashJoinExec::swap_inputs`]
+    /// for more details.
+    pub fn swap_inputs(&self) -> Result<Arc<dyn ExecutionPlan>> {
```
Kontinuation, line 196: Now we have a `swap_inputs` function. @zhangfengcdt

rust/sedona-spatial-join/src/planner/logical_plan_node.rs
```diff
@@ -0,0 +1,129 @@
...
+        )
+    }
+
+    fn with_exprs_and_inputs(
+        &self,
+        mut exprs: Vec<Expr>,
+        mut inputs: Vec<LogicalPlan>,
+    ) -> Result<Self> {
+        if exprs.len() != 1 {
+            return plan_err!("SpatialJoinPlanNode expects 1 expr");
```
paleolimbot:
```suggestion
            return sedona_internal_err!("SpatialJoinPlanNode expects 1 expr");
```

rust/sedona-spatial-join/src/planner/logical_plan_node.rs
```diff
@@ -0,0 +1,129 @@
...
+        }
+        if inputs.len() != 2 {
+            return plan_err!("SpatialJoinPlanNode expects 2 inputs");
```
paleolimbot:
```suggestion
            return sedona_internal_err!("SpatialJoinPlanNode expects 2 inputs");
```

(Unless a user can construct this incorrectly?)

rust/sedona-spatial-join/src/planner/logical_plan_node.rs
```diff
@@ -0,0 +1,129 @@
...
+use datafusion_common::{plan_err, DFSchemaRef, NullEquality, Result};
+use datafusion_expr::logical_plan::UserDefinedLogicalNodeCore;
+use datafusion_expr::{Expr, JoinConstraint, JoinType, LogicalPlan};
+
+/// Logical extension node used as a planning hook for spatial joins.
+///
+/// Carries a join's inputs and filter expression so the physical planner can recognize and plan
+/// a `SpatialJoinExec`.
+#[derive(PartialEq, Eq, Hash)]
+pub(crate) struct SpatialJoinPlanNode {
```
paleolimbot, line 32: Not in this PR, but we can also make this pub and construct it directly from R and Python DataFrame APIs (e.g., `df.sjoin(...)`)

rust/sedona-spatial-join/src/planner/optimizer.rs
```diff
@@ -0,0 +1,231 @@
...
+        };
+        if !ext.spatial_join.enable {
+            return Ok(Transformed::no(plan));
+        }
+
+        let LogicalPlan::Join(join) = &plan else {
+            return Ok(Transformed::no(plan));
+        };
+
+        // v1: only rewrite joins that already have a spatial predicate in `filter`.
```
paleolimbot:
```suggestion
        // only rewrite joins that already have a spatial predicate in `filter`.
```

rust/sedona-spatial-join/src/planner/physical_planner.rs
```diff
@@ -0,0 +1,263 @@
...
+/// Registers a query planner that can produce [`SpatialJoinExec`] from a logical extension node.
+pub fn register_spatial_join_planner(builder: SessionStateBuilder) -> SessionStateBuilder {
+    builder.with_query_planner(Arc::new(SedonaSpatialQueryPlanner))
+}
+
+/// Query planner that enables Sedona's spatial join planning.
+///
+/// Installs an [`ExtensionPlanner`] that recognizes `SpatialJoinPlanNode` and produces
+/// `SpatialJoinExec` when supported and enabled.
+pub struct SedonaSpatialQueryPlanner;
```
paleolimbot: Because we only get one physical planner, this should probably be always be the planner when constructing a `SedonaContext` (i.e., this struct can live in `sedona` and always be the planner). I'm not sure there's a high chance of many more extension logical nodes but that is also where we do things like interpret SQL and potentially pre-process the logical plan (this would be the opportunity to post-process the logical plan) and that should probably be in the same place.

```suggestion
/// Query planner that enables Sedona's spatial join planning.
///
/// Installs an [`ExtensionPlanner`] that recognizes `SpatialJoinPlanNode` and produces
/// `SpatialJoinExec` when supported and enabled.
pub struct SedonaSpatialQueryPlanner { extension_planner: Vec<Arc<ExtensionPlanner>> };
```

rust/sedona-spatial-join/src/planner/optimizer.rs
```diff
@@ -0,0 +1,231 @@
...
+/// Register only the logical spatial join optimizer rule.
+///
+/// This enables building `Join(filter=...)` from patterns like `Filter(CrossJoin)`.
+/// It intentionally does not register any physical plan rewrite rules.
+pub fn register_spatial_join_logical_optimizer(
+    session_state_builder: SessionStateBuilder,
+) -> SessionStateBuilder {
+    session_state_builder
+        .with_optimizer_rule(Arc::new(MergeSpatialProjectionIntoJoin))
+        .with_optimizer_rule(Arc::new(SpatialJoinLogicalRewrite))
```
paleolimbot, line 41: Do we have the ability to ensure that these run before the `pushdown_filters` rule to ensure that filters aren't pushed through a KNN join?

rust/sedona-spatial-join/src/planner.rs
```diff
@@ -0,0 +1,40 @@
...
+
+mod logical_plan_node;
+mod optimizer;
+mod physical_planner;
+mod spatial_expr_utils;
+
+/// Register Sedona spatial join planning hooks.
+///
+/// Enables logical rewrites (to surface join filters) and a query planner extension that can
+/// plan `SpatialJoinExec`.
```
paleolimbot:
```suggestion
/// Register Sedona spatial join planning hooks.
///
/// Enables logical rewrites (to surface join filters) and a query planner extension that can
/// plan `SpatialJoinExec`. This is the primary entry point to leveraging the spatial join
/// implementation provided by this crate and ensures joins created by SQL or using
/// a DataFrame API that meet certain conditions (e.g. contain a spatial predicate as
/// a join condition) are executed using the `SpatialJoinExec`.
```

Just giving this a bit of fanfare since it's the entry point to pretty much everything else.

rust/sedona-spatial-join/src/planner/optimizer.rs
```diff
@@ -0,0 +1,231 @@
...
+use datafusion_expr::logical_plan::Extension;
+use datafusion_expr::{BinaryExpr, Expr, Operator};
+use datafusion_expr::{Filter, Join, JoinType, LogicalPlan};
+use sedona_common::option::SedonaOptions;
+
+/// Register only the logical spatial join optimizer rule.
+///
+/// This enables building `Join(filter=...)` from patterns like `Filter(CrossJoin)`.
+/// It intentionally does not register any physical plan rewrite rules.
+pub fn register_spatial_join_logical_optimizer(
```
paleolimbot: Should these be slightly less public or link to the public entrypoint that should actually be used?

```suggestion
pub(crate) fn register_spatial_join_logical_optimizer(
```

rust/sedona-spatial-join/src/planner/physical_planner.rs
```diff
@@ -0,0 +1,263 @@
...
+use sedona_common::sedona_internal_err;
+
+use crate::exec::SpatialJoinExec;
+use crate::planner::logical_plan_node::SpatialJoinPlanNode;
+use crate::planner::spatial_expr_utils::{is_spatial_predicate_supported, transform_join_filter};
+use crate::spatial_predicate::SpatialPredicate;
+use sedona_common::option::SedonaOptions;
+
+/// Registers a query planner that can produce [`SpatialJoinExec`] from a logical extension node.
+pub fn register_spatial_join_planner(builder: SessionStateBuilder) -> SessionStateBuilder {
```
paleolimbot:
```suggestion
/// Registers a query planner that can produce [`SpatialJoinExec`] from a logical extension node.
pub(crate) fn register_spatial_join_planner(builder: SessionStateBuilder) -> SessionStateBuilder {
```

Should this be slightly less public or link to the function people should more likely be using?

rust/sedona-spatial-join/src/planner/spatial_expr_utils.rs
```diff
@@ -14,566 +14,95 @@
...
-        } else {
-            expressions
-                .into_iter()
-                .reduce(|acc, expr| {
-                    Arc::new(BinaryExpr::new(acc, Operator::And, expr)) as Arc<dyn PhysicalExpr>
-                })
-                .unwrap()
+    fn is_distance_expr(expr: &Expr) -> bool {
+        let Expr::ScalarFunction(datafusion_expr::expr::ScalarFunction { func, .. }) = expr else {
+            return false;
```
paleolimbot, line 91: Perhaps not for this PR, but some of these may fit well in sedona-expr (so we can use them in future optimizer rules)

paleolimbot (approved):
Awesome!

A few suggestions about where some of this can live but feel free to punt on any of this.

I forget exactly how most of our fuzz testing is done but I am assuming that it still holds for the rewrite (I seem to recall it was SQL which should go through the paths that were changed here?)

I think we may need to update the ordering of the rules still to ensure we get proper filter pushdown through KNN joins (or maybe there's already a test for this)

Kontinuation:
I need to clarity that this PR does not change the behavior of filter pushdown for KNN joins. The original problems with KNN joins remain the same, and users still need to use barriers to explicitly workaround that problem.

This PR mainly fixes the correctness of plan properties of SpatialJoinExec, and avoid breaking the plan property requirements during the physical plan -> physical plan transformation, which may result in physical plan sanity check failures or silently producing incorrect results.

The optimization rules for merging filter into Join node and converting Join nodes to SpatialJoinPlanNode all happen after DataFusion's builtin optimization passes, so filter pushdown still behave the same as before.

rust/sedona-spatial-join/src/planner/optimizer.rs
Kontinuation: No, this is not within the scope of this PR. We need to add some early stage passes to convert Join nodes for KNN join to something else even before the logical plan optimization starts.

rust/sedona-spatial-join/src/planner/physical_planner.rs
Kontinuation: Good idea. I have added `extension_planner` and also made `SedonaSpatialQueryPlanner` private.

Kontinuation:
I prefer fixing filter pushdown for KNN join in a separate PR, since this PR is already self-contained and pretty large. We can do the following things in a follow-up PR:

1. Prepending a rule before `PushDownFilter` for converting regular join to SpatialJoinExecPlan to prevent incorrect filter pushdown
2. Disallow join conditions such as `<expr> AND ST_KNN(...)`. We can allow `ST_KNN(...) AND <expr>` and apply `Filter(<expr>)` to the result of KNN join.
3. Preserving the regular join node for non-KNN spatial joins to allow DataFusion's builtin passes to optimize it, and convert it to SpatialJoinExecPlan in the last few optimization rules
4. Adding filter pushdown test cases, validating that filter pushdown happens for non-KNN join, but not for KNN join.

This stage does not support pushing filters down to the query side of the KNN join. We can implement the optimization rules discussed by https://ieeexplore.ieee.org/document/5447873 after fixing the most basic semantic correctness issues.

PR #563, opened 2026-01-30
feat(rust/sedona-spatial-join): Support spatial partitioned spatial join for non-knn predicates

Kontinuation:
This patch adds a probe-side partitioned stream provider for repartitioning the probe side, and create probe streams for specified partitions. This partitioned stream provider is integrated into the spatial join execution flow to support larger-than-memory dataset by breaking the data into smaller partitions, where each partition could be fully loaded into memory.

Currently only non-knn joins are supported. We'll add larger-than-memory KNN join support in subsequent patches.

### DIFF
--- a/rust/sedona-spatial-join/src/probe/partitioned_stream_provider.rs
+++ b/rust/sedona-spatial-join/src/probe/partitioned_stream_provider.rs
@@ -344,6 +344,7 @@ mod tests {
     use datafusion_physical_plan::metrics::ExecutionPlanMetricsSet;
     use futures::TryStreamExt;
     use sedona_geometry::bounding_box::BoundingBox;
+    use sedona_geometry::wkb_factory::wkb_point;
     use sedona_schema::datatypes::WKB_GEOMETRY;
     use std::sync::Arc;
 
@@ -362,13 +363,6 @@ mod tests {
         let geom_array: ArrayRef = Arc::new(BinaryArray::from(geom_values));
         let geom_array = EvaluatedGeometryArray::try_new(geom_array, &WKB_GEOMETRY)?;
         Ok(EvaluatedBatch { batch, geom_array })
-    }
-
-    fn point_wkb(x: f64, y: f64) -> Vec<u8> {
-        let mut buf = vec![1u8, 1, 0, 0, 0];
-        buf.extend_from_slice(&x.to_le_bytes());
-        buf.extend_from_slice(&y.to_le_bytes());
-        buf
     }
 
     fn rect_wkb(min_x: f64, min_y: f64, max_x: f64, max_y: f64) -> Vec<u8> {
@@ -445,10 +439,10 @@ mod tests {
         let batch = sample_batch(
             &[0, 1, 2, 3],
             vec![
-                Some(point_wkb(10.0, 10.0)),
-                Some(point_wkb(60.0, 10.0)),
-                Some(point_wkb(70.0, 20.0)),
-                Some(point_wkb(200.0, 200.0)),
+                Some(wkb_point((10.0, 10.0)).unwrap()),
+                Some(wkb_point((60.0, 10.0)).unwrap()),
+                Some(wkb_point((70.0, 20.0)).unwrap()),
+                Some(wkb_point((200.0, 200.0)).unwrap()),
             ],
         )?;
         let probe_stream = create_probe_stream(vec![batch], Some(Arc::clone(&partitioner)));
@@ -473,7 +467,7 @@ mod tests {
     #[tokio::test]
     async fn requesting_regular_partition_before_first_pass_fails() -> Result<()> {
         let partitioner = sample_partitioner()?;
-        let batch = sample_batch(&[0], vec![Some(point_wkb(60.0, 10.0))])?;
+        let batch = sample_batch(&[0], vec![Some(wkb_point((60.0, 10.0)).unwrap())])?;
         let probe_stream = create_probe_stream(vec![batch], Some(Arc::clone(&partitioner)));
         assert!(probe_stream
             .stream_for(SpatialPartition::Regular(1))
@@ -511,7 +505,10 @@ mod tests {
     async fn non_partitioned_stream_exposes_only_partition_zero() -> Result<()> {
         let batch = sample_batch(
             &[0, 1],
-            vec![Some(point_wkb(10.0, 10.0)), Some(point_wkb(12.0, 12.0))],
+            vec![
+                Some(wkb_point((10.0, 10.0)).unwrap()),
+                Some(wkb_point((12.0, 12.0)).unwrap()),
+            ],
         )?;
         let probe_stream = create_probe_stream(vec![batch], None);
 
--- a/rust/sedona-spatial-join/bench/partitioning/stream_repartitioner.rs
+++ b/rust/sedona-spatial-join/bench/partitioning/stream_repartitioner.rs
@@ -32,7 +32,7 @@ use datafusion_execution::runtime_env::RuntimeEnv;
 use datafusion_physical_plan::metrics::{ExecutionPlanMetricsSet, SpillMetrics};
 use futures::executor::block_on;
 use rand::{rngs::StdRng, Rng, SeedableRng};
-use sedona_geometry::{bounding_box::BoundingBox, interval::IntervalTrait};
+use sedona_geometry::{bounding_box::BoundingBox, interval::IntervalTrait, wkb_factory::wkb_point};
 use sedona_schema::datatypes::WKB_GEOMETRY;
 use sedona_spatial_join::evaluated_batch::{
     evaluated_batch_stream::{in_mem::InMemoryEvaluatedBatchStream, SendableEvaluatedBatchStream},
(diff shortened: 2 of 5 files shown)


rust/sedona-spatial-join/src/utils/join_utils.rs
```diff
@@ -53,6 +53,28 @@ pub(crate) fn need_produce_result_in_final(join_type: JoinType) -> bool {
...
+pub(crate) fn need_probe_multi_partition_bitmap(join_type: JoinType) -> bool {
+    matches!(
+        join_type,
+        JoinType::Right
+            | JoinType::RightAnti
+            | JoinType::RightSemi
+            | JoinType::RightMark
+            | JoinType::Full
+    )
+}
```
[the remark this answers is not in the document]
Kontinuation: The bitmap is only needed by the MULTI partition of probe side, the name already states that.

rust/sedona-spatial-join/src/stream.rs
```diff
@@ -185,49 +213,155 @@ impl SpatialJoinProbeMetrics {
...
+            partition_id,
+            partition: SpatialPartition::Regular(partition_id),
+        }
+    }
+
+    fn multi(partition_id: u32) -> Self {
+        Self {
+            partition_id,
+            partition: SpatialPartition::Multi,
+        }
```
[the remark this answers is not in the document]
Kontinuation: There is no problem. We need to process the MULTI partition after processing each regular partition. The partition_id is needed to track where we are at when processing the MULTI partition.

rust/sedona-spatial-join/src/stream.rs
```diff
@@ -524,6 +759,19 @@ impl SpatialJoinStream {
             spatial_index.merge_probe_stats(stats);
         }
 
+        let visited_probe_side = if matches!(partition_desc.partition, SpatialPartition::Multi) {
+            self.visited_multi_probe_side.clone()
+        } else {
+            None
+        };
```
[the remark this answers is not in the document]
Kontinuation: This is not the case. It only happens when creating the iterator. The hot inner loop happens inside the iterator.

rust/sedona-spatial-join/src/probe/partitioned_stream_provider.rs
```diff
@@ -0,0 +1,500 @@
...
+            }
+            SpatialPartition::Multi => {
+                let spilled = self.slots.get_spilled_partition(partition)?;
+                Ok(Box::pin(
+                    ExternalEvaluatedBatchStream::try_from_spill_files(
+                        Arc::clone(&self.schema),
+                        spilled.into_spill_files(),
+                    )?,
+                ))
+            }
```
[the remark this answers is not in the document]
Kontinuation: It is designed to be consumed multiple times and being asymmetric with regular partitions. There is no problem.

rust/sedona-spatial-join/src/probe/first_pass_stream.rs
```diff
@@ -0,0 +1,236 @@
...
+        }
+    }
+
+    fn transition_to_failed(&mut self, err: DataFusionError) -> DataFusionError {
+        let err_arc = Arc::new(err);
+        let callback_opt = self.callback.take();
+        if let Some(callback) = callback_opt {
+            callback
+                .call(Err(DataFusionError::Shared(err_arc.clone())))
+                .ok();
```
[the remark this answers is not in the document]
Kontinuation: Good idea. I'll use log::warn! for this.

rust/sedona-spatial-join/src/prepare.rs
```diff
@@ -316,6 +292,30 @@ impl SpatialJoinComponentsBuilder {
...
+            let mut partition_bounds = Vec::with_capacity(num_partitions);
+            for k in 0..num_partitions {
+                let partition = SpatialPartition::Regular(k as u32);
+                let partition_bound = merged_spilled_partitions
+                    .spilled_partition(partition)?
+                    .bounding_box()
+                    .cloned()
+                    .unwrap_or(BoundingBox::empty());
+                partition_bounds.push(partition_bound);
+            }
```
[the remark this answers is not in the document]
Kontinuation: This lookup is very cheap (array index access). I don't see a better way of implementing this, and this is not on a performance critical spot, so I'll simply leave it as is.

rust/sedona-spatial-join/src/exec.rs
```diff
@@ -627,7 +609,8 @@ impl SpatialJoinExec {
 
 #[cfg(test)]
 mod tests {
-    use arrow_array::{Array, RecordBatch};
+    use arrow_array::Array;
+    use arrow_array::RecordBatch;
```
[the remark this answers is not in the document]
Kontinuation: I'll fix this.

rust/sedona-spatial-join/src/probe/partitioned_stream_provider.rs
```diff
@@ -0,0 +1,531 @@
...
+        let geom_array = EvaluatedGeometryArray::try_new(geom_array, &WKB_GEOMETRY)?;
+        Ok(EvaluatedBatch { batch, geom_array })
+    }
+
+    fn point_wkb(x: f64, y: f64) -> Vec<u8> {
+        let mut buf = vec![1u8, 1, 0, 0, 0];
+        buf.extend_from_slice(&x.to_le_bytes());
+        buf.extend_from_slice(&y.to_le_bytes());
+        buf
+    }
```
paleolimbot: This test helper appears in many of the test files and I believe `sedona_geometry::wkb_factory` has it as well.

rust/sedona-spatial-join/src/probe/partitioned_stream_provider.rs
```diff
@@ -0,0 +1,531 @@
...
+
+    fn rect_wkb(min_x: f64, min_y: f64, max_x: f64, max_y: f64) -> Vec<u8> {
+        let mut buf = Vec::with_capacity(1 + 4 + 4 + 4 + 5 * 16);
+        buf.push(1u8);
+        buf.extend_from_slice(&3u32.to_le_bytes());
+        buf.extend_from_slice(&1u32.to_le_bytes());
+        buf.extend_from_slice(&5u32.to_le_bytes());
```
paleolimbot, line 373: This one also appears in at least one other test file and if it isn't in sedona_geometry it certainly could be added.

Feel free to open a follow-up for this (deduplicate test utilities in sedona-spatial join, or something)

rust/sedona-spatial-join/src/utils/join_utils.rs
```diff
@@ -93,6 +115,161 @@ pub(crate) fn get_final_indices_from_bit_map(
     (left_indices, right_indices)
 }
 
+pub(crate) fn adjust_indices_with_visited_info(
+    left_indices: UInt64Array,
+    right_indices: UInt32Array,
```
paleolimbot, line 120: Is there a reason that left and right indices are not the same type? (Are left indices global and right indices local to a batch?)

rust/sedona-spatial-join/src/utils/join_utils.rs
```diff
@@ -93,6 +115,161 @@ pub(crate) fn get_final_indices_from_bit_map(
...
+
+    match join_type {
+        JoinType::Right | JoinType::Full => {
+            if !produce_unmatched_probe_rows {
+                Ok((left_indices, right_indices))
+            } else {
+                let unmatched_count = adjust_range
+                    .clone()
+                    .filter(|&i| !bitmap.get_bit(i + offset))
+                    .count();
```
paleolimbot, line 150: I doubt this is a bottleneck, but calling `get_bit()` in a loop is not particularly fast. I am not sure if arrow-rs provides an optimized iterator for this operation (in Arrow C++ there are a few options like VisitSetBitRuns that do a better job).

paleolimbot (approved):
🚀 🚀 !

rust/sedona-spatial-join/src/utils/join_utils.rs
Kontinuation: Exactly. This is a variant of DataFusion's `append_right_indices`, the types of `left_indices` and `right_indices` were unchanged.

rust/sedona-spatial-join/src/utils/join_utils.rs
Kontinuation: Unfortunately Arrow's BooleanBufferBuilder does not provide optimized methods for iterating over bit ranges, other join_utils code inherited from DataFusion also did this, so I sticked to using BooleanBufferBuilder for visited bitset to be consistent with the rest of the code.

It didn't show up as a performance bottleneck when running outer joins before, perhaps the other parts of the join is far more heavy weight than bitmap traversal.

rust/sedona-spatial-join/src/probe/partitioned_stream_provider.rs
Kontinuation: Removed all point_wkb functions and replaced them using `wkb_point`.

rust/sedona-spatial-join/src/probe/partitioned_stream_provider.rs
Kontinuation: I think we need a `wkb_rect` in wkb_factory. This could be useful in lots of places. I'll submit a subsequent PR for this and do some cleaning.

PR #565, opened 2026-01-30
fix(rust/sedona): Fix panic when displaying very long content

### LINKED ISSUE #388: Attempt to `show()` value > 65536 characters wide results in a panic
Fun fact: the `EXPLAIN VERBOSE` of reading Overture buildings is > 65536 characters wide (because this will list all the file paths that make up the dataset and there are hundreds of those...)

```python
import sedona.db

sd = sedona.db.connect()

buildings_url = (
    "s3://overturemaps-us-west-2/release/2025-11-19.0/theme=buildings/type=building/"
)

target_wkt = (
    "POLYGON ((-73.21 44.03, -73.21 43.98, -73.11 43.97, -73.12 44.03, -73.21 44.03))"
)

sd.read_parquet(
    buildings_url,
    options={"aws.skip_signature": True, "aws.region": "us-west-2"},
).to_view("buildings")

sd.sql(f"""
SELECT count(*) FROM buildings
WHERE ST_Intersects(geometry, ST_SetSRID(ST_GeomFromText('{target_wkt}'), 4326))
""").explain("extended").show()
#> thread '<unnamed>' panicked at rust/sedona/src/show.rs:425:46:
#> PanicException: attempt to add with overflow
```

paleolimbot:
Fixes a panic when the content of a table is unexpectedly large. The first issue I hit this was was explain plans.

In addition to at least one issue here, there is quite a lot of unchecked addition in the underlying renderer, which we work around by truncating any u16s or strings to 32,000.

Reprex from original issue now renders without panicing:

```python
import sedona.db

sd = sedona.db.connect()

buildings_url = (
    "s3://overturemaps-us-west-2/release/2026-01-21.0/theme=buildings/type=building/"
)

target_wkt = (
    "POLYGON ((-73.21 44.03, -73.21 43.98, -73.11 43.97, -73.12 44.03, -73.21 44.03))"
)

sd.read_parquet(
    buildings_url,
    options={"aws.skip_signature": True, "aws.region": "us-west-2"},
).to_view("buildings")

sd.sql(f"""
SELECT count(*) FROM buildings
WHERE ST_Intersects(geometry, ST_SetSRID(ST_GeomFromText('{target_wkt}'), 4326))
""").explain("extended").show()
#> ...very, very long output that does not crash
```

Closes #388.

### DIFF
--- a/rust/sedona/src/show.rs
+++ b/rust/sedona/src/show.rs
@@ -536,10 +536,7 @@ impl DisplayColumn {
         // Never return content longer than WIDTH_MAX because comfy table may
         // do arithmetic with it without checking for overflow.
         let content_str = content.to_string();
-        let content_safe: String = content_str
-            .chars()
-            .take(WIDTH_MAX as usize)
-            .collect();
+        let content_safe: String = content_str.chars().take(WIDTH_MAX as usize).collect();
 
         let cell = Cell::new(content_safe).set_delimiter('\0');
         let is_numeric = ArgMatcher::is_numeric();


rust/sedona/src/show.rs
```diff
@@ -521,7 +533,12 @@ impl DisplayColumn {
     /// Create a cell for this column. This is where alignment for numeric
     /// output is applied.
     fn cell<T: ToString>(&self, content: T) -> Cell {
-        let cell = Cell::new(content).set_delimiter('\0');
+        // Never return content longer than WIDTH_MAX because comfy table may
+        // do arithmetic with it without checking for overflow.
+        let content_str = content.to_string();
+        let content_safe: String = content_str.chars().take(WIDTH_MAX as usize).collect();
```
zhangfengcdt, line 539: This seems to be limiting the character length, but not display length (width). They may not be the same or may have issues with multi-byte characters.  Maybe use unicode width string handling?

rust/sedona/src/show.rs
```diff
@@ -316,29 +322,32 @@ impl<'a> DisplayTable<'a> {
...
         // Leftmost border
-        let mut total_size = 1;
-        let mut hidden_count = 0;
+        let mut total_size: u16 = 1;
+        let mut hidden_count: u16 = 0;
 
         for col in &self.columns {
             // Padding on both sides plus separator
-            total_size += 3;
+            total_size = total_size.saturating_add(3);
```
zhangfengcdt, line 330: Not very clear to me why "3" here? Maybe define a constant or add comments?

rust/sedona/src/show.rs
paleolimbot: These are uniocde widths! https://doc.rust-lang.org/std/primitive.str.html#method.chars . (We also have tests for unicode output to ensure the character math works for that).

rust/sedona/src/show.rs
paleolimbot: There is a comment `// Padding on both sides plus separator` but I am happy to clarify in a follow-up if needed!

PR #566, opened 2026-01-30
chore(rust): Update to geo 0.32.0

### LINKED ISSUE #491: rust/sedona-geo: Upgrade to geo 0.32
The dependabot PR for this update ( https://github.com/apache/sedona-db/pull/455 ) is running into a number of issues, at least:

- Concave hull's signature changed so the build fails
- There are a few failing tests in geo-generic-alg

paleolimbot:
Updates the geo crate to 0.32.0

Closes #455

Closes #491

Replaces #549


rust/sedona-geo/src/st_concavehull.rs
```diff
@@ -41,6 +42,10 @@ use crate::to_geo::item_to_geometry;
 /// Geo returns a Polygon for every concave hull computation
 /// whereas the Geos implementation returns a MultiPolygon for
 /// certain geometries concave hull computation.
+///
+/// Note that this does *not* match the PostGIS implementation.
+/// In particular, the GEOS/PostGIS "pctconvex" parameter, which extends
+/// from 0..1 does not match this kernel's "concavity" (0..Infinity).
```
paleolimbot, line 48: The Concave Hull algorithm is completely new, but its old output didn't match either. This kernel isn't enabled by default for this reason, but I added this note because in reading the docs it seems that the parameter doesn't carry the same meaning. If we did expose this we might have to give it another name.

rust/sedona-geo-generic-alg/src/algorithm/line_measures/metric_spaces/euclidean/utils.rs
```diff
@@ -1571,28 +1571,28 @@ mod tests {
...
+    //         let generic_dist = nearest_neighbour_distance(&ls1, &ls2);
+
+    //         assert_relative_eq!(
+    //             concrete_dist,
+    //             generic_dist,
+    //             epsilon = 1e-10,
+    //             max_relative = 1e-10
+    //         );
+    //     }
+    // }
```
paleolimbot, line 1595: @Kontinuation Can you look into these two tests? I am wondering if the mismatch is a bug on our side or a bug on their side (or unrelated and I'm not understanding the test).

The failure is:

```
thread 'algorithm::line_measures::metric_spaces::euclidean::utils::tests::test_random_linestring_to_linestring_distance' panicked at rust/sedona-geo-generic-alg/src/algorithm/line_measures/metric_spaces/euclidean/utils.rs:1588:13:
assert_relative_eq!(concrete_dist, generic_dist, epsilon = 1e-10, max_relative = 1e-10)

    left  = 28.34636916784935
    right = 21.713575661323034
```

(i.e., the algorithm here is giving smaller distances than whatever is producing `concrete_dist`)

Kontinuation: Preliminary investigations indicate that our calculations are correct, and geo 0.32.0 is incorrect.

I'll look into it further and see if I can submit a patch to georust/geo.

Kontinuation: I have submitted https://github.com/georust/geo/pull/1499 to fix this.

PR #573, opened 2026-02-04
feat(rust/sedona-spatial-join): Support partitioned KNN join to handle larger than memory object side

Kontinuation:
This patch breaks the object size of KNN(query, object, K) into smaller partitions, and spills results of KNN queries to a set of spill files. We merge the KNN query result with the nearest-so-far results we obtained when processing the previous partition. The global KNN result will be produced after all partitions were processed.

The tricky part for spilling and merging nearest-so-far KNN query results is implemented in knn_result_merger.rs. It needs to handle lots of edge cases correctly and we were very cautious when implementing this. Fuzz tests were added to verify the correctness of the KNN result merger.

### DIFF
--- a/rust/sedona-spatial-join/src/probe/knn_results_merger.rs
+++ b/rust/sedona-spatial-join/src/probe/knn_results_merger.rs
@@ -1294,7 +1294,7 @@ fn truncate_row_selectors_to_top_k(
 mod test {
     use arrow::compute::take_record_batch;
     use datafusion_physical_plan::metrics::ExecutionPlanMetricsSet;
-    use rand::{rngs::StdRng, seq::SliceRandom, Rng, SeedableRng};
+    use rand::{rngs::StdRng, seq::SliceRandom, RngExt, SeedableRng};
     use rstest::rstest;
 
     use super::*;


rust/sedona-spatial-join/src/partitioning/round_robin.rs
```diff
@@ -0,0 +1,80 @@
...
+    fn num_regular_partitions(&self) -> usize {
+        self.num_partitions
+    }
+
+    fn partition(&self, bbox: &BoundingBox) -> Result<SpatialPartition> {
+        self.partition_no_multi(bbox)
+    }
+
+    fn partition_no_multi(&self, _bbox: &BoundingBox) -> Result<SpatialPartition> {
+        let idx = self.counter.fetch_add(1, Ordering::Relaxed);
```
Kontinuation, line 54: This round robin partitioner is nondeterministic due to the order of concurrent tasks being scheduled. We will address this by making all partitioners non-sync, so that each async task will have its own partitioner with its own mutable internal state. This will be a relatively large refactoring so I'll leave it to the next PR.

rust/sedona-spatial-join/src/stream.rs
```diff
@@ -106,6 +114,12 @@ pub(crate) struct SpatialJoinStream {
     /// This is used for outer joins to ensure that we only emit unmatched rows from the Multi
     /// partition once, after all regular partitions have been processed.
     visited_multi_probe_side: Option<Arc<Mutex<BooleanBufferBuilder>>>,
+    /// KNN results merger. Only used for partitioned KNN join. This value is Some when this spatial join stream
+    /// is for KNN join and the number of partitions is greater than 1, except when in the
+    /// [SpatialJoinStreamState::ProcessProbeBatch] state. The `knn_results_merger` will be moved into the
```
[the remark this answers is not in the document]
Kontinuation: `SpatialJoinStreamState::ProcessProbeBatch` is a valid reference.

rust/sedona-spatial-join/src/probe/knn_results_merger.rs
```diff
@@ -0,0 +1,2224 @@
...
+        kept_prob: f64,
+        rng: &mut StdRng,
+    ) -> Vec<FuzzTestKNNResult> {
+        let mut test_data = Vec::with_capacity(num_rows);
+        let mut next_object_id = 0;
+        for query_id in 0..num_rows {
+            // Generate K objects
+            let knn_objects = (next_object_id..next_object_id + k)
+                .map(|object_id| FuzzKNNResultObject {
+                    object_id,
```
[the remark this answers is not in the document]
Kontinuation: Fixed this by just generate floating point distances without casting.

rust/sedona-spatial-join/src/exec.rs
```diff
@@ -1478,32 +1489,58 @@ mod tests {
         results
     }
 
+    #[rstest]
     #[tokio::test]
-    async fn test_knn_join_correctness() -> Result<()> {
+    async fn test_knn_join_correctness(
+        // TODO: enable point_only = false after https://github.com/georust/geo-index/pull/151
+        // is released
```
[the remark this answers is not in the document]
Kontinuation: Improved comment to clarity the reason

rust/sedona-spatial-join/src/probe/knn_results_merger.rs
```diff
@@ -0,0 +1,2224 @@
...
+
+impl<'a> KNNProbeResult<'a> {
+    fn new(
+        probe_row_index: usize,
+        row_range: Range<usize>,
+        distances: &'a [f64],
+        unfiltered_distances: &'a [f64],
+    ) -> Self {
+        assert_eq!(row_range.len(), distances.len());
+        // Please note that we don't have `unfiltered_distances.len() >= distances.len()` here.
```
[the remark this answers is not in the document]
Kontinuation: Explaining them in detail is a good idea. I have added the comment. The comment suggested by Copilot was all wrong.

rust/sedona-spatial-join/src/stream.rs
```diff
@@ -1037,12 +1091,13 @@ impl SpatialJoinBatchIterator {
                 last_produced_probe_idx: -1,
                 build_batch_positions: Vec::new(),
                 probe_indices: Vec::new(),
```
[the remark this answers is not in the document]
Kontinuation: The code will be refactored to move `distances` and `knn_results_merger` into a struct, that will make the invariant always be enforced.

rust/sedona-spatial-join/src/index/spatial_index.rs
```diff
@@ -213,13 +215,33 @@ impl SpatialIndex {
     /// # Returns
     ///
     /// * `JoinResultMetrics` containing the number of actual matches and candidates processed
+    #[allow(unused)]
```
paleolimbot: Is this going to be used in a subsequent PR or can it be `#[cfg(test)]`?

rust/sedona-spatial-join/src/partitioning/broadcast.rs
```diff
@@ -0,0 +1,72 @@
...
+
+use datafusion_common::Result;
+use sedona_common::sedona_internal_err;
+use sedona_geometry::bounding_box::BoundingBox;
+
+use crate::partitioning::{SpatialPartition, SpatialPartitioner};
+
+/// A partitioner that assigns everything to the Multi partition.
+///
+/// This partitioner is useful when we want to broadcast the data to all partitions.
```
paleolimbot, line 26: I think it is worth expanding on this slightly (e.g., that it is used for the KNN join where we can't know in advance how far away a given number of neighbours will be to assign it definitively to a single partition, or something else if I completely misinterpreted that)

rust/sedona-spatial-join/src/stream.rs
```diff
@@ -896,10 +941,20 @@ struct ProbeProgress {
...
     /// for tracking the progress of producing joined batches
     pos: usize,
+    /// KNN-specific progress. Only used for KNN join.
+    knn: Option<KNNProbeProgress>,
+}
+
+struct KNNProbeProgress {
+    /// Accumulated comparable (e.g. squared) distances of the KNN results.
+    /// Should have the same length as `build_batch_positions`.
+    distances: Vec<f64>,
```
paleolimbot, line 951: Is this Vec going to have a size that is large enough that we need to track its allocation (or is this local to a specific batch or partition guaranteed to be fairly small?)

rust/sedona-spatial-join/src/utils/join_utils.rs
```diff
@@ -300,9 +301,28 @@ pub(crate) fn apply_join_filter_to_indices(
...
+            build_indices.len(),
+            "distances length should match indices length"
+        );
+        let dist_array = arrow_array::Float64Array::from(distances.to_vec());
+        let filtered = compute::filter(&dist_array, mask)?;
+        let filtered = filtered
+            .as_any()
+            .downcast_ref::<arrow_array::Float64Array>()
+            .expect("filtered distance array should be Float64Array");
+        Some(filtered.values().to_vec())
```
paleolimbot, line 317: Would (`zip(distances, mask.values()).filter(|(dist, b| b).map(|(dist, b| dist).collect::Vec<f64>()` be any easier?

rust/sedona-spatial-join/src/exec.rs
```diff
@@ -1478,32 +1489,60 @@ mod tests {
...
     }
 
+    #[rstest]
     #[tokio::test]
-    async fn test_knn_join_correctness() -> Result<()> {
+    async fn test_knn_join_correctness(
+        // TODO: Currently the underlying geo-index KNN implementation has bugs working with non-point
+        // geometries, so this test is restricted to point_only = true. Once
+        // https://github.com/georust/geo-index/pull/151 (fixing non-point KNN support) is
+        // released, add #[values(true, false)] here to also exercise non-point data.
```
paleolimbot: You can also update `Cargo.toml` to point to `d70f049bc02da7364d923c519a484a482c9c0a11` now that the linked PR is merged. We also have one other georust crate specified as a git dependency waiting for a release (geos).

rust/sedona-spatial-join/src/probe/knn_results_merger.rs
```diff
@@ -0,0 +1,2226 @@
...
+use datafusion_execution::disk_manager::RefCountedTempFile;
+use datafusion_execution::runtime_env::RuntimeEnv;
+use datafusion_physical_plan::metrics::SpillMetrics;
+use sedona_common::sedona_internal_err;
+
+use crate::index::spatial_index::DISTANCE_TOLERANCE;
+use crate::utils::spill::{RecordBatchSpillReader, RecordBatchSpillWriter};
+
+/// [UnprocessedKNNResultBatch] represents the KNN results produced by probing the spatial index.
+/// An [UnprocessedKNNResultBatch] may include KNN results for multiple probe rows.
```
paleolimbot, line 42: Thank you for this write up!

rust/sedona-spatial-join/src/probe/knn_results_merger.rs
```diff
@@ -0,0 +1,2226 @@
...
+            unfiltered_probe_indices,
+            unfiltered_distances,
+        }
+    }
+
+    /// Merge the current [UnprocessedKNNResultBatch] with another one, producing a new
+    /// [UnprocessedKNNResultBatch].
+    fn merge(self, other: Self) -> Result<Self> {
+        let concat_array =
+            concat(&[&self.row_array, &other.row_array]).map_err(|e| arrow_datafusion_err!(e))?;
```
paleolimbot:
```suggestion
            concat(&[&self.row_array, &other.row_array])?;
```

rust/sedona-spatial-join/src/probe/knn_results_merger.rs
```diff
@@ -0,0 +1,2226 @@
...
+        } = unprocessed_batch;
+
+        assert_eq!(row_array.len(), probe_indices.len());
+        assert_eq!(probe_indices.len(), distances.len());
+        assert_eq!(unfiltered_probe_indices.len(), unfiltered_distances.len());
+        assert!(probe_indices.len() <= unfiltered_probe_indices.len());
+
+        let compute_range_encoding = |mut indices: Vec<usize>| {
+            let mut offsets = Vec::with_capacity(indices.len() + 1);
+            offsets.push(0);
```
paleolimbot: Unless you really have to make this an inline lambda it's probably better to keep this as a separate function to make it more clear what shared state it depends on.

rust/sedona-spatial-join/src/probe/knn_results_merger.rs
```diff
@@ -0,0 +1,2226 @@
...
+        for result in knn_query_result_iterator {
+            // Only the previous result is guaranteed to be complete.
+            if let Some(result) = prev_result_opt {
+                self.merge_and_append_result(&result)?;
+            }
+
+            prev_result_opt = Some(result);
+        }
+
+        // Assembled this batch. Write to spill file or produce output batch.
```
paleolimbot:
```suggestion
        // Assemble this batch. Write to spill file or produce output batch.
```

rust/sedona-spatial-join/src/probe/knn_results_merger.rs
```diff
@@ -0,0 +1,2226 @@
...
+        if self.index_array_builder.is_empty() {
+            return Ok(None);
+        }
+
+        // Reset builders for building columns required by spilled batches. Building these columns seems to be wasted work
+        // when we only need to produce result batches, but it simplifies the code significantly and the performance impact is minimal.
+        let _ = std::mem::replace(
+            &mut self.index_array_builder,
+            UInt64Array::builder(self.capacity),
+        );
```
paleolimbot: For my own education, is this better than `self.index_array_builder = UInt64Array::builder(self.capacity)`?

rust/sedona-spatial-join/src/probe/knn_results_merger.rs
```diff
@@ -0,0 +1,2226 @@
...
+    }
+
+    #[test]
+    fn test_knn_result_merger_with_empty_partitions() {
+        let k = 5;
+        let include_tie_breaker = false;
+        let num_rows = 100;
+        let num_partitions = 3;
+        let kept_prob = 0.5;
+        let mut rng = StdRng::seed_from_u64(42);
```
paleolimbot: Probably worth mixing up your random seed to not use 42 every single time (there are a lot of 42s in this file and crate generally).

rust/sedona-spatial-join/src/probe/knn_results_merger.rs
```diff
@@ -0,0 +1,2226 @@
...
+/// Each probe row has a unique index. The index must be strictly increasing
+/// across probe rows. The sequence of index across the entire sequence of ingested
+/// [UnprocessedKNNResultBatch] must also be strictly increasing. The index is computed based on
+/// the 0-based index of the probe row in this probe partition.
+///
+/// The KNN results are filtered, meaning that the original KNN results obtained by probing
+/// the spatial index may be further filtered based on some predicates. It is also possible that
+/// all the KNN results for a probe row are filtered out. However, we still need to keep track of the
+/// distances of unfiltered results to correctly compute the top-K distances before filtering. This
+/// is critical for correctly merging KNN results from multiple partitions.
```
paleolimbot, line 58: Are the predicates you are talking about here are all related to the distance (notably: the maximum distance), or does the KNN join also process arbitrary predicates that refer to both join sides and can't be pushed down? Or does that situation turn into >1 join? (Probably worth a line about what filters are getting applied at this stage to this write up).

rust/sedona-spatial-join/src/probe/knn_results_merger.rs
Kontinuation: It does not depend on any internal states. I can easily pull it out as a standalone function.

rust/sedona-spatial-join/src/probe/knn_results_merger.rs
Kontinuation: Removed all redundant calls to arrow_datafusion_err. I forgot why I added that in the first place.

rust/sedona-spatial-join/src/exec.rs
Kontinuation: OK. I have updated the Cargo.toml and the test cases.

I think I need make another round of breaking change to get rid of the "use-geo_0_31" feature flag.

rust/sedona-spatial-join/src/utils/join_utils.rs
Kontinuation: I still prefer using `compute::filter` to make it easy to observe that the computations done for distance is roughly the same as build_indices and probe_indices.

rust/sedona-spatial-join/src/stream.rs
Kontinuation: No, it is local to one probe batch, the size won't exceed probe_batch_size * K.

rust/sedona-spatial-join/src/partitioning/broadcast.rs
Kontinuation: Improved the comment.

rust/sedona-spatial-join/src/index/spatial_index.rs
Kontinuation: No. I think I should replace it using `query_knn_with_distance`.

rust/sedona-spatial-join/src/probe/knn_results_merger.rs
Kontinuation: Fixed comment

rust/sedona-spatial-join/src/probe/knn_results_merger.rs
Kontinuation: Probably not. This is just stupidly copied from the other places where mem::replace were actually required.

rust/sedona-spatial-join/src/probe/knn_results_merger.rs
Kontinuation: Change the tests to rstest with multiple seeds as parameters.

rust/sedona-spatial-join/src/probe/knn_results_merger.rs
Kontinuation: It can be any condition, not necessarily related to the distance.

The writing can be a bit confusing. I have extended the comment to explain it.

PR #575, opened 2026-02-04
feat(python/sedonadb): Implement parameter binding

### LINKED ISSUE #111: feat: add parameterized queries
I am currently writing code like this:

```python
cities.to_view("cities", overwrite=True)
countries.to_view("countries", overwrite=True)

sd.sql("""
select * from cities
join countries
where ST_Intersects(cities.geometry, countries.geometry)
""").show()
```

This could be cool to skip the explicit view registration:

```python
sd.sql("""
select * from cities
join countries
where ST_Intersects(cities.geometry, countries.geometry)
""", cities=cities, countries=countries).show()
```

It would be especially sweet if other objects (like dates & Shapely objects) could be parameterized too!

paleolimbot:
This PR provides the ability to bind arbitrary Python objects as parameters in a query (e.g. `SELECT ST_Envelope($1)`). As discussed in the ticket, most of the work here is actually around the logic to convert arbitrary Python objects, where the heuristic is:

- Have a special-case list (e.g., this handles Shapely objects and GeoDataFrame/GeoSeries)
- Try `obj.__arrow_c_array__()` protocol
- Try `pyarrow.array([obj])`

In general, dataframe-ish objects with exactly one value and anything convertible to arrow of length one works (this is vaugely the same logic as how you can put a subquery in any spot where a scalar is expected). It also works nicely for geometry because GeoPandas objects preserve their CRS but shapely objects don't (so `geo_df.geometry[0]` is lossy but `geo_df.geometry` is not).

As a side effect this also kick starts our expression API with a single expression type `Literal`.

Closes #111.

```python
import pandas as pd
import sedona.db

sd = sedona.db.connect()

sd.sql("SELECT $1 + 1 as two", params=(1, )).show()
#> ┌───────┐
#> │  two  │
#> │ int64 │
#> ╞═══════╡
#> │     2 │
#> └───────┘

sd.sql("SELECT $foofy + 1 as two", params={"foofy": 1}).show()
#> ┌───────┐
#> │  two  │
#> │ int64 │
#> ╞═══════╡
#> │     2 │
#> └───────┘
```

Geometry objects can also be bound here:

```python
import sedona.db
import shapely

sd = sedona.db.connect()

sd.sql("SELECT ST_Buffer($1, 0.5)").with_params(
    shapely.box(1, 2, 3, 4)
).to_pandas().plot()
```

### DIFF
--- a/python/sedonadb/python/sedonadb/expr/literal.py
+++ b/python/sedonadb/python/sedonadb/expr/literal.py
@@ -115,6 +115,8 @@ def _lit_from_series(obj):
     if len(obj) != 1:
         raise ValueError("Can't create SedonaDB literal from Series with length != 1")
 
+    # A column with dtype "geometry" is not always a GeoSeries; however, if the dtype
+    # is geometry, obj.array.crs should still be available to extract the CRS.
     if obj.dtype.name == "geometry":
         first_value = obj.array[0]
         first_wkb = None if first_value is None else first_value.wkb
@@ -169,6 +171,7 @@ SPECIAL_CASED_LITERALS = {
     "shapely.geometry.point.Point": _lit_from_shapely,
     "shapely.geometry.linestring.LineString": _lit_from_shapely,
     "shapely.geometry.polygon.Polygon": _lit_from_shapely,
+    "shapely.geometry.polygon.LinearRing": _lit_from_shapely,
     "shapely.geometry.multipoint.MultiPoint": _lit_from_shapely,
     "shapely.geometry.multilinestring.MultiLineString": _lit_from_shapely,
     "shapely.geometry.multipolygon.MultiPolygon": _lit_from_shapely,
--- a/python/sedonadb/tests/expr/test_literal.py
+++ b/python/sedonadb/tests/expr/test_literal.py
@@ -79,6 +79,14 @@ def test_shapely_literal(wkt):
     assert array == ga.as_wkb([wkt])
 
 
+def test_shapely_linearring():
+    shapely_obj = shapely.from_wkt("LINEARRING (0 0, 1 0, 0 1, 0 0)")
+    literal = lit(shapely_obj)
+
+    array = pa.array(literal)
+    assert array == ga.as_wkb(["LINESTRING (0 0, 1 0, 0 1, 0 0)"])
+
+
 def test_geopandas_literal():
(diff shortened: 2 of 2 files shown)


python/sedonadb/python/sedonadb/expr/literal.py
```diff
@@ -0,0 +1,177 @@
...
+    type = ga.wkb().with_crs(crs)
+    storage = pa.array([wkb], type.storage_type)
+    return type.wrap_array(storage)
+
+
+def _qualified_type_name(obj):
+    return f"{type(obj).__module__}.{type(obj).__name__}"
+
+
+SPECIAL_CASED_LITERALS = {
```
zhangfengcdt, line 162: Would this work for subclass if we match the class names?

and missing LinearRing?

python/sedonadb/python/sedonadb/expr/literal.py
```diff
@@ -0,0 +1,177 @@
...
+        )
+
+    return _resolve_arrow_lit(obj.iloc[0])
+
+
+def _lit_from_series(obj):
+    if len(obj) != 1:
+        raise ValueError("Can't create SedonaDB literal from Series with length != 1")
+
+    if obj.dtype.name == "geometry":
```
zhangfengcdt, line 120: should we also check crs is defined here?

Kontinuation: I'm not sure if dtype.name == "geometry" implies that the series is a GeoPandas GeoSeries, and if can assume that crs is defined in this case. Besides, the PR LGTM.

paleolimbot: This one is because of a fun corner case in Pandas land: `geo_df.iloc[0]` is a `Series` with dtype `geometry`, not a `GeoSeries` (hence checking `obj.array.crs` instead of `obj.crs`).

python/sedonadb/python/sedonadb/expr/literal.py
paleolimbot: Good catch on LinearRing!

I use the class name approach (which doesn't catch subclasses) because in order to do `isinstance(obj, shapely.Geometry)` we need to `import shapely`, which is something I'd rather not do at the module level if it can be avoided. If something comes up where there's no alternative we could require certain dependencies for using parameterized queries.

PR #578, opened 2026-02-06
feat(rust/sedona-geoparquet): Support WKB validation in `read_parquet()`

2010YOUY01:
Follow up to https://github.com/apache/sedona-db/pull/560

## Rationale

`read_parquet()` now supports an override option, `geometry_columns`, to cast Binary columns to geometries. This is an unsafe operation, so adding a validation step is helpful.

This PR currently limits validation to WKB validity. In the long term, additional metadata properties can also be validated. If any entry fails validation, an error will be returned.

Geo columns are inferred from {(Geo)Parquet metadata, user-provided `geometry_columns` override options}. The `validate` option applies uniformly in all cases—if `validate = true`, geometry columns are always validated. I think this approach is more general and simpler.

## Implementation

Propagate the `validate` option to `GeoParquetFileOpener` (which yields the final decoded batches as the data source output), and use it to optionally validate the decoded batches for WKB validity.


rust/sedona-geoparquet/src/file_opener.rs
```diff
@@ -98,11 +106,11 @@ pub(crate) struct GeoParquetFileOpener {
     pub inner: Arc<dyn FileOpener>,
     pub object_store: Arc<dyn ObjectStore>,
     pub metadata_size_hint: Option<usize>,
-    pub predicate: Arc<dyn PhysicalExpr>,
+    pub predicate: Option<Arc<dyn PhysicalExpr>>,
```
2010YOUY01, line 109: refactor 1: before it's required since `GeoParquetFileOpener` is only used when there is geo column, and also there is valid spatial predicate, otherwise it fall back to inner parquet opener.

Now it also have to be used if validation is enabled, and there is no spatial predicate, so making it optional.

rust/sedona-geoparquet/src/file_opener.rs
```diff
@@ -98,11 +106,11 @@ pub(crate) struct GeoParquetFileOpener {
...
     pub file_schema: SchemaRef,
     pub enable_pruning: bool,
     pub metrics: GeoParquetFileOpenerMetrics,
-    pub overrides: Option<HashMap<String, GeoParquetColumnMetadata>>,
+    pub options: TableGeoParquetOptions,
```
2010YOUY01, line 113: refactor 2: before `overrides` lives inside `TableGeoParquetOptions`, the new `validate` flag also lives inside that option, so I think only keeping the option here can make it simpler.

rust/sedona-geoparquet/src/format.rs
```diff
@@ -361,7 +362,7 @@ pub struct GeoParquetFileSource {
     inner: ParquetSource,
     metadata_size_hint: Option<usize>,
     predicate: Option<Arc<dyn PhysicalExpr>>,
-    overrides: Option<HashMap<String, GeoParquetColumnMetadata>>,
+    options: TableGeoParquetOptions,
```
2010YOUY01, line 365: refactor 3: similar to another refactor in `file_opener.rs`

python/sedonadb/tests/test_context.py
```diff
@@ -313,3 +313,79 @@ def test_read_parquet_invalid_azure_option():
...
     ):
         con.read_parquet(url, options={"azure.unknown_option": "value"})
+
+
+def test_read_parquet_validate_wkb_single_valid_row(con, tmp_path):
+    valid_wkb = bytes.fromhex("0101000000000000000000F03F0000000000000040")
+
+    table = pa.table({"id": [1], "geom": [valid_wkb]})
+    path = tmp_path / "single_valid_wkb.parquet"
+    pq.write_table(table, path)
```
paleolimbot: Optional, but there is a io/test_parquet.py that is possibly a better fit for these tests (and the geometry_columns tests) since there are now a lot of them!

python/sedonadb/python/sedonadb/context.py
```diff
@@ -176,9 +177,18 @@ def read_parquet(
...
+                  enabled, inconsistent data may cause undefined behavior.
+            validate:
+                When set to `True`, geometry column contents are validated against
+                their metadata. Metadata can come from the source Parquet file or
+                the user-provided `geometry_columns` option.
+                Only supported properties are validated; unsupported properties are
+                ignored. If validation fails, execution stops with an error.
+
+                Supported validation properties:
+                - WKB encoding
```
```suggestion
                Currently the only property that is validated is the WKB of input geometry
                columns.
```

rust/sedona-geoparquet/src/file_opener.rs
```diff
@@ -158,12 +170,108 @@ impl FileOpener for GeoParquetFileOpener {
...
+                        return exec_err!(
+                            "WKB validation failed for column '{}' at row {}: {}",
+                            column_name,
+                            row_index,
+                            e
+                        );
+                    }
+                }
+            }
+        }
```
paleolimbot, line 240: While we're here, it would be good to handle the `LargeBinary` case (this can happen for reasons outside a GeoParquet producer's control due to various auto-store/load of schemas)

python/sedonadb/tests/test_context.py
2010YOUY01: updated in 99d52a2

rust/sedona-geoparquet/src/file_opener.rs
2010YOUY01: fixed in  99d52a2

2010YOUY01:
> A few optional things to consider...thank you!
> 
> Just curious...have you run into files where this is a problem? I imagine if it does happen the existing error would be rather cryptic (although I believe we're good about validating WKB payload in every operator).

Thank you for the review! I have not encountered corrupted file so far, but my concern is the same: if this error is caught inside operators, the error might get propagated several times and produce confusing error messages.

PR #579, opened 2026-02-06
fix: no-default-features

b4l:
Fixes `cargo check -p sedona --no-default-features`


b4l:
> Can you also add a CI check for this so we don't break it?

Added it in the latest commit. It broke only for the specific package, so I added it explicitly.

PR #580, opened 2026-02-08
chore(ci): Add cargo clippy check to git pre-commit hooks

2010YOUY01:
## Rationale

Now `clippy` check is not available in local pre-commit hooks, this PR:
1. Extract ci's `cargo clippy` check to a local script for consistency
2. Reuse this script for local pre-commit check

I have tested it locally:
```sh
(.venv) yongting@Yongtings-MacBook-Pro-2 ~/C/sedona-db (pre-commit-clippy)> git diff
diff --git a/rust/sedona-geoparquet/src/file_opener.rs b/rust/sedona-geoparquet/src/file_opener.rs
index ae2a56d..a0ca352 100644
--- a/rust/sedona-geoparquet/src/file_opener.rs
+++ b/rust/sedona-geoparquet/src/file_opener.rs
@@ -205,7 +205,7 @@ fn filter_access_plan_using_geoparquet_covering(
     let row_group_indices_to_scan = access_plan.row_group_indexes();

     // What we're about to do is a bit of work, so skip it if we can.
-    if row_group_indices_to_scan.is_empty() {
+    if row_group_indices_to_scan.len() == 0 {
         return Ok(());
     }

(.venv) yongting@Yongtings-MacBook-Pro-2 ~/C/sedona-db (pre-commit-clippy *)> git add .
(.venv) yongting@Yongtings-MacBook-Pro-2 ~/C/sedona-db (pre-commit-clippy +)> git commit -m "test clippy violation"
check illegal windows names..........................(no files to check)Skipped
...
clang-format.........................................(no files to check)Skipped
cargo clippy fix workspace (best effort).................................Failed
- hook id: cargo-clippy-fix-workspace
- files were modified by this hook

    Checking sedona-geo-traits-ext v0.3.0 (/Users/yongting/Code/sedona-db/rust/sedona-geo-traits-ext)
    ...
    Checking sedonadb v0.3.0 (/Users/yongting/Code/sedona-db/python/sedonadb)
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 26.46s

(.venv) yongting@Yongtings-MacBook-Pro-2 ~/C/sedona-db (pre-commit-clippy +*) [0|1]> git diff
diff --git a/rust/sedona-geoparquet/src/file_opener.rs b/rust/sedona-geoparquet/src/file_opener.rs
index a0ca352..ae2a56d 100644
--- a/rust/sedona-geoparquet/src/file_opener.rs
+++ b/rust/sedona-geoparquet/src/file_opener.rs
@@ -205,7 +205,7 @@ fn filter_access_plan_using_geoparquet_covering(
     let row_group_indices_to_scan = access_plan.row_group_indexes();

     // What we're about to do is a bit of work, so skip it if we can.
-    if row_group_indices_to_scan.len() == 0 {
+    if row_group_indices_to_scan.is_empty() {
         return Ok(());
     }
```

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -153,11 +153,7 @@ jobs:
       - name: Clippy
         if: matrix.name == 'clippy'
         run: |
-          cargo clippy --workspace --all-targets --all-features -- -Dwarnings
-          # Clean up clippy artifacts aggressively
-          rm -rf target/debug/deps
-          rm -rf target/debug/incremental
-          rm -rf target/debug/build
+          bash ci/scripts/cargo-clippy.sh --ci-cleanup
 
       - name: Test
         if: matrix.name == 'test'
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -66,3 +66,12 @@ repos:
         types_or: [c, c++]
         # Don't run on vendored files
         exclude: "^c/(sedona-geoarrow-c/src/geoarrow|sedona-geoarrow-c/src/nanoarrow|sedona-tg/src/tg)/.*"
+
+  - repo: local
+    hooks:
+      - id: cargo-clippy-fix-workspace
+        name: cargo clippy fix workspace (best effort)
+        entry: bash -c 'bash ci/scripts/cargo-clippy.sh --fix || true'
+        language: system
+        pass_filenames: false
+        files: '(^|/)(Cargo\.toml|Cargo\.lock|.*\.rs)$'
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -67,11 +67,13 @@ repos:
         # Don't run on vendored files
         exclude: "^c/(sedona-geoarrow-c/src/geoarrow|sedona-geoarrow-c/src/nanoarrow|sedona-tg/src/tg)/.*"
 
+  # Slow hooks, only run on `git push`
   - repo: local
     hooks:
       - id: cargo-clippy-fix-workspace
         name: cargo clippy fix workspace (best effort)
-        entry: bash -c 'bash ci/scripts/cargo-clippy.sh --fix || true'
+        entry: bash -c 'bash ci/scripts/cargo-clippy.sh --fix'
         language: system
+        stages: [pre-push]
         pass_filenames: false
         files: '(^|/)(Cargo\.toml|Cargo\.lock|.*\.rs)$'
--- a/ci/scripts/cargo-clippy.sh
+++ b/ci/scripts/cargo-clippy.sh
@@ -1,0 +1,67 @@
+#!/usr/bin/env bash
+#
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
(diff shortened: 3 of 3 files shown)


paleolimbot (commented):
Thank you for opening!

This has been proposed before, although this is tricky for me personally because when I work on the R or Python bindings I often don't want to run a command that builds the entire project (i.e., I've always preferred a pre-commit that can run in a trivial amount of time since I run it frequently). I use an IDE configuration to run clippy and I think it can be installed as a git pre-commit hook independently if somebody prefers?

Happy to hear dissenting opinions!

2010YOUY01:
> Thank you for opening!
> 
> This has been proposed before, although this is tricky for me personally because when I work on the R or Python bindings I often don't want to run a command that builds the entire project (i.e., I've always preferred a pre-commit that can run in a trivial amount of time since I run it frequently). I use an IDE configuration to run clippy and I think it can be installed as a git pre-commit hook independently if somebody prefers?
> 
> Happy to hear dissenting opinions!

Thank you for the review! I agree it would be better to separate fast and slow checks.

WDYT making clippy a pre-push hook: it only triggers on `git push`.

I made the change in [301e4a5](https://github.com/apache/sedona-db/pull/580/commits/301e4a54ae124dcc10ea5c70ca5f128c9b651836), and have it tested locally:
```sh
# setup
pre-commit install --hook-type pre-push

(.venv) yongting@Yongtings-MacBook-Pro-2 ~/C/sedona-db (pre-commit-clippy +)> git commit --amend -m "Review: make clippy pre-push hook instead"
...
clang-format.........................................(no files to check)Skipped
[pre-commit-clippy 301e4a5] Review: make clippy pre-push hook instead
 Date: Tue Feb 10 13:04:33 2026 +0800
 1 file changed, 3 insertions(+), 1 deletion(-)

(.venv) yongting@Yongtings-MacBook-Pro-2 ~/C/sedona-db (pre-commit-clippy)> git push -f origin pre-commit-clippy
...
clang-format.........................................(no files to check)Skipped
cargo clippy fix workspace (best effort).............(no files to check)Skipped        <--- Only runs on `git push`
```

petern48:
> (I do a lot of pushing to docs, R, and Python and do a lot of branch switching and my IDE is already running clippy as I do that)

just out of curiosity, is the concern that adding it to pre-commit would make clippy run twice? i always thought your concern was that you didn't want clippy to run very often, but it sounds like it does?

FWIW, my personal workflow is usually to install clippy in pre-commit and just use `git commit --no-verify` to skip it in most cases, other than before pushing. Though I'm also someone who writes a concise number of commits, so I don't need to type out `--no-verity` super often.

I'm a fan of adding instructions to the contrib guide to optionally make it more automatic.

2010YOUY01:
Yes that sounds good.

Here is some additional information: my final goal is to have an easy way to reproduce the full CI locally, so for small fixes we don’t have to go back and forth checking the GitHub CI runner status.

I usually use plain shell scripts for this purpose, as I find them much more flexible than pre-commit configuration files.

How about maintaining CI jobs like:

```
cargo-fmt.sh
cargo-clippy.sh
test.sh
```

Each one can support different options, for example:

```
./cargo-clippy.sh          # check only
./cargo-clippy.sh --fix    # auto-fix and then check
./test.sh --fast           # quick tests, finish in minutes
./test.sh                  # full test suite, takes longer
```

Then local CI scripts, GitHub CI configuration, and pre-commit hooks can all call these shared scripts, while still allowing us to customize our local development workflow based on personal preferences.

I’m trying to set this up and hopefully sync it with upstream later.

PR #586, opened 2026-02-10
feat(c/sedona-libgpuspatial): Add Rust Wrapper

pwrliang:
This PR adds Rust wrappers and tests for c/sedona-libgpuspatial. `libgpuspatial.rs` contains the Rust wrappers for `GpuSpatialIndexFloat2D` and `GpuSpatialRefiner`; `lib.rs` contains a high-level struct `GpuSpatial` that supports all functions of libgpuspatial.

### DIFF
--- a/c/sedona-libgpuspatial/Cargo.toml
+++ b/c/sedona-libgpuspatial/Cargo.toml
@@ -43,7 +43,6 @@ thiserror = { workspace = true }
 log = { workspace = true }
 geo = { workspace = true }
 sedona-schema = { workspace = true }
-nvml-wrapper = "0.10.0"
 
 [dev-dependencies]
 wkt = { workspace = true }
--- a/c/sedona-libgpuspatial/Cargo.toml
+++ b/c/sedona-libgpuspatial/Cargo.toml
@@ -40,12 +40,12 @@ which = "8.0"
 arrow-array = { workspace = true, features = ["ffi"] }
 arrow-schema = { workspace = true }
 thiserror = { workspace = true }
-log = { workspace = true }
-geo = { workspace = true }
+geo-types = { workspace = true }
 sedona-schema = { workspace = true }
 
 [dev-dependencies]
 wkt = { workspace = true }
+geo = { workspace = true }
 sedona-expr = { path = "../../rust/sedona-expr" }
 sedona-geos = { path = "../sedona-geos" }
 sedona-testing = { path = "../../rust/sedona-testing" }
--- a/c/sedona-libgpuspatial/Cargo.toml
+++ b/c/sedona-libgpuspatial/Cargo.toml
@@ -35,3 +35,18 @@ gpu = []
 bindgen = "0.72.1"
 cmake = "0.1"
 which = "8.0"
+
+[dependencies]
+arrow-array = { workspace = true, features = ["ffi"] }
+arrow-schema = { workspace = true }
+thiserror = { workspace = true }
+log = { workspace = true }
+geo = { workspace = true }
+sedona-schema = { workspace = true }
+nvml-wrapper = "0.10.0"
+
+[dev-dependencies]
+wkt = { workspace = true }
+sedona-expr = { path = "../../rust/sedona-expr" }
+sedona-geos = { path = "../sedona-geos" }
+sedona-testing = { path = "../../rust/sedona-testing" }
--- a/c/sedona-libgpuspatial/libgpuspatial/include/gpuspatial/gpuspatial_c.h
+++ b/c/sedona-libgpuspatial/libgpuspatial/include/gpuspatial/gpuspatial_c.h
@@ -54,7 +54,7 @@ void GpuSpatialRuntimeCreate(struct GpuSpatialRuntime* runtime);
 
 struct GpuSpatialIndexConfig {
   /** Pointer to an initialized GpuSpatialRuntime struct */
-  struct GpuSpatialRuntime* runtime;
+  const struct GpuSpatialRuntime* runtime;
   /** How many threads will concurrently call Probe method */
   uint32_t concurrency;
 };
@@ -124,7 +124,7 @@ int GpuSpatialIndexFloat2DCreate(struct SedonaFloatIndex2D* index,
 
 struct GpuSpatialRefinerConfig {
   /** Pointer to an initialized GpuSpatialRuntime struct */
-  struct GpuSpatialRuntime* runtime;
+  const struct GpuSpatialRuntime* runtime;
   /** How many threads will concurrently call Probe method */
   uint32_t concurrency;
   /** Whether to compress the BVH structures to save memory */
--- a/c/sedona-libgpuspatial/libgpuspatial/include/gpuspatial/gpuspatial_c.h
+++ b/c/sedona-libgpuspatial/libgpuspatial/include/gpuspatial/gpuspatial_c.h
@@ -84,30 +84,23 @@ struct SedonaFloatIndex2D {
    * @return 0 on success, non-zero on failure
    */
   int (*finish_building)(struct SedonaFloatIndex2D* self);
+
   /**
    * Probe the spatial index with the given rectangles, each rectangle is represented by 4
    * floats: [min_x, min_y, max_x, max_y] Points can also be probed by providing [x, y, x,
    * y] but points and rectangles cannot be mixed in one Probe call. The results of the
-   * probe will be stored in the context.
+   * probe will be stored in the context. The callback function will be called for each
+   * batch of results, with the build and probe indices of the candidate pairs in the
+   * batch. The user_data pointer will be passed to the callback
    *
    * @return 0 on success, non-zero on failure
    */
   int (*probe)(struct SedonaFloatIndex2D* self, struct SedonaSpatialIndexContext* context,
-               const float* buf, uint32_t n_rects);
-  /** Get the build indices buffer from the context
-   *
-   * @return A pointer to the buffer and its length
-   */
-  void (*get_build_indices_buffer)(struct SedonaSpatialIndexContext* context,
-                                   uint32_t** build_indices,
-                                   uint32_t* build_indices_length);
-  /** Get the probe indices buffer from the context
-   *
-   * @return A pointer to the buffer and its length
-   */
-  void (*get_probe_indices_buffer)(struct SedonaSpatialIndexContext* context,
-                                   uint32_t** probe_indices,
-                                   uint32_t* probe_indices_length);
+               const float* buf, uint32_t n_rects,
+               void (*callback)(const uint32_t* build_indices,
+                                const uint32_t* probe_indices, uint32_t length,
+                                void* user_data),
+               void* user_data);
   /** Get the last error message from either the index
    *
    * @return A pointer to the error message string
--- a/c/sedona-libgpuspatial/libgpuspatial/include/gpuspatial/gpuspatial_c.h
+++ b/c/sedona-libgpuspatial/libgpuspatial/include/gpuspatial/gpuspatial_c.h
@@ -71,10 +71,10 @@ struct SedonaFloatIndex2D {
   void (*create_context)(struct SedonaSpatialIndexContext* context);
   /** Destroy a previously created context */
   void (*destroy_context)(struct SedonaSpatialIndexContext* context);
-  /** Push rectangles for building the spatial index, each rectangle is represented by 4
-   * floats: [min_x, min_y, max_x, max_y].
+  /** Push rectangles for building the spatial index.
+   * @param buf each rectangle is represented by 4 floats: [min_x, min_y, max_x, max_y].
    * Points can also be indexed by providing degenerated rectangles [x, y, x, y].
-   *
+   * @param n_rects The number of rectangles in the buffer
    * @return 0 on success, non-zero on failure
    */
   int (*push_build)(struct SedonaFloatIndex2D* self, const float* buf, uint32_t n_rects);
@@ -84,22 +84,27 @@ struct SedonaFloatIndex2D {
    * @return 0 on success, non-zero on failure
    */
   int (*finish_building)(struct SedonaFloatIndex2D* self);
-
   /**
-   * Probe the spatial index with the given rectangles, each rectangle is represented by 4
-   * floats: [min_x, min_y, max_x, max_y] Points can also be probed by providing [x, y, x,
-   * y] but points and rectangles cannot be mixed in one Probe call. The results of the
-   * probe will be stored in the context. The callback function will be called for each
-   * batch of results, with the build and probe indices of the candidate pairs in the
-   * batch. The user_data pointer will be passed to the callback
+   * Probe the spatial index with the given rectangles.
+   * @param buf The buffer of rectangles to probe, stored in the same format as the build
+   * rectangles.
+   * @param n_rects The number of rectangles in the probe buffer.
+   * @param callback The callback function to call for each batch of results.
+   * The callback should return 0 to continue receiving results, or non-zero to stop the
+   * probe early. The callback will be called with arrays of build and probe indices
+   * corresponding to candidate pairs of rectangles that intersect.
+   * The user-provided callback is required to return a value that will be further passed
+   * to the probe function to indicate whether there's an error during the callback
+   * execution.
+   * @param user_data The user_data pointer will be passed to the callback
    *
    * @return 0 on success, non-zero on failure
    */
   int (*probe)(struct SedonaFloatIndex2D* self, struct SedonaSpatialIndexContext* context,
                const float* buf, uint32_t n_rects,
-               void (*callback)(const uint32_t* build_indices,
-                                const uint32_t* probe_indices, uint32_t length,
-                                void* user_data),
+               int (*callback)(const uint32_t* build_indices,
+                               const uint32_t* probe_indices, uint32_t length,
+                               void* user_data),
                void* user_data);
   /** Get the last error message from either the index
    *
--- a/c/sedona-libgpuspatial/libgpuspatial/src/gpuspatial_c.cc
+++ b/c/sedona-libgpuspatial/libgpuspatial/src/gpuspatial_c.cc
@@ -193,22 +193,25 @@ struct GpuSpatialIndexFloat2DExporter {
 
   static int CProbe(self_t* self, SedonaSpatialIndexContext* context, const float* buf,
                     uint32_t n_rects,
-                    void (*callback)(const uint32_t* build_indices,
-                                     const uint32_t* probe_indices, uint32_t length,
-                                     void* user_data),
+                    int (*callback)(const uint32_t* build_indices,
+                                    const uint32_t* probe_indices, uint32_t length,
+                                    void* user_data),
                     void* user_data) {
+    auto* p_ctx = static_cast<context_t*>(context->private_data);
     // Do not use SafeExecute because this method is thread-safe and we don't want to set
     // last_error for the whole index if one thread encounters an error
     try {
       auto* rects = reinterpret_cast<const spatial_index_t::box_t*>(buf);
-      auto& buff = static_cast<context_t*>(context->private_data)->payload;
+      auto& buff = p_ctx->payload;
       use_index(self).Probe(rects, n_rects, &buff.build_indices, &buff.probe_indices);
-      callback(buff.build_indices.data(), buff.probe_indices.data(),
-               buff.build_indices.size(), user_data);
-      return 0;
+
+      return callback(buff.build_indices.data(), buff.probe_indices.data(),
+                      buff.build_indices.size(), user_data);
     } catch (const std::exception& e) {  // user should call context_get_last_error
+      p_ctx->last_error = std::string(e.what());
       return EINVAL;
     } catch (...) {
+      p_ctx->last_error = "Unknown internal error";
       return EINVAL;
     }
   }
--- a/c/sedona-libgpuspatial/libgpuspatial/src/gpuspatial_c.cc
+++ b/c/sedona-libgpuspatial/libgpuspatial/src/gpuspatial_c.cc
@@ -159,8 +159,6 @@ struct GpuSpatialIndexFloat2DExporter {
     out->push_build = &CPushBuild;
     out->finish_building = &CFinishBuilding;
     out->probe = &CProbe;
-    out->get_build_indices_buffer = &CGetBuildIndicesBuffer;
-    out->get_probe_indices_buffer = &CGetProbeIndicesBuffer;
     out->get_last_error = &CGetLastError;
     out->context_get_last_error = &CContextGetLastError;
     out->release = &CRelease;
@@ -194,12 +192,25 @@ struct GpuSpatialIndexFloat2DExporter {
   }
 
   static int CProbe(self_t* self, SedonaSpatialIndexContext* context, const float* buf,
-                    uint32_t n_rects) {
-    return SafeExecute(static_cast<context_t*>(context->private_data), [&] {
+                    uint32_t n_rects,
+                    void (*callback)(const uint32_t* build_indices,
+                                     const uint32_t* probe_indices, uint32_t length,
+                                     void* user_data),
+                    void* user_data) {
+    // Do not use SafeExecute because this method is thread-safe and we don't want to set
+    // last_error for the whole index if one thread encounters an error
+    try {
       auto* rects = reinterpret_cast<const spatial_index_t::box_t*>(buf);
       auto& buff = static_cast<context_t*>(context->private_data)->payload;
       use_index(self).Probe(rects, n_rects, &buff.build_indices, &buff.probe_indices);
-    });
+      callback(buff.build_indices.data(), buff.probe_indices.data(),
+               buff.build_indices.size(), user_data);
+      return 0;
+    } catch (const std::exception& e) {  // user should call context_get_last_error
+      return EINVAL;
+    } catch (...) {
+      return EINVAL;
+    }
   }
 
   static void CGetBuildIndicesBuffer(struct SedonaSpatialIndexContext* context,
--- a/c/sedona-libgpuspatial/src/error.rs
+++ b/c/sedona-libgpuspatial/src/error.rs
@@ -18,6 +18,7 @@ use arrow_schema::ArrowError;
 use std::fmt;
 use thiserror::Error;
 
+/// Errors that can occur during GPU spatial operations.
 #[derive(Error, Debug)]
 pub enum GpuSpatialError {
     Arrow(ArrowError),
--- a/c/sedona-libgpuspatial/src/error.rs
+++ b/c/sedona-libgpuspatial/src/error.rs
@@ -1,0 +1,60 @@
+// Licensed to the Apache Software Foundation (ASF) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership.  The ASF licenses this file
+// to you under the Apache License, Version 2.0 (the
+// "License"); you may not use this file except in compliance
+// with the License.  You may obtain a copy of the License at
+//
+//   http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+use arrow_schema::ArrowError;
+use std::fmt;
+use thiserror::Error;
+
+#[derive(Error, Debug)]
+pub enum GpuSpatialError {
+    Arrow(ArrowError),
+    Init(String),
+    PushBuild(String),
+    FinishBuild(String),
+    Probe(String),
+    Refine(String),
+}
+
+impl From<ArrowError> for GpuSpatialError {
+    fn from(value: ArrowError) -> Self {
+        GpuSpatialError::Arrow(value)
+    }
+}
+
+impl fmt::Display for GpuSpatialError {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        match self {
+            GpuSpatialError::Arrow(error) => {
+                write!(f, "{error}")
+            }
+            GpuSpatialError::Init(errmsg) => {
+                write!(f, "Initialization failed: {}", errmsg)
+            }
+            GpuSpatialError::PushBuild(errmsg) => {
+                write!(f, "Push build failed: {}", errmsg)
+            }
+            GpuSpatialError::FinishBuild(errmsg) => {
+                write!(f, "Finish building failed: {}", errmsg)
+            }
+            GpuSpatialError::Probe(errmsg) => {
+                write!(f, "Probe failed: {}", errmsg)
+            }
+            GpuSpatialError::Refine(errmsg) => {
+                write!(f, "Refine failed: {}", errmsg)
+            }
+        }
+    }
+}
--- a/c/sedona-libgpuspatial/src/error.rs
+++ b/c/sedona-libgpuspatial/src/error.rs
@@ -56,6 +56,9 @@ impl fmt::Display for GpuSpatialError {
             GpuSpatialError::Refine(errmsg) => {
                 write!(f, "Refine failed: {}", errmsg)
             }
+            GpuSpatialError::GpuNotAvailable => {
+                write!(f, "GPU is not available")
+            }
         }
     }
 }
--- a/c/sedona-libgpuspatial/src/error.rs
+++ b/c/sedona-libgpuspatial/src/error.rs
@@ -26,6 +26,7 @@ pub enum GpuSpatialError {
     FinishBuild(String),
     Probe(String),
     Refine(String),
+    GpuNotAvailable,
 }
 
 impl From<ArrowError> for GpuSpatialError {
--- a/c/sedona-libgpuspatial/src/lib.rs
+++ b/c/sedona-libgpuspatial/src/lib.rs
@@ -24,7 +24,7 @@ mod libgpuspatial;
 mod libgpuspatial_glue_bindgen;
 mod options;
 mod predicate;
-// Re-export to the users
+
 pub use error::GpuSpatialError;
 pub use options::GpuSpatialOptions;
 pub use predicate::GpuSpatialRelationPredicate;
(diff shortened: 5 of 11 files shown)


c/sedona-libgpuspatial/src/libgpuspatial.rs
```diff
@@ -0,0 +1,523 @@
...
+        build_data_type: &DataType,
+        probe_data_type: &DataType,
+    ) -> Result<(), GpuSpatialError> {
+        let ffi_build_schema = FFI_ArrowSchema::try_from(build_data_type)?;
+        let ffi_probe_schema = FFI_ArrowSchema::try_from(probe_data_type)?;
+
+        if let Some(init_schema_fn) = self.refiner.init_schema {
+            // Use pointer casting instead of transmute for safer FFI
+            let ffi_build_ptr = &ffi_build_schema as *const _ as *const ArrowSchema;
+            let ffi_probe_ptr = &ffi_probe_schema as *const _ as *const ArrowSchema;
```
paleolimbot: These functions should take `&mut self`, because they must not be called concurrently (i.e., the caller must own a single mutable reference). In general you should not have to do `as *const _ as *mut _` (use `&mut self.refiner as *mut _`).

c/sedona-libgpuspatial/src/libgpuspatial.rs
```diff
@@ -0,0 +1,523 @@
...
+
+    pub fn push_build(&self, array: &ArrayRef) -> Result<(), GpuSpatialError> {
+        // Keep ffi_array alive until the C function returns
+        let (ffi_array, _ffi_schema) = arrow_array::ffi::to_ffi(&array.to_data())?;
+
+        if let Some(push_build_fn) = self.refiner.push_build {
+            let ffi_array_ptr = &ffi_array as *const _ as *const ArrowArray;
+            unsafe {
+                check_ffi_call(
+                    || push_build_fn(&self.refiner as *const _ as *mut _, ffi_array_ptr as *mut _),
```
paleolimbot: On the interface side, the refiner callback should accept a `const ArrowArray*` if it doesn't intend on releasing it. These is the same pattern used by `ArrowArrayViewSetArray()`, which only reads from its array input and will never release it.

c/sedona-libgpuspatial/src/libgpuspatial.rs
```diff
@@ -0,0 +1,523 @@
...
+    pub fn refine(
+        &self,
+        probe_array: &ArrayRef,
+        predicate: GpuSpatialRelationPredicateWrapper,
+        build_indices: &mut Vec<u32>,
+        probe_indices: &mut Vec<u32>,
+    ) -> Result<(), GpuSpatialError> {
+        let (ffi_array, _ffi_schema) = arrow_array::ffi::to_ffi(&probe_array.to_data())?;
+
+        if let Some(refine_fn) = self.refiner.refine {
```
paleolimbot: Should this code be responsible for reserving some capacity on `build_indices` and/or `probe_indices`?

c/sedona-libgpuspatial/src/libgpuspatial.rs
```diff
@@ -0,0 +1,523 @@
...
+            2 => Ok(Self::Touches),
+            3 => Ok(Self::Contains),
+            4 => Ok(Self::Covers),
+            5 => Ok(Self::Intersects),
+            6 => Ok(Self::Within),
+            7 => Ok(Self::CoveredBy),
+            _ => Err("Invalid GpuSpatialPredicate value"),
+        }
+    }
+}
```
paleolimbot: The details of how this is passed to C shouldn't affect the public API of this crate or be used to construct these...if you need an integer from this you can use a private method in `impl GpuSpatialRelationPredicateWrapper`.

```suggestion
#[derive(Debug, PartialEq, Copy, Clone)]
pub enum GpuSpatialRelationPredicateWrapper {
    Equals,
    Disjoint,
    Touches,
    Contains,
    Covers,
    Intersects,
    Within,
    CoveredBy,
}
```

c/sedona-libgpuspatial/src/libgpuspatial.rs
```diff
@@ -0,0 +1,523 @@
...
+            if !err_ptr.is_null() {
+                CStr::from_ptr(err_ptr).to_string_lossy().into_owned()
+            } else {
+                "Unknown error (null error message)".to_string()
+            }
+        } else {
+            "Unknown error (get_last_error not available)".to_string()
+        };
+
+        log::error!("GpuSpatial FFI Error: {}", error_string);
```
paleolimbot: The returned error should be sufficient? (The caller can always log this if they would like to?)

```suggestion
```

c/sedona-libgpuspatial/src/libgpuspatial.rs
```diff
@@ -0,0 +1,523 @@
...
+use std::sync::{Arc, Mutex};
+
+// ----------------------------------------------------------------------
+// Helper Functions
+// ----------------------------------------------------------------------
+/// Helper to handle the common pattern of calling a C function returning an int status,
+/// checking if it failed, and retrieving the error message if so.
+///
+/// T: The type of the object (Runtime, Index, Refiner) being operated on.
+unsafe fn check_ffi_call<T, F, ErrMap>(
```
paleolimbot: Rust will let you put these at the bottom of the file before they are used (which usually helps the readability of the file).

c/sedona-libgpuspatial/src/libgpuspatial.rs
```diff
@@ -0,0 +1,523 @@
...
+        Ok(())
+    }
+
+    pub fn create_context(&self, ctx: &mut SedonaSpatialIndexContext) {
+        if let Some(create_context_fn) = self.index.create_context {
+            unsafe {
+                create_context_fn(ctx as *mut _);
+            }
+        }
+    }
```
paleolimbot: It would be nice for these contexts to get managed as part of the wrapper (e.g., how these structs keep a reference to an `Arc` of the runtime engine to ensure it is kept alive while required).

c/sedona-libgpuspatial/src/libgpuspatial.rs
```diff
@@ -0,0 +1,523 @@
...
+
+    pub unsafe fn probe(
+        &self,
+        ctx: &mut SedonaSpatialIndexContext,
+        buf: *const f32,
+        n_rects: u32,
+    ) -> Result<(), GpuSpatialError> {
```
paleolimbot: I think this comment got lost in the last review, but unless there's a good reason to avoid this pattern, this should use the same signature as `refine()` (i.e., the caller provides the buffers into which the prove callback can populate the index pairs).

c/sedona-libgpuspatial/src/lib.rs
```diff
@@ -15,6 +15,628 @@
...
+            GpuSpatialRelationPredicate::Disjoint => write!(f, "disjoint"),
+            GpuSpatialRelationPredicate::Touches => write!(f, "touches"),
+            GpuSpatialRelationPredicate::Contains => write!(f, "contains"),
+            GpuSpatialRelationPredicate::Covers => write!(f, "covers"),
+            GpuSpatialRelationPredicate::Intersects => write!(f, "intersects"),
+            GpuSpatialRelationPredicate::Within => write!(f, "within"),
+            GpuSpatialRelationPredicate::CoveredBy => write!(f, "coveredby"),
+        }
+    }
+}
```
paleolimbot: This is duplicated in libgpuspatial.rs (perhaps you can move this and `GpuSpatialOptions` to a new module `options.rs`).

c/sedona-libgpuspatial/src/lib.rs
```diff
@@ -15,6 +15,628 @@
 // specific language governing permissions and limitations
 // under the License.
 
-// Module declarations
+use arrow_schema::DataType;
+use geo::Rect;
+
+#[cfg(gpu_available)]
+pub mod error;
```
paleolimbot:
```suggestion
pub mod error;
```

c/sedona-libgpuspatial/src/lib.rs
```diff
@@ -15,6 +15,628 @@
...
+            GpuSpatialRelationPredicate::Intersects => write!(f, "intersects"),
+            GpuSpatialRelationPredicate::Within => write!(f, "within"),
+            GpuSpatialRelationPredicate::CoveredBy => write!(f, "coveredby"),
+        }
+    }
+}
+
+// Re-export Error type from the sys module abstraction
+pub use sys::GpuSpatialError;
+pub type Result<T> = std::result::Result<T, GpuSpatialError>;
```
paleolimbot:
```suggestion
```

c/sedona-libgpuspatial/src/lib.rs
```diff
@@ -15,6 +15,628 @@
...
+mod sys {
+    use super::libgpuspatial;
+    use super::libgpuspatial_glue_bindgen;
+    use super::*;
+
+    use nvml_wrapper::Nvml;
+    use std::sync::{Arc, Mutex};
+
+    // Re-export the error from the parent module so the public API works
+    pub use super::error::GpuSpatialError;
```
paleolimbot: I feel like I'm missing something with the errors but it's probably best to have them not depend on the gpu availability if at all possible.

c/sedona-libgpuspatial/src/lib.rs
```diff
@@ -15,6 +15,628 @@
...
+    use libgpuspatial_glue_bindgen::SedonaSpatialIndexContext;
+
+    // -- Thread Safety Setup --
+    unsafe impl Send for SedonaSpatialIndexContext {}
+    unsafe impl Send for libgpuspatial_glue_bindgen::GpuSpatialRuntime {}
+    unsafe impl Sync for libgpuspatial_glue_bindgen::GpuSpatialRuntime {}
+    unsafe impl Send for libgpuspatial_glue_bindgen::SedonaFloatIndex2D {}
+    unsafe impl Send for libgpuspatial_glue_bindgen::SedonaSpatialRefiner {}
+    unsafe impl Sync for libgpuspatial_glue_bindgen::SedonaFloatIndex2D {}
+    unsafe impl Sync for libgpuspatial_glue_bindgen::SedonaSpatialRefiner {}
```
paleolimbot: I don't think these are correct `unsafe impl`s. The wrappers are the appropriate place to do the `unsafe impl Send`, and I am not sure you have implemented any that are `unsafe impl Sync`. It is probably better to use a `Mutex` or `RwLock` to ensure something is `impl Sync`.

c/sedona-libgpuspatial/src/lib.rs
```diff
@@ -15,6 +15,628 @@
...
+            )
+        }
+    }
+
+    pub fn is_gpu_available() -> bool {
+        match Nvml::init() {
+            Ok(instance) => instance.device_count().map(|c| c > 0).unwrap_or(false),
+            Err(_) => false,
+        }
+    }
```
paleolimbot: If at all possible I would like to avoid `Nvml` here. I think having this be a function that is exposed from the C API will simplify the build/link portion of this (i.e., I'd like to constrain the place where we get CUDA linking right in exactly one place that is not Rust!)

c/sedona-libgpuspatial/src/error.rs
```diff
@@ -0,0 +1,60 @@
...
+use thiserror::Error;
+
+#[derive(Error, Debug)]
+pub enum GpuSpatialError {
+    Arrow(ArrowError),
+    Init(String),
+    PushBuild(String),
+    FinishBuild(String),
+    Probe(String),
+    Refine(String),
```
paleolimbot, line 29:
```suggestion
    Refine(String),
    GpuNotAvailable,
```

(Maybe this can solve the multiple errors issue?)

paleolimbot (commented):
Thank you! I really appreciate this as a separate PR. All my comments here are mostly just about how to organize this such that the users of these structs don't have to worry about safety and can be composed.

For the index, I think we can organize this as a few structs that explicitly declare their thread safety properties (the wrapper and the builder and the context wrapper are Send; the SharedFloatIndex2D is Send and Sync).

```rust
struct FloatIndex2DBuilder { inner: FloatIndex2DWrapper }

impl FloatIndex2DBuilder {
  pub fn push_build(&mut self)
  pub fn clear(&mut self)

  // Consumes self so that we can turn inner into an Arc
  pub fn finish(self) -> SharedFloatIndex2D
}

struct SharedFloatIndex2D { inner: Arc<FloatIndex2DWrapper> }
unsafe impl Send for SharedFloatIndex2D {}
unsafe impl Sync for SharedFloatIndex2D {}

impl SharedFloatIndex2D {
  fn create_context(&self) -> FloatIndex2D
}

struct FloatIndex2DContext { inner: Arc<FloatIndex2DWrapper>, context: SedonaSpatialIndexContext }
unsafe impl Send for FloatIndex2DContext {}

impl FloatIndex2DContext {
  fn probe(&self)
}

impl Drop for FloatIndex2D {
  // Call destroy context using inner
}

// not pub
struct FloatIndex2DWrapper { inner: SedonaFloatIndex2D }
unsafe impl Send for FloatIndex2DWrapper {}

// The current GpuSpatialIndexFloat2DWrapper
```

The refiner is simpler (it is just Send) and your current wrapper is great (I left a comment about marking methods `mut`...the caller can manage mutable access depending on how it will be used).

A few other general comments to think about for the next revision:

- When you're confident in the interface to be used outside this crate, anything marked `pub` needs documentation. For things that are internal to the crate, you can use `pub(crate)`.
- The names of the user-facing structs shouldn't have `Gpu` in them (they are indexes and refiners implemented in C). For at least the refiner I think I will be able to implement this for Geography (also C++) since we will need that in the join as well.

c/sedona-libgpuspatial/src/libgpuspatial.rs
pwrliang: Since this is a refinement operation, the sizes of build_indices and probe_indices will only decrease, so no reservation is needed.

c/sedona-libgpuspatial/src/libgpuspatial.rs
pwrliang: Do you suggest that the caller should manage the buffer that stores the probe results (build indices and probe indices)? This may not work for the probe interface because the buffer size is unknown to the caller.

c/sedona-libgpuspatial/src/lib.rs
pwrliang: I need two of these as the runtime is a singleton and it is declared as `static GLOBAL_GPUSPATIAL_RUNTIME: Mutex<Option<Arc<Mutex<GpuSpatialRuntimeWrapper>>>> =
        Mutex::new(None);`
```rust
    unsafe impl Send for libgpuspatial_glue_bindgen::GpuSpatialRuntime {}
    unsafe impl Sync for libgpuspatial_glue_bindgen::GpuSpatialRuntime {}
```

pwrliang:
Thanks for your suggestions @paleolimbot I think I have fixed most of the issues except the one related to the design of `probe` interface. Could you elaborate on the current design flaw?

c/sedona-libgpuspatial/src/lib.rs
```diff
@@ -15,6 +15,537 @@
...
+        Mutex::new(None);
+
+    // -- The Actual Implementation Struct --
+    pub struct SpatialImpl {
+        runtime: Option<Arc<Mutex<GpuSpatialRuntimeWrapper>>>,
+        // Index state is either building (Builder) or built (Shared)
+        index_builder: Option<FloatIndex2DBuilder>,
+        index: Option<SharedFloatIndex2D>,
+        refiner: Option<GpuSpatialRefinerWrapper>,
+    }
```
Kontinuation: I have some suggestions about the overall design of SpatialImpl. I am not familiar with the special complexities brought by GPU programming, so there might be some misunderstanding.

1. `SpatialImpl` needs to call `init()` to work properly. I think a more Rusty idiom here is RAII: we implement a `try_new` that accepts `options: GpuSpatialOptions`, and once the `SpatialImpl` is created it is ready to work. No two-step construction.
2. `index` does not need to be a member of `SpatialImpl`. `index_finish_building` can just consume the builder and return a `SharedFloatIndex2D` value to the caller. `probe` doesn't actually use any other field of `SpatialImpl` — it only needs the index — so it could take the index as a parameter, or even become a method on `SharedFloatIndex2D` directly.

After doing this, `SpatialImpl` mostly won't need `Option` members anymore. `runtime` and `refiner` get set in `try_new`, `index` is gone (returned to the caller). The only one that still needs to be `Option` is `index_builder`, since it's consumed by `finish` — but that's a clear lifecycle, not a "was init called?" check. We get rid of all the `.ok_or_else(|| GpuSpatialError::Init(...))` ceremony in the refiner and probe methods.

If someone wants to defer initialization, they can use `Option<SpatialImpl>`. The requirement of "only use it after initializing" just becomes transparent.

Kontinuation: Just discovered that Dewey has already gave a more comprehensive suggestion. https://github.com/apache/sedona-db/pull/586#pullrequestreview-3781255055

pwrliang:
@Kontinuation @paleolimbot Thank you for your suggestions. I have refactored both index and refiner with the builder pattern. There are no Option fields anymore. The resource allocation also follows the RAII idiom. Originally, I thought we could reuse the index and refiner after the execution of the spatial join, so there would be less memory allocation overhead. It may be an anti-pattern in Rust, but I'm happy with the changes.

c/sedona-libgpuspatial/src/options.rs
```diff
@@ -0,0 +1,25 @@
...
+// under the License.
+
+pub struct GpuSpatialOptions {
+    pub cuda_use_memory_pool: bool,
+    pub cuda_memory_pool_init_percent: i32,
+    pub concurrency: u32,
+    pub device_id: i32,
+    pub compress_bvh: bool,
+    pub pipeline_batches: u32,
+}
```
paleolimbot, line 32: Can you add documentation to these public items? I know this is repetitive but this is the place that other SedonaDB developers will interact with this and it is worth a brief note about what these are and where to look for more details would be helpful.

c/sedona-libgpuspatial/src/predicate.rs
```diff
@@ -0,0 +1,61 @@
...
+    Contains,
+    Covers,
+    Intersects,
+    Within,
+    CoveredBy,
+}
+
+impl GpuSpatialRelationPredicate {
+    /// Internal helper to convert the Rust enum to the C-compatible integer.
+    #[allow(dead_code)] // not used if the GPU feature is disabled
```
paleolimbot: I think this will work instead of allowing dead code:

```suggestion

#[cfg(gpu_available)]
impl GpuSpatialRelationPredicate {
    /// Internal helper to convert the Rust enum to the C-compatible integer.
```

c/sedona-libgpuspatial/Cargo.toml
```diff
@@ -35,3 +35,17 @@ gpu = []
 bindgen = "0.72.1"
 cmake = "0.1"
 which = "8.0"
+
+[dependencies]
+arrow-array = { workspace = true, features = ["ffi"] }
+arrow-schema = { workspace = true }
+thiserror = { workspace = true }
+log = { workspace = true }
+geo = { workspace = true }
```
paleolimbot: If we can get away with not requiring `geo` here except in the development dependencies. You can use `geo-types` if all you need is `Rect` (much lighter).

c/sedona-libgpuspatial/src/libgpuspatial.rs
```diff
@@ -0,0 +1,583 @@
...
+// Spatial Index - Builder
+// ----------------------------------------------------------------------
+
+/// Builder for the Spatial Index. This struct has exclusive ownership
+/// and is not thread-safe (Send but not Sync) because building is a
+/// single-threaded operation.
+pub struct FloatIndex2DBuilder {
+    inner: FloatIndex2DWrapper,
+}
+
```
paleolimbot, line 130: (Apologies if I missed this somewhere else)

```suggestion

unsafe impl Send for FloatIndex2DBuilder {}

```

c/sedona-libgpuspatial/src/libgpuspatial.rs
```diff
@@ -0,0 +1,583 @@
...
+    pub fn clear(&mut self) {
+        if let Some(clear_fn) = self.inner.index.clear {
+            unsafe {
+                clear_fn(&mut self.inner.index as *mut _);
+            }
+        }
+    }
+
+    pub unsafe fn push_build(
```
paleolimbot: The other functions here use an unsafe block rather than mark their functions as unsafe, which I think is better than `unsafe fn` here (i.e., the caller of this function shouldn't need an unsafe block)

c/sedona-libgpuspatial/src/libgpuspatial.rs
paleolimbot: Got it! That's a good reason for the context to manage the results. Perhaps add that to the documentation to the C API if it's not there already since it wasn't clear to me.

c/sedona-libgpuspatial/src/libgpuspatial.rs
```diff
@@ -0,0 +1,583 @@
...
+/// This struct is Send (can be moved between threads) but NOT Sync.
+pub struct FloatIndex2DContext {
+    inner: Arc<FloatIndex2DWrapper>, // Shared reference to the index wrapper to ensure it lives as long as the context
+    context: SedonaSpatialIndexContext, // The actual C context struct that holds thread-local state and result buffers
+}
+
+unsafe impl Send for FloatIndex2DContext {}
+
+impl FloatIndex2DContext {
+    fn get_indices_buffer_helper(
```
paleolimbot: I think this is the appropriate place for `probe()` so that the caller doesn't need to keep track of the shared index AND the context.

I think you can also have the input and output here be more Rust-y...this file is the right level of abstraction to handle creating slices from pointers and vice versa.

```suggestion
impl FloatIndex2DContext {
    pub fn probe(&mut self, &[(f64, f64, f64, f64)]) -> Result<(&[u32], &[u32]), GpuSpatialError> {
        todo!()
    }

    fn get_indices_buffer_helper(
```

c/sedona-libgpuspatial/src/libgpuspatial.rs
```diff
@@ -0,0 +1,583 @@
...
+    }
+}
+
+impl Drop for FloatIndex2DContext {
+    fn drop(&mut self) {
+        if let Some(destroy_context_fn) = self.inner.index.destroy_context {
+            unsafe {
+                destroy_context_fn(&mut self.context);
+            }
+        }
```
paleolimbot: Make sure all of these functions do something loud and error-y for the case where `self.inner.index.destroy_context` is `None`. It's unlikely but if it happens we want to know about it! Using `.expect()` is probably best (I think some of our other wrappers return an internal error when they can but either is fine).

c/sedona-libgpuspatial/src/libgpuspatial.rs
```diff
@@ -0,0 +1,583 @@
...
+            unsafe {
+                release_fn(&mut self.refiner as *mut _);
+            }
+        }
+    }
+}
+
+pub struct GpuSpatialRefinerBuilder {
+    inner: RefinerWrapper,
+}
```
paleolimbot: I am not sure you need both a Builder and a Wrapper here because to my reading a refiner can never be queried in parallel.

In other words, I think you can just call this `Refiner` and add `pub fn refine(&mut self)`.

c/sedona-libgpuspatial/src/lib.rs
```diff
@@ -15,6 +15,437 @@
...
+        pub fn runtime(&self) -> Arc<Mutex<GpuSpatialRuntimeWrapper>> {
+            self.runtime.clone()
+        }
+    }
+
+    pub struct IndexBuilderImpl {
+        inner: IndexBuilderInner,
+    }
+
+    impl IndexBuilderImpl {
```
paleolimbot: Do you need this extra level of wrapper? I think the implementations you have in libgpuspatial.rs are high-level enough that this extra level isn't needed.

In other words, I think `pub use FloatIndex2D` and `pub use Refiner` would let you avoid this level of indirection.

c/sedona-libgpuspatial/src/lib.rs
```diff
@@ -15,6 +15,437 @@
...
+        ) -> Result<()> {
+            Err(GpuSpatialError::GpuNotAvailable)
+        }
+    }
+}
+
+/// Builder for creating a GPU Spatial Index.
+pub struct GpuSpatialIndexBuilder {
+    inner: sys::IndexBuilderImpl,
+}
```
paleolimbot: I think you can avoid this level of indirection with `pub use sys::{Float2DIndexBuilder, Refiner}`.

paleolimbot (commented):
Thank you! This is much closer...all comments are purely organizational about how to ensure callers can just use Rust safety/idioms and how to minimize duplication of the wrapper code.

> Originally, I thought we could reuse the index and refiner after the execution of the spatial join, so there would be less memory allocation overhead.

I may be misinterpreting some of this so feel free to correct me, but I think that the building of the spatial index and the querying of the spatial index are always distinct from our perspective, so the pattern of Builder -> SharedIndex with one context per partition will work, and the overhead of disposing the index and creating the index will only happen once per query (which I think is OK). It is possible to reuse it between queries even with the current setup (you can move out of an Arc if you hold the only reference) but I think this is more work than it is worth for now.

For the refiner, I think there is no getting around that it is `Send` but not `Sync`: all its methods require `&mut self` and concurrent access has to be managed at the Rust level. There are a few ways to do that (e.g., one refiner per partition if that's an option).

c/sedona-libgpuspatial/src/libgpuspatial.rs
pwrliang: Refiner can be used in parallel, just like the index. The geometries from the build side are pushed to the refiner, and will be shared by all partitions on the stream side.

paleolimbot: Ah, I see...but we don't need the Context because there because the indices are owned by the caller for the `refine()` call. I had thought that the purpose of the index was to reduce the number of geometries whose full definitions needed to be loaded into the GPU.

If the only purpose of the context is to accumulate indices, it would be nice to eliminate it because it complicates the interface considerably. Would a callback work?

```c
static int CProbe(self_t* self, SedonaSpatialIndexContext* context, const float* buf, (*callback)(const uint32_t* build_indices, uint32_t* probe_indices, void* private_data), void* private_data);
```

...this is roughly how GEOS' index is probed as well and eliminates the need to accumulate all of the output pairs in memory at once (potentially n_build times n_probe / 2 ish). Then you don't need to separate the builder for the index.

c/sedona-libgpuspatial/src/libgpuspatial.rs
```diff
@@ -0,0 +1,607 @@
...
+    }
+}
+
+struct RefinerWrapper {
+    refiner: SedonaSpatialRefiner,
+    _runtime: Arc<Mutex<GpuSpatialRuntimeWrapper>>,
+}
+
+unsafe impl Send for RefinerWrapper {}
+unsafe impl Sync for RefinerWrapper {}
```
paleolimbot:
```suggestion
unsafe impl Send for RefinerWrapper {}
```

I think the `unsafe impl Sync` would go on the `GpuSpatialRefinerBuilder`.

c/sedona-libgpuspatial/src/libgpuspatial.rs
```diff
@@ -0,0 +1,607 @@
...
+        } else {
+            return Err(GpuSpatialError::FinishBuild(
+                "finish_building function is None".to_string(),
+            ));
+        }
+
+        Ok(GpuSpatialRefinerWrapper {
+            inner: Arc::new(self.inner),
+        })
+    }
```
paleolimbot, line 433: I think that you can safely add `fn refine(&self, ...)` here without separating the builder (and you can impl unsafe Sync for this if that can be done safely from multiple threads).

paleolimbot (commented):
Apologies for misunderstanding the concurrency of the refiner! If my callback suggestion works for the index probe, this could be just:

```rust
struct FloatIndex2D { inner: FloatIndex2DWrapper }
unsafe impl Send for FloatIndex2D {}
unsafe impl Sync for FloatIndex2D {}

impl FloatIndex2D {
  pub fn push_build(&mut self)
  pub fn clear(&mut self)
  pub fn finish_building(&mut self)
  pub fn probe(&self, &[(f32, f32, f32, f32)], impl FnMut(&[uint32], &[uint32]) -> bool);
}

struct Refiner { inner: RefinerWrapper }
unsafe impl Send for Refiner {}
unsafe impl Sync for Refiner {}

impl Refiner {
  pub fn init_schema(&mut self)
  pub fn push_build(&mut self)
  pub fn clear(&mut self)
  pub fn finish_building(&mut self)
  pub fn refine(&self, array: &ArrayRef, build_indices: &mut Vec<u32>, probe_indices: &mut Vec<u32>);
}
```

pwrliang:
Thank you! I have changed the signature of the probe, making it take a callback function to return results. `libgpuspatial.rs` has also been refactored to remove the builders.

c/sedona-libgpuspatial/src/predicate.rs
```diff
@@ -0,0 +1,61 @@
...
+    Disjoint,
+    Touches,
+    Contains,
+    Covers,
+    Intersects,
+    Within,
+    CoveredBy,
+}
+
+#[allow(dead_code)] // not used if the GPU feature is disabled
```
paleolimbot:
```suggestion
#[cfg(not(gpu_available))] // not used if the GPU feature is disabled
```

c/sedona-libgpuspatial/src/libgpuspatial.rs
```diff
@@ -0,0 +1,502 @@
...
+                GpuSpatialError::FinishBuild,
+            )
+        }
+    }
+
+    pub fn probe(
+        &self,
+        buf: *const f32,
+        n_rects: u32,
+    ) -> Result<(Vec<u32>, Vec<u32>), GpuSpatialError> {
```
paleolimbot:
```suggestion
    pub fn probe(
        &self,
        rects: &[(f32, f32, f32, f32)],
    ) -> Result<(Vec<u32>, Vec<u32>), GpuSpatialError> {
       let n_rects = rects.len() as u32;
       let buf: *const f32 = unsafe { transmute(rects.as_ptr()) };
```

c/sedona-libgpuspatial/src/lib.rs
```diff
@@ -15,6 +15,305 @@
...
+    use super::libgpuspatial;
+    use super::*;
+    use libgpuspatial::GpuSpatialRuntimeWrapper;
+    use std::sync::{Arc, Mutex};
+
+    pub type Result<T> = std::result::Result<T, GpuSpatialError>;
+
+    // Global Runtime State
+    unsafe impl Send for GpuSpatialRuntimeWrapper {}
+    unsafe impl Sync for GpuSpatialRuntimeWrapper {}
```
paleolimbot: These should go right below the struct definition for `GpuSpatialRuntimeWrapper` (makes it easier to spot members or functions of the struct that might not be Send or Sync)

c/sedona-libgpuspatial/src/lib.rs
```diff
@@ -15,6 +15,305 @@
...
+                    return Err(GpuSpatialError::Init(format!(
+                        "Runtime conflict: Initialized on Device {}, requested Device {}.",
+                        existing_runtime.device_id, options.device_id
+                    )));
+                }
+                // Return the existing one
+                return Ok(Self {
+                    runtime: existing_runtime.clone(),
+                });
+            }
```
paleolimbot, line 68: I think a `OnceLock<Arc<GpuSpatialRuntimeWrapper>>` is the Rusty way to do this. In action:

https://github.com/apache/sedona-db/blob/7eac903e08c84ae9aa6396d2ff1e8101d1e6b8df/rust/sedona-functions/src/st_setsrid.rs#L434-L448

c/sedona-libgpuspatial/src/libgpuspatial.rs
```diff
@@ -0,0 +1,502 @@
...
+        unsafe {
+            release_fn(&mut self.index as *mut _);
+        }
+    }
+}
+
+pub struct FloatIndex2D {
+    inner: FloatIndex2DWrapper,
+}
+
```
paleolimbot, line 130:
```suggestion
unsafe impl Send for FloatIndex2D;
unsafe impl Sync for FloatIndex2D;
```

c/sedona-libgpuspatial/src/libgpuspatial.rs
```diff
@@ -0,0 +1,502 @@
...
+        let release_fn = self.refiner.release.expect("release function is None");
+        unsafe {
+            release_fn(&mut self.refiner as *mut _);
+        }
+    }
+}
+pub struct Refiner {
+    inner: RefinerWrapper,
+}
+
```
paleolimbot, line 330:
```suggestion
unsafe impl Send for Refiner;
unsafe impl Sync for Refiner;
```

c/sedona-libgpuspatial/src/lib.rs
```diff
@@ -15,6 +15,305 @@
...
+                runtime: arc_wrapper,
+            })
+        }
+    }
+
+    pub struct GpuSpatialIndex {
+        inner: libgpuspatial::FloatIndex2D,
+    }
+
+    impl GpuSpatialIndex {
```
paleolimbot, line 99: I think eventually we will want to collapse all of this wrapper code into the `FloatIndex2D` but no need to do that now. One good reason to do this is that it means you only have to document it once. If we keep this wrapper for this PR, we still need to document both (although you can link to the inner documentation from the documentation for this struct).

c/sedona-libgpuspatial/src/libgpuspatial.rs
```diff
@@ -0,0 +1,502 @@
...
+    }
+    Ok(())
+}
+
+unsafe extern "C" fn probe_callback_wrapper(
+    build_indices: *const u32,
+    probe_indices: *const u32,
+    length: u32,
+    user_data: *mut c_void,
+) {
```
paleolimbot: Not for this PR, but we should expose the visitor interface via Rust because it is so much more memory efficient if there are any polygons whose bbox is not selective.

c/sedona-libgpuspatial/libgpuspatial/include/gpuspatial/gpuspatial_c.h
```diff
@@ -84,30 +84,23 @@ struct SedonaFloatIndex2D {
...
-   *
-   * @return A pointer to the buffer and its length
-   */
-  void (*get_probe_indices_buffer)(struct SedonaSpatialIndexContext* context,
-                                   uint32_t** probe_indices,
-                                   uint32_t* probe_indices_length);
+               const float* buf, uint32_t n_rects,
+               void (*callback)(const uint32_t* build_indices,
+                                const uint32_t* probe_indices, uint32_t length,
+                                void* user_data),
```
paleolimbot: We should let the callback return an error code. Right now you only visit exactly one pre-allocated buffer but hopefully we can split that up into multiple calls for huge probe results and returning an error code will allow that process to be cancelled.

```suggestion
               int (*callback)(const uint32_t* build_indices,
                               const uint32_t* probe_indices, uint32_t length,
                               void* user_data),
```

c/sedona-libgpuspatial/libgpuspatial/src/gpuspatial_c.cc
```diff
@@ -194,12 +192,25 @@ struct GpuSpatialIndexFloat2DExporter {
...
+    // Do not use SafeExecute because this method is thread-safe and we don't want to set
+    // last_error for the whole index if one thread encounters an error
+    try {
       auto* rects = reinterpret_cast<const spatial_index_t::box_t*>(buf);
       auto& buff = static_cast<context_t*>(context->private_data)->payload;
       use_index(self).Probe(rects, n_rects, &buff.build_indices, &buff.probe_indices);
-    });
+      callback(buff.build_indices.data(), buff.probe_indices.data(),
+               buff.build_indices.size(), user_data);
+      return 0;
```
paleolimbot:
```suggestion
      return callback(buff.build_indices.data(), buff.probe_indices.data(),
                      buff.build_indices.size(), user_data);
```

c/sedona-libgpuspatial/libgpuspatial/src/gpuspatial_c.cc
```diff
@@ -194,12 +192,25 @@ struct GpuSpatialIndexFloat2DExporter {
...
+    } catch (const std::exception& e) {  // user should call context_get_last_error
+      return EINVAL;
+    } catch (...) {
+      return EINVAL;
+    }
```
paleolimbot, line 216: You should be able to set  the context error string here safely, correct? If `use_index(self).Probe(...)` already does this, you can add a comment here. If the process of probing never returns a meaningful error message, you can set the context error message to "Unknown error" or something.

c/sedona-libgpuspatial/libgpuspatial/src/gpuspatial_c.cc
```diff
@@ -194,12 +192,25 @@ struct GpuSpatialIndexFloat2DExporter {
   }
 
   static int CProbe(self_t* self, SedonaSpatialIndexContext* context, const float* buf,
-                    uint32_t n_rects) {
-    return SafeExecute(static_cast<context_t*>(context->private_data), [&] {
+                    uint32_t n_rects,
+                    void (*callback)(const uint32_t* build_indices,
+                                     const uint32_t* probe_indices, uint32_t length,
+                                     void* user_data),
```
paleolimbot: This callback should return an error code (this will let long-running probe operations be cancelled if or when you split up the output of huge probe results).

```suggestion
                    int (*callback)(const uint32_t* build_indices,
                                    const uint32_t* probe_indices, uint32_t length,
                                    void* user_data),
```

paleolimbot (commented):
Thank you for continuing to work on this!

I think the Rust interface you've landed on here is great! A few suggestions on details, but the main thing this needs before merging is documentation for the `FloatIndex2D`, `Refiner`, GPU runtime, and wrappers in `lib.rs` if you choose to keep those for the time being.

c/sedona-libgpuspatial/libgpuspatial/src/gpuspatial_c.cc
pwrliang: Thanks, forgot to set the error message.

c/sedona-libgpuspatial/src/libgpuspatial.rs
pwrliang: Should we also annotate the public interfaces (GpuSpatialIndex and GpuSpatialRefiner) with these traits?

c/sedona-libgpuspatial/src/lib.rs
pwrliang: I need `get_or_try_init`, which is unstable in the latest Rust stable release. Can we replace this later?
```rust
            let runtime_ref = GLOBAL_GPUSPATIAL_RUNTIME.get_or_try_init(|| {
                let out_path = PathBuf::from(env!("OUT_DIR"));
                let ptx_root = out_path.join("share/gpuspatial/shaders");

                let ptx_root_str = ptx_root
                    .to_str()
                    .ok_or_else(|| GpuSpatialError::Init("Invalid PTX path".to_string()))?;

                // Return the raw Wrapper here (not an Arc)
                libgpuspatial::GpuSpatialRuntimeWrapper::try_new(
                    options.device_id,
                    ptx_root_str,
                    options.cuda_use_memory_pool,
                    options.cuda_memory_pool_init_percent,
                )
            })?;
```

pwrliang:
Thank you for the review. I think the above issues have been addressed.

c/sedona-libgpuspatial/src/libgpuspatial.rs
paleolimbot: If a use case pops up where it is needed you can...I believe that Rust will automatically know that a struct whose members are all `Send`/`Sync` is also `Send`/`Sync` but I'm not clear on the details. You will get an error implementing the next level up if you need to.

paleolimbot (approved):
Thank you for your continued effort on this!

PR #589, opened 2026-02-10
feat(rust/sedona-schema): Add Int8, UInt64, Int64 band data types, don't panic in BandMetadataRef

Kontinuation:
## Summary
- Add three new `BandDataType` variants: `Int8`, `UInt64`, `Int64`
- Derive `Hash` and `Copy` on both `BandDataType` and `StorageType` enums for ergonomic use
- Update match arms in `BandMetadataRefImpl::data_type()` to decode the new ordinals
- Update data_type() and storage_type() to return Result, don't panic when observing unknown values
- Extend test coverage and test data generators (`generate_random_band_data`, `get_nodata_value_for_type`) to handle all 10 band data types

### DIFF
--- a/rust/sedona-raster/src/array.rs
+++ b/rust/sedona-raster/src/array.rs
@@ -133,19 +133,24 @@ impl<'a> BandMetadataRef for BandMetadataRefImpl<'a> {
         }
     }
 
-    fn storage_type(&self) -> StorageType {
-        match self.storage_type_array.value(self.band_index) {
+    fn storage_type(&self) -> Result<StorageType, ArrowError> {
+        let value = self.storage_type_array.value(self.band_index);
+        let storage_type = match value {
             0 => StorageType::InDb,
             1 => StorageType::OutDbRef,
-            _ => panic!(
-                "Unknown storage type: {}",
-                self.storage_type_array.value(self.band_index)
-            ),
-        }
+            _ => {
+                return Err(ArrowError::InvalidArgumentError(format!(
+                    "Unknown storage type: {}",
+                    value
+                )))
+            }
+        };
+        Ok(storage_type)
     }
 
-    fn data_type(&self) -> BandDataType {
-        match self.datatype_array.value(self.band_index) {
+    fn data_type(&self) -> Result<BandDataType, ArrowError> {
+        let value = self.datatype_array.value(self.band_index);
+        let band_data_type = match value {
             1 => BandDataType::UInt8,
             2 => BandDataType::UInt16,
             3 => BandDataType::Int16,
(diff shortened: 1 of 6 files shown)


rust/sedona-schema/src/raster.rs
```diff
@@ -88,19 +88,23 @@ impl RasterSchema {
 
 /// Band data type enumeration for raster bands.
 ///
+/// Values align with GDALDataType ordinals for non-complex types (excluding Unknown).
 /// Only supports basic numeric types.
 /// In future versions, consider support for complex types used in
 /// radar and other wave-based data.
```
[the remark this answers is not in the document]
Kontinuation: Will improve the comment to make it more clear

rust/sedona-raster/src/array.rs
```diff
@@ -146,13 +146,16 @@ impl<'a> BandMetadataRef for BandMetadataRefImpl<'a> {
...
+            3 => BandDataType::Int16,
+            4 => BandDataType::UInt32,
+            5 => BandDataType::Int32,
+            6 => BandDataType::Float32,
+            7 => BandDataType::Float64,
+            12 => BandDataType::UInt64,
+            13 => BandDataType::Int64,
+            14 => BandDataType::Int8,
             _ => panic!(
                 "Unknown band data type: {}",
```
paleolimbot: Should this return an error or is this validated before we get here? (i.e., do we risk a panic for a corrupted raster array?)

rust/sedona-raster/src/builder.rs
```diff
@@ -605,6 +605,7 @@ mod tests {
         // Test all BandDataType variants
         let test_cases = vec![
             (BandDataType::UInt8, vec![1u8, 2u8, 3u8, 4u8]),
+            (BandDataType::Int8, vec![255u8, 254u8, 253u8, 252u8]), // -1, -2, -3, -4 as i8
```
paleolimbot, line 608: If possible this may be a good time to add test for a corrupted band identifier to make sure it doesn't panic

rust/sedona-schema/src/raster.rs
```diff
@@ -88,19 +88,22 @@ impl RasterSchema {
 
 /// Band data type enumeration for raster bands.
 ///
-/// Only supports basic numeric types.
-/// In future versions, consider support for complex types used in
-/// radar and other wave-based data.
+/// Ordinals match GDALDataType for real-valued pixel types only.
+/// GDT_Unknown (0) and complex types (CInt16=8, CInt32=9, CFloat32=10, CFloat64=11)
+/// are intentionally omitted.
```
paleolimbot: You could handle these with an `Other(usize)` variant (whose internal representation is opaque bytes).

rust/sedona-schema/src/raster.rs
```diff
@@ -88,19 +88,22 @@ impl RasterSchema {
...
 #[repr(u16)]
-#[derive(Clone, Debug, PartialEq, Eq)]
+#[derive(Clone, Debug, PartialEq, Eq, Hash, Copy)]
 pub enum BandDataType {
```
paleolimbot: This is probably more flexible without the `#[repr(u16)]` since at some point we may need a variant to handle other types of data (e.g., generic fixed size). It's great to align our representation to GDAL's but it is probably better scoped as `try_to_gdal_band_data_type() -> u16` and `try_from_gdal_band_data_type(gdal_type: u16)`.

Kontinuation: The values here were happened to be the same as GDAL band data types. We'll have functions for converting between then rather than implicitly relying on it. https://github.com/Kontinuation/sedona-db/blob/0e73e17a79332cf7e58b9881084d86dd3f3531c1/rust/sedona-raster-gdal/src/gdal_common.rs#L50-L88

I decided to simply define the values without GDAL awareness just as before.

rust/sedona-raster/src/builder.rs
Kontinuation: Added `test_invalid_band_metadata_returns_err` for testing this

rust/sedona-raster/src/array.rs
Kontinuation: Sure. I found that storage_type also need to be fixed, I have changed both storage_type() and data_type() to return Result<...>.

rust/sedona-schema/src/raster.rs
Kontinuation: We don't need `Other` variant for now since we decided to not enforce the values of `BandDataType` to match GDAL. I believe that it is a better design.

PR #590, opened 2026-02-10
refactor(rust/sedona-raster-functions): extract CachedCrsToSRIDMapping and simplify SRID/CRS logic

Kontinuation:
## Summary

- Extract `CachedCrsToSRIDMapping` struct into `sedona-schema::crs` for reusable CRS-to-SRID caching, replacing ad-hoc `HashMap` caching in `ST_SRID` and duplicated logic in `RS_SRID`
- Refactor `RS_SRID` and `RS_CRS` in `sedona-raster-functions` to use early-return patterns instead of deeply nested match arms
- Add `crs_str_ref()` to `RasterRefImpl` for zero-copy CRS string access
- Make `with_global_proj_engine` generic (`pub fn ... -> Result<R>`) and public for external use
- Add test coverage for rasters with missing CRS and CRS without SRID

## Dependencies

- Depends on #588 (`pr2-better-band-metadata-ref`)

### DIFF
--- a/c/sedona-proj/src/lib.rs
+++ b/c/sedona-proj/src/lib.rs
@@ -19,5 +19,5 @@ mod proj;
 mod proj_dyn_bindgen;
 pub mod register;
 pub mod sd_order_lnglat;
-pub mod st_transform;
+mod st_transform;
 pub mod transform;
--- a/c/sedona-proj/src/st_transform.rs
+++ b/c/sedona-proj/src/st_transform.rs
@@ -389,7 +389,10 @@ pub fn configure_global_proj_engine(builder: ProjCrsEngineBuilder) -> Result<()>
 
 /// Do something with the global thread-local PROJ engine, creating it if it has not
 /// already been created.
-pub fn with_global_proj_engine<R, F: FnMut(&CachingCrsEngine<ProjCrsEngine>) -> Result<R>>(
+pub(crate) fn with_global_proj_engine<
+    R,
+    F: FnMut(&CachingCrsEngine<ProjCrsEngine>) -> Result<R>,
+>(
     mut func: F,
 ) -> Result<R> {
     PROJ_ENGINE.with(|engine_cell| {
@@ -425,10 +428,6 @@ pub fn with_global_proj_engine<R, F: FnMut(&CachingCrsEngine<ProjCrsEngine>) -> 
 /// or never set to use all default settings.
 static PROJ_ENGINE_BUILDER: RwLock<Option<ProjCrsEngineBuilder>> =
     RwLock::<Option<ProjCrsEngineBuilder>>::new(None);
-
-thread_local! {
-    static PROJ_CACHE_SIZE: OnceCell<usize> = const { OnceCell::new() };
-}
 
 // CrsTransform backed by PROJ is not thread safe, so we define the cache as thread-local
 // to avoid race conditions.


rust/sedona-raster-functions/src/rs_srid.rs
```diff
@@ -97,45 +96,17 @@ impl SedonaScalarKernel for RsSrid {
...
-                        }
-                    }
-                }
-            }
+            let Some(raster) = raster_opt else {
+                builder.append_null();
+                return Ok(());
+            };
+
+            let maybe_crs = raster.crs_str_ref();
```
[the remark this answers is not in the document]
Kontinuation: This is intentional. crs_str_ref() returns reference with longer lifetime.

rust/sedona-raster-functions/src/rs_srid.rs
```diff
@@ -167,26 +138,14 @@ impl SedonaScalarKernel for RsCrs {
...
-                },
-            }
+            let Some(raster) = raster_opt else {
+                builder.append_null();
+                return Ok(());
+            };
+
+            // This is similar to ST_CRS: if no CRS is set, return "0"
+            let crs_str = raster.crs().unwrap_or("0");
+            builder.append_value(crs_str);
```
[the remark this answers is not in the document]
Kontinuation: I think we should make RS_CRS consistent with ST_CRS. If this is not the case before, we have fixed it here.

c/sedona-proj/src/lib.rs
```diff
@@ -19,5 +19,5 @@ mod proj;
 mod proj_dyn_bindgen;
 pub mod register;
 pub mod sd_order_lnglat;
-mod st_transform;
+pub mod st_transform;
```
paleolimbot: Can this be reverted? (I don't see a reference to `with_global_proj_engine()` in this PR?)

We should probably move the PROJ engine to its own public module and/or abstract it so we can switch to something faster for simple transforms.

c/sedona-proj/src/st_transform.rs
```diff
@@ -427,6 +426,10 @@ pub(crate) fn with_global_proj_engine(
 static PROJ_ENGINE_BUILDER: RwLock<Option<ProjCrsEngineBuilder>> =
     RwLock::<Option<ProjCrsEngineBuilder>>::new(None);
 
+thread_local! {
+    static PROJ_CACHE_SIZE: OnceCell<usize> = const { OnceCell::new() };
+}
```
paleolimbot: Is this used in this PR?

rust/sedona-schema/src/crs.rs
```diff
@@ -101,6 +108,55 @@ pub fn deserialize_crs_from_obj(crs_value: &serde_json::Value) -> Result<Crs> {
     Ok(Some(Arc::new(projjson)))
 }
 
+/// Translating CRS into integer SRID with a cache to avoid expensive CRS deserialization.
+pub struct CachedCrsToSRIDMapping {
+    cache: HashMap<Cow<'static, str>, u32>,
+}
```
paleolimbot, line 114: Not for this PR, but the abbreviated CRS should probably have this too (I have a batch-local cache for st_setsrid):

https://github.com/apache/sedona-db/blob/cb713ef27b4e8a09e7b825a70c20c40be757ab38/rust/sedona-functions/src/st_setsrid.rs#L510-L542

c/sedona-proj/src/lib.rs
Kontinuation: Good catch. We may need to make it pub later, but definitely not in this PR. I'll revert it.

c/sedona-proj/src/st_transform.rs
Kontinuation: No. I'll remove it.

rust/sedona-schema/src/crs.rs
Kontinuation: I'll handle it when implementing RS_SetSRID and RS_SetCRS.

PR #591, opened 2026-02-10
feat(rust/sedona-functions): Add raster display support to SD_Format

Kontinuation:
## Summary

- Add raster formatting to `SD_Format`, displaying raster metadata (dimensions, geotransform, band count, outdb status) as a human-readable string
- Previously `SD_Format` returned an error for raster types; now it produces output like: `Raster[w=1, h=2, ul=(1.000000, 2.000000), scale=(0.000000, -0.000000), skew=(0.000000, 0.000000), bands=1, outdb=false]`
- Respects the `width_hint` option for truncation
- Add `sedona-raster` dependency to `sedona-functions`

## Example

```
> SELECT RS_Example();
┌────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┐
│                                                        rs_example()                                                        │
│                                                           raster                                                           │
╞════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════╡
│ Raster[w=64, h=32, ul=(43.080000, 79.070000), scale=(2.000000, 2.000000), skew=(1.000000, 1.000000), bands=3, outdb=false] │
└────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┘
```

### DIFF
--- a/rust/sedona-functions/src/sd_format.rs
+++ b/rust/sedona-functions/src/sd_format.rs
@@ -617,16 +617,13 @@ mod tests {
         let formatted = result.as_string::<i32>();
 
         // Index 0: valid raster (no skew)
-        assert_eq!(
-            formatted.value(0),
-            "RASTER [1x2/1] @ [1 1.6 1.1 2] / OGC:CRS84"
-        );
+        assert_eq!(formatted.value(0), "[1x2/1] @ [1 1.6 1.1 2] / OGC:CRS84");
         // Index 1: null raster should produce null output
         assert!(formatted.is_null(1));
         // Index 2: valid raster (with skew)
         assert_eq!(
             formatted.value(2),
-            "RASTER [3x4/1] @ [3 2.4 3.84 4.24] skew=(0.06, 0.08) / OGC:CRS84"
+            "[3x4/1] @ [3 2.4 3.84 4.24] skew=(0.06, 0.08) / OGC:CRS84"
         );
     }
 
@@ -641,11 +638,11 @@ mod tests {
         let formatted = result.as_string::<i32>();
 
         // Index 0: valid raster (no skew)
-        assert!(formatted.value(0).starts_with("RASTER [1x2/"));
+        assert!(formatted.value(0).starts_with("[1x2/"));
         // Index 1: null raster should produce null output
         assert!(formatted.is_null(1));
         // Index 2: valid raster (with skew)
-        assert!(formatted.value(2).starts_with("RASTER [3x4/"));
+        assert!(formatted.value(2).starts_with("[3x4/"));
     }
 
     #[test]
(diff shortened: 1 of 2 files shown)


rust/sedona-functions/src/sd_format.rs
```diff
@@ -356,6 +359,73 @@ fn list_view_value_to_formatted_value<OffsetSize: OffsetSizeTrait>(
     ))
 }
 
+fn raster_value_to_formatted_value(
+    columnar_value: &ColumnarValue,
+    maybe_width_hint: Option<usize>,
+) -> Result<ColumnarValue> {
+    match columnar_value {
```
paleolimbot, line 365: The logic for displaying individual raster items should probably live in sedona-raster

rust/sedona-functions/src/sd_format.rs
```diff
@@ -356,6 +359,73 @@ fn list_view_value_to_formatted_value<OffsetSize: OffsetSizeTrait>(
...
+                    metadata.height(),
+                    metadata.upper_left_x(),
+                    metadata.upper_left_y(),
+                    metadata.scale_x(),
+                    metadata.scale_y(),
+                    metadata.skew_x(),
+                    metadata.skew_y(),
+                    bands.len(),
+                    has_outdb_band
+                );
```
paleolimbot: At least for rasters that are not rotated or skewed, what I would really like to know are the bounds and CRS. Can this output be more like:

```
[WxH/nbands] @ [xmin ymin xmax ymax] / crs <outdb>
```

That may be a little more compact. Even for skewed data the 2D bounds are probably more useful output for human reading.

(Drawing from the last time I had to decide this: https://paleolimbot.github.io/wk/reference/grd.html )

paleolimbot (approved):
Thank you!

rust/sedona-functions/src/sd_format.rs
Kontinuation: OK. I believe that this format requires the CRS to be abbreviated when it is possible. I'll submit the patch for refactoring the CRS abbreviation cache for ST_SetCRS before implementing this raster display format.

rust/sedona-raster/src/display.rs
```diff
@@ -0,0 +1,163 @@
...
+
+        let has_outdb = bands
+            .iter()
+            .any(|band| matches!(band.metadata().storage_type(), Ok(StorageType::OutDbRef)));
+
+        // Write: RASTER [WxH/nbands] @ [xmin ymin xmax ymax]
+        write!(
+            f,
+            "RASTER [{width}x{height}/{nbands}] @ [{xmin} {ymin} {xmax} {ymax}]"
+        )?;
```
paleolimbot: In SedonaDB we have the good fortune of getting to print the type name at the top of the column, so I think we can get seven characters of width back for each item:

```suggestion
        // Write: [WxH/nbands] @ [xmin ymin xmax ymax]
        write!(
            f,
            "[{width}x{height}/{nbands}] @ [{xmin} {ymin} {xmax} {ymax}]"
        )?;
```

PR #599, opened 2026-02-11
feat(rust/sedona): Add SedonaFairSpillPool memory pool and CLI memory limit support

Kontinuation:
## Summary

- Add `SedonaFairSpillPool`, a new memory pool that reserves a configurable fraction of total memory for unspillable consumers, preventing spillable operators from exhausting all available memory (addresses [datafusion#17334](https://github.com/apache/datafusion/issues/17334) in the Sedona context)
- Add `--memory-limit`, `--mem-pool-type`, and `--unspillable-reserve-ratio` CLI arguments to `sedona-cli` for configuring memory pool behavior
- Refactor `SedonaContext::new_local_interactive()` to expose `new_local_interactive_with_runtime_env()`, allowing callers to inject a custom `RuntimeEnv` with pre-configured memory pools

## Motivation

When running out-of-core spatial joins, spillable operators (e.g., `SpatialJoinExec`) could consume all available memory, causing unspillable operators (e.g., `RepartitionExec`'s merge consumer) to fail with OOM errors. The `SedonaFairSpillPool` mitigates this by reserving a configurable portion (default 20%) of the memory pool exclusively for unspillable allocations.


sedona-cli/src/main.rs
```diff
@@ -187,3 +237,80 @@ fn parse_command(command: &str) -> Result<String, String> {
...
+        m.insert("mb", ByteUnit::MiB);
+        m.insert("g", ByteUnit::GiB);
+        m.insert("gb", ByteUnit::GiB);
+        m.insert("t", ByteUnit::TiB);
+        m.insert("tb", ByteUnit::TiB);
+        m
+    });
+
+    static SUFFIX_REGEX: LazyLock<regex::Regex> =
+        LazyLock::new(|| regex::Regex::new(r"^([0-9]+)([a-z]+)?$").unwrap());
```
paleolimbot:
```suggestion
        LazyLock::new(|| regex::Regex::new(r"^([0-9.]+)\s*([a-z]+)?$").unwrap());
```

(e.g., `9.5 gb`)

sedona-cli/src/main.rs
```diff
@@ -187,3 +237,80 @@ fn parse_command(command: &str) -> Result<String, String> {
...
+            ByteUnit::KiB => 1 << 10,
+            ByteUnit::MiB => 1 << 20,
+            ByteUnit::GiB => 1 << 30,
+            ByteUnit::TiB => 1 << 40,
+        }
+    }
+}
+
+fn parse_size_string(size: &str, label: &str) -> Result<usize, String> {
+    static BYTE_SUFFIXES: LazyLock<HashMap<&'static str, ByteUnit>> = LazyLock::new(|| {
```
paleolimbot, line 263: Can you copy DataFusion's test for this?

https://github.com/apache/datafusion/blob/ecf3b502cfd8c5baeaabb97737605ef66549c753/datafusion-cli/src/main.rs#L464-L510

This could also go in `rust/sedona` (we'll need it in R and Python, too?)

rust/sedona/src/pool_type.rs
```diff
@@ -0,0 +1,48 @@
...
+// specific language governing permissions and limitations
+// under the License.
+
+use std::{
+    fmt::{self, Display, Formatter},
+    str::FromStr,
+};
+
+#[derive(PartialEq, Debug, Clone)]
+pub enum PoolType {
```
paleolimbot, line 24: It seems like we will also need this in R and Python...should it go in `rust/sedona`?

rust/sedona/src/memory_pool.rs
```diff
@@ -0,0 +1,279 @@
...
+};
+use datafusion_common::{resources_datafusion_err, DataFusionError, Result};
+use parking_lot::Mutex;
+
+pub const DEFAULT_UNSPILLABLE_RESERVE_RATIO: f64 = 0.2;
+
+/// A [`MemoryPool`] implementation similar to DataFusion's [`datafusion::execution::memory_pool::FairSpillPool`],
+/// but with the following changes:
+///
+/// It implements a reservation mechanism for unspillable memory consumers. This addresses an issue
```
2010YOUY01:
```suggestion
/// Spillable and non-spillable operators use logically separate portions of the memory pool, 
/// controlled by unspillable_reserve_ratio, instead of sharing a single pool as in
/// DataFusion’s default FairSpillPool, which can  lead to the following issues:
```
I’m trying to explain the “what” before the “why.”

2010YOUY01 (approved):
This looks like a good idea. If it works well, we could use it in DataFusion upstream. Looking forward to the results using it in the spilling queries!

One potential follow-up: we could set configurations through SQL `SET ...` statements like https://datafusion.apache.org/user-guide/configs.html#runtime-configuration-settings

sedona-cli/src/main.rs
Kontinuation: Added test, also augmented with decimal numbers and too large numbers.

rust/sedona/src/pool_type.rs
Kontinuation: Moved this to `rust/sedona`.

PR #600, opened 2026-02-11
refactor(rust/sedona-raster-functions): Split monolithic benchmark into per-function benchmark files

Kontinuation:
## Summary

- Split `benches/native-raster-functions.rs` into separate per-function benchmark files: `rs_envelope`, `rs_geotransform`, `rs_rastercoordinate`, `rs_size`, `rs_srid`, and `rs_worldcoordinate`
- Updated `Cargo.toml` `[[bench]]` entries to match the new file names
- Added multi-size benchmarks (64x64, 256x256, 1024x1024) for `rs_envelope`, `rs_size` (width/height)

## Rationale

We will add more raster functions later, having a monolithic benchmark file does not scale, so we'd better break it into several smaller files.


Kontinuation:
Most of the benchmarks in this module are just a few lines, splitting them to multiple files may not be a good idea.

PR #604, opened 2026-02-12
feat(rust/sedona-pointcloud): add laz chunk statistics

b4l:


b4l:
I added some tests to ensure no faulty pruning.

rust/sedona-pointcloud/src/laz/format.rs
```diff
@@ -200,9 +202,28 @@ impl FileFormat for LazFormat {
 
     async fn create_physical_plan(
         &self,
-        _state: &dyn Session,
+        state: &dyn Session,
         conf: FileScanConfig,
     ) -> Result<Arc<dyn ExecutionPlan>, DataFusionError> {
+        let mut source = conf
```
b4l, line 208: @paleolimbot, these changes are necessary to actually cache the metadata. Maybe this should be considered for the Parquet reader as well.

paleolimbot: Thank you! We do in general have cache issues 😬  (Peter tried something like this a while ago but we didn't notice a difference at the time: https://github.com/apache/sedona-db/pull/294 ).

paleolimbot (approved):
Thank you!

Issue #554, opened 2026-01-26
rust: Handle transforms where the number or type of dimensions changes


paleolimbot:
There are a few functions (`ST_Force2D`, `ST_Force3D`, `ST_Force3D`, `ST_Force3DZ`, `ST_Force3DM`, `ST_Force4D`) that not only modify the coordinates present in the geometry but (potentially) change the number or types of dimensions.

As noted in https://github.com/apache/sedona-db/pull/544, ST_Transform can potentially do this, too (e.g., transforming from Lon/Lat to a geocentric CRS like `'EPSG:4978'`, where a 2D coordinate becomes 3D).

This could be handled with the `CrsTransform` (by including the input and output `Dimensions` and ensuring there is a hook available for all combinations of dimensions that might occur. For ST_Transform with PROJ specifically, we can query the number of axes of the input/output to at least detect changes (we probably have to assume that 3D is always XYZ because I'm not sure we have the information to guess otherwise).

paleolimbot:
Closed in #606 and #620.

PR #606, opened 2026-02-12
feat(rust/sedona-functions): Implement `ST_Force2D()` and `ST_Force3D()`

yutannihilation:
Part of #554

This pull request enhances `CrsTransform` to change the number or types of dimensions by introducing these two changes

- `transform_coord_3d()` takes the dimension of the input coordinate because otherwise there's no way to distinguish if the z value of `coord` is from the original input or is filled.
  - I think `transform_coord()` doesn't need this, but I'm not fully sure
- new `output_dim()` returns the dimension of the output

```rust
pub trait CrsTransform: std::fmt::Debug {
    // same
    fn transform_coord(...) -> Result<(), SedonaGeometryError>;

    fn transform_coord_3d(
        &self,
        coord: &mut (f64, f64, f64),
        input_dims: Dimensions, // new
    ) -> Result<(), SedonaGeometryError>;

    // new
    fn output_dim(&self) -> Option<Dimensions>;
}
```

This pull request implements only `ST_Force2D()` and `ST_Force3D()`. In order to implement the variants related to M dimension, `ST_Force3DM` and `ST_Force4D`, probably `CrsTransform` needs to provide more methods that handles M dimension. But, probably these will be used only by `ST_Force3DM` and `ST_Force4D`, and I'm not sure if it's worth making the trait complex.

### DIFF
--- a/rust/sedona-functions/src/st_force_dim.rs
+++ b/rust/sedona-functions/src/st_force_dim.rs
@@ -330,7 +330,8 @@ mod tests {
                 Some("POINT Z EMPTY"),
                 Some("POINT (1 2)"),
                 Some("POINT Z (3 4 5)"),
-                Some("POINT ZM (8 9 10 11)"),
+                Some("POINT M (6 7 8)"),
+                Some("POINT ZM (9 10 11 12)"),
             ],
             &sedona_type,
         );
@@ -342,7 +343,8 @@ mod tests {
                 Some("POINT Z EMPTY"),
                 Some("POINT Z (1 2 0)"),
                 Some("POINT Z (3 4 5)"),
-                Some("POINT Z (8 9 10)"),
+                Some("POINT Z (6 7 0)"),
+                Some("POINT Z (9 10 11)"),
             ],
             &WKB_GEOMETRY,
         );
@@ -364,15 +366,24 @@ mod tests {
                 None,
                 Some("POINT EMPTY"),
                 Some("POINT (1 2)"),
+                Some("POINT (1 2)"),
                 Some("POINT Z (3 4 5)"),
-                Some("POINT (6 7)"),
-                Some("POINT ZM (8 9 10 11)"),
+                Some("POINT M (6 7 8)"),
+                Some("POINT ZM (9 10 11 12)"),
             ],
             &sedona_type,
         );
         let z = create_array!(
             Float64,
-            [Some(9.0), Some(9.0), Some(9.0), Some(9.0), None, Some(9.0)]
+            [
+                Some(9.0),
+                Some(9.0),
+                Some(9.0),
+                None,
+                Some(9.0),
+                Some(9.0),
+                Some(9.0)
+            ]
         );
 
         let expected = create_array(
@@ -380,9 +391,10 @@ mod tests {
                 None,
                 Some("POINT Z EMPTY"),
                 Some("POINT Z (1 2 9)"),
-                Some("POINT Z (3 4 5)"),
                 None,
-                Some("POINT Z (8 9 10)"),
+                Some("POINT Z (3 4 5)"),
+                Some("POINT Z (6 7 9)"),
+                Some("POINT Z (9 10 11)"),
             ],
             &WKB_GEOMETRY,
         );


rust/sedona-functions/src/st_force_dim.rs
```diff
@@ -0,0 +1,394 @@
...
+                    };
+                    transform(wkb, &trans, &mut builder)
+                        .map_err(|e| DataFusionError::External(Box::new(e)))?;
+                    builder.append_value([]);
+                }
+                _ => {
+                    builder.append_null();
+                }
+            }
+            i += 1;
```
paleolimbot: I'm not sure this is strictly better but it is how we do this in other places:

```suggestion
        let mut z_iter = z_array.iter();
        executor.execute_wkb_void(|maybe_wkb| {
            match (maybe_wkb, z_array.next().unwrap()) {
                (Some(wkb), z) => {
                    let trans = Force3DTransform { z };
                    transform(wkb, &trans, &mut builder)
                        .map_err(|e| DataFusionError::External(Box::new(e)))?;
                    builder.append_value([]);
                }
                _ => {
                    builder.append_null();
                }
            }
```

rust/sedona-functions/src/st_force_dim.rs
```diff
@@ -0,0 +1,394 @@
...
+impl CrsTransform for Force3DTransform {
+    fn output_dim(&self) -> Option<geo_traits::Dimensions> {
+        Some(geo_traits::Dimensions::Xyz)
+    }
+
+    fn transform_coord(
+        &self,
+        _coord: &mut (f64, f64),
+    ) -> std::result::Result<(), SedonaGeometryError> {
+        unreachable!()
```
paleolimbot:
```suggestion
        sedona_internal_err!("Unexpected call to transform_coord()")
```

rust/sedona-functions/src/st_force_dim.rs
```diff
@@ -0,0 +1,394 @@
...
+        tester.assert_return_type(WKB_GEOMETRY);
+
+        let points = create_array(
+            &[
+                None,
+                Some("POINT EMPTY"),
+                Some("POINT Z EMPTY"),
+                Some("POINT (1 2)"),
+                Some("POINT Z (3 4 5)"),
+                Some("POINT ZM (8 9 10 11)"),
```
paleolimbot, line 300: Should a POINT M work here?

rust/sedona-functions/src/st_force_dim.rs
```diff
@@ -0,0 +1,394 @@
...
+        let tester = ScalarUdfTester::new(st_force3d_udf().into(), vec![sedona_type.clone()]);
+        tester.assert_return_type(WKB_GEOMETRY);
+
+        let points = create_array(
+            &[
+                None,
+                Some("POINT EMPTY"),
+                Some("POINT Z EMPTY"),
+                Some("POINT (1 2)"),
+                Some("POINT Z (3 4 5)"),
```
paleolimbot, line 332: Should POINT M work here?

rust/sedona-functions/src/st_force_dim.rs
```diff
@@ -0,0 +1,394 @@
...
+        );
+        tester.assert_return_type(WKB_GEOMETRY);
+
+        let points = create_array(
+            &[
+                None,
+                Some("POINT EMPTY"),
+                Some("POINT (1 2)"),
+                Some("POINT Z (3 4 5)"),
+                Some("POINT (6 7)"),
```
paleolimbot: Should POINT M work here?

rust/sedona-functions/src/st_force_dim.rs
yutannihilation: Ah, thanks for catching, I tend to forget an arrow array has `.iter()`...

yutannihilation:
I see, then I'll address the M dimension variants in another pull request.

> feel free to file a follow-up ticket to track support of M values in transforms.

This pull request doesn't close https://github.com/apache/sedona-db/issues/554, so I don't think another ticket is needed :)

PR #607, opened 2026-02-12
feat(rust/sedona-geoparquet): Ensure GeoParquet configuration options for read and write can be passed via SQL

paleolimbot:
This PR makes some tweaks to how configuration options are passed around so that they work in SQL. For example, reading with a geometry columns override. The motivation behind this change was actually to expose some important generic Parquet write options that affect queryable output (notably: row group size and compression). A side effect of exposing those options was ensuring all our wrappers around Parquet things are synchronized and now things like this work in SQL (including the CLI):

```sql
CREATE EXTERNAL TABLE test
STORED AS PARQUET
LOCATION '...'
OPTIONS ('geometry_columns' '{"geometry": {"encoding": "WKB", "crs": "EPSG:3857"}}');
```

Writing GeoParquet 1.1:

```sql
COPY (
    SELECT * FROM epic_table
) TO 'foofy.parquet'
OPTIONS (GEOPARQUET_VERSION '1.1')
```

### DIFF
--- a/python/sedonadb/python/sedonadb/dataframe.py
+++ b/python/sedonadb/python/sedonadb/dataframe.py
@@ -343,7 +343,7 @@ class DataFrame:
         sort_by: Optional[Union[str, Iterable[str]]] = None,
         single_file_output: Optional[bool] = None,
         geoparquet_version: Literal["1.0", "1.1", None] = None,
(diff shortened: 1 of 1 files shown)


zhangfengcdt (approved):
LGTM

PR #610, opened 2026-02-13
feat(rust/sedona-spatial-join): Auto-repartition probe side for balancing spatial join workload

Kontinuation:
## Rationale

We found that spatial join is very sensitive to the skewness of probe side data. There will be stragglers when the probe side is not evenly distributed, which increases the latency of queries. Repartitioning the probe side automatically showed performance improvements in SpatialBench queries involving spatial joins, and trading the overhead of RepartitionExec for workload balance is a reasonable choice, especially for queries that does not utilize the partitioning scheme of the probe side.

## Summary

- Inserts a `RepartitionExec` with `RoundRobinBatch(N)` partitioning as the parent of the probe side input to `SpatialJoinExec`, breaking spatial locality in sorted/skewed datasets to balance workloads during out-of-core spatial join.
- Adds a new `repartition_probe_side` option to `SpatialJoinOptions` (default: `true`) to control this behavior.
- Correctly handles non-KNN predicates with input swapping (`should_swap`) and KNN predicates where `probe_side` can be `Left` or `Right`.

## Changes

- **`rust/sedona-common/src/option.rs`**: Added `repartition_probe_side: bool` (default `true`) to `SpatialJoinOptions`.
- **`rust/sedona-spatial-join/src/planner/physical_planner.rs`**: Added repartition logic in `plan_extension()` and a new `repartition_probe_side()` helper function. The repartition is applied to the pre-swap input that will become the probe side after any potential swap.

## Testing

All existing tests pass (332 unit tests, 169 integration tests, 3 doc-tests for `sedona-spatial-join`; 8 tests for `sedona-common`). Clippy clean.

### DIFF
--- a/rust/sedona-spatial-join/src/planner/physical_planner.rs
+++ b/rust/sedona-spatial-join/src/planner/physical_planner.rs
@@ -306,7 +306,13 @@ fn repartition_probe_side(
     let probe_plan = match spatial_predicate {
         SpatialPredicate::KNearestNeighbors(knn) => match knn.probe_side {
             JoinSide::Left => &mut physical_left,
-            _ => &mut physical_right,
+            JoinSide::Right => &mut physical_right,
+            JoinSide::None => {
+                // KNNPredicate::probe_side is asserted not to be None in its constructor;
+                // treat this as a debug-only invariant violation and default to right.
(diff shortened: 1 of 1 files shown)


paleolimbot (approved):
Thank you!

PR #612, opened 2026-02-13
feat(rust/sedona-raster-functions): add RS_BandPixelType, RS_BandNoDataValue UDFs

Kontinuation:
## Summary

- Add `RS_BandPixelType` and `RS_BandNoDataValue` raster UDFs to `sedona-raster-functions`
- Each function supports per-band queries with an optional band index parameter (defaults to band 1)
- Includes unit tests and benchmark entries in `native-raster-functions.rs`

### DIFF
--- a/rust/sedona-raster-functions/src/rs_band_accessors.rs
+++ b/rust/sedona-raster-functions/src/rs_band_accessors.rs
@@ -243,7 +243,7 @@ fn get_nodata_value(
 #[cfg(test)]
 mod tests {
     use super::*;
-    use arrow_array::{Array, Float64Array, Int32Array, StringArray, StructArray};
+    use arrow_array::{Array, Float64Array, Int32Array, Int64Array, StringArray, StructArray};
     use datafusion_expr::ScalarUDF;
     use sedona_raster::builder::RasterBuilder;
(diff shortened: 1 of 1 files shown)


rust/sedona-raster-functions/src/rs_band_accessors.rs
```diff
@@ -0,0 +1,413 @@
...
+    }
+
+    fn invoke_batch(
+        &self,
+        arg_types: &[SedonaType],
+        args: &[ColumnarValue],
+    ) -> Result<ColumnarValue> {
+        let executor = RasterExecutor::new(arg_types, args);
+        let band_index_array = args[1].clone().into_array(executor.num_iterations())?;
+        let band_index_array = band_index_array.as_primitive::<Int32Type>();
```
paleolimbot: Probably better to use DataFusion's downcasting here (`as_int32_array()`, checks the downcast and returns an internal error in debug mode).

Kontinuation: I have found that we need to cast the band index here before calling `as_int32_array`. I have switched to calling `as_int32_array` and added the cast.

paleolimbot (approved):

PR #616, opened 2026-02-13
chore(docs/reference/sql): Migrate function documentation defined in Rust into Markdown

paleolimbot:
This PR removes documentation defined in rust/sedona-functions and places it in docs/reference/sql (as the one source of truth for function documentation). 

This was done by a combination of Opus and Codex, although I had to review all the descriptions to ensure we didn't have trivially repeated sections and the examples ran.

I am not sure I'll get there before 0.3 is released, but I'll open a follow-up to add `sedona-function-definitions`, which can do cross checking and make the documentation available from SQL.

After this PR we can canonically add function documentation to `docs/reference/sql` when adding functions rather than the Rust source code.

### DIFF
--- a/rust/sedona-functions/src/st_force_dim.rs
+++ b/rust/sedona-functions/src/st_force_dim.rs
@@ -443,19 +443,15 @@ mod tests {
     fn udf_metadata() {
         let st_force2d: ScalarUDF = st_force2d_udf().into();
         assert_eq!(st_force2d.name(), "st_force2d");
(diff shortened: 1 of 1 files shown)


yutannihilation:
Sorry for making a conflict on `st_force_dim.rs`. I created a pull request to resolve (but feel free to ignore if it's easier to do by yourself).

https://github.com/paleolimbot/sedona-db/pull/1

PR #619, opened 2026-02-14
feat(rust/sedona-geometry): Let CrsTransform handle M coordinate

yutannihilation:
This is a followup of #606. Before addressing `ST_Force3DM()` and `ST_Force4D()`, I want to confirm if my approach is correct.

This pull request changes `CrsTransform` to have two additional methods to handle M coordinate (accordingly, `transform_coord_3d()` is renamed to `transform_coord_xyz`). Maybe I could create a separate `transform_m()` method, but I thought it's better to do the transformation in a single call.

- `transform_xym()`
- `transform_xyzm()`

This also requires `transform_and_write_coords()` to extract and fill M coordinate.

### DIFF
--- a/c/sedona-proj/src/transform.rs
+++ b/c/sedona-proj/src/transform.rs
@@ -230,7 +230,7 @@ impl CrsTransform for ProjTransform {
         Ok(())
     }
 
-    fn transform_coord_3d(
+    fn transform_coord_xyz(
         &self,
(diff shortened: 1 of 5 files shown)


yutannihilation:
Thanks!

> I also wonder if we've made the transform slower

This was my main concern. But, I agree with you that we can optimize later.

PR #620, opened 2026-02-15
feat(rust/sedona-functions): Implement ST_Force3DM and ST_Force4D

yutannihilation:
Part of https://github.com/apache/sedona-db/issues/554

This pull request implements `ST_Force3DM` and `ST_Force4D`.

I tried to unify all `ST_Force*`, but it seems it would make handling of optional args tricky (you can see the attempt at d9f46eda391fbfaee65d7d07f76a988e2c3b2999). So, this simply duplicate and tweak the code based on the existing `ST_Force2D` and `ST_Force3D`.

### DIFF
--- a/python/sedonadb/tests/functions/test_functions.py
+++ b/python/sedonadb/tests/functions/test_functions.py
@@ -1527,14 +1527,14 @@ def test_st_force_dim(eng, geom, expected_2d, expected_3d):
         (
             "POINT EMPTY",
             5,
-            "POINT ZM (nan nan nan nan)",
-            "POINT ZM (nan nan nan nan)",
+            "POINT M (nan nan nan)",
+            "POINT M (nan nan nan)",
         ),
-        ("POINT (0 1)", 5, "POINT ZM (0 1 0 0)", "POINT ZM (0 1 0 5)"),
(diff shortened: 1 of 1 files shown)


yutannihilation:
The suspicious diff of `rust/sedona-functions/src/st_force_dim.rs` is because the current file uses CRLF for line break. This was my fault when I created this file on Windows. Sorry...

rust/sedona-geometry/src/transform.rs
```diff
@@ -458,7 +458,8 @@ where
     match input_dims {
         // If the input doesn't have M coordinate, fill with 0.
         Dimensions::Xy | Dimensions::Xyz | Dimensions::Unknown(_) => (coord.x(), coord.y(), 0.0),
-        Dimensions::Xym | Dimensions::Xyzm => (coord.x(), coord.y(), coord.nth_or_panic(2)),
+        Dimensions::Xym => (coord.x(), coord.y(), coord.nth_or_panic(2)),
+        Dimensions::Xyzm => (coord.x(), coord.y(), coord.nth_or_panic(3)),
```
yutannihilation, line 462: I noticed this logic added in #619 was wrong. Fixed now.

PR #630, opened 2026-02-17
feat(rust/sedona-raster-functions): add RS_SetSRID/RS_SetCRS with batch-local cache refactoring

Kontinuation:
## Summary

- Add `RS_SetSRID` and `RS_SetCRS` raster functions that set the CRS on raster values, with null-input semantics: null SRID/CRS input nulls the entire raster, while SRID=0/CRS="0" clears the CRS but preserves the raster.
- Introduce `CachedSRIDToCrs` and `CachedCrsNormalization` batch-local cache structs in `sedona-schema/src/crs.rs`, shared by both ST_ (geometry) and RS_ (raster) setter functions.
- Refactor `ST_SetSRID`/`ST_SetCRS` to use the new shared cache types instead of inline caching.

### RS_SetSRID / RS_SetCRS
- Kernel structs carry `engine: Option<Arc<dyn CrsEngine + Send + Sync>>` matching the ST_ pattern
- Both `_with_engine_udf(engine)` and `_udf()` (delegates with `None`) constructors
- `extract_input_nulls()` captures the null buffer from the original input argument before any conversion, then `replace_raster_crs()` merges input nulls into the struct's null buffer via `NullBuffer::union()`
- 17 tests including 4 null-input tests covering scalar and array cases

### Batch-local caches
- `CachedSRIDToCrs`: caches SRID integer → CRS string conversion (e.g. 4326 → "OGC:CRS84")
- `CachedCrsNormalization`: caches CRS string → abbreviated `authority:code` form via deserialization
- Both use `#[derive(Default)]` and are used identically by ST_ and RS_ functions

### DIFF
--- a/rust/sedona-schema/src/crs.rs
+++ b/rust/sedona-schema/src/crs.rs
@@ -151,10 +151,10 @@ impl CachedCrsToSRIDMapping {
     }
 }
 
-/// Batch-local cache for converting integer SRIDs to CRS strings.
+/// Cache for converting integer SRIDs to CRS strings.
 ///
 /// Maps SRID integers to their CRS string representation with caching to avoid
-/// repeated validation/deserialization of the same SRID within a batch:
+/// repeated validation/deserialization of the same SRID:
 /// - `0` → `None` (no CRS)
 /// - `4326` → `Some("OGC:CRS84")`
 /// - other → `Some("EPSG:{srid}")`, validated once via the caller-provided closure
@@ -198,10 +198,10 @@ impl CachedSRIDToCrs {
     }
 }
 
-/// Batch-local cache for normalizing CRS strings to their abbreviated form.
+/// Cache for normalizing CRS strings to their abbreviated form.
 ///
 /// Maps CRS input strings to their abbreviated `authority:code` representation
-/// with caching to avoid repeated deserialization of the same CRS string within a batch:
+/// with caching to avoid repeated deserialization of the same CRS string:
 /// - `"0"` or `""` → `None` (no CRS)
 /// - other → deserialized and abbreviated to `authority:code` if possible
 #[derive(Default)]


rust/sedona-schema/src/crs.rs
```diff
@@ -157,6 +151,108 @@ impl CachedCrsToSRIDMapping {
     }
 }
 
+/// Batch-local cache for converting integer SRIDs to CRS strings.
+///
+/// Maps SRID integers to their CRS string representation with caching to avoid
+/// repeated validation/deserialization of the same SRID within a batch:
```
paleolimbot:
```suggestion
/// Cache for converting integer SRIDs to CRS strings.
///
/// Maps SRID integers to their CRS string representation with caching to avoid
/// repeated validation/deserialization of the same SRID:
```

(Currently it's batch-local but it doesn't have to be)

rust/sedona-schema/src/crs.rs
```diff
@@ -157,6 +151,108 @@ impl CachedCrsToSRIDMapping {
...
+        Ok(result)
+    }
+}
+
+/// Batch-local cache for normalizing CRS strings to their abbreviated form.
+///
+/// Maps CRS input strings to their abbreviated `authority:code` representation
+/// with caching to avoid repeated deserialization of the same CRS string within a batch:
+/// - `"0"` or `""` → `None` (no CRS)
+/// - other → deserialized and abbreviated to `authority:code` if possible
```
paleolimbot:
```suggestion
/// Cache for normalizing CRS strings to their abbreviated form.
///
/// Maps CRS input strings to their abbreviated `authority:code` representation
/// with caching to avoid repeated deserialization of the same CRS string:
/// - `"0"` or `""` → `None` (no CRS)
/// - other → deserialized and abbreviated to `authority:code` if possible
```

rust/sedona-raster-functions/src/rs_setsrid.rs
```diff
@@ -0,0 +1,722 @@
...
+fn broadcast_string_view(array: &StringViewArray, len: usize) -> ArrayRef {
+    if array.len() == len {
+        return Arc::new(array.clone());
+    }
+    debug_assert_eq!(array.len(), 1);
+    if array.is_null(0) {
+        Arc::new(StringViewArray::new_null(len))
+    } else {
+        let value = array.value(0);
+        Arc::new(std::iter::repeat_n(Some(value), len).collect::<StringViewArray>())
```
paleolimbot, line 267: I think there is an optimization for string views with respect to appending repeated values:

https://github.com/apache/datafusion/blob/6798dff0d7e1c52d6bea74b141d4df6ad9cc1516/datafusion/common/src/scalar/mod.rs#L3120-L3127

Kontinuation: The method optimized for inserting repeated values (`try_append_value_n`) was introduced in arrow-rs 57.2.0, we need to upgrade to a later version of DataFusion and arrow-rs to use them. See https://github.com/apache/datafusion/pull/19797

I would like to stick to the current method since it is the way DataFusion 51 uses to append repeated values: https://github.com/apache/datafusion/blob/51.0.0/datafusion/common/src/scalar/mod.rs#L2960C13-L2964

PR #633, opened 2026-02-17
chore(rust): Remove documentation field from SedonaScalarUDF and SedonaAggregateUDF

yutannihilation:
A small followup of #616 

Now that we have `.qmd` documentations, I guess we never use `documentation` field in Rust. This pull request removes it so that we don't need to type `None` everytime.


paleolimbot (approved):
Thank you!

PR #638, opened 2026-02-18
refactor(rust/sedona-functions): Remove stub functions

paleolimbot:
Therse were previously used as homes for documentation; however, we no longer need them for that! This makes sedona-functions a bit simpler (only contains functions implemented there). We get to delete ~1000 lines of code and 30 files!

This also needed a slight tweak to the function metadata renderer (see inline comment) and making some changes for aggregate functions that we'd already made for scalar functions (we can just create them as they are added to the function set instead of adding the stubs first).


docs/reference/sql/_render_meta.py
```diff
@@ -187,6 +187,17 @@ def to_str(v):
...
+            quote_type = v["c"][0]["t"]
+            if quote_type == "SingleQuote":
+                quote_char = "'"
+            elif quote_type == "DoubleQuote":
+                quote_char = '"'
+            else:
+                raise ValueError(f"Unhandled quote type in ast {v}")
+
+            content = "".join(to_str(item) for item in v["c"][1])
+            return f"{quote_char}{content}{quote_char}"
```
paleolimbot, line 200: This is an unfortunate consequence of how the build system is implemented: we get a Pandoc AST here because that's how Quarto filters give us the data. Functionally it means that we have to handle Pandoc syntax elements for anything that shows up in the description or argument names/descriptions.

zhangfengcdt (approved):

PR #641, opened 2026-02-19
feat(rust/sedona-spatial-join): Automatic query-side filter pushdown for KNN joins

Kontinuation:
## Summary

- Adds a `KnnQuerySideFilterPushdown` optimizer rule that automatically pushes query-side-only filters below the `SpatialJoinPlanNode` extension node for KNN inner joins
- Only handles `INNER JOIN` (conservative start); outer join support can be added later
- Updates docs to document the automatic pushdown behavior and clarify when `barrier()` is still needed

## Background

Previously, KNN joins blocked ALL filter pushdown (both query-side and object-side) because the `SpatialJoinPlanNode` extension node's default `prevent_predicate_push_down_columns()` returns all columns. Object-side pushdown must remain blocked (it changes KNN candidate sets), but query-side pushdown is safe and should be automatic.

DataFusion's built-in `PushDownFilter` pushes the same predicate to ALL children of an extension node, so a query-side filter like `h.stars >= 4` would fail when applied to the object-side child that doesn't have column `h.stars`. This requires a custom optimizer rule instead.

## Implementation

The `KnnQuerySideFilterPushdown` rule:
1. Pattern matches `Filter(predicate, Extension(SpatialJoinPlanNode))` where the join filter contains `ST_KNN`
2. Uses `find_knn_query_side()` to determine which child is the query side (from the first argument of `ST_KNN`)
3. Splits the filter predicate into conjuncts; pushes query-side-only conjuncts below the extension node; keeps the rest above
4. Runs **before** DataFusion's `PushDownFilter` so the pushed-down filters are further optimized into scan nodes in the same pass

## Testing

- 12 unit tests for `find_st_knn_call` and `find_knn_query_side`
- 3 integration tests verifying correct plan structure (filter pushed into query-side child, object-side filters stay above)

Depends on #635

### DIFF
--- a/rust/sedona-spatial-join/src/planner/spatial_expr_utils.rs
+++ b/rust/sedona-spatial-join/src/planner/spatial_expr_utils.rs
@@ -2369,10 +2369,7 @@ mod tests {
     #[test]
     fn test_find_st_knn_call_simple() {
         let st_knn_udf = create_dummy_st_knn_udf();
-        let expr = Expr::ScalarFunction(ScalarFunction {
-            func: st_knn_udf,
-            args: vec![col("l.geom"), col("r.geom"), lit(5i32), lit(false)],
-        });
+        let expr = st_knn_udf.call(vec![col("l.geom"), col("r.geom"), lit(5i32), lit(false)]);
 
         let result = find_st_knn_call(&expr);
         assert!(result.is_some());
@@ -2382,10 +2379,7 @@ mod tests {
     #[test]
     fn test_find_st_knn_call_nested_in_and() {
         let st_knn_udf = create_dummy_st_knn_udf();
-        let knn_expr = Expr::ScalarFunction(ScalarFunction {
-            func: st_knn_udf,
-            args: vec![col("l.geom"), col("r.geom"), lit(5i32), lit(false)],
-        });
+        let knn_expr = st_knn_udf.call(vec![col("l.geom"), col("r.geom"), lit(5i32), lit(false)]);
 
         // ST_KNN(...) AND l.id > 5
         let and_expr = Expr::BinaryExpr(datafusion_expr::expr::BinaryExpr {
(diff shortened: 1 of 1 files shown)


rust/sedona-spatial-join/src/planner/spatial_expr_utils.rs
```diff
@@ -2309,4 +2365,196 @@ mod tests {
...
         assert!(names.is_empty());
     }
+
+    #[test]
+    fn test_find_st_knn_call_simple() {
+        let st_knn_udf = create_dummy_st_knn_udf();
+        let expr = Expr::ScalarFunction(ScalarFunction {
+            func: st_knn_udf,
+            args: vec![col("l.geom"), col("r.geom"), lit(5i32), lit(false)],
+        });
```
paleolimbot: Optional, but if you have datafusion as a dev dependency already you can also use `ctx.parse_sql_expr()` to create these (might make it easier to spot errors or create more complex test cases in the future).

rust/sedona-spatial-join/src/planner/spatial_expr_utils.rs
```diff
@@ -2309,4 +2365,196 @@ mod tests {
...
+        assert_eq!(result.unwrap().func.name(), "st_knn");
+    }
+
+    #[test]
+    fn test_find_st_knn_call_nested_in_and() {
+        let st_knn_udf = create_dummy_st_knn_udf();
+        let knn_expr = Expr::ScalarFunction(ScalarFunction {
+            func: st_knn_udf,
+            args: vec![col("l.geom"), col("r.geom"), lit(5i32), lit(false)],
+        });
```
paleolimbot: `.call()` is probably a bit more compact since there are lot of these:

```suggestion
        let knn_expr = st_knn_udf.call(vec![col("l.geom"), col("r.geom"), lit(5i32), lit(false)]);
```

rust/sedona-spatial-join/src/planner/spatial_expr_utils.rs
Kontinuation: It would require registering UDFs with a SessionContext, which adds complexity for what are currently straightforward unit tests. I have replaced these using `st_knn_udf.call`.

rust/sedona-spatial-join/src/planner/spatial_expr_utils.rs
Kontinuation: Fixed.

PR #642, opened 2026-02-19
feat(python/sedonadb): Add sort by geometry support in to_parquet()

### LINKED ISSUE #370: Improve order-by-geometry interface
In https://github.com/apache/sedona-db/pull/360 we added a basic but awkward means by which to "sort by geometry". There are some improvements we need to make once we have some experience with it:

> Maybe it's just me, but this feels a little awkward. I would definitely expect the default ordering of sd_order to have meaning. https://github.com/apache/sedona-db/pull/360#discussion_r2566209846

This is a great point...we can improve this situation by (1) erroring (i.e., refuse to sort if we're not sure it will be meaningful) and (2) removing the need for s2geography for computing the ordering (it's in most of our builds but not R yet). For most input/current released builds, the ordering will be meaningful.

> Why do we call it SD_Order? If this is a UDF, why not ST?
> Can we add docs together with the PR if possible?

A missing piece from the initial PR to the final interface here is the logical optimizer rule that transforms `ORDER BY geometry` to `ORDER BY sd_order(geometry)`. We have `sd_format()` as well and will likely add `sd_cast()` soon as they are called in a similar (internal) way.

When we add the optimizer rule, we'll need to add a section on sorting (probably separate from the function docs)

paleolimbot:
This PR adds support for geometry columns in `to_parquet(..., sort_by="...")`. This is currently a special case for the Python to_parquet function...we could insert an optimizer rule that does this so that we can support ORDER BY geometry in SQL, too; however, at the point we need SQL the workaround is not too bad. This at least allows writing mostly optimized GeoParquet 1.1 without SQL.

Closes #370.

```python
import geopandas
import sedona.db

sd = sedona.db.connect()

sd.funcs.table.sd_random_geometry(
    "Point", 10000, seed=948, bounds=[-50, -50, 50, 50]
).to_view("pts", overwrite=True)
df = sd.sql("SELECT id, ST_SetSRID(geometry, 4326) AS geometry FROM pts")
df.to_parquet(
    "sorted.parquet",
    sort_by="geometry",
    geoparquet_version="1.1",
    max_row_group_size=100000,
)
```


python/sedonadb/src/dataframe.rs
```diff
@@ -189,13 +189,52 @@ impl InternalDataFrame {
...
+                if geometry_column_names.contains(name.as_str()) {
+                    // Create the call sd_order(column). If we're ordering by geometry but don't have
+                    // the required feature for high quality sort output, give an error. This is mostly
+                    // an issue when using maturin develop because geography is not a default feature.
+                    if has_geography {
+                        let order_udf_opt: Option<ScalarUDF> = ctx
+                            .inner
+                            .functions
+                            .scalar_udf("sd_order")
+                            .map(|udf_opt| udf_opt.clone().into());
```
zhangfengcdt: Can you use a reference instead of clone instead, by just call the order udf directly?

zhangfengcdt (commented):
Just a minor comment ...

python/sedonadb/src/dataframe.rs
paleolimbot: Good point...I updated this to clone an existing Arc rather than clone the SedonaScalarUDF.

PR #644, opened 2026-02-19
chore: Keep builds passing after yanked 11.0.0 geos release

paleolimbot:
The 11.0.0 release was yanked to fix a releas issue ( https://github.com/georust/geos/issues/216 ), so we need to go back to a git dependency.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2885,9 +2885,8 @@ dependencies = [
 
 [[package]]
 name = "geos"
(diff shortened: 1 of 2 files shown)


oreilles:
Sorry for noticing too late, 11.0.0 has been yanked because the geos submodule dependency had not been published.

oreilles:
11.0.1 has just been released. 👍

PR #646, opened 2026-02-19
feat(rust/sedona-geoparquet): Ensure metadata cache is used in GeoParquet wrapper code

### LINKED ISSUE #250: Use metadata cache when fetching metadata in geoparquet reader

paleolimbot:
In #251 we tried to use the file metadata cache and found that it actually slowed down queries. @yutannihilation  kindly benchmarked the effect of the cache against DuckDB to demonstrate that the file cache there is effective for queries against large tables. @b4l kindly showed how to do this in #604.

This PR pipes through the requisite options to ensure the cache is used for GeoParquet reads. This is especially important because we need to pull two extra copies of the metadata after DataFusion has already pulled it: if we don't use the cached version, we issue three requests where we could have issued one. For most queries this is done in parallel/async in a non-blocking way and is hard to notice; however, remote tables with large numbers of Parquet files do very badly.

A secondary issue is that the default size of the cache is not well-equiped to deal with Overture buildings, which we were using to benchmark this. The buildings data requires almost 900 megabytes of cache space and because it is a least-recently used cache being queried roughly in order three times, if the cache size is even a little bit smaller than the full size of the dataset then it is 0% useful. The increase we see in time is probably because of contention on the mutex guarding the in-memory cache.

```python
import re
import os
os.environ["AWS_SKIP_SIGNATURE"] = "true"
os.environ["AWS_DEFAULT_REGION"] = "us-west-2"
import sedona.db

sd = sedona.db.connect()

sd.sql("SET datafusion.runtime.metadata_cache_limit = '900M'").execute()

# 16s on main, 10s on this PR with a big enough cache
sd.read_parquet(
    "s3://overturemaps-us-west-2/release/2026-02-18.0/theme=buildings/type=building/"
).to_view("buildings", overwrite=True)

# Second time: 16s on main, 0s with this PR
sd.read_parquet(
    "s3://overturemaps-us-west-2/release/2026-02-18.0/theme=buildings/type=building/"
).to_view("buildings", overwrite=True)
```

I took the opportunity to redo the Overture buildings documentation page to include this and a few other improvements we added in the last few months.

Closes #250.


yutannihilation:
Thanks for this!

Sorry for a noob question. Even if I include `sd.sql("SET datafusion.runtime.metadata_cache_limit = '900M'").execute()`, the benchmark still doesn't improve much. Is this expected at the moment (until the new version of DataFusion...?)? Maybe the difference is that DuckDB somehow holds a persistent cache among sessions while SedonaDB doesn't?

Benchmark results (seconds):
| Engine | Median | Mean | Min | Max | Runs | row_count | max_confidence |
|---|---:|---:|---:|---:|---:|---:|---:|
| DuckDB | 0.591740 | 2.654925 | 0.575550 | 6.797485 | 3 | 7471 | 0.999544084072 |
| SedonaDB | 9.257312 | 8.987796 | 8.034394 | 9.671681 | 3 | 7471 | 0.999544084072 |

python/sedonadb/tests/test_udf.py
```diff
@@ -237,42 +237,3 @@ def __datafusion_scalar_udf__(self):
...
-    udf_impl = udf.arrow_udf(pa.binary(), [udf.STRING, udf.NUMERIC])(some_udf)
-
-    # Register with our session
-    con.register_udf(udf_impl)
-
-    # Create a datafusion session, fetch our udf and register with the other session
-    datafusion_ctx = datafusion.SessionContext()
-    datafusion_ctx.register_udf(
-        datafusion.ScalarUDF.from_pycapsule(con._impl.scalar_udf("some_udf"))
-    )
```
paleolimbot, line 253: I added https://github.com/apache/sedona-db/issues/655 to track this...you have to try pretty hard to trigger this failing functionality so I just removed the tests for now.

PR #648, opened 2026-02-20
feat(rust/sedona-pointcloud): add optional round robin partitioning and parallel statistics extraction

b4l:
This contains two optional features that greatly improve the performance of the LAS/LAZ listing table provider.

* Round-robin partitioning: The default way to partition a dataset to enable parallel reading by DataFusion is through splitting files by byte ranges into the number of target partitions. For selective queries on (partially) ordered datasets that support pruning, this can result in unequal resource use, as all the work is done on one partition while the rest is pruned. Additionally, this breaks the existing locality in the input when it is converted, as data from all partitions ends up in each output row group. This approach addresses these issues by partitioning the dataset using a round-robin scheme across sequential chunks. This improves selective query performance by more than half.
* Parallel statistics extraction: While the method to infer the schema, adopted from the Parquet reader, uses concurrency (metadata fetch concurrency), it is not parallel. Extracting statistics in parallel can substantially improve the extraction process by a factor of the number of cores available.

### DIFF
--- a/rust/sedona-pointcloud/src/las/options.rs
+++ b/rust/sedona-pointcloud/src/las/options.rs
@@ -127,8 +127,8 @@ extensions_options! {
     /// LAS/LAZ configuration options
     ///
     /// * `geometry encoding`: plain (x, y, z), wkb or native (geoarrow)
-    /// * `collect statistics`: extract las/laz chunk statistics (reqires a full scan on registration)
-    /// * `parallel statistics extration`: extract statistics in parallel
+    /// * `collect statistics`: extract las/laz chunk statistics (requires a full scan on registration)
+    /// * `parallel statistics extraction`: extract statistics in parallel
     /// * `persist statistics`: store statistics in a sidecar file for future reuse (requires write access)
     /// * `round robin partitioning`: read chunks in parallel with round robin instead of byte range (default)
     /// * `extra bytes`: las extra byte attributes handling, ignore, keep as binary blob, or typed


rust/sedona-pointcloud/src/las/source.rs
```diff
@@ -149,6 +152,40 @@ impl FileSource for LasSource {
         self.extension.as_str()
     }
 
+    fn repartitioned(
+        &self,
```
paleolimbot, line 155: Can you add some of the text you have in this PR to this section so that future readers have some background on why this is necessary?

rust/sedona-pointcloud/src/options.rs
```diff
@@ -80,7 +80,9 @@ extensions_options! {
     pub struct PointcloudOptions {
         pub geometry_encoding: GeometryEncoding, default = GeometryEncoding::default()
         pub collect_statistics: bool, default = false
+        pub parallel_statistics_extraction: bool, default = false
         pub persist_statistics: bool, default = false
+        pub round_robin_partitioning: bool, default = false
```
paleolimbot: All of these would benefit from a brief summary docstring of when these values should be modified (e.g., use round robin partitioning when running queries with selective workloads that benefit from parallelization).

paleolimbot (approved):
Sorry it took me a while to circle back to this...I had a release TODO list and lost track of a few things. Feel free to ping me if this happens again!

This looks great! I added some optional suggestions of where you put the nice text that you have in the PR description into the code for future readers.

rust/sedona-pointcloud/src/las/source.rs
b4l: done in https://github.com/apache/sedona-db/pull/648/commits/c6f34a3a8fd786e8038805fcb4b3b6a8532fe96a

rust/sedona-pointcloud/src/options.rs
b4l: done in https://github.com/apache/sedona-db/pull/648/commits/c6f34a3a8fd786e8038805fcb4b3b6a8532fe96a

b4l:
@paleolimbot, no worries, I guess you have a full plate already. Added some documentation and reworked the options to be self-contained in the las module for now, which seems more concise.

paleolimbot (approved):
Thank you!

PR #632, opened 2026-02-17
feat(python/sedonadb): Write GDAL/OGR via pyogrio

paleolimbot:
This PR adds a thin wrapper around `pyogrio.raw.write_arrow()` that fills in some default arguments with information we already have helpers to get (e.g., geometry column name and its crs).

Closes #358.

```python
import sedona.db

sd = sedona.db.connect()

url = "https://github.com/geoarrow/geoarrow-data/releases/download/v0.2.0/ns-water_water-point.fgb"
sd.read_pyogrio(url).to_pyogrio("foofy.fgb")
```

Because the arrow interface also powers some calls to `ogr2ogr`, I did a quick check to make sure it was not too far off (and it is about the same):

```python
# curl -L https://github.com/geoarrow/geoarrow-data/releases/download/v0.2.0/ns-water_elevation.fgb -o input.fgb
%time sd.read_pyogrio("input.fgb").to_pyogrio("foofy.fgb")
#> CPU times: user 5.51 s, sys: 1.3 s, total: 6.81 s
#> Wall time: 7 s

# Similar wall time to ogr2ogr, which does approximately the same thing
# ogr2ogr foofy.fgb input.fgb  6.51s user 1.58s system 93% cpu 8.697 total
```

One hiccup that I will solve in a follow-up PR is that the version of GDAL in current pip pyogrio doesn't support string or binary views, so some pretty common operations like `read_parquet().to_pyogrio()` will error ( https://github.com/OSGeo/gdal/issues/13942 ). There are some other libraries that have this restriction too and so we can provide a function to insert the appropriate casts to eliminate exotic types (e.g., string view, run end encoded, dictionary).

PR #650, opened 2026-02-20
feat(rust/sedona-functions): Add sd_simplifystorage utility

### LINKED ISSUE #647: Add utility to cast column storage to non-view/dictionary/run-end encoded types
In #632 we add write support for GDAL; however, it won't work for many inputs because GDAL doesn't support newer Arrow types. This is also true for some other libraries and we can add a utility to cast away those types to simpler ones when we need to.

paleolimbot:
This PR adds a scalar function and data frame utility to remove view and dictionary types from the output for compatibility with other tools. This is required to support writing to GDAL/OGR in #632 for data frames derived from Parquet (which have string views turned on by default) but also may be useful otherwise. This is implemented as a scalar function with a minimal wrapper to help with Python:

```python
import sedona.db

sd = sedona.db.connect()

df = sd.read_parquet("https://github.com/geoarrow/geoarrow-data/releases/download/v0.2.0/ns-water_water-point.parquet")
df.schema
#> SedonaSchema with 8 fields:
#>   OBJECTID: int64<Int64>
#>   FEAT_CODE: utf8<Utf8View>
#>   ZVALUE: float64<Float64>
#>   PT_CLASS: int32<Int32>
#>   NAMEID_1: utf8<Utf8View>
#>   NAME_1: utf8<Utf8View>
#>   HID: utf8<Utf8View>
#>   geometry: geometry<WkbView({...})>

df._simplify_storage_types().schema
#> SedonaSchema with 8 fields:
#>   OBJECTID: int64<Int64>
#>   FEAT_CODE: utf8<Utf8>
#>   ZVALUE: float64<Float64>
#>   PT_CLASS: int32<Int32>
#>   NAMEID_1: utf8<Utf8>
#>   NAME_1: utf8<Utf8>
#>   HID: utf8<Utf8>
#>   geometry: geometry<Wkb({...})>
```

Closes #647.

### DIFF
--- a/rust/sedona/src/projected_reader.rs
+++ b/rust/sedona/src/projected_reader.rs
@@ -116,7 +116,7 @@ mod test {
     use arrow_array::{create_array, ArrayRef};
     use arrow_schema::{DataType, Field};
 
-    use crate::{context::SedonaContext};
+    use crate::context::SedonaContext;
 
     use super::*;
 


python/sedonadb/python/sedonadb/dataframe.py
```diff
@@ -527,6 +527,11 @@ def __repr__(self) -> str:
         else:
             return super().__repr__()
 
+    def _simplify_storage_types(self):
```
zhangfengcdt, line 625: should this be public api instead (i.e., removing _)?

zhangfengcdt:
[minor] Would it be better called sd_simplify_storage to be more consistent with the naming in st functions, such as st_force_dim, st_line_merge?

Do we consider lazy evaluation, or does this materialize the entire dataset immediately as what datafusion cast operations?

python/sedonadb/python/sedonadb/dataframe.py
paleolimbot: I merged in the GDAL write PR and added a test...hopefully the intent of this is more clear now (we only need this when we specifically call the pyogrio write function so I kept it internal for now).

PR #681, opened 2026-03-03
feat(c/sedona-gdal): add crate with dynamically loaded GDAL bindings

Kontinuation:
## Summary

- Add the `sedona-gdal` crate providing runtime-loaded GDAL FFI bindings via `libloading`, following the same pattern as `sedona-proj`.
- Contains the `SedonaGdalApi` function-pointer struct, dynamic symbol loading (`dyn_load`), the `GdalApi` handle with `call_gdal_api!` macro, error types, and global API registration.
- Integrates into the workspace (`Cargo.toml`, `rust/sedona/Cargo.toml`) and CI (`.github/workflows/rust.yml` adds `libgdal-dev`).

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -128,7 +128,7 @@ jobs:
       - name: Install dependencies
         shell: bash
         run: |
-          sudo apt-get update && sudo apt-get install -y libgeos-dev
+          sudo apt-get update && sudo apt-get install -y libgeos-dev libgdal-dev
 
       - name: Check
         if: matrix.name == 'check'
--- a/c/sedona-gdal/Cargo.toml
+++ b/c/sedona-gdal/Cargo.toml
@@ -1,0 +1,45 @@
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+[package]
+name = "sedona-gdal"
+version.workspace = true
+license.workspace = true
+keywords.workspace = true
+categories.workspace = true
+authors.workspace = true
+homepage.workspace = true
+repository.workspace = true
+description.workspace = true
+readme.workspace = true
+edition.workspace = true
+rust-version.workspace = true
+
+[lib]
+crate-type = ["staticlib", "cdylib", "lib"]
+
+[dependencies]
+gdal-sys = { version = "0.12.0", optional = true }
+libloading = { workspace = true }
+thiserror = { workspace = true }
+
+[features]
+default = ["gdal-sys"]
+gdal-sys = ["dep:gdal-sys"]
+
+[dev-dependencies]
+sedona-testing = { workspace = true }
--- a/c/sedona-gdal/Cargo.toml
+++ b/c/sedona-gdal/Cargo.toml
@@ -29,9 +29,6 @@ readme.workspace = true
 edition.workspace = true
 rust-version.workspace = true
 
-[lib]
-crate-type = ["staticlib", "cdylib", "lib"]
-
 [dependencies]
 gdal-sys = { version = "0.12.0", optional = true }
 libloading = { workspace = true }
--- a/c/sedona-gdal/src/dyn_load.rs
+++ b/c/sedona-gdal/src/dyn_load.rs
@@ -182,16 +182,10 @@ fn load_all_symbols(lib: &Library, api: &mut SedonaGdalApi) -> Result<(), GdalIn
         api,
         MEMDatasetCreate,
         [
-            // Linux (char**)
+            // Linux and macOS
             b"_ZN10MEMDataset6CreateEPKciii12GDALDataTypePPc\0",
-            // macOS (char**)
-            b"__ZN10MEMDataset6CreateEPKciii12GDALDataTypePPc\0",
-            // MSVC (char**)
+            // MSVC
             b"?Create@MEMDataset@@SAPEAV1@PEBDHHHW4GDALDataType@@PEAPEAD@Z\0",
-            // Linux (const char**)
-            b"_ZN10MEMDataset6CreateEPKciii12GDALDataTypePPKc\0",
-            // macOS (const char**)
-            b"__ZN10MEMDataset6CreateEPKciii12GDALDataTypePPKc\0",
         ]
     );
 
@@ -250,10 +244,12 @@ fn current_process_library() -> Result<Library, GdalInitLibraryError> {
 }
 
 #[cfg(not(any(unix, windows)))]
-compile_error!(
-    "sedona-gdal: current_process_library() is not implemented for this platform. \
-     Only Unix and Windows are supported."
-);
+fn current_process_library() -> Result<Library, GdalInitLibraryError> {
+    Err(GdalInitLibraryError::Invalid(
+        "current_process_library() is not implemented for this platform. \
+    Only Unix and Windows are supported.",
+    ))
+}
 
 #[cfg(test)]
 mod test {
--- a/c/sedona-gdal/src/dyn_load.rs
+++ b/c/sedona-gdal/src/dyn_load.rs
@@ -163,6 +163,9 @@ fn load_all_symbols(lib: &Library, api: &mut SedonaGdalApi) -> Result<(), GdalIn
     load_fn!(lib, api, GDALFPolygonize);
     load_fn!(lib, api, GDALPolygonize);
 
+    // --- Version ---
+    load_fn!(lib, api, GDALVersionInfo);
+
     // --- Config ---
     load_fn!(lib, api, CPLSetThreadLocalConfigOption);
 
--- a/c/sedona-gdal/src/dyn_load.rs
+++ b/c/sedona-gdal/src/dyn_load.rs
@@ -1,0 +1,252 @@
+// Licensed to the Apache Software Foundation (ASF) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership.  The ASF licenses this file
+// to you under the Apache License, Version 2.0 (the
+// "License"); you may not use this file except in compliance
+// with the License.  You may obtain a copy of the License at
+//
+//   http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+use std::path::Path;
+
+use libloading::Library;
+
+use crate::errors::GdalInitLibraryError;
+use crate::gdal_dyn_bindgen::SedonaGdalApi;
+
+/// Load a single symbol from the library and write it into the given field.
+///
+/// We load as a raw `*const ()` pointer and transmute to the target function pointer
+/// type. This is the standard pattern for dynamic symbol loading where the loaded
+/// symbol's signature is known but cannot be expressed as a generic parameter to
+/// `Library::get` (because each field has a different signature).
+///
+/// On failure returns a `GdalInitLibraryError` with the symbol name and the
+/// underlying OS error message.
+macro_rules! load_fn {
+    ($lib:expr, $api:expr, $name:ident) => {
+        // The target types here are too verbose to annotate for each call site
+        #[allow(clippy::missing_transmute_annotations)]
+        {
+            $api.$name = Some(unsafe {
+                let sym = $lib
+                    .get::<*const ()>(concat!(stringify!($name), "\0").as_bytes())
+                    .map_err(|e| {
+                        GdalInitLibraryError::LibraryError(format!(
+                            "Failed to load symbol {}: {}",
+                            stringify!($name),
+                            e
+                        ))
+                    })?;
+                std::mem::transmute(sym.into_raw().into_raw())
+            });
+        }
+    };
+}
+
+/// Try to load a symbol under one of several names (e.g. for C++ mangled symbols).
+/// Writes the first successful match into `$api.$field`. Returns an error only if
+/// *none* of the names resolve.
+macro_rules! load_fn_any {
+    ($lib:expr, $api:expr, $field:ident, [$($name:expr),+ $(,)?]) => {{
+        let mut found = false;
+        $(
+            if !found {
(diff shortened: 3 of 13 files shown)


c/sedona-gdal/src/gdal_api.rs
```diff
@@ -0,0 +1,101 @@
...
+            } else {
+                CStr::from_ptr(msg_ptr).to_string_lossy().into_owned()
+            }
+        };
+        unsafe { call_gdal_api!(self, CPLErrorReset) };
+        GdalError::CplError {
+            class: default_err_class,
+            number: err_no,
+            msg: err_msg,
+        }
```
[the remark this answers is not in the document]
Kontinuation: Clarified that it will always return an error, even if the error number is 0.

c/sedona-gdal/src/register.rs
```diff
@@ -0,0 +1,152 @@
...
+            ));
+        };
+        gdal_all_register();
+    }
+
+    let _ = GDAL_API.set(api);
+    Ok(GDAL_API.get().expect("GDAL API should be set"))
+}
+
+pub fn configure_global_gdal_api(shared_library: PathBuf) -> Result<(), GdalInitLibraryError> {
```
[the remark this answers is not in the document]
Kontinuation: I think we should load GDAL lazily just like `configure_global_proj_engine`. It will allow user to call `configure_gdal` after importing sedonadb. I'll work on refactoring it later.

c/sedona-gdal/src/dyn_load.rs
```diff
@@ -0,0 +1,252 @@
...
+        if !found {
+            return Err(GdalInitLibraryError::LibraryError(format!(
+                "Failed to resolve {} under any known mangled name",
+                stringify!($field),
+            )));
+        }
+    }};
+}
+
+/// Populate all 64 function-pointer fields of [`SedonaGdalApi`] from the given
```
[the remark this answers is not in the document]
Kontinuation: Removed the exact number from comment

c/sedona-gdal/src/lib.rs
```diff
@@ -0,0 +1,34 @@
...
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+// --- FFI layer ---
+pub(crate) mod dyn_load;
+pub mod gdal_dyn_bindgen;
```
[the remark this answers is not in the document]
Kontinuation: We still need to expose the constants defined in this file. However, we can mark `SedonaGdalApi` as pub(crate) to forbid calling into GDAL unsafe APIs from other crates.

c/sedona-gdal/src/gdal_api.rs
```diff
@@ -0,0 +1,101 @@
...
+
+/// Invoke a function pointer from the `SedonaGdalApi` struct.
+///
+/// # Panics
+///
+/// Panics if the function pointer is `None`. This is unreachable in correct usage
+/// because all function pointers are guaranteed to be `Some` after successful
+/// `sedona_gdal_dyn_api_init` (the C loader fails if any symbol is missing),
+/// and you cannot obtain a `&GdalApi` without successful initialization
+/// (it's behind `OnceLock`).
```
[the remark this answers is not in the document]
Kontinuation: Fixed comment

c/sedona-gdal/Cargo.toml
```diff
@@ -0,0 +1,45 @@
...
+homepage.workspace = true
+repository.workspace = true
+description.workspace = true
+readme.workspace = true
+edition.workspace = true
+rust-version.workspace = true
+
+[lib]
+crate-type = ["staticlib", "cdylib", "lib"]
+
```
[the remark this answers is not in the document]
Kontinuation: Yes, it is unnecessary, removed.

c/sedona-gdal/src/gdal_api.rs
```diff
@@ -0,0 +1,101 @@
...
+            }
+        };
+        unsafe { call_gdal_api!(self, CPLErrorReset) };
+        GdalError::CplError {
+            class: default_err_class,
+            number: err_no,
+            msg: err_msg,
+        }
+    }
+}
```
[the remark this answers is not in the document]
Kontinuation: Added tests

c/sedona-gdal/src/dyn_load.rs
```diff
@@ -0,0 +1,252 @@
...
+    let lib = current_process_library()?;
+    let mut api = SedonaGdalApi::default();
+    load_all_symbols(&lib, &mut api)?;
+    Ok((lib, api))
+}
+
+/// Open a handle to the current process image.
+#[cfg(unix)]
+fn current_process_library() -> Result<Library, GdalInitLibraryError> {
+    Ok(libloading::os::unix::Library::this().into())
```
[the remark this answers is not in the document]
Kontinuation: Add `compile_error!` for unsupported platforms.

c/sedona-gdal/src/dyn_load.rs
```diff
@@ -0,0 +1,318 @@
...
+            b"__ZN10MEMDataset6CreateEPKciii12GDALDataTypePPc\0",
+            // Linux (const char**)
+            b"_ZN10MEMDataset6CreateEPKciii12GDALDataTypePPKc\0",
+            // macOS (const char**)
+            b"__ZN10MEMDataset6CreateEPKciii12GDALDataTypePPKc\0",
+            // MSVC (char**)
+            b"?Create@MEMDataset@@SAPEAV1@PEBDHHH4GDALDataType@@PEAPEAD@Z\0",
+            // MSVC (const char**)
+            b"?Create@MEMDataset@@SAPEAV1@PEBDHHH4GDALDataType@@PEAPEBAD@Z\0",
+        ]
```
paleolimbot, line 192: We chatted a bit offline about how this is kind of a hack but one that we're going to move forward with because dynamically loading GDAL is so useful (and `MEMDatasetCreate()` is essential for performance because of a global GDAL locking issue).

Is the difference between Linux and MacOS actually GCC vs clang or is this actually MacOS vs Linux?

c/sedona-gdal/src/dyn_load.rs
```diff
@@ -0,0 +1,318 @@
...
+            ))
+        })?
+        .into())
+}
+
+#[cfg(not(any(unix, windows)))]
+compile_error!(
+    "sedona-gdal: current_process_library() is not implemented for this platform. \
+     Only Unix and Windows are supported."
+);
```
paleolimbot: I don't think we need to fail compilation, just fail at runtime if somebody tries to use a function that requires GDAL and didn't configure it any other way.

c/sedona-gdal/src/errors.rs
```diff
@@ -0,0 +1,77 @@
...
+
+    #[error("StrUtf8Error")]
+    StrUtf8Error(#[from] std::str::Utf8Error),
+
+    #[error(transparent)]
+    IntConversionError(#[from] TryFromIntError),
+
+    #[error("Buffer length {0} does not match raster size {1:?}")]
+    BufferSizeMismatch(usize, (usize, usize)),
+}
```
paleolimbot: I see a number of places where these aren't used consistently (e.g., checking the library version does not use any of null pointer, strutf8, intostring or FfiNul). We should probably use them consistently everywhere or more coarsely split them up (I would personally just use fewer error variants until some use case arises where you need to treat them differently).

c/sedona-gdal/src/gdal_api.rs
```diff
@@ -0,0 +1,102 @@
...
+            func($($arg),*)
+        } else {
+            panic!("{} function not available", stringify!($func))
+        }
+    };
+}
+
+// Re-export for use in other modules in this crate
+#[allow(unused_imports)]
+pub(crate) use call_gdal_api;
```
paleolimbot: Can this be removed?

c/sedona-gdal/src/errors.rs
```diff
@@ -0,0 +1,77 @@
...
+
+pub type Result<T> = std::result::Result<T, GdalError>;
```
paleolimbot:
```suggestion
```

We don't do this elsewhere

.github/workflows/rust.yml
```diff
@@ -128,7 +128,7 @@ jobs:
       - name: Install dependencies
         shell: bash
         run: |
-          sudo apt-get update && sudo apt-get install -y libgeos-dev
+          sudo apt-get update && sudo apt-get install -y libgeos-dev libgdal-dev
```
paleolimbot, line 131: We should also update the contributor guide because I think the general dev setup will now fail without a system GDAL install.

c/sedona-gdal/src/lib.rs
```diff
@@ -0,0 +1,35 @@
...
+pub mod gdal_api;
+pub mod register;
+
+// --- Re-exports for convenient use ---
+pub use errors::GdalError;
+pub use gdal_api::GdalApi;
+pub use register::{
+    configure_global_gdal_api, get_global_gdal_api, is_gdal_api_configured, with_global_gdal_api,
+    GdalApiBuilder,
+};
```
paleolimbot: Until we start formalizing public APIs  for internal crates I think we can skip public reexports

c/sedona-gdal/src/register.rs
```diff
@@ -0,0 +1,314 @@
...
+
+static GDAL_API: OnceLock<GdalApi> = OnceLock::new();
+
+/// Get a reference to the global GDAL API, initializing it if not already done.
+///
+/// On first call, reads the builder set by [`configure_global_gdal_api`] (or uses
+/// [`GdalApiBuilder::default()`] if none was configured) and calls its `build()`
+/// method to create the [`GdalApi`]. The result is stored in a process-global
+/// `OnceLock` and reused for all subsequent calls.
+pub fn get_global_gdal_api() -> Result<&'static GdalApi, GdalInitLibraryError> {
```
paleolimbot: Probably the `with_xxx()`  pattern is the one you want to expose?

```suggestion
fn get_global_gdal_api() -> Result<&'static GdalApi, GdalInitLibraryError> {
```

c/sedona-gdal/src/global.rs
```diff
@@ -0,0 +1,314 @@
...
+    global_builder.replace(builder);
+    Ok(())
+}
+
+pub fn is_gdal_api_configured() -> bool {
+    GDAL_API.get().is_some()
+}
+
+pub fn with_global_gdal_api<F, R>(func: F) -> Result<R, GdalInitLibraryError>
+where
```
paleolimbot, line 199: Can we add documentation to these since they are the access points?

c/sedona-gdal/src/register.rs
```diff
@@ -0,0 +1,314 @@
...
+/// We use `GDALVersionInfo("VERSION_NUM")` instead of `GDALCheckVersion` because
+/// the latter performs an **exact** major.minor match and rejects newer versions
+/// (e.g. GDAL 3.12 fails a check for 3.4), whereas we need a **minimum** version
+/// check (>=).
+#[cfg(feature = "gdal-sys")]
+fn check_gdal_version() -> Result<(), GdalInitLibraryError> {
+    use std::ffi::CStr;
+
+    // Matches the GDAL_COMPUTE_VERSION(maj,min,rev) macro: maj*1000000 + min*10000 + rev*100
+    let min_version_num = MIN_GDAL_VERSION_MAJOR * 1_000_000 + MIN_GDAL_VERSION_MINOR * 10_000;
```
paleolimbot: It might be worth keeping the minimum version requirements closer to this line so that it's more clear what that is.

Also, should this check also be performed for external GDAL and not just when it's loaded from the process?

paleolimbot (commented):
Great!

c/sedona-gdal/src/dyn_load.rs
Kontinuation: It is not related to the compiler, clang and gcc produces ABI compatible binaries. Mach-O adds the `_` prefix to the symbol names, but that does not seem to affect `dlsym`.

I'll remove the macOS variants with underscore prefixes. The symbol name should work for both Linux and macOS.

c/sedona-gdal/src/dyn_load.rs
Kontinuation: Replaced compile_error with a current_process_library function directly returning `Err(GdalInitLibraryError::Invalid)`.

c/sedona-gdal/src/errors.rs
Kontinuation: Alright, these are needed by upcoming PRs. I'll remove unused enums from current PR.

c/sedona-gdal/src/gdal_api.rs
Kontinuation: Removed. We'll mark it `#[macro_export]` in upcoming PRs.

c/sedona-gdal/src/errors.rs
Kontinuation: Removed

c/sedona-gdal/src/lib.rs
Kontinuation: Removed these reexports

c/sedona-gdal/src/register.rs
Kontinuation: I thought that we can expose both. `with_global_gdal_api` has the style as the sedona-proj API so I'll prefer `with_global_gdal_api` and mark `get_global_gdal_api` as non-public.

c/sedona-gdal/src/global.rs
Kontinuation: Documentation added

c/sedona-gdal/src/register.rs
Kontinuation: I have refactored this to work with both compile-time-linked gdal-sys and dynamically loaded GDAL library.

c/sedona-gdal/src/gdal_api.rs
Kontinuation: Maybe we don't need to expose the low-level APIs to other crates, since we'll built higher level safe APIs within this crate, so we can just make it private and prevent other crates from directly calling into raw GDAL C APIs.

Kontinuation: I still decided to export this to allow other modules to call the raw GDAL C APIs directly. This is the sole purpose of exposing `with_global_gdal_api`; otherwise, both could remain private.

.github/workflows/rust.yml
Kontinuation: Updated contributor guide.

c/sedona-gdal/src/gdal_api.rs
paleolimbot: Probably it would make more sense to expose a higher-level global (e.g., like `with_global_proj_engine()`) and make both private but if you do that now everything will give you an unused warning 🙂 . Either is OK for now.

c/sedona-gdal/src/register.rs
paleolimbot: Probably a higher level global makes more sense later (and perhaps even later, a session-scoped higher level abstraction like the CrsEngine to prevent a mess of feature flags when doing raster joins).

c/sedona-gdal/src/gdal_api.rs
Kontinuation: I'm going to add a high-level entrypoint `struct Gdal` and expose `with_global_gdal` in https://github.com/apache/sedona-db/pull/682/changes#diff-ebf7d28b4d19ad8b7ee49c23ed29a66ac42c4a91c6f66b74966de036f1e5a066R207-R219. We can expose both high level and low level APIs similar to gdal and gdal-sys.

paleolimbot (approved):
Thank you!

PR #682, opened 2026-03-03
feat(sedona-gdal): add high-level safe GDAL API wrappers

Kontinuation:
## Summary

- Add safe Rust wrappers on top of the dynamically loaded GDAL bindings from #681.
- Includes: `Dataset`, `Driver`/`DriverManager`, `RasterBand`, `SpatialRef`, `Geometry`, `Feature`, `Layer`, VRT dataset builder, VSI virtual filesystem, `rasterize`, `rasterize_affine`, `polygonize`, `GeoTransform`, `CslStringList`, `MemDatasetBuilder`, and supporting types.
- APIs use the [georust/gdal](https://github.com/georust/gdal) v0.19.0 crate as a reference (see file-level doc comments for attribution). We only implemented the code we actually need for now. We'll extend these high-level APIs as we implement more RS functions.

Depends on #681.

### DIFF
--- a/c/sedona-gdal/src/gdal.rs
+++ b/c/sedona-gdal/src/gdal.rs
@@ -27,12 +27,12 @@ use crate::dataset::Dataset;
 use crate::driver::{Driver, DriverManager};
 use crate::errors::Result;
 use crate::gdal_api::GdalApi;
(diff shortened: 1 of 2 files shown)


Kontinuation:
This PR was split into smaller pieces, and most of them were already merged into main.

PR #683, opened 2026-03-03
refactor(c/sedona-s2geography): Move s2geography UDFs to extension ABI

paleolimbot:
This is a PR that migrates UDFs to use the sedona-extension ABI instead of one specifically invented for Geography (history being that the geography one came first and that I added the extension one after some experience writing the first one). The benefit here is that we get to remove almost all of the wrapper code on the Rust side because we already test the sedona-extension wrapper in isolation. This forces a tiny bit of wrapper code on to C++ but so far it is not too bad.

This PR updates the s2geography submodule to point to the main branch after https://github.com/paleolimbot/s2geography/pull/75 and a few others (I also updated the CMake which was simplified considerably...among other things, this means we don't have to consider building/linking our own geoarrow-c or nanoarrow).


c/sedona-s2geography/src/register.rs
```diff
@@ -14,35 +14,20 @@
...
-            scalar_kernel::st_line_locate_point_impl(),
-        ),
-        ("st_length", scalar_kernel::st_length_impl()),
-        ("st_symdifference", scalar_kernel::st_sym_difference_impl()),
-        ("st_maxdistance", scalar_kernel::st_max_distance_impl()),
-        ("st_perimeter", scalar_kernel::st_perimeter_impl()),
-        ("st_shortestline", scalar_kernel::st_shortest_line_impl()),
-        ("st_union", scalar_kernel::st_union_impl()),
-    ]
+pub fn scalar_kernels() -> Result<Vec<(&'static str, ScalarKernelRef)>> {
```
zhangfengcdt, line 29: I like this change and now match based on names. Since this is a public function and entry to the kernel functions, could we add some document here? such as "Initialize s2geography scalar kernels via extension ABI. lazy loading, return named pairs ..."

zhangfengcdt (approved):
Nice work and much cleaner approach I believe. Just one minor comment, otherwise, lgtm.

PR #689, opened 2026-03-05
feat(python/sedonadb): add Python GDAL configuration API

Kontinuation:
## Summary

- Adds a `configure_gdal()` Python API to the `sedonadb` package that automatically locates and loads GDAL's shared library at import time
- Supports multiple presets: `auto`, `pyogrio`, `rasterio`, `conda`, `homebrew`, `system` — covering common GDAL installation methods
- Exposes `gdal_version()` to query the loaded GDAL library version from Python
- Fixes PROJ library discovery to be platform-aware (Linux `.so` / macOS `.dylib` / Windows `.dll` + conda `Library/bin/` layout)
- Adds `test_gdal.py` pytest that verifies GDAL loads successfully and reports a version >= 3.x

## Dependencies

> **Stacked PR**: This PR depends on #681 (`feat/sedona-gdal-ffi`) and should be merged after it.


python/sedonadb/python/sedonadb/context.py
```diff
@@ -604,3 +604,184 @@ def _configure_proj_prefix(prefix: str):
...
+        elif preset == "conda":
+            _configure_gdal_conda()
+            return
+        elif preset == "homebrew":
+            prefix = os.environ.get("HOMEBREW_PREFIX", "/opt/homebrew")
+            shared_library = Path(prefix) / "lib" / _gdal_lib_name()
+        elif preset == "system":
+            shared_library = _gdal_lib_name()
+        elif preset == "auto":
+            tried = ["pyogrio", "rasterio", "conda", "homebrew", "system"]
```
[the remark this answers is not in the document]
Kontinuation: Not a real concern for us: Python 3.8+ changed `ctypes.CDLL` to no longer search CWD by default on Windows, and Python < 3.8 is EOL. More importantly, `configure_proj` uses the exact same pattern (`_configure_proj_system()` passes bare `"proj.dll"` on Windows), and `system` is the last fallback in `auto` — it only runs if pyogrio, rasterio, conda, and homebrew all failed. No change needed; if we ever want to harden this, we should do it for both PROJ and GDAL together.

python/sedonadb/python/sedonadb/__init__.py
```diff
@@ -15,16 +15,17 @@
...
 
 __all__ = ["connect", "options"]
 
-# Attempt to configure PROJ on import. This will warn if PROJ
-# can't be configured but should never error. The auto-configured
+# Attempt to configure PROJ and GDAL on import. This will warn if PROJ
+# or GDAL can't be configured but should never error. The auto-configured
 # value can be overridden as long as the call to configure_proj()
-# occurs before actually creating a transform.
+# and configure_gdal() occurs before actually creating a transform.
```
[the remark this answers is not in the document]
Kontinuation: Good catch — fixed in bef2d663. The comment now correctly distinguishes between PROJ transforms and GDAL-backed operations.

python/sedonadb/python/sedonadb/context.py
```diff
@@ -604,3 +604,184 @@ def _configure_proj_prefix(prefix: str):
...
+
+
+def configure_gdal(
+    preset: Optional[
+        Literal["auto", "pyogrio", "rasterio", "conda", "homebrew", "system"]
+    ] = None,
+    *,
+    shared_library: Optional[Union[str, Path]] = None,
+    verbose: bool = False,
+) -> None:
```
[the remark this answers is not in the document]
Kontinuation: `configure_proj()` has the same structure and also has no dedicated Python-level unit tests — it is tested implicitly through integration tests. Adding monkeypatched tests for GDAL without equivalent tests for PROJ would be inconsistent. This could be a follow-up for both APIs.

Kontinuation:
I asked a coding agent to work on multi-platform CIs, and it pinpointed the problem loading C++ symbol on Windows, but didn't come up with a correct fix. I have worked on a proper fix manually and now it should be fine.

The mangled name of MEMDataset::Create works for GDAL>=3.5, I have downloaded every pyogrio versions from 0.5.0 to 0.12.1 to verify that. Apparently a better solution is to add MEMCreate to the upstream GDAL C API.

python/sedonadb/python/sedonadb/__init__.py
```diff
@@ -15,16 +15,18 @@
...
 from sedonadb import _lib
-from sedonadb.context import connect, configure_proj
+from sedonadb.context import connect, configure_proj, configure_gdal, gdal_version
 
 __version__ = _lib.sedona_python_version()
 
 __features__ = _lib.sedona_python_features()
 
-__all__ = ["connect", "options"]
+__all__ = ["connect", "options", "gdal_version"]
```
paleolimbot: The version will pretty much never be accessible from this module (because of the `apache-sedona` wrapper)...probably just as useful to keep it in `context`.

python/sedonadb/python/sedonadb/context.py
```diff
@@ -585,22 +590,234 @@ def _configure_proj_pyproj():
...
+    on package load but may need additional configuration (particularly if the
+    automatic configuration fails).
+
+    This function may be called at any time; however, once a GDAL-backed
+    operation has been performed, subsequent configuration has no effect.
+
+    Args:
+        preset: One of:
+            - None: Use a custom `shared_library` path.
+            - auto: Try all presets in the order pyogrio, rasterio, conda,
```
```suggestion
            - auto: Try all presets in the order rasterio, pyogrio, conda,
```

paleolimbot (approved):
Thank you!

The proj and gdal configurations would probably benefit from alignment (e.g., pull both PROJ and GDAL from rasterio if possible) but no need to do that here. I'm not sure this will add significant import time, but if it does, it may be worth deferring the configuration in the future.

python/sedonadb/python/sedonadb/__init__.py
Kontinuation: Removed `gdal_version` in https://github.com/apache/sedona-db/pull/689/changes/73956bde818c4026f2cbcb7865c270204fc2d0e2

python/sedonadb/python/sedonadb/context.py
Kontinuation: Fixed in https://github.com/apache/sedona-db/pull/689/changes/73956bde818c4026f2cbcb7865c270204fc2d0e2

paleolimbot (approved):
Thank you!

PR #695, opened 2026-03-09
feat(sedona-gdal): add geometry and spatial ref primitives

Kontinuation:
## Summary
- add standalone geometry, geotransform, spatial reference, and VSI wrappers
- provide the primitive GDAL wrapper layer that later dataset and raster operation PRs build on
- keep the module surface explicit by avoiding wrapper re-export aliases

## Stack
- foundation is already merged via #696, so this branch now targets `main` directly

## Testing
- `cargo test -p sedona-gdal`


c/sedona-gdal/src/vsi.rs
```diff
@@ -0,0 +1,186 @@
...
+use std::ffi::CString;
+
+use crate::errors::{GdalError, Result};
+use crate::gdal_api::{call_gdal_api, GdalApi};
+
+/// Creates a new VSI in-memory file from a given buffer.
+///
+/// The data is copied into GDAL-allocated memory (via `VSIMalloc`) so that
+/// GDAL can safely free it with `VSIFree` when ownership is taken.
+pub fn create_mem_file(api: &'static GdalApi, file_name: &str, data: &[u8]) -> Result<()> {
```
Kontinuation, line 88: The same function in georust/gdal is defined as

```rust
pub fn create_mem_file(api: &'static GdalApi, file_name: &str, data: &[u8]) -> Result<()>
```

See [src/vsi.rs#L40-L70](https://github.com/georust/gdal/blob/500753686b27c982cfd20ad4ab459b0cfafdba79/src/vsi.rs#L40-L70). This can be super problematic. The mem file buffer is allocated by Rust allocator and will be freed by the native GDAL library. There's no guarantee that Rust and GDAL will use the same or compatible allocator, Deallocation failures or memory corruption could happen when the mem file was dropped.

c/sedona-gdal/src/geo_transform.rs
```diff
@@ -0,0 +1,152 @@
...
+    fn invert(&self) -> errors::Result<GeoTransform>;
+}
+
+impl GeoTransformEx for GeoTransform {
+    /// Pure-Rust equivalent of GDAL's `GDALApplyGeoTransform`.
+    fn apply(&self, x: f64, y: f64) -> (f64, f64) {
+        let geo_x = self[0] + x * self[1] + y * self[2];
+        let geo_y = self[3] + x * self[4] + y * self[5];
+        (geo_x, geo_y)
+    }
```
paleolimbot, line 58: I think this is the third way in our code base that we have this (once in sedona-raster, once using a separate rust crate, and here). It may be worth just using the rust dependency rate instead of maintaining tests for this.

c/sedona-gdal/src/spatial_ref.rs
```diff
@@ -0,0 +1,172 @@
...
+            unsafe { call_gdal_api!(self.api, OSRRelease, self.c_srs) };
+        }
+    }
+}
+
+impl SpatialRef {
+    /// Create a new SpatialRef from a WKT string.
+    pub fn from_wkt(api: &'static GdalApi, wkt: &str) -> Result<Self> {
+        let c_wkt = CString::new(wkt)?;
+        let c_srs = unsafe { call_gdal_api!(api, OSRNewSpatialReference, c_wkt.as_ptr()) };
```
paleolimbot, line 54: Does this also handle other types of input or should we expose the version that does (maybe called SetFromUserInput?).

Do we have to set the axis mapping strategy to avoid flipped lon/lat definitions for EPSG:4326?

c/sedona-gdal/src/vsi.rs
```diff
@@ -0,0 +1,209 @@
...
+        return Err(GdalError::UnlinkMemFile {
+            file_name: file_name.to_string(),
+        });
+    }
+
+    Ok(())
+}
+
+/// Copies the bytes of the VSI in-memory file, taking ownership and freeing the GDAL memory.
+pub fn get_vsi_mem_file_bytes_owned(api: &'static GdalApi, file_name: &str) -> Result<Vec<u8>> {
```
paleolimbot: Maybe this could return a proxy to avoid the copy?

```rust
struct VSIBuffer {
   data: mut* u8,
   length: usize
}

impl Drop for VSIBuffer { ... }
impl AsRef<[u8]> for VSIBuffer { ... }
```

We could even attach this to BinaryView without copying I believe if we're using this for output somehow.

paleolimbot (approved):
Thank you!

c/sedona-gdal/src/vsi.rs
Kontinuation: Yes. `pub struct VSIBuffer` seems to be a better API. I have added `get_vsi_mem_file_buffer_owned` that returns `Result<VSIBuffer>`.

c/sedona-gdal/src/spatial_ref.rs
Kontinuation: I have added another constructor `SpatialRef::from_definition` that calls `OSRSetFromUserInput` under the hood. The function is also brought from georust/gdal.

Actually we do not need to construct SpatialRef objects directly when implementing RS functions. We call [`GDALSetProjection`](https://gdal.org/en/stable/api/raster_c_api.html#_CPPv417GDALSetProjection12GDALDatasetHPKc), which calls `SetFromUserInput` internally. It also forces GIS traditional axis order by [calling `oSRS.SetAxisMappingStrategy(OAMS_TRADITIONAL_GIS_ORDER)`](https://github.com/OSGeo/gdal/blob/19348d2770d32597d6d3875b20ec48b41fe33f1d/gcore/gdaldataset.cpp#L1509-L1525) before loading the projection, so EPSG:4326 will have lon-lat axis order just as OGC:CRS84.

I have added `axis_mapping_strategy`, `set_axis_mapping_strategy`, `data_axis_to_srs_axis_mapping` and `epsg_treats_as_lat_long` to SpatialRef for feature completeness. SpatialRef does not implicitly enforce axis order, users are free to set the axis order by calling `set_axis_mapping_strategy` themselves.

c/sedona-gdal/src/geo_transform.rs
Kontinuation: I have evaluated the approach of using `glam` to implement `GeoTransformEx` and found that it is less maintainable than the current approach.

1. We need to carefully map geo transform parameters to the matrix elements and translation vector of DAffine2, which still need all the tests to verify that we are implement it correctly.
2. The code for detecting close-to-zero determinant values cannot be eliminated after switching to glam, since glam does not expose an interface for computing inverse affine transformation while reporting non-inversible affine transformation as error. The code is still verbose after the refactoring.
3. These functions were directly translated from `GDALApplyGeoTransform` and `GDALInvGeoTransform` in GDAL source code, so the maintainer could easily verify that the behavior is identical to the original GDAL code.

The reason why we don't reuse the code in [`affine_transformation.rs`](https://github.com/apache/sedona-db/blob/cff68404ede9732e19c9cba5845aff9d09cdefc7/rust/sedona-raster/src/affine_transformation.rs) is because we want c/sedona-gdal to be a standalone module.

Another approach is to define fn apply as `apply(&self, api: &'static GdalApi, x: f64, y: f64)` and call into the `GDALApplyGeoTransform` of dynamically loaded GDAL. This requires the caller to pass in an awkward `api` parameter for this simple function, and the function is stable and simple enough to be replicated in Rust to avoid another level of indirection.

paleolimbot (approved):
Thank you!

PR #697, opened 2026-03-09
feat(sedona-gdal): add convenience facade and mem builder

Kontinuation:
## Summary
- add the high-level `Gdal` facade and `with_global_gdal` convenience entry point
- add the MEM dataset builder on top of the lower-level dataset and raster wrappers
- keep the top-level API explicit by importing concrete raster/vector modules instead of relying on wrapper re-export aliases

Depends on #698.

## Testing
- `cargo test -p sedona-gdal`

### DIFF
--- a/c/sedona-gdal/src/dataset.rs
+++ b/c/sedona-gdal/src/dataset.rs
@@ -300,11 +300,11 @@ impl Dataset {
     ///
     /// # Safety
     ///
-    /// `data_ptr` must point to valid band data that outlives this dataset.
+    /// `data_ptr` must point to valid mutable band data that outlives this dataset.
     pub unsafe fn add_band_with_data(
         &self,
         data_type: RustGdalDataType,
-        data_ptr: *const u8,
+        data_ptr: *mut u8,
         pixel_offset: Option<i64>,
         line_offset: Option<i64>,
     ) -> Result<()> {
--- a/c/sedona-gdal/src/dataset.rs
+++ b/c/sedona-gdal/src/dataset.rs
@@ -38,10 +38,17 @@ pub struct Dataset {
     c_dataset: GDALDatasetH,
 }
 
+<<<<<<< HEAD
 // SAFETY: `Dataset` has unique ownership of its GDAL dataset handle and only moves
 // that ownership across threads. The handle is closed exactly once on drop, and this
 // wrapper does not provide shared concurrent access, so `Send` is sound while `Sync`
 // remains intentionally unimplemented.
+=======
+// SAFETY: `Dataset` carries an opaque GDAL dataset handle plus an ownership flag.
+// Moving the wrapper across threads only transfers ownership of that handle; it does
+// not permit concurrent shared access. The handle is closed at most once on drop when
+// `owned` is true, so `Send` is sound while `Sync` remains intentionally unimplemented.
+>>>>>>> b522280f (test(sedona-gdal): add layer iteration coverage)
 unsafe impl Send for Dataset {}
 
 impl Drop for Dataset {
--- a/c/sedona-gdal/src/dataset.rs
+++ b/c/sedona-gdal/src/dataset.rs
@@ -38,17 +38,10 @@ pub struct Dataset {
     c_dataset: GDALDatasetH,
 }
 
-<<<<<<< HEAD
 // SAFETY: `Dataset` has unique ownership of its GDAL dataset handle and only moves
 // that ownership across threads. The handle is closed exactly once on drop, and this
 // wrapper does not provide shared concurrent access, so `Send` is sound while `Sync`
 // remains intentionally unimplemented.
-=======
-// SAFETY: `Dataset` carries an opaque GDAL dataset handle plus an ownership flag.
-// Moving the wrapper across threads only transfers ownership of that handle; it does
-// not permit concurrent shared access. The handle is closed at most once on drop when
-// `owned` is true, so `Send` is sound while `Sync` remains intentionally unimplemented.
->>>>>>> b522280f (test(sedona-gdal): add layer iteration coverage)
 unsafe impl Send for Dataset {}
 
 impl Drop for Dataset {
--- a/c/sedona-gdal/src/gdal.rs
+++ b/c/sedona-gdal/src/gdal.rs
@@ -51,7 +51,7 @@ pub struct Gdal {
 }
 
 impl Gdal {
-    /// Create a new `Gdal` instance wrapping the given API reference.
+    /// Create a `Gdal` wrapper for the given API reference.
     pub(crate) fn new(api: &'static GdalApi) -> Self {
         Self { api }
     }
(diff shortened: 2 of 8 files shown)


c/sedona-gdal/src/mem.rs
```diff
@@ -0,0 +1,439 @@
...
+
+    /// Add a zero-copy band from a raw data pointer.
+    /// Use default contiguous row-major pixel and line offsets.
+    ///
+    /// # Safety
+    ///
+    /// The caller must ensure `data_ptr` points to a valid buffer of at least
+    /// `height * width * data_type.byte_size()` bytes, and that the buffer
+    /// outlives the built [`Dataset`].
+    pub unsafe fn add_band(self, data_type: GdalDataType, data_ptr: *const u8) -> Self {
```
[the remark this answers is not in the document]
Kontinuation: Addressed in 44091077. I tightened the unsafe docs for the MEM builder to cover alignment requirements, dataset lifetime, and the possibility of GDAL writing through attached DATAPOINTER bands. I also made the external-band pointer contract explicit by switching this API from `*const u8` to `*mut u8`.

c/sedona-gdal/src/mem.rs
```diff
@@ -0,0 +1,439 @@
...
+
+/// Nodata value for a raster band.
+///
+/// GDAL has three separate APIs for setting nodata depending on the band data type:
+/// - [`f64`] for most types (UInt8 through Float64, excluding Int64/UInt64)
+/// - [`i64`] for Int64 bands
+/// - [`u64`] for UInt64 bands
+///
+/// This enum encapsulates all three variants so callers don't need to match on
+/// the band type when setting nodata.
```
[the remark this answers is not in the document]
Kontinuation: Addressed docs-wise in 44091077. I removed the stronger claim that callers do not need to match on band type and reworded this to describe the three GDAL nodata representations more directly. I am not adding validation here: we already have a mixed-type test (`test_mem_builder_mixed_owned_and_external_bands`) that uses `Nodata::U64(255)` with a `UInt8` band and round-trips successfully, so this behavior is intentional rather than an unchecked mismatch bug.

c/sedona-gdal/src/gdal.rs
```diff
@@ -0,0 +1,237 @@
...
+    // -- Info ----------------------------------------------------------------
+
+    /// Return the name of the loaded GDAL library.
+    pub fn name(&self) -> &str {
+        self.api.name()
+    }
+
+    /// Fetch GDAL version information for a standard request key.
+    pub fn version_info(&self, request: &str) -> String {
+        self.api.version_info(request)
```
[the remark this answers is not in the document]
Kontinuation: I am leaving this as-is. `version_info()` is intended for GDAL-defined request keys such as `RELEASE_NAME` and `VERSION_NUM`, so an interior NUL here is invalid programmer input and results in a panic rather than an unsoundness issue. I do not think that warrants widening the facade API to `Result<String>` or silently truncating the request.

c/sedona-gdal/src/mem.rs
```diff
@@ -0,0 +1,441 @@
...
+    ) -> Self {
+        self.bands.push(MemBand {
+            data_type,
+            data_ptr,
+            pixel_offset,
+            line_offset,
+            nodata,
+        });
+        self
+    }
```
paleolimbot, line 161: This is so cool

c/sedona-gdal/src/gdal.rs
```diff
@@ -0,0 +1,237 @@
...
+
+    // -- Info ----------------------------------------------------------------
+
+    /// Return the name of the loaded GDAL library.
+    pub fn name(&self) -> &str {
+        self.api.name()
+    }
+
+    /// Fetch GDAL version information for a standard request key.
+    pub fn version_info(&self, request: &str) -> String {
```
paleolimbot, line 68: These are all a bit light on the documentation (maybe they could link to the underlying implementation's docs which I seem to remember have a bit more depth?). Also OK to defer since this is mostly intended to be for internal usage.

c/sedona-gdal/src/gdal.rs
```diff
@@ -0,0 +1,237 @@
...
+
+/// High-level convenience wrapper around [`GdalApi`].
+///
+/// Stores a `&'static GdalApi` reference and provides ergonomic methods that
+/// delegate to the various constructors and free functions in the crate.
+pub struct Gdal {
+    api: &'static GdalApi,
+}
+
+impl Gdal {
```
paleolimbot, line 53: I do like the idea of a high-level object that dishes out other safe high-level objects, although putting it all here involves repeating some documentation (or omitting it) and I'm not sure it's needed given that the constructors it refers to are pub anyway. Reexporting the structs or functions that are part of the intended external API might be another way to expose the API without loosing the docs?

c/sedona-gdal/src/mem.rs
```diff
@@ -0,0 +1,441 @@
...
+
+    /// Set the dataset geotransform coefficients.
+    pub fn geo_transform(mut self, gt: [f64; 6]) -> Self {
+        self.geo_transform = Some(gt);
+        self
+    }
+
+    /// Set the dataset projection definition string.
+    pub fn projection(mut self, wkt: impl Into<String>) -> Self {
+        self.projection = Some(wkt.into());
```
paleolimbot: Is this actually WKT or can this by anything? (It may be less error prone to accept anything or an explicit spatialref object)

c/sedona-gdal/src/dataset.rs
```diff
@@ -300,11 +300,11 @@ impl Dataset {
     ///
     /// # Safety
     ///
-    /// `data_ptr` must point to valid band data that outlives this dataset.
+    /// `data_ptr` must point to valid mutable band data that outlives this dataset.
     pub unsafe fn add_band_with_data(
         &self,
         data_type: RustGdalDataType,
-        data_ptr: *const u8,
+        data_ptr: *mut u8,
```
paleolimbot, line 307: Is this actually required to be mutable? (This seems like a strange constraint, and also one that will result in it not being particularly useful to us)

c/sedona-gdal/src/mem.rs
```diff
@@ -0,0 +1,441 @@
...
+    /// Use default contiguous row-major pixel and line offsets.
+    ///
+    /// # Safety
+    ///
+    /// The caller must ensure `data_ptr` points to a valid buffer of at least
+    /// `height * width * data_type.byte_size()` bytes, is properly aligned for
+    /// `data_type`, and outlives the built [`Dataset`].
+    pub unsafe fn add_band(self, data_type: GdalDataType, data_ptr: *mut u8) -> Self {
+        self.add_band_with_options(data_type, data_ptr, None, None, None)
+    }
```
paleolimbot, line 136: Are you sure the data pointer has to be mutable? (This would preclude backing a GDAL mem dataset with Arrow data?)

paleolimbot (approved):
A few optional things that may be worth considering. Thank you!

c/sedona-gdal/src/gdal.rs
Kontinuation: Added "See also" style pointers to more comprehensive docs.

c/sedona-gdal/src/gdal.rs
Kontinuation: This `Gdal` facade is to avoid passing the `api: GdalApi` value every time we call a GDAL safe API. I have also found that exposing `with_global_gdal` instead of `with_global_gdal_api` makes the users of the gdal crate much easier to discover GDAL APIs, since all almost all GDAL APIs were grouped in the same place instead of scattering in different modules.

c/sedona-gdal/src/mem.rs
Kontinuation: It does not have to be a WKT. I have renamed the parameter to `projection` to avoid confusion. It calls `GDALSetProjection` to set the projection, which handles X/Y axis order automatically. The unique semantics of `GDALSetProjection` makes it a bit different from `GDALSetSpatialRef`, and I believe that `GDALSetProjection` is a more suitable API for sedona-db's axis order convention.

c/sedona-gdal/src/dataset.rs
Kontinuation: `*mut u8` is required because this crate exposes safe APIs like `RasterBand::write`, `rasterize`, and `rasterize_affine` that can write through the attached MEM band into the caller-provided backing buffer. If the builder accepted `*const u8`, safe Rust could later trigger writes into immutable memory.

Actually one of the most important use case of the MEM dataset is for holding the result of rasterize, or the intermediate mask raster as a result of rasterize when running zonal stats. The underlying buffer will be mutated in such use cases.

paleolimbot (approved):
Thank you!

PR #699, opened 2026-03-09
feat(sedona-gdal): add dataset and vector/raster wrappers

Kontinuation:
## Summary
- add safe wrappers for GDAL datasets, drivers, raster bands, features, and layers
- wire the core raster/vector modules that higher-level operations build on
- keep the wrapper surface explicit by removing module re-export aliases from this layer

## Stack
- foundation and primitives are already merged into `main` via #696 and #695
- this branch is now rebased directly onto current `main`
- downstream dependency: #698

## Testing
- `cargo test -p sedona-gdal`

### DIFF
--- a/c/sedona-gdal/src/dataset.rs
+++ b/c/sedona-gdal/src/dataset.rs
@@ -23,6 +23,7 @@ use std::ffi::{CStr, CString};
 use std::ptr;
 
 use crate::cpl::CslStringList;
+use crate::driver::Driver;
 use crate::errors::{GdalError, Result};
 use crate::gdal_api::{call_gdal_api, GdalApi};
 use crate::gdal_dyn_bindgen::*;
@@ -243,7 +244,7 @@ impl Dataset {
     /// Create a copy of this dataset to a new file using the given driver.
     pub fn create_copy(
         &self,
-        driver: &crate::driver::Driver,
+        driver: &Driver,
         filename: &str,
         options: &[&str],
     ) -> Result<Dataset> {
--- a/c/sedona-gdal/src/dataset.rs
+++ b/c/sedona-gdal/src/dataset.rs
@@ -39,6 +39,10 @@ pub struct Dataset {
     owned: bool,
 }
 
+// SAFETY: `Dataset` carries an opaque GDAL dataset handle plus an ownership flag.
+// Moving the wrapper across threads only transfers ownership of that handle; it does
+// not permit concurrent shared access. The handle is closed at most once on drop when
+// `owned` is true, so `Send` is sound while `Sync` remains intentionally unimplemented.
 unsafe impl Send for Dataset {}
 
 impl Drop for Dataset {
--- a/c/sedona-gdal/src/dataset.rs
+++ b/c/sedona-gdal/src/dataset.rs
@@ -36,18 +36,17 @@ use crate::vector::layer::Layer;
 pub struct Dataset {
     api: &'static GdalApi,
     c_dataset: GDALDatasetH,
-    owned: bool,
 }
 
-// SAFETY: `Dataset` carries an opaque GDAL dataset handle plus an ownership flag.
-// Moving the wrapper across threads only transfers ownership of that handle; it does
-// not permit concurrent shared access. The handle is closed at most once on drop when
-// `owned` is true, so `Send` is sound while `Sync` remains intentionally unimplemented.
+// SAFETY: `Dataset` has unique ownership of its GDAL dataset handle and only moves
+// that ownership across threads. The handle is closed exactly once on drop, and this
+// wrapper does not provide shared concurrent access, so `Send` is sound while `Sync`
+// remains intentionally unimplemented.
 unsafe impl Send for Dataset {}
 
 impl Drop for Dataset {
     fn drop(&mut self) {
-        if self.owned && !self.c_dataset.is_null() {
+        if !self.c_dataset.is_null() {
             unsafe { call_gdal_api!(self.api, GDALClose, self.c_dataset) };
         }
     }
@@ -100,33 +99,12 @@ impl Dataset {
             return Err(api.last_cpl_err(CE_Failure as u32));
         }
 
-        Ok(Self {
-            api,
-            c_dataset,
-            owned: true,
-        })
+        Ok(Self { api, c_dataset })
     }
 
-    /// Create a new owned Dataset from a C handle.
-    pub(crate) fn new_owned(api: &'static GdalApi, c_dataset: GDALDatasetH) -> Self {
-        Self {
-            api,
-            c_dataset,
-            owned: true,
-        }
-    }
-
-    /// Wrap an existing C dataset handle (non-owning).
-    ///
-    /// # Safety
-    ///
-    /// The caller must ensure the handle is valid and outlives this `Dataset`.
-    pub unsafe fn from_c_dataset(api: &'static GdalApi, c_dataset: GDALDatasetH) -> Self {
-        Self {
-            api,
-            c_dataset,
-            owned: false,
-        }
+    /// Create a new Dataset from an owned C handle.
+    pub(crate) fn new(api: &'static GdalApi, c_dataset: GDALDatasetH) -> Self {
+        Self { api, c_dataset }
     }
 
     /// Return the raw C dataset handle.
@@ -269,11 +247,7 @@ impl Dataset {
         if c_ds.is_null() {
             return Err(self.api.last_cpl_err(CE_Failure as u32));
         }
-        Ok(Dataset {
-            api: self.api,
-            c_dataset: c_ds,
-            owned: true,
-        })
+        Ok(Dataset::new(self.api, c_ds))
     }
 
     /// Create a new vector layer.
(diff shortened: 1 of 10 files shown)


c/sedona-gdal/src/vector/layer.rs
```diff
@@ -0,0 +1,118 @@
...
+    }
+}
+
+/// Iterator over features in a layer.
+pub struct FeatureIterator<'a> {
+    layer: &'a Layer<'a>,
+}
+
+impl<'a> Iterator for FeatureIterator<'a> {
+    type Item = Feature<'a>;
```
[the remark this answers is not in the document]
Kontinuation: Thanks. I adjusted `Layer::features` to take `&mut self` so callers cannot create multiple iterators from the same layer borrow at once, which avoids the interference concern for this API shape. I am leaving the lifetime structure as-is for now and will revisit it only if we run into a concrete borrow ergonomics issue later.

c/sedona-gdal/src/dataset.rs
```diff
@@ -0,0 +1,462 @@
...
+    }
+
+    /// Set the geo-transform.
+    pub fn set_geo_transform(&self, gt: &[f64; 6]) -> Result<()> {
+        let rv = unsafe {
+            call_gdal_api!(
+                self.api,
+                GDALSetGeoTransform,
+                self.c_dataset,
+                gt.as_ptr() as *mut f64
```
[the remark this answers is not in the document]
Kontinuation: I checked the actual GDAL signature here: `GDALSetGeoTransform` takes `const double *`, and the corresponding `gdal-sys` binding also uses `*const f64`. So this does not need a wrapper-side change; the earlier concern came from the dynamic binding signature, not from the GDAL API contract.

c/sedona-gdal/src/dataset.rs
```diff
@@ -0,0 +1,462 @@
...
+use crate::spatial_ref::SpatialRef;
+use crate::vector::layer::Layer;
+
+/// A GDAL dataset.
+pub struct Dataset {
+    api: &'static GdalApi,
+    c_dataset: GDALDatasetH,
+    owned: bool,
+}
+
```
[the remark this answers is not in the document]
Kontinuation: Fixed. I added a SAFETY comment on `Dataset` explaining why moving the owned handle across threads is sound while `Sync` remains intentionally unimplemented.

c/sedona-gdal/src/raster/rasterband.rs
```diff
@@ -0,0 +1,368 @@
...
+            let path = fixture("tinymarble.tif");
+            let dataset = Dataset::open_ex(api, &path, GDAL_OF_READONLY, None, None, None).unwrap();
+            let rb = dataset.rasterband(1).unwrap();
+            let rv = rb
+                .read_as::<u8>((20, 30), (4, 4), (2, 2), Some(ResampleAlg::Average))
+                .unwrap();
+            assert_eq!(rv.shape, (2, 2));
+            // Average resampling; exact values are GDAL-version-dependent.
+            // Verify shape and that results differ from the non-resampled full read.
+            assert_eq!(rv.data().len(), 4);
```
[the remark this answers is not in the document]
Kontinuation: Good catch. I took the low-risk option here and updated the comment so it matches the actual assertion instead of adding a value-sensitive resampling check that could become GDAL-version-dependent.

c/sedona-gdal/src/vector/layer.rs
```diff
@@ -0,0 +1,118 @@
...
+            )
+        }
+    }
+
+    /// Iterate over all features.
+    pub fn features(&self) -> FeatureIterator<'_> {
+        self.reset_reading();
+        FeatureIterator { layer: self }
+    }
+}
```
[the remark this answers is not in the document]
Kontinuation: Fixed. I added a dedicated round-trip test for layer iteration/reset behavior that inserts a few point features into an in-memory vector dataset and then exercises `feature_count`, iterator exhaustion, `reset_reading`, and repeated `features()` calls.

c/sedona-gdal/src/dataset.rs
```diff
@@ -0,0 +1,458 @@
...
+use crate::raster::types::{DatasetOptions, GdalDataType as RustGdalDataType};
+use crate::spatial_ref::SpatialRef;
+use crate::vector::layer::Layer;
+
+/// A GDAL dataset.
+pub struct Dataset {
+    api: &'static GdalApi,
+    c_dataset: GDALDatasetH,
+    owned: bool,
+}
```
paleolimbot, line 39: Can you expand a little on the ownership here and why we need a non-owned version of this? (I trust that you need this, I have just not seen an `owned` flag that behaves in this way before).

c/sedona-gdal/src/dataset.rs
```diff
@@ -0,0 +1,458 @@
...
+                GDALSetProjection,
+                self.c_dataset,
+                c_projection.as_ptr()
+            )
+        };
+        if rv != CE_None {
+            return Err(self.api.last_cpl_err(rv as u32));
+        }
+        Ok(())
+    }
```
paleolimbot, line 201: Given we have a wrapper for SpatialRef, do we need these?

c/sedona-gdal/src/dataset.rs
```diff
@@ -0,0 +1,458 @@
...
+            options: None,
+        }
+    }
+}
+
+#[cfg(all(test, feature = "gdal-sys"))]
+mod tests {
+    use crate::driver::DriverManager;
+    use crate::global::with_global_gdal_api;
+
```
paleolimbot, line 369: These tests don't cover any of the vector layer features. Should those features be removed for now (I can work on GDAL vector layer stuff later so we can use it to back a datasource and add the associated tests then)

c/sedona-gdal/src/vector/layer.rs
```diff
@@ -0,0 +1,208 @@
...
+use crate::gdal_api::{call_gdal_api, GdalApi};
+use crate::gdal_dyn_bindgen::*;
+use crate::vector::feature::{Feature, FieldDefn};
+
+/// An OGR layer (borrowed from a Dataset).
+pub struct Layer<'a> {
+    api: &'static GdalApi,
+    c_layer: OGRLayerH,
+    _dataset: PhantomData<&'a Dataset>,
+}
```
paleolimbot, line 35: Unless you need this for the raster work I think we should remove it for now. The options exposed by georust/gdal are not really sufficient for what we would need to do with them to make use of this (and we need a different ownership model so that we can return `Box<dyn RecordBatchReader>`). I'm happy to port over the vector stuff later.

c/sedona-gdal/src/vector/feature.rs
```diff
@@ -0,0 +1,213 @@
...
+use crate::gdal_api::{call_gdal_api, GdalApi};
+use crate::gdal_dyn_bindgen::*;
+use crate::vector::geometry::Envelope;
+
+/// An OGR feature.
+pub struct Feature<'a> {
+    api: &'static GdalApi,
+    c_feature: OGRFeatureH,
+    _lifetime: PhantomData<&'a ()>,
+}
```
paleolimbot, line 35: Unless you need this for raster work I think we can port this separately as well. We would almost certainly use the Arrow interface instead of the feature interface.

paleolimbot (approved):
Thank you!

The vector bits seem lightly tested or not tested...unless you need them for something specific I think we should just port them separately (perhaps more closely targeted to exactly what we need to do with them).

c/sedona-gdal/src/dataset.rs
Kontinuation: I believe that non-owned dataset is not used by our code. georust/gdal also does not support Dataset wrapping non-owned GDAL dataset handles. I'll going to remove it.

c/sedona-gdal/src/dataset.rs
Kontinuation: `GDALSetProjection` is pretty convenient and forces lon/lat axis order automatically, so I'd prefer keeping and using it.

c/sedona-gdal/src/vector/layer.rs
Kontinuation: It is required to make raster work. Raster operations such as [polygonize](https://gdal.org/en/stable/api/gdal_alg.html#_CPPv414GDALPolygonize15GDALRasterBandH15GDALRasterBandH9OGRLayerHiPPc16GDALProgressFuncPv) and clipping need to work with geometry layers and features.

c/sedona-gdal/src/dataset.rs
Kontinuation: I have removed the `owned` field and always treat the dataset as owned.

c/sedona-gdal/src/dataset.rs
Kontinuation: I have added basic vector layer tests in the last commit https://github.com/apache/sedona-db/pull/699/commits/873364eddf29418e3eddf551abff1faf8787790d.

paleolimbot (approved):
Thank you!

PR #700, opened 2026-03-09
fix(python/sedonadb): support DuckDB 1.5 spatial tests

Kontinuation:
## Summary
- guard the DuckDB test helper's GeoArrow registration call so it still works across DuckDB 1.4 and 1.5
- switch DuckDB test result collection to the DB-API Arrow path, which avoids the DuckDB 1.5 GeoParquet relation export failure
- validate the focused Python regression suite against both DuckDB 1.4.4 and 1.5.0


paleolimbot (approved):
Thanks!

I'm going to merge this for CI on other PRs!

PR #717, opened 2026-03-17
feat(c/libgpuspatial): Change the interface of SpatialRefiner

pwrliang:
This PR addresses a thread-safety issue in `RTSpatialRefiner` that occurs during concurrent execution.

### The Problem: Race Conditions in Schema Initialization
In the original design, the `init_schema` method was responsible for initializing the `ArrowArrayView` for both the **build** and **stream** sides. 

When `RTSpatialRefiner` is invoked across multiple threads, the shared state within the stream-side `ArrowArrayView` is modified concurrently. This leads to a race condition where threads overwrite or corrupt the view, resulting in undefined behavior during spatial refinement.

### Fix ###
In this PR, the `init_schema` is replaced with `init_build_schema`. In the `CRefine` method, the caller is required to pass a schema for the stream side to create a local `ArrowArrayView` object.

### DIFF
--- a/c/sedona-libgpuspatial/libgpuspatial/include/gpuspatial/gpuspatial_c.h
+++ b/c/sedona-libgpuspatial/libgpuspatial/include/gpuspatial/gpuspatial_c.h
@@ -155,9 +155,13 @@ struct SedonaSpatialRefiner {
   /** Clear all built geometries from the refiner */
   int (*clear)(struct SedonaSpatialRefiner* self);
 
-  int (*init_schema)(struct SedonaSpatialRefiner* self,
-                     const struct ArrowSchema* build_schema,
-                     const struct ArrowSchema* probe_schema);
+  /** Initialize the spatial refiner with the schema of the build geometries
+   *
+   * @param build_schema The Arrow schema of the build geometries;
+   * @return 0 on success, non-zero on failure
+   */
+  int (*init_build_schema)(struct SedonaSpatialRefiner* self,
+                           const struct ArrowSchema* build_schema);
 
   /** Push geometries for building the spatial refiner
    *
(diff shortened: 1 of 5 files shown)


pwrliang:
Hi @paleolimbot please take a look.

c/sedona-libgpuspatial/libgpuspatial/include/gpuspatial/gpuspatial_c.h
```diff
@@ -155,9 +155,13 @@ struct SedonaSpatialRefiner {
...
-  int (*init_schema)(struct SedonaSpatialRefiner* self,
-                     const struct ArrowSchema* build_schema,
-                     const struct ArrowSchema* probe_schema);
+  /** Initialize the spatial refiner with the schema of the build geometries
+   *
+   * @param build_schema The Arrow schema of the build geometries;
+   * @return 0 on success, non-zero on failure
+   */
+  int (*init_build_schema)(struct SedonaSpatialRefiner* self,
+                           const struct ArrowSchema* build_schema);
```
paleolimbot, line 164: Rather than changing the interface, how about keeping a copy of the `nanoarrow::UniqueSchema` for the probe side and using it to recreate the probe side ArrowArrayView for each call to `refine()` (i.e., the change you made below)?

I still think it is good inferface-wise to give as much information as possible as early as possible.

paleolimbot (commented):
Thank you! You are definitely correct that reusing the array views between threads isn't going to work 😬 . I have a suggestion about how to go about this but let me know if this is not going to work.

c/sedona-libgpuspatial/libgpuspatial/include/gpuspatial/gpuspatial_c.h
pwrliang: That would work, but could we avoid binding the build and probe schemas together? Since the `GPUSpatialIndex` design follows the CPU-based spatial index interfaces, it doesn't require any knowledge of the probe side to build the index until it is actually used. Requiring the initialization of both schemas makes the design overly complicated, as it forces us to pass the probe schema from the creation of SpatialJoinExec all the way down to `GPUSpatialIndexBuilder`. Furthermore, the CPU implementation doesn't require the probe schema at all. I don't see any benefit to passing the schemas as a pair.

paleolimbot (approved):
Point taken...thanks! I may need this to handle geographies but if I do (and I need the refiner to know what it needs to handle early) I'll circle back. We definitely know the schema somewhere by the point we instantiate a refiner but I can definitely appreciate that it's not available there today.

PR #718, opened 2026-03-17
feat(c/libgpuspatial): Upgrade RAPIDS RMM to 25.12

pwrliang:
RAPIDS RMM is a memory management library used by c/libgpuspatial. The current RMM version we used is 25.06, which has a use-after-free bug https://github.com/rapidsai/rmm/pull/1966.
This PR upgrades RMM to 25.12, which solves that bug, but it also has another bug (https://github.com/rapidsai/rmm/issues/2303). I have patched up RMM 25.12 in this PR. Hopefully, we can upgrade the RMM to its next release version, and the patch will not be needed.


paleolimbot (approved):
Thank you!

PR #721, opened 2026-03-17
feat(c/libgpuspatial): Fix synchronization bugs and improve logging

pwrliang:
This PR fixes synchronization bugs of RTSpatialRefiner and improves logging for profiling purposes.

PR #719, opened 2026-03-17
feat(c/sedona-libgpuspatial): Improve building robustness

pwrliang:
This PR improves the robustness of building libgpuspatial by finding both `lib` and `lib64`. This PR resolves the build failure on CentOS.

Issue #702, opened 2026-03-09
dev/release: Simplify cargo publish


paleolimbot:
Our cargo publish sequence is currently quite complicated because we have circular dev dependencies. It's possible to work around this but it is annoying and not great practice!

I think we should:

- Add some check for circular dependencies in the dependency graph. The only culprits I know about are `sedona` and `sedona-testing` in dev-dependencies.
- I don't know how sedona-raster became a dependency of sedona-functions but we should probably move that
- sedona depending on sedona-testing (for the random geometry generator) needs to get fixed.
- Separate out a crate for high-level benchmarks and/or high level tests. DataFusion has most of its tests in the `datafusion` crate's `tests` directory and a completely separate benchmark setup, perhaps for this reason. I do like keeping *most* of the tests file-local and I'm sure we can find a way to separate out the logic we need (maybe a high- and low- level testing crate) to keep that (even if we have to move some higher level integration tests or benchmarks).
- Many of our benchmarks could be dynamic (e.g., we could have a CLI that passes some ArgSpecs as strings).

paleolimbot:
Bumping to the next release so we can get this one out the door!

PR #722, opened 2026-03-17
feat(rust/sedona-spatial-join): Integrate libgpuspatial into sedona-spatial-join

pwrliang:
This PR integrates libgpuspatial by providing a `GPUSpatialIndex` and `GPUSpatialIndexBuilder`. The PR should pass the CI after its predecessor PRs (https://github.com/apache/sedona-db/pull/717, https://github.com/apache/sedona-db/pull/718, https://github.com/apache/sedona-db/pull/719, https://github.com/apache/sedona-db/pull/721) are merged.


pwrliang:
@paleolimbot @Kontinuation Could you take a look when you are available?

rust/sedona-spatial-join/src/index/gpu_spatial_index_builder.rs
```diff
@@ -0,0 +1,327 @@
...
+    fn add_batch(&mut self, indexed_batch: EvaluatedBatch) -> Result<()> {
+        let in_mem_size = indexed_batch.in_mem_size()?;
+        self.indexed_batches.push(indexed_batch);
+        self.record_memory_usage(in_mem_size);
+        Ok(())
+    }
+    fn merge_stats(&mut self, stats: GeoStatistics) -> &mut Self {
+        self.stats.merge(&stats);
+        self
+    }
```
Kontinuation: GPU join does not make use of geo statistics. We can remove `merge_stats` and `stats` field altogether.

rust/sedona-spatial-join/src/index/gpu_spatial_index_builder.rs
```diff
@@ -0,0 +1,327 @@
...
+            let all_record_batches: Vec<&RecordBatch> = self
+                .indexed_batches
+                .iter()
+                .map(|batch| &batch.batch)
+                .collect();
+            let schema = all_record_batches[0].schema();
+            let batch =
+                arrow::compute::concat_batches(&schema, all_record_batches).map_err(|e| {
+                    DataFusionError::Execution(format!("Failed to concatenate left batches: {}", e))
+                })?;
```
Kontinuation: There is no need to `.map_err` for `ArrowError`, it could be converted to DataFusionError::ArrowError automatically.

rust/sedona-spatial-join-gpu/src/index/gpu_spatial_index_builder.rs
```diff
@@ -0,0 +1,327 @@
...
+        // Concat indexed batches into a single batch to reduce build time
+        let total_rows: usize = self
+            .indexed_batches
+            .iter()
+            .map(|batch| batch.batch.num_rows())
+            .sum();
+
+        let sedona_type = self.indexed_batches[0].geom_array.sedona_type.clone();
+
+        if self.options.gpu.concat_build {
```
Kontinuation: Do we have tests covering the `gpu.concat_build: false` case?

rust/sedona-spatial-join/src/operand_evaluator.rs
```diff
@@ -116,10 +142,35 @@ impl EvaluatedGeometryArray {
...
+                    #[cfg(feature = "gpu")]
+                    {
+                        use wkb::reader::GeometryType;
+                        // For point geometries, we can directly cast f64 to f32 without expanding the box.
+                        // This enables libgpuspatial to treat the Rect as point for faster processing.
+                        if wkb.geometry_type() == GeometryType::Point {
+                            Some(Rect::new(
+                                coord!(x: min.x as f32, y: min.y as f32),
+                                coord!(x: max.x as f32, y: max.y as f32),
+                            ))
```
Kontinuation: Will this introduce false negatives since the resulting bounding box is not guaranteed to cover the original bounding box?

paleolimbot (commented):

My main concern with this PR is that the GPU details are very entangled with the non-GPU details. How about:

- `sedona-spatial-join-common` (defines the traits you nicely separated in the last PR and a `SedonaSpatialJoinExtension` trait)
- `sedona-spatial-join-gpu` (implements the traits in sedona-spatial-join-common)
- `sedona-spatial-join` (uses an `Arc<dyn SedonaSpatialJoinExtension>` object to decide if an extension join should be used in place of the normal join). The hook to register a join extension can be a global function for now and you can use a mock join extension built from the default implementation index/refiner to test it.
- `sedona` (when built with the gpu feature, registers the Gpu join with sedona-spatial-join)

There are a few things I think this will help with:

- We may want to add other join accelerators for other hardware (e.g., Apple GPU) or other data types (Geography x H3 cell join), or possibly Raster/Vector join.
- It will make it easier for you to have clear ownership over a subdirectory so you can iterate faster.
- It makes the `sedona` crate the single place where features are assembled (to avoid two levels of feature flags, the consequences of which you've started to identify here with the "all except GPU" feature).

I do think it's important to get this right early while we are all here and have time to dedicate to this!

pwrliang:
> Thank you for working on this!
> 
> My main concern with this PR is that the GPU details are very entangled with the non-GPU details. How about:
> 
> * `sedona-spatial-join-common` (defines the traits you nicely separated in the last PR and a `SedonaSpatialJoinExtension` trait)
> * `sedona-spatial-join-gpu` (implements the traits in sedona-spatial-join-common)
> * `sedona-spatial-join` (uses an `Arc<dyn SedonaSpatialJoinExtension>` object to decide if an extension join should be used in place of the normal join). The hook to register a join extension can be a global function for now and you can use a mock join extension built from the default implementation index/refiner to test it.
> * `sedona` (when built with the gpu feature, registers the Gpu join with sedona-spatial-join)
> 
> There are a few things I think this will help with:
> 
> * We may want to add other join accelerators for other hardware (e.g., Apple GPU) or other data types (Geography x H3 cell join), or possibly Raster/Vector join.
> * It will make it easier for you to have clear ownership over a subdirectory so you can iterate faster.
> * It makes the `sedona` crate the single place where features are assembled (to avoid two levels of feature flags, the consequences of which you've started to identify here with the "all except GPU" feature).
> 
> I do think it's important to get this right early while we are all here and have time to dedicate to this!

Thanks for the suggestions. If I understand correctly, this design implies that `sedona-spatial-join-gpu` will execute end-to-end spatial joins. This requires several components to work together, such as the optimizer, planner, and `SpatialJoinStream`. We would either need to export all these components to the GPU module or copy/paste the code. My concern is that this design could introduce a lot of redundant code. Since `libgpuspatial` functions as a refinement backend (similar to GEOS, GEO, or TG), I'm not sure why adding one more backend requiring such significant architectural changes.

rust/sedona-spatial-join/src/operand_evaluator.rs
pwrliang: Thanks for mentioning this. When one side contains only points, their coordinates in FP64 are downcast to FP32, which may result in false negatives. However, on the other side containing rectangles, I called `next_after` to produce a conservative bounding box. This compensates for potential false-negative issues.

paleolimbot: For Geography joins we will also have to play similar games (i.e., create our own EvaluatedGeometryArray)...I will try to find an extension point here so that we can keep this logic separated.

paleolimbot: I implemented this in https://github.com/apache/sedona-db/pull/737 (the ability to override the Rect creation for a SpatialJoinProvider); however, it's worth double checking that these rectangles are not used anywhere that this assumption might cause problems.

rust/sedona-common/src/option.rs
```diff
@@ -114,6 +114,9 @@ config_namespace! {
 
         /// Options for debugging or testing spatial join
         pub debug : SpatialJoinDebugOptions, default = SpatialJoinDebugOptions::default()
+
+        /// GPU acceleration options
+        pub gpu: GpuOptions, default = GpuOptions::default()
```
paleolimbot: These config options can also live in sedona-spatial-join-gpu (see how sedona-pointcloud does this to ensure the options are injected at the appropriate time).

rust/sedona-spatial-join/src/index/default_spatial_index.rs
```diff
@@ -283,6 +283,7 @@ impl SpatialIndex for DefaultSpatialIndex {
         self.inner.schema.clone()
     }
 
+    #[allow(dead_code)]
```
paleolimbot:
```suggestion
```

I think this can be removed now because the trait is pub?

rust/sedona/src/context.rs
```diff
@@ -149,11 +149,24 @@ impl SedonaContext {
         let mut planner = SedonaQueryPlanner::new();
         #[cfg(feature = "spatial-join")]
         {
-            use sedona_spatial_join::physical_planner::DefaultSpatialJoinPhysicalPlanner;
+            #[cfg(not(feature = "gpu"))]
```
paleolimbot:
```suggestion
```

I believe that we always want the default spatial join in place here...this is how fallback works when the GPU planner returns `None`.

rust/sedona-spatial-join-gpu/Cargo.toml
```diff
@@ -0,0 +1,75 @@
...
+[features]
+gpu = ["sedona-libgpuspatial/gpu"]
+
+[dependencies]
+arrow = { workspace = true }
+arrow-array = { workspace = true }
+arrow-schema = { workspace = true }
+async-trait = { workspace = true }
+datafusion = { workspace = true }
+datafusion-common = { workspace = true }
```
paleolimbot, line 42: Can any of these dependencies be removed?

rust/sedona-spatial-join-gpu/src/physical_planner.rs
```diff
@@ -0,0 +1,342 @@
...
+        &self,
+        args: &PlanSpatialJoinArgs<'_>,
+    ) -> Result<Option<Arc<dyn ExecutionPlan>>> {
+        if !is_spatial_predicate_supported(
+            args.spatial_predicate,
+            &args.physical_left.schema(),
+            &args.physical_right.schema(),
+        )? {
+            return Ok(None);
+        }
```
paleolimbot: Should you check the fallback option here and error if it is false?

rust/sedona-spatial-join-gpu/src/join_provider.rs
```diff
@@ -0,0 +1,128 @@
...
+                if let Some(rect) = wkb.bounding_rect() {
+                    let min = rect.min();
+                    let max = rect.max();
+                    // Why conservative bounding boxes prevent false negatives:
+                    // 1. P32 = round_nearest(P64), so P32 is the closest possible float to P64.
+                    // 2. Min32 = round_down(Min64), guaranteeing Min32 <= Min64.
+                    // 3. Max32 = round_up(Max64), guaranteeing Max32 >= Max64.
+                    // If P64 is inside Box64 (Min64 <= P64 <= Max64), P32 cannot fall outside Box32.
+                    // If P32 < Min32, it would mean Min32 is closer to P64 than P32 is, which
+                    // contradicts P32 being the nearest float. Therefore, false negatives are impossible.
```
paleolimbot, line 115: The only thing I am not sure about here is the case where spilling occurs, which I think is the only place these rectangles are used except to pass them to the SpatialIndex. They're used to split batches up into partitions (where the approximation is fine) but I think the bounds of each partition are calculated using them as well. No need to solve that here, just pointing it out since I recently spent some time with those internals.

You may want to allude to why the false negatives are prevented specifically for points with a reference to how they are sent to the GPU (not obvious from the current comment).

rust/sedona-spatial-join-gpu/src/index/gpu_spatial_index_builder.rs
```diff
@@ -0,0 +1,315 @@
...
+    }
+
+    pub(crate) fn estimate_extra_memory_usage(
+        geo_stats: &GeoStatistics,
+        _spatial_predicate: &SpatialPredicate,
+        _options: &SpatialJoinOptions,
+    ) -> usize {
+        let num_geoms = geo_stats.total_geometries().unwrap_or(0) as usize;
+        // Each geometry requires 4 f32 values to store the bounding rectangle (min_x, min_y, max_x, max_y)
+        num_geoms * (4 * 4)
```
paleolimbot, line 127: If these are stored on the GPU you might not have to count them (either way it may be good to leave a note here about how much memory might be required on the GPU given this estimate in addition to the CPU memory for the index + refiner if there is any).

rust/sedona-spatial-join-gpu/src/index/gpu_spatial_index_builder.rs
```diff
@@ -0,0 +1,315 @@
...
+    for (batch_idx, batch) in batches.iter().enumerate() {
+        for row_idx in 0..batch.num_rows() {
+            indices.push((batch_idx, row_idx));
+        }
+    }
+
+    // 3. Concatenate Geometry Arrays using the interleave method
+    let geom_arrays: Vec<&EvaluatedGeometryArray> = batches.iter().map(|b| &b.geom_array).collect();
+
+    let concatenated_geom_array = EvaluatedGeometryArray::interleave(&geom_arrays, &indices)?;
```
paleolimbot, line 152: No need to solve it here, but it might be faster if we provided an `EvaluatedGeometryArray::concatenate()`

rust/sedona-spatial-join-gpu/src/index/gpu_spatial_index_builder.rs
```diff
@@ -0,0 +1,315 @@
...
+                SpatialRelationType::CoveredBy => true,
+                SpatialRelationType::Touches => true,
+                SpatialRelationType::Crosses => false,
+                SpatialRelationType::Overlaps => false,
+                SpatialRelationType::Equals => true,
+            },
+            SpatialPredicate::Distance(_) => false,
+            SpatialPredicate::KNearestNeighbors(_) => false,
+        }
+    }
```
paleolimbot: This is different than the list you have in planner.rs...is that intentional?

rust/sedona-spatial-join-gpu/src/index/gpu_spatial_index.rs
```diff
@@ -0,0 +1,765 @@
...
+            ),
+            indexed_batches: vec![],
+            data_id_to_batch_pos: vec![],
+            visited_build_side: None,
+            probe_threads_counter,
+        })
+    }
+
+    #[allow(clippy::too_many_arguments)]
+    pub fn new(
```
paleolimbot: Instead of the `allow()` here, it's usually more readable to just remove the `new()` constructor and construct these using the named struct syntax.

rust/sedona-spatial-join-gpu/src/index/gpu_spatial_index.rs
```diff
@@ -0,0 +1,765 @@
...
+    }
+}
+
+#[async_trait]
+impl SpatialIndex for GPUSpatialIndex {
+    fn schema(&self) -> SchemaRef {
+        self.schema.clone()
+    }
+
+    #[allow(dead_code)] // This is used for tests
```
paleolimbot:
```suggestion
```

rust/sedona-spatial-join-gpu/src/index/gpu_spatial_index.rs
```diff
@@ -0,0 +1,765 @@
...
+            range.end,
+        ))
+    }
+    fn need_more_probe_stats(&self) -> bool {
+        false
+    }
+
+    fn merge_probe_stats(&self, stats: GeoStatistics) {
+        let _ = stats;
+    }
```
paleolimbot:
```suggestion
    fn merge_probe_stats(&self, _stats: GeoStatistics) {}
```

rust/sedona-spatial-join-gpu/src/index/gpu_spatial_index.rs
```diff
@@ -0,0 +1,765 @@
...
+
+#[cfg(test)]
+#[cfg(feature = "gpu")]
+mod tests {
+    use crate::index::gpu_spatial_index_builder::GPUSpatialIndexBuilder;
+    use arrow_array::RecordBatch;
+    use arrow_schema::{DataType, Field, SchemaRef};
+    use datafusion_common::JoinType;
+    use datafusion_physical_expr::expressions::Column;
+    use futures::Stream;
```
paleolimbot, line 272: For #702 we have to move all of these tests into `rust/sedona/tests` to avoid circular dev dependencies, where we can deduplicate the common pieces with sedona-spatial-join.

paleolimbot (commented):
Thank you!

All just details from me. Looking forward to having this merged!

rust/sedona-spatial-join-gpu/Cargo.toml
pwrliang: I have removed some of dependencies, but these are required as they are used by `SpatialIndex` trait

rust/sedona-spatial-join-gpu/src/index/gpu_spatial_index_builder.rs
pwrliang: This line estimates the size of temporary memory space for an array of bounding boxes, which will be fed to the GPU library to build a on-device index. The memory is allocated until the finishing of a spatial join.

pwrliang:
I have addressed most of the issues. I will run an end-to-end test on a machine with a GPU and post performance numbers on a subset of SpatialBench.

pwrliang:
Instance: g6.2xlarge (NVIDIA L4)
Spilling is disabled.

SF=1
| Query | sedonadb_gpu | datafusion batch size |
|-------|--------------|-----------------------|
| q2    | 0.43s        | 100K                  |
| q4    | 0.60s        | 100K                  |
| q6    | 0.67s        | 100K                  |
| q9    | 0.03s        | 100K                  |
| q10   | 2.53s        | 2M                    |
| q11   | 4.29s        | 2M                    |

SF=10

| Query | sedonadb_gpu | datafusion batch size |
|-------|--------------|-----------------------|
| q2    | 1.91s        | 100K                  |
| q4    | 1.23s        | 100K                  |
| q6    | 2.70s        | 100K                  |
| q9    | 0.12s        | 100K                  |
| q10   | 8.52s        | 2M                    |
| q11   | 25.51s       | 2M                    |

I noticed performance regressions compared to my previous implementation.

Screenshot 2026-04-12 at 21 00 20


After my investigation, the performance degradation is due to the introduction of `ProbeShuffleExec` node and the new implementation of `GpuSpatialIndexBuilder` with `EvaluatedGeometryArray::interleave`. The former can be addressed by setting `sedona.spatial_join.repartition_probe_side = false` and the latter can be improved by providing the `EvaluatedGeometryArray::concatenate()` method as suggested.

rust/sedona-spatial-join-gpu/src/join_provider.rs
paleolimbot: Can you add a comment referencing that this is an optimization for points and you're referring to some of this rounding occurring in the spatial index (or something else if I've misunderstood). As it is this comment doesn't make sense and somebody is likely to come through here and make it "more correct" if it is not explained properly.

rust/sedona-spatial-join-gpu/src/physical_planner.rs
```diff
@@ -0,0 +1,365 @@
...
+        let (physical_left, physical_right) = if args.join_options.repartition_probe_side {
+            repartition_probe_side(
+                args.physical_left.clone(),
+                args.physical_right.clone(),
+                args.spatial_predicate,
+                should_swap,
+            )?
+        } else {
+            (args.physical_left.clone(), args.physical_right.clone())
+        };
```
paleolimbot: You're in control of this code now, so maybe you just want to turn this off for the GPU join since it's slowing you down? You could then remove the partitioner below to keep your join planner a bit more focused.

rust/sedona-spatial-join-gpu/src/index/gpu_spatial_index_builder.rs
paleolimbot: Since you've taken the time to diagnose this as significant, it is worth opening an issue and adding a comment here in case neither of us remembers 🙂

rust/sedona-spatial-join-gpu/src/physical_planner.rs
```diff
@@ -0,0 +1,365 @@
...
+    Ok((physical_left, physical_right))
+}
+
+pub fn is_spatial_predicate_supported(
+    spatial_predicate: &SpatialPredicate,
+    left_schema: &Schema,
+    right_schema: &Schema,
+) -> Result<bool> {
+    fn is_geometry_type_supported(expr: &Arc<dyn PhysicalExpr>, schema: &Schema) -> Result<bool> {
+        let return_field = expr.return_field(schema)?;
```
paleolimbot, line 168: I don't think you need both `is_spatial_predicate_supported()` and `is_spatial_predicate_supported_on_gpu()`?

rust/sedona-spatial-join-gpu/src/physical_planner.rs
```diff
@@ -0,0 +1,365 @@
...
+    use super::*;
+    use arrow_schema::{DataType, Field};
+    use datafusion_physical_expr::expressions::Column;
+    use sedona_query_planner::spatial_predicate::{KNNPredicate, SpatialRelationType};
+    use sedona_schema::datatypes::{WKB_GEOGRAPHY, WKB_GEOMETRY};
+
+    #[test]
+    fn test_is_spatial_predicate_supported() {
+        // Planar geometry field
+        let geom_field = WKB_GEOMETRY.to_storage_field("geom", false).unwrap();
```
paleolimbot: I am not sure you need these (or if you do, perhaps tune them to your GPU-specific requirements?)

rust/sedona-spatial-join/src/exec.rs
```diff
@@ -128,12 +129,22 @@ impl SpatialJoinExec {
         join_type: &JoinType,
         projection: Option<Vec<usize>>,
         options: &SpatialJoinOptions,
+        join_provider: Arc<dyn SpatialJoinProvider>,
     ) -> Result<Self> {
```
paleolimbot: How about:

```rust
    pub fn with_spatial_join_provider(mut self, join_provider: Arc<dyn SpatialJoinProvider>) -> Self {
        self.join_provider = join_provider;
        self
    }
```

...to reduce the code changes for the common path of creating the default join and remove the clippy allow?

rust/sedona-spatial-join/src/join_provider.rs
```diff
@@ -66,7 +66,7 @@ pub trait SpatialJoinProvider: std::fmt::Debug + Send + Sync {
 
 /// Default implementation of the [SpatialJoinProvider]
 #[derive(Debug)]
-pub(crate) struct DefaultSpatialJoinProvider;
+pub struct DefaultSpatialJoinProvider;
```
paleolimbot: With the change I suggested above I think this can stay crate-local

rust/sedona-spatial-join-gpu/src/index/gpu_spatial_index_builder.rs
pwrliang: https://github.com/apache/sedona-db/issues/763

rust/sedona-spatial-join-gpu/src/physical_planner.rs
pwrliang: Sorry about that, I thought I had addressed this issue.

rust/sedona-spatial-join/src/exec.rs
pwrliang: Thank you, this is a very good suggestion.

rust/sedona-spatial-join-gpu/src/physical_planner.rs
pwrliang: Copy-paste them without thinking, I have removed them.

rust/sedona-spatial-join-gpu/src/physical_planner.rs
pwrliang: I have removed the this exec node

rust/sedona-spatial-join-gpu/src/join_provider.rs
pwrliang: Thanks for the suggestion. The following comments have been added
```rust
// This piece of code checks whether the underlying geometry is a point
// By representing the point with an MBR with the same min corner and max corner,
// libgpuspatial treats the MBR as a point, which triggers an optimized point query
// instead of using rect-rect query for high performance
// Ref: https://github.com/apache/sedona-db/blob/9187f8b8c4ca52b64837fab5fddd377703f7331b/c/sedona-libgpuspatial/libgpuspatial/src/rt_spatial_index.cu#L374
```

paleolimbot (approved):
Thank you!

PR #735, opened 2026-03-24
refactor(rust/sedona-query-planner): Move query planner and utilities to dedicated crate

paleolimbot:
This PR separates the planning component from the sedona-spatial-join crate into sedona-query-planner. There are a few things related to planning (e.g., optimizer rules, statistics propagation) that are in our future and a few things regarding joins in our future (GPU joins, Geography joins) and I would like to separate out the execution to make it easier to experiment without touching the internals of sedona-spatial-join (which has been carefully optimized!).

This is inspired by https://github.com/apache/sedona-db/pull/722#pullrequestreview-3983152094 but is a standalone improvement regardless, particularly so that there's an obvious place to put other optimizer rules and analyzer code. I'll stack a PR on top of this to show how it will help more cleanly separate the GPU join, too.


rust/sedona-query-planner/src/spatial_join_factory.rs
```diff
@@ -0,0 +1,249 @@
...
+/// Implementations decide whether they can handle a given spatial predicate and,
+/// if so, produce an appropriate [`ExecutionPlan`].
+pub trait SpatialJoinFactory: std::fmt::Debug + Send + Sync {
+    /// Given the provided arguments, produce an [ExecutionPlan] implementing
+    /// the join operation or `None` if this implementation cannot execute the join
+    fn plan_spatial_join(
+        &self,
+        args: &PlanSpatialJoinArgs<'_>,
+    ) -> Result<Option<Arc<dyn ExecutionPlan>>>;
+}
```
paleolimbot: This is the new piece of code that allows the query planner to live here (with the default join injected in the sedona crate when we construct the session state).

rust/sedona-spatial-join/src/factory.rs
```diff
@@ -0,0 +1,314 @@
...
+            &args.physical_right.schema(),
+        )? {
+            return Ok(None);
+        }
+
+        let should_swap = !matches!(
+            args.spatial_predicate,
+            SpatialPredicate::KNearestNeighbors(_)
+        ) && args.join_type.supports_swap()
+            && should_swap_join_order(args.physical_left.as_ref(), args.physical_right.as_ref())?;
```
paleolimbot: This is the implementation of the factory for the default join (the only modification here is using `args`...previously this was `plan_extension()`).

rust/sedona/src/context.rs
```diff
@@ -143,11 +146,17 @@ impl SedonaContext {
...
         #[cfg(feature = "spatial-join")]
         {
-            state_builder = sedona_spatial_join::register_planner(state_builder)?;
+            use sedona_spatial_join::factory::DefaultSpatialJoinFactory;
+
+            planner = planner.with_spatial_join_factory(Arc::new(DefaultSpatialJoinFactory::new()));
         }
 
+        state_builder = register_spatial_join_logical_optimizer(state_builder)?;
+        state_builder = state_builder.with_query_planner(Arc::new(planner));
```
paleolimbot, line 160: This is how it is used when creating the session

rust/sedona-query-planner/src/spatial_join_factory.rs
Kontinuation: I suggest renaming it to `SpatialJoinPhysicalPlanner` or `SpatialJoinPlanner` since the main function of this trait is producing physical plans. `SpatialJoinFactory` is a bit vague since there are lots of components related to spatial join.

Kontinuation (approved):

rust/sedona-spatial-join/src/index/spatial_index.rs
pwrliang: `#[cfg(test)]` does not work cross-crate. We may resort to `#[allow(dead_code)]`

rust/sedona-spatial-join/src/index.rs
pwrliang, line 25: Shall we also change the visibility to pub for `spatial_index` and `spatial_index_builder` as `SpatialIndex` trait has been declared as `pub` in this PR?

rust/sedona-spatial-join/src/lib.rs
```diff
@@ -18,25 +18,25 @@
 pub mod evaluated_batch;
 pub mod exec;
 mod index;
```
pwrliang: We may expose `index` to `pub`

rust/sedona-spatial-join/src/index/spatial_index_builder.rs
pwrliang: We may change the visiblity to `pub` or add some methods to update these variables, otherwise we cannot use `SpatialJoinBuildMetrics` out of the crate.

pwrliang (commented):
Thanks for this PR. I have some issues with visibility. I have annotated them, not sure whether there're more.

rust/sedona-spatial-join/src/index/spatial_index.rs
paleolimbot: I just removed it (now that it's pub there's no warning!)

PR #737, opened 2026-03-26
refactor(rust/sedona-spatial-join): Allow SpatialIndexBuilder and EvaluatedGeometryArray to be configurable

paleolimbot:
This PR is primarily in preparation for a Geography join, although it also extracts the configurable pieces required for a GPU join as well. In combination with #735 this (hopefully with tweaks) allows both the GPU join and the Geography join to be implemented in separate crates to minimize the disruption to sedona-spatial-join and unify the code pathways between all three of the default join, the GPU join, and the Geography join (and should in theory enable all of them to work with the partitioning approach for larger than memory input if I'm reading the code correctly).

The key piece that the Geography join requires is the ability to override the Cartesian bounding performed in `EvaluatedGeometryArray::try_new()`. Other than that I think we could use the default index at least to get started. The GPU join definitely needs to override the SpatialIndex and SpatialIndexBuilder so I also included this change. There are probably some details I'm missing for both but hopefully this takes care of most of the disruption.

I did this by introducing a `trait SpatialJoinProvider` and providing a default implementation such that all of the current spatial join pathways run through it (I'll highlight this inline shortly). The part that could especially use review is the part where I add the `Rect`s to the spilled batches. This was primarily to avoid having pass the `SpatialJoinProvider` through the machinery that reads the spilled batches; however, it also avoids unnecessarily recomputing bounds (which is probably great for large polygons but maybe not so great for points).

### DIFF
--- a/rust/sedona-spatial-join/src/evaluated_batch.rs
+++ b/rust/sedona-spatial-join/src/evaluated_batch.rs
@@ -19,7 +19,6 @@ use arrow_array::RecordBatch;
 use arrow_schema::SchemaRef;
 use datafusion_common::Result;
 use datafusion_expr::ColumnarValue;
-use geo::Rect;
 use wkb::reader::Wkb;
 
 use crate::{
@@ -61,12 +60,8 @@ impl EvaluatedBatch {
         wkbs[idx].as_ref()
     }
 
-    pub fn rects(&self) -> &Vec<Option<Rect<f32>>> {
-        &self.geom_array.rects
-    }
-
     pub fn distance(&self) -> &Option<ColumnarValue> {
-        &self.geom_array.distance
+        self.geom_array.distance()
     }
 }
 
--- a/rust/sedona-spatial-join/src/evaluated_batch.rs
+++ b/rust/sedona-spatial-join/src/evaluated_batch.rs
@@ -18,8 +18,6 @@
 use arrow_array::RecordBatch;
 use arrow_schema::SchemaRef;
 use datafusion_common::Result;
-use datafusion_expr::ColumnarValue;
-use wkb::reader::Wkb;
 
 use crate::{
     operand_evaluator::EvaluatedGeometryArray, utils::arrow_utils::get_record_batch_memory_size,
@@ -53,15 +51,6 @@ impl EvaluatedBatch {
 
     pub fn num_rows(&self) -> usize {
         self.batch.num_rows()
-    }
-
-    pub fn wkb(&self, idx: usize) -> Option<&Wkb<'_>> {
-        let wkbs = self.geom_array.wkbs();
-        wkbs[idx].as_ref()
-    }
-
-    pub fn distance(&self) -> &Option<ColumnarValue> {
-        self.geom_array.distance()
     }
 }
 
--- a/rust/sedona-spatial-join/src/evaluated_batch/spill.rs
+++ b/rust/sedona-spatial-join/src/evaluated_batch/spill.rs
@@ -158,7 +158,7 @@ impl EvaluatedBatchSpillWriter {
                     rect.min().x(),
                     rect.min().y(),
                     rect.max().x(),
-                    rect.max().y,
+                    rect.max().y(),
                 ]);
                 rect_builder.append(true);
             } else {
--- a/rust/sedona-spatial-join/src/evaluated_batch/spill.rs
+++ b/rust/sedona-spatial-join/src/evaluated_batch/spill.rs
@@ -250,6 +250,8 @@ pub(crate) fn spilled_batch_to_evaluated_batch(
         None
     };
 
+    // Mark
+
     // Create EvaluatedGeometryArray
     let mut geom_array = EvaluatedGeometryArray::try_new(geom_array, &sedona_type)?;
     geom_array.distance = distance;
--- a/rust/sedona-spatial-join/src/evaluated_batch/spill.rs
+++ b/rust/sedona-spatial-join/src/evaluated_batch/spill.rs
@@ -291,7 +291,7 @@ pub(crate) fn spilled_batch_to_evaluated_batch(
         .as_any()
         .downcast_ref::<Float32Array>()
         .ok_or_else(|| {
-            sedona_internal_datafusion_err!("Expected rect column to be FixedSizeListArray")
+            sedona_internal_datafusion_err!("Expected rect column child to be Float32Array")
         })?;
     let rect_vec = (0..rect_array.len())
         .map(|i| {
--- a/rust/sedona-spatial-join/src/evaluated_batch/spill.rs
+++ b/rust/sedona-spatial-join/src/evaluated_batch/spill.rs
@@ -123,7 +123,7 @@ impl EvaluatedBatchSpillWriter {
         // Store dist into a Float64Array
         let mut dist_builder = arrow::array::Float64Builder::with_capacity(num_rows);
         let geom_array = &evaluated_batch.geom_array;
-        match &geom_array.distance {
+        match geom_array.distance() {
             Some(ColumnarValue::Scalar(scalar)) => match scalar {
                 ScalarValue::Float64(dist_value) => {
                     for _ in 0..num_rows {
@@ -152,7 +152,7 @@ impl EvaluatedBatchSpillWriter {
         // Store rect into a FixedSizeList array
         let mut rect_builder =
             arrow::array::FixedSizeListBuilder::new(arrow::array::Float32Builder::new(), 4);
-        for rect_opt in &geom_array.rects {
+        for rect_opt in geom_array.rects() {
             if let Some(rect) = rect_opt {
                 rect_builder.values().append_slice(&[
                     rect.min().x(),
@@ -171,7 +171,7 @@ impl EvaluatedBatchSpillWriter {
         // Assemble the final spilled RecordBatch
         let columns = vec![
             Arc::new(data_struct_array) as ArrayRef,
-            Arc::clone(&geom_array.geometry_array),
+            Arc::clone(geom_array.geometry_array()),
             Arc::new(dist_array) as ArrayRef,
             Arc::new(rect_array) as ArrayRef,
         ];
@@ -315,9 +315,9 @@ pub(crate) fn spilled_batch_to_evaluated_batch(
         .collect::<Vec<_>>();
 
     // Create EvaluatedGeometryArray
-    let mut geom_array =
-        EvaluatedGeometryArray::try_new_with_rects(geom_array, rect_vec, &sedona_type)?;
-    geom_array.distance = distance;
+    let geom_array =
+        EvaluatedGeometryArray::try_new_with_rects(geom_array, rect_vec, &sedona_type)?
+            .with_distance(distance);
 
     Ok(EvaluatedBatch { batch, geom_array })
 }
@@ -393,10 +393,11 @@ mod tests {
     fn create_test_evaluated_batch() -> Result<EvaluatedBatch> {
         let batch = create_test_record_batch()?;
         let (geom_array, sedona_type) = create_test_geometry_array()?;
-        let mut geom_array = EvaluatedGeometryArray::try_new(geom_array, &sedona_type)?;
 
-        // Add distance as a scalar value
-        geom_array.distance = Some(ColumnarValue::Scalar(ScalarValue::Float64(Some(10.0))));
+        // With distance as ScalarValue
+        let geom_array = EvaluatedGeometryArray::try_new(geom_array, &sedona_type)?.with_distance(
+            Some(ColumnarValue::Scalar(ScalarValue::Float64(Some(10.0)))),
+        );
 
         Ok(EvaluatedBatch { batch, geom_array })
     }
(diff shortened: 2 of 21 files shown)


rust/sedona-spatial-join/src/join_provider.rs
```diff
@@ -0,0 +1,95 @@
...
+    /// Create a new [EvaluatedGeometryArray]
+    ///
+    /// Use [EvaluatedGeometryArray::try_new_with_rects] for custom computation of
+    /// bounding rectangles for specific items.
+    fn try_new_evaluated_array(
+        &self,
+        geometry_array: ArrayRef,
+        sedona_type: &SedonaType,
+    ) -> Result<EvaluatedGeometryArray>;
+}
```
paleolimbot, line 65: This is the extension point that is the primary motivation for this PR. Implementing it in this way allows all of the existing join pathways to run through this trait to reduce the test responsibility for custom joins.

rust/sedona-spatial-join/src/partitioning/stream_repartitioner.rs
```diff
@@ -645,115 +643,14 @@ pub(crate) fn interleave_evaluated_batch(
         return sedona_internal_err!("interleave_evaluated_batch requires at least one batch");
     }
     let batch = interleave_record_batch(record_batches, indices)?;
-    let geom_array = interleave_geometry_array(geom_arrays, indices)?;
+    let geom_array = EvaluatedGeometryArray::interleave(geom_arrays, indices)?;
```
paleolimbot, line 646: I did this to avoid having to pipe through the `SpatialJoinEvaluator` here; however, I think it is also positive (the rectangles from the input EvaluatedGeometryArray are used instead of being recomputed).

rust/sedona-spatial-join/src/index/spatial_index_builder.rs
```diff
@@ -41,8 +42,10 @@ pub(crate) trait SpatialIndexBuilder {
         options: &SpatialJoinOptions,
     ) -> usize;
 
+    fn operand_evaluator(&self) -> Arc<dyn OperandEvaluator>;
```
paleolimbot: I needed this because the OperandEvaluator is one of the places where EvaluatedGeometryArrays are created and this eliminated one more place where I had to pipe through the SpatialJoinProvider

rust/sedona-spatial-join/src/evaluated_batch/spill.rs
```diff
@@ -70,7 +73,12 @@ impl EvaluatedBatchSpillWriter {
             Field::new("data", DataType::Struct(data_inner_fields.clone()), false);
         let geom_field = sedona_type.to_storage_field("geom", true)?;
         let dist_field = Field::new("dist", DataType::Float64, true);
-        let spill_schema = Schema::new(vec![data_struct_field, geom_field, dist_field]);
+        let rect_field = Field::new(
+            "rect",
+            DataType::new_fixed_size_list(DataType::Float32, 4, true),
+            true,
+        );
+        let spill_schema = Schema::new(vec![data_struct_field, geom_field, dist_field, rect_field]);
```
paleolimbot, line 81: Spilling the rectangles here was primarily to avoid having to pipe the SpatialIndexProvider through the machinery that reads the batches, which is a verbose change. This is potentially faster because it avoid recomputing bounds and potentially slower because it increases the size of the spilled batch.

rust/sedona-spatial-join/src/partitioning/stream_repartitioner.rs
```diff
@@ -645,115 +643,14 @@ pub(crate) fn interleave_evaluated_batch(
...
     Ok(EvaluatedBatch { batch, geom_array })
 }
 
-fn interleave_geometry_array(
-    geom_arrays: &[&EvaluatedGeometryArray],
-    indices: &[(usize, usize)],
-) -> Result<EvaluatedGeometryArray> {
-    if geom_arrays.is_empty() {
-        return sedona_internal_err!("interleave_geometry_array requires at least one batch");
-    }
```
paleolimbot, line 658: This was just moved to EvaluatedGeometryArray (with its tests)

rust/sedona-spatial-join/src/prepare.rs
pwrliang: We may use `SpatialJoinProvider` to create an instance of `SpatialIndexBuilder`

rust/sedona-spatial-join/src/index/spatial_index_builder.rs
pwrliang, line 42: We should change the visibility of `SpatialJoinBuildMetrics` to `pub`, so we can use `SpatialJoinProvider` out of the crate

rust/sedona-spatial-join/src/join_provider.rs
```diff
@@ -0,0 +1,95 @@
...
+    SpatialPredicate,
+};
+
+/// Provider for join internals
+///
+/// This trait provides an extension point for overriding the evaluation
+/// details of a spatial join. In particular it allows plugging in a custom
+/// index for accelerated joins on specific hardware (e.g., GPU) and a custom
+/// bounder for specific data types (e.g., geography).
+pub(crate) trait SpatialJoinProvider: std::fmt::Debug + Send + Sync {
```
pwrliang, line 41: Can we change the visibility of `SpatialJoinProvider`, so the GPU join can exist in another crate?

pwrliang (commented):
Thanks for the PR. The pluggable design for the spatial-join module is elegant. I have a few questions about the visibility of `SpatialJoinProvider` and the instantiation of `SpatialJoinBuilder`, but otherwise, everything looks good. I'm looking forward to get this PR merged.

rust/sedona-spatial-join/src/prepare.rs
paleolimbot: Good catch! 😬

rust/sedona-spatial-join/src/index/spatial_index_builder.rs
paleolimbot: I agree! I didn't do that here because it causes a huge conflict with https://github.com/apache/sedona-db/pull/735. When that PR is rebased on main after this one, we can update the visibilities.

rust/sedona-spatial-join/src/join_provider.rs
2010YOUY01: I have a question: why does promoting a low-level API help testing? A more common approach is to maintain end-to-end behavioral tests that can be reused across different specialized join implementations.

(I might be mistaken for this part, as I’ve only skimmed the code.) The existing spatial join does not seem to rely directly on this primitive. Instead, it uses `OperandEvaluator` to build `EvaluatedGeometryArray`. For predicates like spatial distance, one side’s geometries are expanded when creating `EvaluatedGeometryArray` so that the R-tree probe logic remains consistent across different spatial predicates.

As a result, this proposed API does not appear to be used directly in the implementation. Is it still useful for testing?

2010YOUY01: Another point, fully from a software engineering perspective: I’m wondering if the responsibility of this `SpatialJoinProvider` should be limited to providing a unified interface for creating `SpatialIndexBuilder`. The `try_new_evaluated_array()` API seems more like a spatial-index-related utility, so placing it within `SpatialIndexProvider` might make it easier to maintain and avoid scattering similar functionality across different modules.

Kontinuation:
> @Kontinuation if you have bandwidth, do you have an opinion on whether or not it is worth threading the `SpatialJoinProvider` through the spill reader machinery is worth it? (It's a bit ugly but can be done). (Also any critiques of the approach are very welcome!)
> Details

I think the performance drop for compressed spilling is tolerable considering the benefit of less dependency on other components. I also wonder will writing bounding boxes to spill files always be better when working with geography data, since the equibox of geography could be computationally intensive.

I also wonder is it overly complex to support geography by implementing a new `SpatialJoinProvider`. We know the sedona_type of the geometries in the `EvaluatedGeometryArray`, so we can support different ways to evaluate bounding boxes within `EvaluatedGeometryArray` itself, and even serialize the spilled batches in its mostly optimized form (with or without rects) according to the value of sedona_type.

rust/sedona-spatial-join/src/join_provider.rs
paleolimbot: > I have a question: why does promoting a low-level API help testing?

If we have to maintain a "default join" and "custom join" pathway, we have add tests for the "custom join" pathway. Here they're essentially both running through a "custom join" pathway, which doesn't remove the need for integration tests for the GPU and Geography joins, but does in theory limit their low-level testing responsibility to building and querying the index. It's not a perfect system but in the previous PR implementing the join there was `#[cfg(feat = "gpu")]` sprinkled throughout various pieces of sedona-spatial join and it wasn't clear which ones of those were tested.

>  A more common approach is to maintain end-to-end behavioral tests that can be reused across different specialized join implementations.

We can/will do this too. We have to move some of the tests here into `rust/sedona/tests` anyway because we have a circular dependency problem at the moment (https://github.com/apache/sedona-db/issues/702). We also have some in Python that can be parameterized.

> The try_new_evaluated_array() API seems more like a spatial-index-related utility

That's a great point...I think this can be moved to the SpatialIndexBuilder for this PR. I am not 100% on exactly what we need to do to make the evaluatedgeometryarray work with Geography yet...it may be this can be isolated to the DefaultSpatialIndex with some more (possibly disruptive) effort.

rust/sedona-spatial-join/src/index/default_spatial_index.rs
```diff
@@ -299,37 +292,6 @@ impl SpatialIndex for DefaultSpatialIndex {
         &self.inner.indexed_batches[batch_idx].batch
     }
 
-    /// This method implements [`SpatialIndex::query`], which is a two-phase spatial join:
-    /// 1. **Filter phase**: Uses the R-tree index with the probe geometry's bounding rectangle
-    ///    to quickly identify candidate geometries that might satisfy the spatial predicate
-    /// 2. **Refinement phase**: Evaluates the exact spatial predicate on candidates to determine
-    ///    actual matches
-    fn query(
-        &self,
```
paleolimbot, line 308: This wasn't used (only `query_batch()` was), so I removed it

rust/sedona-spatial-join/src/index/spatial_index_builder.rs
```diff
@@ -16,38 +16,25 @@
...
-pub(crate) trait SpatialIndexBuilder {
-    /// Estimate the amount of memory required by the R-tree index and evaluating spatial predicates.
-    /// The estimated memory usage does not include the memory required for holding the build side
-    /// batches.
-    fn estimate_extra_memory_usage(
-        &self,
-        geo_stats: &GeoStatistics,
-        spatial_predicate: &SpatialPredicate,
-        options: &SpatialJoinOptions,
-    ) -> usize;
```
paleolimbot, line 42: I moved this to the `SpatialJoinProvider` because we don't need to instantiate an actual ready-to-go builder just to estimate hypothetical memory usage

rust/sedona-spatial-join/src/evaluated_batch.rs
```diff
@@ -55,19 +52,6 @@ impl EvaluatedBatch {
...
-        wkbs[idx].as_ref()
-    }
-
-    pub fn rects(&self) -> &Vec<Option<Rect<f32>>> {
-        &self.geom_array.rects
-    }
-
-    pub fn distance(&self) -> &Option<ColumnarValue> {
-        &self.geom_array.distance
-    }
```
paleolimbot, line 70: I moved these to the `EvaluatedGeometryArray` as part of the process of figuring out the exact responsibility of these and mediating field access of the evaluated array behind methods.

rust/sedona-spatial-join/src/operand_evaluator.rs
```diff
@@ -30,38 +31,53 @@ use sedona_geo_generic_alg::BoundingRect;
...
+/// expand to support more compact or efficient serialization/deserialization of the
+/// evaluated array when spilling.
+pub(crate) trait EvaluatedGeometryArrayFactory: fmt::Debug + Send + Sync {
+    /// Create a new [EvaluatedGeometryArray]
+    fn try_new_evaluated_array(
+        &self,
+        geometry_array: ArrayRef,
+        sedona_type: &SedonaType,
+    ) -> Result<EvaluatedGeometryArray>;
+}
```
paleolimbot, line 55: I added this because it most precisely encapsulates what custom joins need to control. SpatialIndexBuilders and SpatialIndexes ultimately consume EvaluatedGeometryArrays that are created by other parts of the machinery. What exactly "eagerly evaluated" looks like depends on what the index will ultimately need.

I think there are some improvements possible here by better encapsulating what questions are being asked of this object by non-SpatialIndex internals (e.g., update the spatial analyzer stats, repartition, serialize to record batch, deserialize from record batch) but that is a future project once we have a working Geography and GPU join with the current structure.

rust/sedona-spatial-join/src/operand_evaluator.rs
```diff
@@ -73,32 +89,34 @@ pub(crate) trait OperandEvaluator: fmt::Debug + Send + Sync {
...
     /// The array of geometries produced by evaluating the geometry expression.
-    pub geometry_array: ArrayRef,
+    geometry_array: ArrayRef,
     /// The rects of the geometries in the geometry array. The length of this array is equal to the number of geometries.
     /// The rects will be None for empty or null geometries.
-    pub rects: Vec<Option<Rect<f32>>>,
+    rects: Vec<Option<Rect<f32>>>,
     /// The distance value produced by evaluating the distance expression.
-    pub distance: Option<ColumnarValue>,
+    distance: Option<ColumnarValue>,
```
paleolimbot, line 119: I made these all accessible by methods to better track places that these are fetched or set

paleolimbot (commented):
I pushed a little harder with the `OperandEvaluator` and `EvaluatedGeometryArray` to better tease out the extendable pieces. I left some inline comments on specifics...I do think the result is a little cleaner and makes it more clear what is going on. SpatialJoinProvider is one level above the builder: it can estimate memory and creates the EvaluatedGeometryArrayFactory, which is used by separate internals of the spatial join but whose arrays are ultimately passed to the builder and index.

> I also wonder is it overly complex to support geography by implementing a new SpatialJoinProvider.

A geography join I do think can be implemented by a simply a EvaluatedGeometryArrayFactory, but the internals of the evaluated array will need some updating to accommodate wraparounds. The index seems ready to go though!

> The existing spatial join does not seem to rely directly on this primitive. Instead, it uses OperandEvaluator to build EvaluatedGeometryArray

This is primarily what I tweaked in the last set of changes...evaluating physical expressions is definitely internal to the join shuffling, but what "eagerly evaluated" is is index-specific (with the caveat that they are eagerly evaluated long before the index ever sees them, sometimes with a spill in between).

PR #747, opened 2026-03-31
test(rust/sedona-raster-functions): Add multi-band regression tests for RS_* raster functions

james-willis:
## Summary

- Adds `generate_multi_band_raster()` helper to `sedona-testing` — 3-band raster with heterogeneous types (UInt8/UInt16/Float32) and different nodata values
- Adds multi-band tests to existing test modules:
  - `rs_numbands`: verifies band count = 3
  - `rs_band_accessors`: verifies RS_BandPixelType and RS_BandNoDataValue return correct per-band values across heterogeneous bands

Pre-work for the N-Dimensional Raster Type Extension. The migration will replace the raster Arrow schema, traits, builder, and reader — these tests cover gaps in multi-band coverage so behavioral changes are caught.

## Test plan

- [x] `cargo test -p sedona-raster-functions` — 143 tests pass (140 existing + 3 new)
- [x] `cargo test -p sedona-testing test_generate_multi_band` — round-trip test passes
- [x] `cargo clippy` — no new warnings


rust/sedona-raster-functions/src/lib.rs
```diff
@@ -33,3 +33,6 @@ pub mod rs_size;
 pub mod rs_spatial_predicates;
 pub mod rs_srid;
 pub mod rs_worldcoordinate;
+
+#[cfg(test)]
+mod regression_tests;
```
paleolimbot: It's a bit more more rust-y to put these in `<crate>/tests` (we do this for some other crates as well)

paleolimbot (approved):
Thank you!

There's one CI fix needed (`cargo clippy --fix` should do it, or you can configure an editor like VSCOde to use cargo clippy as the linting command). Hopefully CI will run automatically for you after this merges.

PR #749, opened 2026-04-01
feat: N-dimensional raster type extension (Phase 1)

james-willis:
## Summary

Upgrades SedonaDB's raster type from 2D tiles to N-dimensional chunks with named dimensions, enabling support for climate model outputs, hyperspectral imagery, Zarr/NetCDF datacubes, and other multi-dimensional geospatial datasets.

This is a schema-and-API migration. All 33 existing `RS_*` functions are migrated mechanically and produce identical outputs on 2D inputs; no new SQL functions are added in this PR. GDAL-backed functions added in #787 are ported to read from the new schema and gate on `BandRef::is_2d()` — N-D bands return a clear `Execution` error pending MDArray/Zarr support.

This PR is rebased onto post-#787 `main` and is the **destructive replacement** of #787's band schema. Because the schema swap and the GDAL loader port have to land together to keep `main` compiling, the PR is structured as four review-friendly commits but **only the tip is required to be CI-green** — intermediate commits intentionally don't build and the PR is not bisectable across commits 1–3.

This PR depends on #812 (GDAL outdb-URI parser); merge order: #812 → #749.

> **Note on `utils.rs`** — the GDAL indb-loader helpers added in #811 (`append_as_indb_raster`, `dataset_to_indb_raster`) are deleted in this PR because they target the legacy `BandMetadata` / `StorageType` / `band_data_writer` API that the N-D port removes. They are reintroduced against the canonical N-D schema in the follow-up PR #818, which is already drafted and stacked on top of this branch.

## Commits

1. **Schema swap** — replace #787's band schema with the canonical N-D schema. Removes `nodata_value`, `storage_type`, `outdb_url`, `outdb_band_id`; adds `dim_names`, `source_shape`, nullable `view`, `outdb_uri`, `outdb_format`. Top-level gains `spatial_dims` / `spatial_shape`.
2. **Trait surface + `is_2d`** — `RasterRef` and `BandRef` accessors for the N-D schema; `BandRef::is_2d()` default method returns true iff `dim_names == ["y","x"]` with the identity view (used by GDAL-backed paths to refuse N-D input cleanly).
3. **Reader / builder / RS_*** **migration** — identity-view reader and builder, identity-view default via null `view` row, all 33 `RS_*` functions ported. `RS_BandPath` keeps its existing fragment-stripping (format-agnostic display) and is unchanged.
4. **Port GDAL loader to canonical schema** — rewrite `sedona-raster-gdal` to read from `outdb_uri` + the #812 parser instead of `storage_type` / `outdb_url` / `outdb_band_id`. VSI normalization, dataset cache, and RasterIO bodies are byte-for-byte unchanged — only the schema-read sites move. Each GDAL-backed SQL function gates on `band.is_2d()` at entry.

## Equivalence mapping (vs #787)

| #787 field | New schema |
| --- | --- |
| `storage_type` (InDb / OutDbRef enum) | `data.is_empty()` — InDb bands carry their bytes in `data`; OutDb bands have an empty `data` buffer and resolve through `outdb_uri` + `outdb_format`. |
| `outdb_url` | `outdb_uri` (no rename, same string) |
| `outdb_band_id` | encoded inside `outdb_uri` (`<uri>#band=N` or GDAL native subdataset URI). Parsed only inside the GDAL format driver via #812's `parse_outdb_source`. Format-agnostic surfaces (incl. `RS_BandPath`) treat `outdb_uri` as opaque. |
| `nodata_value` | `nodata: Binary` (typed bytes; a null row means "no nodata") |

Top-level adds `spatial_dims: List<Utf8>` and `spatial_shape: List<UInt64>` (e.g. `["y","x"]` and `[height, width]`).

## Schema (canonical)

```
raster: Struct<
  crs:           Utf8View (nullable),
  transform:     List<Float64>,            // 6-element GDAL GeoTransform
  spatial_dims:  List<Utf8View>,           // ["x","y"] today
  spatial_shape: List<Int64>,              // [width, height] today
  bands: List<Struct<
    name:         Utf8 (nullable),
    dim_names:    List<Utf8>,              // 2D bands: ["y","x"]
    source_shape: List<UInt64>,            // C-order extent of source
    data_type:    UInt32 (BandDataType discriminant),
    nodata:       Binary (nullable),
    view:         List<Struct<source_axis,start,step,steps: Int64>>
                  (nullable; null = canonical identity view),
    outdb_uri:    Utf8 (nullable),         // null = InDb; set = OutDbRef
    outdb_format: Utf8View (nullable),     // GDAL driver hint
    data:         BinaryView (non-nullable; empty for outdb-only)
  >>
>
```

The `view` field is defined in the schema so it is part of the canonical N-D layout and round-trips through Arrow IPC, but in this PR every writer emits a null view row (canonical identity) and the reader treats a non-null view row as a hard read error. Encoding identity as a null row keeps the common "no slice applied" case free in Arrow space; consumers see the same shape and bytes regardless of which path produced the band. The composition path that decodes non-null view rows lands in #813.

## Traits

- `RasterRef` — `spatial_dims`, `spatial_shape`, `transform`, `crs`, `num_bands`, `band(i)`, `width`/`height`/`x_dim`/`y_dim` (derived).
- `BandRef` — `name`, `dim_names`, `source_shape`, `shape` (visible, derived), `view`, `data_type`, `nodata`, `outdb_uri`, `outdb_format`, `nd_buffer`, `contiguous_data` returning `Cow<[u8]>`, `is_2d` (default).
- `NdBuffer` — zero-copy buffer view used at the numpy / Arrow C Data Interface boundary.

## Builder

- `start_raster` / `start_band` for full N-D; `start_raster_2d` / `start_band_2d` for legacy 2D.
- `start_band` writes a null view (canonical identity) — no caller change.
- `finish_band` validates each band's visible shape against the raster's `spatial_shape` along the spatial dims.

## Backward compatibility

Legacy 2D rasters are represented as bands with `dim_names = ["y","x"]`, `source_shape = [height, width]`, and an identity view. All existing `RS_*` outputs are byte-identical on 2D inputs; every test that passed against #787's GDAL functions passes against the ported loader.

## Crates modified

- `sedona-schema` — N-D raster schema definition.
- `sedona-raster` — traits, Arrow array reader, builder, affine transform, display.
- `sedona-testing` — raster test helpers.
- `sedona-raster-functions` — all `RS_*` function implementations migrated; `RS_BandPath` unchanged.
- `sedona-raster-gdal` — loader reads `outdb_uri` + #812 parser; SQL functions gate on `is_2d()`.

## Out of scope (follow-ups)

- **GDAL indb-loader port (#818)** — reintroduces `append_as_indb_raster` / `dataset_to_indb_raster` (deleted here, see note above) against the N-D schema.
- View machinery — `validate_view`, `start_band_with_view`, view-decoding in the reader, view-aware `contiguous_data`, and the explicit-view test cluster — landing in #813 between this PR and #750.
- Functions that produce non-identity views (slice/crop, broadcast, axis reorder, heterogeneous-source stack) — coming in #750 on top of #813.
- N-D-aware GDAL via MDArray (lifts the `is_2d()` guard for selected functions).
- Zarr / non-GDAL N-D backends.
- View-aware fast paths in arithmetic functions.

## Test plan

- [x] `sedona-schema`: tests pass; verifies the `view` field is nullable and the view-struct field order matches the indices module.
- [x] `sedona-raster`: tests pass — identity round-trip via `start_band` (null view row); on-disk schema invariant that identity bands serialise as null view rows; reader rejects non-null view rows with a clear error; `BandRef::is_2d` truth-table tests; builder visible-shape validation against `spatial_shape`.
- [x] `sedona-raster-functions`: tests pass — all existing function tests produce identical outputs after migration.
- [x] `sedona-raster-gdal`: 46/46 pass — every #787 SQL-function test passes against the ported loader; new tests cover N-D rejection at `raster_ref_to_gdal_mem` and the VRT path; one end-to-end test confirms `path#band=2` selects band 2 of a two-band GeoTIFF.
- [x] `cargo clippy --all-targets -- -D warnings` clean on the affected crates.
- [x] `cargo fmt --all --check` clean.


rust/sedona-raster-functions/src/rs_bandpath.rs
```diff
@@ -225,12 +218,18 @@ mod tests {
...
+            "geotiff://s3://bucket/raster_0.tif#band=1"
+        );
+        // Raster 1: null raster → null
         assert!(string_array.is_null(1));
-        // Raster 2, band 2: OutDbRef -> URL
-        assert_eq!(string_array.value(2), "s3://bucket/raster_2.tif");
+        // Raster 2, band 2: OutDbRef → URI
+        assert_eq!(
+            string_array.value(2),
+            "geotiff://s3://bucket/raster_2.tif#band=3"
```
james-willis: need to think through the right behavior here.

rust/sedona-schema/src/raster.rs
```diff
@@ -16,34 +16,33 @@
...
-            Field::new(column::SCALE_X, DataType::Float64, false),
-            Field::new(column::SCALE_Y, DataType::Float64, false),
-            Field::new(column::SKEW_X, DataType::Float64, false),
-            Field::new(column::SKEW_Y, DataType::Float64, false),
-        ]))
+    /// Affine transform schema — 6-element GDAL GeoTransform:
+    /// `[origin_x, scale_x, skew_x, origin_y, skew_y, scale_y]`
+    pub fn transform_type() -> DataType {
+        DataType::List(FieldRef::new(Field::new("item", DataType::Float64, false)))
     }
```
paleolimbot, line 54: Just highlighting for myself + others that this is the main change.

The main conceptual change here seems to be that the height and width in 2D space are now members of the bands. Even though it might involve some duplication, I wonder if a top-level shape (whose ordering is always x, y, and possibly z in some future) would still be appropriate. The bands would then all be parsed according to the x/y dim name and validated against the top-level declaration.

X_DIM + Y_DIM should probably also be a list to allow for future Z. Lists are kind of a pain in Rust but interacting with these in Python is fairly easy to do and this PR is probably the only one that needs to consider that detail.

rust/sedona-schema/src/raster.rs
```diff
@@ -55,32 +54,37 @@ impl RasterSchema {
...
-            Field::new(column::DATA, Self::band_data_type(), false),
+            Field::new(column::NAME, DataType::Utf8, true),
+            Field::new(column::DIM_NAMES, Self::dim_names_type(), false),
+            Field::new(column::SHAPE, Self::shape_type(), false),
+            Field::new(column::DATATYPE, DataType::UInt32, false),
+            Field::new(column::NODATA, DataType::Binary, true),
+            Field::new(column::STRIDES, Self::strides_type(), false),
+            Field::new(column::OFFSET, DataType::UInt64, false),
+            Field::new(column::OUTDB_URI, DataType::Utf8, true),
+            Field::new(column::DATA, DataType::BinaryView, false),
```
paleolimbot: Highlighting changes here to check my understanding...the main change is that bands have their own dimensions, with the addition of dim_names. The spatial dimensions of all the bands have to match. Here outdb_uri typically refers to a portion of a raster (e.g., COG tile) rather than a full array.

This also introduces strides/offset, which comes from pybuffer ( https://docs.python.org/3/c-api/buffer.html ) terminology. This allows some (but not all) subsets to be propagated without modifying `DATA` or `OUTDB_URI`. In particular, because OUTDB_URI doesn't have to be modified, we can delay loading tiles for longer.

I am a tiny bit worried that this won't be sufficient to avoid loading all of the things we want to avoid loading, but also we maybe don't have the full picture on what the subset specification looks like yet. Maybe this can be `subset: Utf8` and we can sort out the exact subset specifications we'll allow and how to resolve them later (e.g., GDAL has its own way to specify this as well).

paleolimbot (commented):
I'll take a closer look at the implementation tomorrow, but wanted to get a preliminary review of the schema out!

rust/sedona-raster/src/affine_transformation.rs
```diff
@@ -124,139 +132,123 @@ pub fn to_raster_coordinate(
...
-                height: 20,
-                upperleft_x: 100.0,
-                upperleft_y: 200.0,
-                scale_x: 1.0,
-                scale_y: 0.0,
-                skew_x: 0.0,
-                skew_y: 0.0,
-            },
-        };
+        let raster = TestRaster::new(100.0, 200.0, 1.0, -2.0, 0.25, 0.5);
```
paleolimbot: For what it's worth the named struct construction here is more readable than the revised `TestRaster::new()`

rust/sedona-raster/src/affine_transformation.rs
```diff
@@ -324,11 +301,6 @@ mod tests {
         };
         let result = a.inv_transform(0.0, 0.0);
         assert!(result.is_err());
-        assert!(result
-            .err()
-            .unwrap()
-            .to_string()
-            .contains("determinant is zero."));
```
paleolimbot, line 331: Just checking that this expectation was moved and not deleted

rust/sedona-raster/src/display.rs
```diff
@@ -113,51 +93,3 @@ impl fmt::Display for RasterDisplay<'_> {
...
 }
-
-#[cfg(test)]
-mod tests {
-    use super::*;
-    use crate::array::RasterStructArray;
-    use sedona_testing::rasters::generate_test_rasters;
-
-    #[test]
-    fn display_non_skewed_raster() {
```
paleolimbot, line 124: Are these tests moved to a different place or should they be re-added?

rust/sedona-raster/src/outdb_uri.rs
```diff
@@ -0,0 +1,133 @@
...
+/// or making the entire outdb_uri a JSON string for those formats.
+#[derive(Debug, PartialEq)]
+pub struct OutDbUri<'a> {
+    /// Loader scheme (e.g., "geotiff", "zarr")
+    pub scheme: &'a str,
+    /// External resource path (e.g., "s3://bucket/file.tif")
+    pub path: &'a str,
+    /// Loader-specific fragment (e.g., "band=1"), or None if absent
+    pub fragment: Option<&'a str>,
+}
```
paleolimbot: There's a sturcture for this already that we probably want to reuse (I think it's called the ObjectStoreUrl). It's based on the url crate I believe, which would also be an improvement over rolling our own here. I think this would give you the extra parameter information you'll need without having to maintain it here.

rust/sedona-testing/src/rasters.rs
```diff
@@ -292,8 +215,6 @@ pub fn generate_multi_band_raster() -> StructArray {
     builder.finish().unwrap()
 }
 
-/// Determine if this tile contains a corner of the overall grid and return its position
-/// Returns Some(position) if this tile contains a corner, None otherwise
```
paleolimbot, line 298: Is there a reason these comments were removed?

rust/sedona-testing/src/rasters.rs
```diff
@@ -578,65 +492,33 @@ mod tests {
...
+        let raster1 = rsa1.get(0).unwrap();
+        let raster2 = rsa2.get(0).unwrap();
         assert_raster_equal(&raster1, &raster2);
     }
 
     #[test]
-    fn test_generate_multi_band_raster() {
-        let struct_array = generate_multi_band_raster();
-        let raster_array = RasterStructArray::new(&struct_array);
-        assert_eq!(raster_array.len(), 1);
```
paleolimbot, line 606: Is this test no longer relevant?

paleolimbot (commented):
I did another round...there are a few places where tests or comments were removed that were still useful.

I do think it's worth iterating a tiny bit on the schema, and importantly, I think we should wait until @Kontinuation has finished the GDAL read portion of the previous set of raster refactoring because (1) this was a lot of work and Kristin's time is valuable and (2) these are important use cases to consider when making the new type work!

rust/sedona-schema/src/raster.rs
james-willis: in zarr, each variable (here band) can have any permutation of dimensions. so band 1 might be xyzt, 2 might be yzt, 3 might be xy

rust/sedona-raster/src/affine_transformation.rs
james-willis: Fixed — replaced with a `test_raster(origin_x, origin_y, scale_x, scale_y, skew_x, skew_y)` free function so the parameter names are visible at the call site.

rust/sedona-raster/src/affine_transformation.rs
james-willis: Confirmed — the expectation is preserved. Same test logic, just uses the new constructor.

rust/sedona-raster/src/display.rs
james-willis: Re-added all three tests and the doc comments.

rust/sedona-testing/src/rasters.rs
james-willis: No good reason — accidentally dropped during the rewrite. Restored.

rust/sedona-testing/src/rasters.rs
james-willis: Still relevant — the equivalent test is `test_raster_different_metadata` which now panics with "Raster transforms do not match" instead of "Raster upper left x does not match" (same data, different field name in the assertion).

rust/sedona-schema/src/raster.rs
james-willis: > X_DIM + Y_DIM should probably also be a list to allow for future Z. Lists are kind of a pain in Rust but interacting with these in Python is fairly easy to do and this PR is probably the only one that needs to consider that detail.

Sorry I'm not clear what you want me to do. Should I decompose this into 2 instances of some GeoTransformDimension type? 

I adopted this norm to match GDAL. I my assumption was that X and Y are somewhat privileged in the system. I hadn't given much thought to affine transformations outside of the xy plane.

rust/sedona-schema/src/raster.rs
james-willis: > The main change is that bands have their own dimensions, with the addition of dim_names. The spatial dimensions of all the bands have to match. 

Yes

> Here outdb_uri typically refers to a portion of a raster (e.g., COG tile) rather than a full array.

You're probably right. That's up to the reader implementation, e.g. the geotiff reader can implement reading tiles, whole arrays, or both.

> In particular, because OUTDB_URI doesn't have to be modified, we can delay loading tiles for longer.

I will think about this more before I respond. I need to think about how I intend for striding and outdb to interact. I think I had considered striding only an indb operation but that will add complexity to operations that dont bring the raster in but would benefit from stride-on-read

rust/sedona-raster/src/outdb_uri.rs
james-willis: I will check it out

james-willis: Instead of the composite scheme (`geotiff://s3://bucket/file.tif#band=1`), how about we split into two fields:

  - outdb_uri — standard resource URL that works directly with DataFusion's object store registry, e.g. `s3://bucket/file.tif#band=1`. The fragment carries driver-specific parameters (band index, chunk coords, etc.). These are client-side only so won't collide with query params that need to pass to the server (e.g. for s3 presigned URLs).
  - outdb_format— driver name ("geotiff", "zarr") for dispatch. Nullable — null means in-memory.

  This lets the loader look up pre-configured ObjectStore instances from the session via `ObjectStoreUrl::parse()`, giving us credential management and connection reuse for free. Each format driver parses its own fragment. The custom `parse_outdb_uri` goes away.

rust/sedona-schema/src/raster.rs
james-willis: I intend to solve this issue by having has-a relationships between BandRef implementations. Imagine the case:
```sql
SELECT RS_Width(RS_Downsample(RS_Rechunk(rasterCol))) FROM zarr_table
```

We never need the actual pixels read here so if we can keep this all lazy it'll be a big win.

The `rasterCol` would be composed of RasterRefs that contains lazy ZarrBandRefs.

### RS_Rechunk 
two possibilities:
1. The simple lazy approach: a lazy StridedBandRef encodes the subsetting and points to the original ZarrBandRef. This reference to the original ZarrBandRef is used when `nd_buffer` or `contiguous_data` is called.
2. A new ZarrBandRef that encodes the subsetting bounds into the outdb_uri. This assumes the reader supports subsetting.

### RS_Downsamples
this one changes the dimensions of the dataset. obviously an eager implementation is pretty simple, but the composition approach allows us to continue to be lazy.

A CoursenBandRef implements `shape` from the metadata of its child BandRef. It would only perform an actual coarsen if a data retrieval method is called. In the coarsen case, being lazy with the child BandRef also can reduce peak memory consumption because if the child can stream the input data, we never hold both the original and result in memory at once.

### RS_Width
This is trivial - it returns a scalar so all it does is call the `shape` method(s) of the input BandRef.

## Limitations
This stacking only works when theres no serialization between the functions (e.g. a shuffle) we are stacking together. serialization from this composition solution is lossy. 

When a shuffle occurs we need to compress all the stacking down into something that can be represented in the arrow metadata. I think these cases are fairly limited and can be further reduces by using Deferred Expressions (Dewey's idea), which is an optimization we can do sometime later.

james-willis: Thinking about this more, I need to correct something in my earlier stacking comment. I said stacking breaks at *shuffles*, but that was too optimistic — in DataFusion, stacking breaks at **every UDF boundary**,
  shuffle or not. Every scalar UDF receives `ColumnarValue` (Arrow) and returns `ColumnarValue`, so `Arc<dyn BandRef>` wrappers can't live across `RS_A(RS_B(...))` without extra machinery. The stacking model is
  still the right *internal* data model — it's what lets a single kernel compose lazy views — but "stacking survives normal UDF chaining for free" was wrong.

  That correction actually makes the strides/offset story cleaner, because now it's the same question in both places: *how do we compress a stack into something Arrow can carry to the next UDF?*

  **View-producing stacks compress losslessly.** Any composition of `StridedBandRef`-shaped operations — `RS_Slice`, `RS_SliceRange`, `RS_DimToBand`, transposes, constant broadcasts (stride=0) — collapses exactly into four integers (strides, offset, shape) plus an unmodified `outdb_uri`. So `RS_Slice(RS_DimToBand(zarr_raster, 'wavelength'), 'time', 0)` serializes to the next UDF with `outdb_uri` preserved, strides/offset/shape updated, empty `data`. Full laziness survives the Arrow round-trip; the downstream `ZarrBandRef` reconstitutes on read. This is exactly what strides/offset buys us, and it's why I want to keep those fields as typed ints rather than a subset string.

  **Compute-producing stacks compress lossily.** The moment a `CoarsenBandRef` (or Normalize, MapAlgebra, Resample) is in the stack, we're producing new pixel values — there's no four-integer view that represents "coarsened." At the next UDF boundary it has to materialize. Under Phase 1 that means `RS_Width(RS_Coarsen(r, 2))` pays full materialization cost even though `Width` only needs `shape`. A `subset: Utf8` doesn't help here either — a subset DSL can encode ranges/strides/index lists, but not "coarsened."

  **The clean future-work answer is operator fusion at plan time.** An `ExprRewriter` rule detects chains like `RS_Width(RS_Coarsen(RS_Rechunk(r)))` and rewrites them to a single `RasterPipeline` meta-UDF. Inside
  one kernel, the stacking model works exactly as I first described — `StridedBandRef` → `CoarsenBandRef` → `.shape()` — and Arrow serialization only happens at the outermost boundary. Fusion is what makes a
  multi-UDF chain behave like the single-UDF-internal case where stacking was always going to work. No schema change, no trait change; can land whenever the compute-chain pattern shows up as a real bottleneck.

  So the revised picture:

  - Phase 1 as drafted: view-producing stacks stay lazy across UDF boundaries via strides/offset + preserved `outdb_uri`. Compute-producing stacks materialize at the boundary.
  - Future: operator fusion lets compute-producing stacks stay lazy too, without changing the schema we land now.

  On `subset: Utf8` specifically — tempting, but the trade-offs point the other way:

  - Parser/validator surface: any DSL needs parsing, error messages, docs, version stability. Strides/offset are four integers.
  - No laziness advantage over strides/offset for view cases, no help at all for compute cases — the real axis is fusion, not subset expressiveness.
  - Interop: strides/offset is the pybuffer / numpy / Arrow FFI vocabulary — a numpy consumer memory-maps our buffer directly. A DSL needs translation.
  - The one thing strides/offset can't express is non-contiguous index lists (e.g. `y ∈ {2, 3, 5}`); I'd handle that with materialize-then-stride when it comes up.

  If we hit a case strides/offset can't express *and* fusion can't handle, we can revisit. But for Phase 1 I'd rather keep the schema narrow and treat fusion as deferred work

rust/sedona-schema/src/raster.rs
paleolimbot: > Sorry I'm not clear what you want me to do. 

- Change `Field::new(column::X_DIM, DataType::Utf8View, false), Field::new(column::Y_DIM, DataType::Utf8View, false)` to  `Field::new(column::DIMS, DataType::new_list(DataType::Utf8View), false)`. This be of length 2 for now and you don't have to change your Rust wrapper.
- Your Arrow representation of the transform is already Z-ready (it would have 12 values instead of 6 in the future where Z is supported).
- Add `Field::new(column::SHAPE, DataType::new_list(DataType::Int64), false)`, which is the number of values in each of `column::DIMS`, in the same order. I may be reading it wrong but I think you currently have to peek into the first band for this. I think it is clearer to keep the spatial source of truth at the raster level and then validate every band against that (the transform and the height/width are very closely related to each other and I think having them at the same level makes sense). Also having a zero-band raster is surprisingly useful (e.g., gdal uses this as a way to specify a target grid to use for the result of an operation like warp).

rust/sedona-schema/src/raster.rs
paleolimbot: Deferring fusing of operators to a future phase is a great idea.

Concretely, how about removing `STRIDES` and `OFFSET` for now or making it clear they are for in-memory bands only? I think they are very useful and can avoid some copies for an in-memory band, but it's less clear what they mean for an outdb band because the exact remote layout of bytes isn't something the Arrow representation is really responsible for.

rust/sedona-raster/src/outdb_uri.rs
paleolimbot: > outdb_format— driver name ("geotiff", "zarr") for dispatch. Nullable — null means in-memory.

"format" works for me and I think is good given that it's also the datafusion terminology (the GDAL terminology for this would be "driver"). If we're dreaming big here you could fold in-memory into this (GDAL has the MEM driver; ObjectStore has a MemoryObjectStore, and there's no logical difference between contiguous remote bytes and continuous local bytes except how long it takes to get them).

paleolimbot: No need to go down the rabbit hole, but when the Scheme and the Format are separated it does maybe make a bit more sense (the GeoTiff format has tile and band as its parameters, the Zarr format has a chunk, memory has an offset and stride; all of those can exist in memory, as a local file, or as a remote file).

james-willis: "Format" terminology locked in.

  On the dream-big version — I want to take half of it now. The part that falls out for free if I use `ObjectStoreUrl` (per your other suggestion) is **scheme orthogonal to format**: `mem://` joins `s3://`,
  `file://`, `http://` as a valid scheme via DataFusion's `ObjectStoreRegistry`, so any format (GeoTIFF, Zarr, …) can live at any scheme including memory. Testing outdb paths without network becomes: register a
  `MemoryObjectStore` as `mem://` on session init, insert fixture bytes, reference them by `mem://fixture.tif` with `outdb_format="geotiff"`. A future caching tier (remote GeoTIFF → local mem copy) is just a URI
  rewrite, no schema change.

  The part I want to defer is making `"memory"` itself a *format*. That requires moving in-band bytes out of `data: BinaryView` into a ref-counted store with lifetime hooks tied to scan/batch cleanup — real work,
  not scoped to Phase 1. So in-memory bands keep `data: BinaryView` and leave `outdb_uri` / `outdb_format` null. `mem://` is only in the outdb path.

  Concrete plan for the follow-up commit:

  - Replace the custom `parse_outdb_uri` with `ObjectStoreUrl::parse()`.
  - Add `outdb_format: Utf8` (nullable — null means use `data` / `BinaryView`).
  - Document `outdb_uri` as accepting any scheme DataFusion's `ObjectStoreRegistry` knows about, explicitly calling out `mem://`.
  - No auto-registration of an `InMemory` store for now — session owners opt in if they want it.

paleolimbot: I think that misses the point...mem is definitely a format (just like how in GDAL it's a driver) referring to a contiguous array with parameters that control how its elements are organized into an ndarray (offsets/strides or the GDAL parameterization of that with line/pixel/band offsets).

(GDAL also has the concept of a memory store, which it calls vsimem).

The follow up plan is fine though (except calling out `mem://`).

james-willis: mem as a format and mem as a location are orthogonal concepts and functions.  so I'm saying I'm going to punt on mem as a format but implement mem as a location. 

mem as a format is more complex vs just having a binaryView field. do you feel that we should replace BinaryView with a mem format + mem location instead?

paleolimbot: I think that a NULL outdb_uri always means that the bytes are inlined into the BinaryView and a non-NULL outdb_uri means they have to be fetched (copied). The thing that may be useful to prevent future churn is to keep the parameters (band for geotiff, chunk for zarr, offsets/strides for mem) in the same place.

james-willis: > I think that a NULL outdb_uri always means that the bytes are inlined into the BinaryView and a non-NULL outdb_uri means they have to be fetched (copied).

I was thinking the opposite: null `data` field indicates you should go fetch data from outdb_uri. This allows us to leave outdb uri populated for cases where we want to have indb representation but give the engine the ability to evict the indb data, knowing it can be reconstructed from the uri + format. 

> The thing that may be useful to prevent future churn is to keep the parameters (band for geotiff, chunk for zarr, offsets/strides for mem) in the same place.

Are you saying you want a dedicated `format_parameters: Utf8` field? current plan is to put them in the URI fragment (the stuff after the #):

`https://website.com/some/dataset?websiteParam=websiteValue#driverParam=driverValue`

james-willis: still thinking about what to dowith strides/offset, will offer my thoughts on that later. trying to decide if theyre special or not

rust/sedona-raster-functions/src/rs_band_accessors.rs
```diff
@@ -120,14 +120,15 @@ fn get_pixel_type(
...
             if band_index < 1 || band_index > num_bands as i32 {
                 builder.append_null();
                 return Ok(());
             }
-            let band = raster.bands().band(band_index as usize)?;
-            let dt = band.metadata().data_type()?;
-            builder.append_value(dt.pixel_type_name());
+            match raster.band_data_type((band_index - 1) as usize) {
+                Some(dt) => builder.append_value(dt.pixel_type_name()),
+                None => builder.append_null(),
```
james-willis: not totally sure if we should handle this branch or throw. _should_ generally be unreachable - we don't expect the arrow to contain datatype values that don't exist.

rust/sedona-raster-gdal/src/gdal_common.rs
```diff
@@ -176,21 +201,39 @@ pub unsafe fn raster_ref_to_gdal_mem<R: RasterRef + ?Sized>(
...
+            .map_err(|e| arrow_datafusion_err!(e))?;
+        let bytes: &[u8] = match &band_data {
+            Cow::Borrowed(b) => b,
+            Cow::Owned(_) => {
+                return exec_err!(
+                    "Internal: contiguous_data must be borrowed for is_2d bands; got owned"
+                );
+            }
+        };
+        let data_ptr = bytes.as_ptr();
```
james-willis: Note to self. This whole section and the contiguous data thing will need some thought from others and me.

rust/sedona-schema/src/raster.rs
james-willis: done: added a concept of views that will be fleshed out in a subsequent PR: each dimension has start, step (step size) and (number of ) steps. This lets us logically describe the crop and enables reversing dimensions and boardcasting.

The next PR after this will implement that functionality.

rust/sedona-raster/src/outdb_uri.rs
james-willis: we settled on start/step/steps for the interface for lazy cropping

rust/sedona-raster-functions/src/executor.rs
```diff
@@ -724,7 +733,7 @@ mod tests {
                 match raster_opt {
                     None => builder.append_null(),
                     Some(raster) => {
-                        let width = raster.metadata().width();
+                        let width = raster.width().unwrap();
```
paleolimbot: Probably `width()` should return a Result so non-test uses can do `raster.width()?` instead of unwrapping

rust/sedona-raster-functions/src/rs_band_accessors.rs
paleolimbot: `sedona_internal_err!()` is a good choice for violations of internal assumptions anywhere you can return a `Result<>` (better to know and get an issue report than a silent null)

rust/sedona-raster-functions/src/rs_band_accessors.rs
```diff
@@ -224,16 +225,19 @@ fn get_nodata_value(
...
-            let band = raster.bands().band(band_index as usize)?;
-            let band_meta = band.metadata();
-            match band_meta.nodata_value_as_f64()? {
-                None => builder.append_null(),
-                Some(val) => builder.append_value(val),
+            let idx = (band_index - 1) as usize;
+            match (raster.band_nodata(idx), raster.band_data_type(idx)) {
+                (Some(bytes), Some(dt)) => {
+                    let val = nodata_bytes_to_f64(bytes, &dt)
+                        .map_err(datafusion_common::DataFusionError::from)?;
```
paleolimbot: Can you make sure this is the version that errors for i64 and u64?

rust/sedona-raster-functions/src/rs_convexhull.rs
```diff
@@ -107,8 +107,18 @@ impl SedonaScalarKernel for RsConvexHull {
 /// of the raster in world coordinates. Due to skew/rotation in the affine
 /// transformation, each corner must be computed individually.
 fn write_convexhull_wkb(raster: &dyn RasterRef, out: &mut impl std::io::Write) -> Result<()> {
-    let width = raster.metadata().width() as i64;
-    let height = raster.metadata().height() as i64;
+    let Some(width) = raster.width() else {
+        return Err(DataFusionError::Execution(
+            "Raster has no spatial dimensions; cannot determine width".into(),
+        ));
+    };
```
paleolimbot: `height()` and `width()` should return `Result<>` so this error doesn't have to be repeated

rust/sedona-raster-functions/src/rs_envelope.rs
```diff
@@ -105,8 +105,18 @@ impl SedonaScalarKernel for RsEnvelope {
 /// derives the min/max X and Y to produce an axis-aligned bounding box.
 /// For skewed/rotated rasters, this differs from the convex hull.
 fn write_envelope_wkb(raster: &dyn RasterRef, out: &mut impl std::io::Write) -> Result<()> {
-    let width = raster.metadata().width() as i64;
-    let height = raster.metadata().height() as i64;
+    let Some(width) = raster.width() else {
+        return Err(DataFusionError::Execution(
+            "Raster has no spatial dimensions; cannot determine width".into(),
+        ));
+    };
```
paleolimbot: e.g., here

rust/sedona-raster-functions/src/rs_georeference.rs
```diff
@@ -158,13 +158,13 @@ fn format_georeference(
...
-            let skew_y = metadata.skew_y();
-            let upper_left_x = metadata.upper_left_x();
-            let upper_left_y = metadata.upper_left_y();
+            let t = raster.transform();
+            let scale_x = t[1];
+            let scale_y = t[5];
+            let skew_x = t[2];
+            let skew_y = t[4];
+            let upper_left_x = t[0];
+            let upper_left_y = t[3];
```
paleolimbot: This is safer in the original interface where the accessors are named.

It may result in less churn to keep `raster.metadata()`...if you run into specific functions where you loose performance because of slightly aggressive parsing you can address that there (I would be surprised if you did).

rust/sedona-raster-functions/src/rs_geotransform.rs
```diff
@@ -162,22 +162,18 @@ impl SedonaScalarKernel for RsGeoTransform {
             match raster_opt {
                 None => builder.append_null(),
                 Some(raster) => {
-                    let metadata = raster.metadata();
+                    let t = raster.transform();
```
paleolimbot: That would also eliminate this change

rust/sedona-raster-gdal/src/lib.rs
```diff
@@ -31,15 +31,10 @@ mod gdal_common;
 // Temporary until https://github.com/apache/sedona-db/issues/804 is resolved.
 #[allow(dead_code)]
 mod gdal_dataset_provider;
-
-mod utils;
-
-#[cfg(test)]
```
paleolimbot, line 37: You can probably leave these and remove them later when the functionality is actually replaced if needed. This should be less disruptive to the existing PRs.

rust/sedona-raster/Cargo.toml
```diff
@@ -40,3 +40,4 @@ sedona-schema = { workspace = true }
 [dev-dependencies]
 sedona-testing = { workspace = true }
 approx = { workspace = true }
+arrow-ipc = { workspace = true }
```
paleolimbot, line 44: Just double checking that you need IPC here?

rust/sedona-testing/src/rasters.rs
paleolimbot: I see, you just moved the test in the file.

rust/sedona-schema/src/raster.rs
```diff
@@ -16,34 +16,54 @@
...
+/// describing the visible region of that buffer.
+///
+/// `spatial_dims` + `spatial_shape` are the raster-level source of truth for
+/// the spatial grid — today length 2 (`["x","y"]`, `[width, height]`),
+/// Z-ready for a future 3D phase. All bands must contain every name in
+/// `spatial_dims` in their own `dim_names`, with the band's *visible* size
+/// for that dim matching `spatial_shape`.
+///
+/// Legacy 2D rasters are represented as bands with `dim_names=["y","x"]` and
+/// `source_shape=[height, width]`.
```
paleolimbot:
```suggestion
/// 2D rasters are represented as bands with `dim_names=["y","x"]` and
/// `source_shape=[height, width]`.
```

rust/sedona-schema/src/raster.rs
```diff
@@ -116,6 +166,23 @@ impl BandDataType {
...
+            1 => Some(BandDataType::UInt8),
+            2 => Some(BandDataType::UInt16),
+            3 => Some(BandDataType::Int16),
+            4 => Some(BandDataType::UInt32),
+            5 => Some(BandDataType::Int32),
+            6 => Some(BandDataType::Float32),
+            7 => Some(BandDataType::Float64),
+            8 => Some(BandDataType::UInt64),
+            9 => Some(BandDataType::Int64),
+            10 => Some(BandDataType::Int8),
```
paleolimbot, line 185: I forget the history on these values but this seems like a funny order. Do we have to make this pub?

rust/sedona-schema/src/raster.rs
```diff
@@ -55,29 +75,59 @@ impl RasterSchema {
...
+        DataType::List(FieldRef::new(Field::new(
+            "item",
+            DataType::Struct(Fields::from(vec![
+                Field::new("source_axis", DataType::Int64, false),
+                Field::new("start", DataType::Int64, false),
+                Field::new("step", DataType::Int64, false),
+                Field::new("steps", DataType::Int64, false),
+            ])),
+            false,
+        )))
```
paleolimbot, line 130: 👍

rust/sedona-raster/src/traits.rs
```diff
@@ -15,120 +15,305 @@
...
+/// `source_shape` (the natural extent of `buffer`) with its `view`
+/// (the per-axis `(source_axis, start, step, steps)` slice spec). Stride
+/// can be zero (broadcast) or negative (reverse iteration), and may not be
+/// C-order. Consumers that need a flat row-major buffer should use
+/// `BandRef::contiguous_data()` instead.
+#[derive(Debug)]
+pub struct NdBuffer<'a> {
+    pub buffer: &'a [u8],
+    pub shape: &'a [u64],
+    pub strides: &'a [i64],
```
```suggestion
    pub shape: Vec<u64>,
    pub strides: Vec<i64>,
```

I think we'll be happier only considering the lifetime of the buffer here instead of linking all three to the same lifetime

rust/sedona-raster/src/traits.rs
```diff
@@ -15,120 +15,305 @@
...
-    fn metadata(&self) -> &dyn BandMetadataRef;
-    /// Raw band data as bytes (zero-copy access)
-    fn data(&self) -> &[u8];
-}
+    // -- Dimension metadata --
 
-/// Trait for accessing individual band metadata
-pub trait BandMetadataRef {
-    /// No-data value as raw bytes (None if null)
-    fn nodata_value(&self) -> Option<&[u8]>;
```
paleolimbot, line 102: I don't think we need to remove these (doing so caused some churn). You've changed the storage type of the bands but these things are mostly still valid to access in exactly the same way.

If you change `outdb_url` to return a `String` you can build the old string that was expected, and you can use your url parsing to lazily compute the band id if relevant.

rust/sedona-raster/src/traits.rs
```diff
@@ -15,120 +15,305 @@
...
+    /// success, or an error when the raw bytes have an unexpected length.
+    ///
+    /// # Warning
+    ///
+    /// For 64-bit integer bands (`Int64`, `UInt64`), the conversion to `f64`
+    /// is lossy when the magnitude exceeds 2^53 — values outside
+    /// `[-9_007_199_254_740_992, 9_007_199_254_740_992]` will be rounded to
+    /// the nearest representable double. Use `nodata()` directly to recover
+    /// the exact bytes if you need full integer precision.
+    fn nodata_as_f64(&self) -> Result<Option<f64>, ArrowError> {
```
paleolimbot: This should error for i64 and u64 band types (or, alternatively you can check the bounds of the actual value and error on a lossy conversion). This can lead to silently incorrect results very easily and is very difficult to trace. It is also reasonable that somebody would chose the extreme value (`0xffffffffffffffff`) to represent nodata as is common in other band types.

rust/sedona-raster/src/traits.rs
```diff
@@ -15,120 +15,305 @@
...
-}
 
-/// Trait for accessing raster metadata (dimensions, geotransform, bounding box, etc.)
-pub trait MetadataRef {
-    /// Width of the raster in pixels
-    fn width(&self) -> u64;
-    /// Height of the raster in pixels
-    fn height(&self) -> u64;
-    /// X coordinate of the upper-left corner
-    fn upper_left_x(&self) -> f64;
```
paleolimbot, line 64: Here as well: we've changed the storage but you we don't need to change the existing interface (to avoid churn in this PR you can `panic!()` if you don't have a width or height and follow up with the version that updates these to return `

You can also update `metadata()` to return `struct RasterMetadata` (i.e., concrete values instead of a trait) since I don't think there's ever much performance saved by lazily parsing any of these things) for compatibility and keep the rest of your top-level RasterRef changes.

rust/sedona-raster/src/traits.rs
```diff
@@ -15,120 +15,305 @@
...
+    /// lifetime reason as `band_outdb_uri`.
+    fn band_nodata(&self, index: usize) -> Option<&[u8]> {
+        let _ = index;
+        None
+    }
+
+    /// CRS string (PROJJSON, WKT, or authority code). None if not set.
     fn crs(&self) -> Option<&str>;
-    /// Bands accessor
-    fn bands(&self) -> &dyn BandsRef;
```
paleolimbot, line 54: You could have this return `Vec<Band<'_>>` to maintain compatibility. I happen to think that `raster.band(0).nodata_value()` / `raster.band_by_name("foofy").nodata_value()` is nicer than `raster.band_nodata(0)` (because then you can write functions that operate on bands in terms of a `&dyn BandRef`) and will never be measurably slower but also doesn't matter much.

rust/sedona-raster/src/builder.rs
```diff
@@ -17,325 +17,580 @@
...
+    /// `spatial_dims` names the raster-level spatial dimensions (today always
+    /// length 2, e.g. `["x","y"]`). `spatial_shape` gives their sizes in the
+    /// same order. Every band added to this raster must contain each name in
+    /// `spatial_dims` within its own `dim_names`, with matching size.
     pub fn start_raster(
         &mut self,
-        metadata: &dyn MetadataRef,
+        transform: &[f64; 6],
+        spatial_dims: &[&str],
+        spatial_shape: &[i64],
```
paleolimbot, line 192: If you keep the `MetadataRef` you don't have to update this function

rust/sedona-raster/src/builder.rs
```diff
@@ -17,325 +17,580 @@
...
+
+        Ok(())
+    }
 
+    /// Convenience: start a 2D raster with the legacy 8-parameter interface.
+    ///
+    /// Sets `spatial_dims=["x","y"]`, `spatial_shape=[width, height]`, and
+    /// builds the 6-element GDAL transform from the individual parameters.
+    #[allow(clippy::too_many_arguments)]
+    pub fn start_raster_2d(
```
paleolimbot: You can probably un-rename this to avoid changes. Most functions and tests will be only building 2D rasters anyway.

rust/sedona-raster/src/builder.rs
```diff
@@ -17,325 +17,580 @@
...
+        self.current_width = width;
+        self.current_height = height;
         Ok(())
     }
 
-    /// Start a new band - this must be called before writing band data
-    pub fn start_band(&mut self, band_metadata: BandMetadata) -> Result<(), ArrowError> {
-        // Append band metadata
-        match band_metadata.nodata_value {
-            Some(nodata) => self.band_nodata.append_value(&nodata),
```
paleolimbot, line 168: If you keep the BandMetadata you can avoid this change, too

rust/sedona-raster/src/builder.rs
```diff
@@ -17,325 +17,580 @@
...
 
         Ok(())
     }
 
-    /// Get direct access to the BinaryViewBuilder for writing the current band's data
-    /// Must be called after start_band() to write data to the current band
+    /// Convenience: start a 2D band with `dim_names=["y","x"]` and `shape=[height, width]`.
+    ///
+    /// Must be called after `start_raster_2d` which sets the current width/height.
+    pub fn start_band_2d(
```
paleolimbot, line 381: I think it's reasonable to `start_band_nd()` be the new interface but keep the old interface for existing code

rust/sedona-raster/src/builder.rs
```diff
@@ -350,500 +605,765 @@ impl RasterBuilder {
...
+        // Schema invariant: a band built via start_band serialises with a
+        // null view row, and the null must survive an Arrow IPC round-trip.
+        // If a future change accidentally writes a non-null empty list
+        // instead, downstream readers (DuckDB, PyArrow, sedona-py) will
+        // disagree about whether the view is identity.
+        use arrow_array::RecordBatch;
+        use arrow_ipc::reader::StreamReader;
+        use arrow_ipc::writer::StreamWriter;
+        use arrow_schema::Schema;
+        use std::io::Cursor;
```
paleolimbot: These can probably move `mod test`

rust/sedona-raster/src/builder.rs
```diff
@@ -350,500 +605,765 @@ impl RasterBuilder {
...
     }
 
     #[test]
-    fn test_band_access_errors() {
-        // Create a simple raster with one band
+    fn test_contiguous_data_identity_via_start_band_is_borrowed() {
+        // Canonical identity: the row's view list is null, and the read path
+        // synthesises the identity view. Should still hand the underlying
+        // bytes back without copying.
+        use std::borrow::Cow;
```
paleolimbot: This can move to `mod test`

paleolimbot (commented):
Thank you for working on this! This is a solid schema design that will be the foundation of a solid nd spatial raster framework that will be flexible enough to deliver the type of performance we are hoping for in the native translation of Sedona raster functions to Rust.

Most of my comments are about minimizing the diff to avoid disrupting existing work. With a few small changes I think you avoid almost all disruption of existing + in progress work and minimize the footprint of this PR considerably. Basically, just keep `RasterMetadata` and `BandMetadata` and `bands()` around, even if they are just shims returning concrete `structs` you'll remove later. For what it's worth I frequently do this with PRs into larger projects using VSCode's pull request extension which nicely lets you roll through your PR diff to minimize it.

rust/sedona-raster-functions/src/executor.rs
james-willis: done

rust/sedona-raster-functions/src/rs_band_accessors.rs
james-willis: done

rust/sedona-raster-functions/src/rs_band_accessors.rs
james-willis: done

rust/sedona-raster-functions/src/rs_convexhull.rs
james-willis: done

rust/sedona-raster-functions/src/rs_envelope.rs
james-willis: done

rust/sedona-raster/Cargo.toml
james-willis: yes, we use it in the tests

rust/sedona-schema/src/raster.rs
james-willis: done

rust/sedona-raster/src/traits.rs
james-willis: done

rust/sedona-raster/src/traits.rs
james-willis: done

rust/sedona-raster/src/builder.rs
james-willis: renamed the *_2d methods back to their original name.


I made a note to myself to come back later and give them a _2d suffix. I think its good to give them a clear name but more important to right now keep the code churn lower on this PR.

rust/sedona-raster-gdal/src/lib.rs
james-willis: I had to remove them so that the project compiles. 

The alternative to deleting it is to pull https://github.com/apache/sedona-db/pull/818 into this PR.

I'd rather do 818 as a fast follow to this PR. but can pull it into this PR if you prefer.

rust/sedona-raster/src/builder.rs
james-willis: done

rust/sedona-raster/src/builder.rs
james-willis: done

rust/sedona-raster/src/builder.rs
james-willis: done

rust/sedona-schema/src/raster.rs
james-willis: order is historic, we didn't change it. I think int8 was added later.

It's public because the raster reader in `sedona-raster::array` decodes the `band.data_type` Arrow column through this method.

james-willis:
**Design note on `raster_ref_to_gdal_mem` lifetime plumbing**

The GDAL backend in this PR uses the shim's `BandRef::data() -> &[u8]` rather than `BandRef::contiguous_data() -> Result<Cow<'_, [u8]>>`. That's deliberate, and it lets us drop the `(Dataset, Vec<Vec<u8>>)` tuple return / `_owned_band_bytes: Vec<Vec<u8>>` field on `RasterDataset`.

**Why it works in this PR**: every band is identity-viewed, so the bytes GDAL points at are a borrow into the StructArray's backing buffer. `RasterDataset` already keeps the source raster alive via `_source_raster: PhantomData<&'a dyn RasterRef>`, so the pointer is valid for the dataset's lifetime — no extra owned-bytes plumbing needed. Non-identity views error at read time here, so `Cow::Owned` cannot appear.

**What changes when the view machinery lands**: once non-identity views are supported, `contiguous_data()` may return `Cow::Owned` for sliced/permuted bands (the reader materializes a new buffer). That `Vec<u8>` will need to outlive the GDAL dataset that points into it. Bringing back the `_owned_band_bytes: Vec<Vec<u8>>` field on `RasterDataset` (and the matching tuple return from `raster_ref_to_gdal_mem`) is the right way to do that — earlier revisions of this PR carried that plumbing for exactly this reason.

Short version: if you're picking this code up later to add view support, expect to re-thread owned band bytes from the reader, through `raster_ref_to_gdal_mem`, and onto `RasterDataset`.

rust/sedona-raster/src/traits.rs
```diff
@@ -73,62 +108,550 @@ pub trait MetadataRef {
...
+    /// view, or a `contiguous_data` error) — corresponds to main's
+    /// infallible `BandRef::data() -> &[u8]` which only ever ran against
+    /// identity-view InDb bands.
+    fn data(&self) -> &[u8] {
+        // Default impl forwards through nd_buffer's borrowed slice. This
+        // only borrows the underlying band buffer for identity-view InDb
+        // bands; everything else is a corrupt-shape call site.
+        self.nd_buffer()
+            .expect("BandRef::data() requires an in-db band with bytes")
+            .buffer
```
paleolimbot: Should this also check for a non-identity view and panic?

rust/sedona-raster/src/traits.rs
```diff
@@ -217,4 +777,208 @@ mod tests {
...
+        // Locks in the documented warning: nodata bytes for Int64 values
+        // beyond f64's 53-bit mantissa silently round on conversion.
+        // The expected f64 is hard-coded — deriving it via `as f64` would
+        // mean the test invokes the same primitive cast it claims to test.
+        let big = (1i64 << 53) + 1; // 2^53 + 1; not representable in f64
+        let bytes = big.to_le_bytes();
+        let val = nodata_bytes_to_f64(&bytes, &BandDataType::Int64).unwrap();
+        assert_eq!(val, 9007199254740992.0_f64);
+        assert_ne!(val as i64, big);
+    }
```
```suggestion
```

rust/sedona-raster/src/traits.rs
```diff
@@ -73,62 +108,550 @@ pub trait MetadataRef {
...
+        nodata_bytes_to_f64_lossless(bytes, &self.data_type()).map(Some)
     }
 }
 
 /// Convert raw nodata bytes to f64 given a [`BandDataType`].
 ///
 /// The bytes are expected to be in little-endian order and exactly match the
 /// byte size of the data type.
-fn nodata_bytes_to_f64(bytes: &[u8], dt: &BandDataType) -> Result<f64, ArrowError> {
+pub fn nodata_bytes_to_f64(bytes: &[u8], dt: &BandDataType) -> Result<f64, ArrowError> {
```
```suggestion
fn nodata_bytes_to_f64(bytes: &[u8], dt: &BandDataType) -> Result<f64, ArrowError> {
```

rust/sedona-raster/src/array.rs
```diff
@@ -15,445 +15,431 @@
...
+    fn ndim(&self) -> usize {
+        self.view_entries.len()
     }
 
-    #[inline(always)]
-    fn upper_left_x(&self) -> f64 {
-        self.upper_left_x_array.value(self.index)
+    fn dim_names(&self) -> Vec<&str> {
+        let start = self.dim_names_list.value_offsets()[self.band_row] as usize;
+        let end = self.dim_names_list.value_offsets()[self.band_row + 1] as usize;
```
paleolimbot: Worth double checking that `value_offsets()` takes into account the global `offset` of `dim_names_list` here.

rust/sedona-raster/src/array.rs
```diff
@@ -15,445 +15,431 @@
...
+        let data_type_value = self.band_datatype_array.value(band_row);
+        let data_type = BandDataType::try_from_u32(data_type_value)?;
+
+        // Only the canonical identity view (null view row) is written today.
+        // A non-null view row would require the view → byte-stride composition
+        // path that is deferred to a follow-up; reject it here so callers see
+        // a clean "no band" rather than a panic.
+        if !self.band_view_list.is_null(band_row) {
+            return None;
+        }
```
paleolimbot: Should this be `sedona_internal_err!()` so it's harder to forget about?

rust/sedona-raster/src/array.rs
```diff
@@ -15,445 +15,431 @@
...
-        // current is 1-based, compare against len() + 1
-        if self.current <= self.bands.len() {
-            let band = self.bands.band(self.current).ok(); // Convert Result to Option
-            self.current += 1;
-            band
-        } else {
-            None
+        let dtype_size = data_type.byte_size() as i64;
+        // C-order byte strides over the source_shape:
+        //   byte_strides[k] = dtype_size * Π_{j>k} source_shape[j]
```
paleolimbot: Is this needed?

```suggestion
```

paleolimbot (approved):
A few optional nits probably worth considering now but in general this look great! Thank you!

rust/sedona-raster/src/array.rs
james-willis: done

rust/sedona-raster/src/traits.rs
james-willis: done

rust/sedona-raster/src/traits.rs
james-willis: done

rust/sedona-raster/src/array.rs
james-willis: done

rust/sedona-raster/src/traits.rs
james-willis: done

rust/sedona-raster/src/array.rs
```diff
@@ -230,11 +233,16 @@ impl<'a> RasterRef for RasterRefImpl<'a> {
...
+        // path, which is not yet implemented. Surface it loudly here rather
+        // than silently returning None, so callers see a clear invariant
+        // violation instead of an out-of-range-looking miss.
+        assert!(
+            self.band_view_list.is_null(band_row),
+            "{}",
+            sedona_common::sedona_internal_datafusion_err!(
+                "non-null view row at band {band_row}: view composition is not yet implemented"
+            )
+        );
```
paleolimbot: Is this intentional? Maybe:

```suggestion
        if !self.band_view_list.is_null(band_row) {
            return sedona_internal_err!(
                "non-null view row at band {band_row}: view composition is not yet implemented"
            );
        }
```

paleolimbot (approved):
Thank you!

rust/sedona-raster/src/array.rs
james-willis: It was intentional; this method returns an Option not a Result.

Though now I'm wondering if it should be optional. the original idea was to return none for band numbers that don't exist but maybe the old behavior to return a Err was better.

paleolimbot: An Err is definitely better (particularly if it already was that way) but the internal err within the assert is the particularly strange bit. If you need to make this an option you can do:

```rust
        if !self.band_view_list.is_null(band_row) {
            painic!(
                "non-null view row at band {band_row}: view composition is not yet implemented"
            );
        }
```

(but Python release builds will just crash on a panic without a user facing error)

james-willis: Ill update this to return a Result instead of an Option

james-willis: > the internal err within the assert is the particularly strange bit.

I think it'll do we what we want: when the assert fails it'll throw with the message from the Result the macro gives because we implement the Display trait. That was the idea: get the correct error string without fixing the return type issues

paleolimbot (approved):

PR #750, opened 2026-04-03
feat: add N-D raster dimension query and manipulation functions

james-willis:
## Summary

Adds 8 new `RS_*` functions for querying and manipulating N-dimensional raster data. Independent of #813 — these functions produce only identity-view bands and use only the schema / reader / builder surface that landed in #749. Purely additive in `sedona-raster-functions` — no schema or trait churn outside that crate.

> **Note: lazy implementations are deferred.** Every slice/dim function in this PR eagerly materializes its output as a fresh identity-view band. The view machinery from #813 is intentionally unused here. Converting these to lazy view composition (so `RS_Slice(RS_Slice(...))` fuses into a single composed view, etc.) is tracked as a follow-up — see the bottom of this description for the rationale and scope. `RS_BandToDim` will stay eager regardless; it requires multi-source view machinery that isn't on the roadmap.

## Conceptual model

The vocabulary follows **xarray**'s labeled-N-D-array model with SQL-friendly verb naming. For a reader coming from xarray / NumPy / PostGIS:

| `RS_*` function | xarray analog | NumPy analog | PostGIS analog |
|---|---|---|---|
| `RS_NumDimensions` | `.ndim` | `.ndim` | — (PostGIS rasters are 2-D + bands) |
| `RS_DimNames` | `.dims` | — | — |
| `RS_DimSize(name)` | `.sizes[name]` | `.shape[i]` | — |
| `RS_Shape` | `.shape` | `.shape` | — |
| `RS_Slice(name, i)` | `.isel({name: i})` (drops dim) | `arr[..., i, ...]` | — |
| `RS_SliceRange(name, a, b)` | `.isel({name: slice(a, b)})` (keeps dim) | `arr[..., a:b, ...]` | — |
| `RS_DimToBand(name)` | partial inverse of `.expand_dims()` | — | — |
| `RS_BandToDim(name)` | partial match to `.stack()` along a new axis | — | — |

We picked `RS_Slice` / `RS_SliceRange` rather than `RS_Isel` (xarray's name) because SQL function names read better when self-documenting in query logs: `RS_Slice(raster, 'time', 5)` over `RS_Isel(raster, 'time', 5)`. The dim ↔ band conversions (`RS_DimToBand` / `RS_BandToDim`) are SedonaDB-specific verbs for crossing the "bands axis" vs "extra dimensions" boundary that's part of the raster schema.

### Dimension query functions (`rs_dimensions.rs`)

- **RS_NumDimensions(raster [, band])** → Int32 — number of dimensions
- **RS_DimNames(raster [, band])** → List\<Utf8\> — ordered dimension names
- **RS_DimSize(raster, dim_name [, band])** → Int64 — size of a named dimension (null if missing)
- **RS_Shape(raster [, band])** → List\<Int64\> — full shape array

When the band argument is omitted, defaults to band 0 and verifies all bands agree — returns an error if bands have different dimensionality.

### Slice functions (`rs_slice.rs`)

- **RS_Slice(raster, dim_name, index)** → Raster — reduce a dimension by picking one index
- **RS_SliceRange(raster, dim_name, start, end)** → Raster — narrow a dimension to [start, end)

### Dimension ↔ band functions (`rs_dim_band.rs`)

- **RS_DimToBand(raster, dim_name)** → Raster — promote a dimension into separate bands
- **RS_BandToDim(raster, dim_name)** → Raster — collapse bands into a new dimension

All slice/dim functions error on spatial dimensions (`x_dim`/`y_dim`).

None of these functions are GDAL-backed — none touch `sedona-raster-gdal/`.

## Lazy implementations: deferred

`RS_Slice`, `RS_SliceRange`, and `RS_DimToBand` are all candidates for lazy view-composition — instead of walking the input bytes and writing a fresh band, the output would be a non-identity-view band that references the input's source bytes through a composed `[ViewEntry]`. That's tracked as a follow-up rather than included here for two reasons:

1. **The builder API needed for clean view composition doesn't exist yet.** The conversion needs a builder method along the lines of `with_view(input: &dyn BandRef, view: &[ViewEntry], ...)` — "create a new view into an existing raster." The current `start_band_with_view` from #813 only handles "construct a band with raw bytes and a view in one step"; it doesn't compose against an input band's existing view. Designing the unified API, deprecating the current surface, and migrating call sites is a focused follow-up — better as its own change than bundled with the function additions here.

2. **Memory trade-off depends on workload.** Lazy slicing in this model stores the *source* bytes in the output band's `data` column, not the visible-region subset. For a small slice of a huge axis (one time index out of 10,000), the output carries the full 10,000-element payload + a view that says "only this element is visible." That's a win for `slice → metadata` and chained-slice workloads, a loss for `slice → write to disk`. Worth measuring on real workloads before flipping the default.

`RS_BandToDim` is structurally eager regardless — it stacks multiple bands into one new dim, and PR-D's view model is single-source per band. Multi-source view machinery isn't planned.

## Test plan

- [x] 31 new tests across 3 files (19 dimension queries + 12 slice/dim-band)
- [x] All 174 tests pass in `sedona-raster-functions` (143 existing + 31 new)
- [x] `cargo clippy --all-targets -- -D warnings` clean
- [x] `cargo fmt --all --check` clean
- [x] Round-trip test: `RS_DimToBand` then `RS_BandToDim` recovers original data


rust/sedona-raster-functions/src/rs_band_accessors.rs
```diff
@@ -224,16 +225,19 @@ fn get_nodata_value(
...
-            let band = raster.bands().band(band_index as usize)?;
-            let band_meta = band.metadata();
-            match band_meta.nodata_value_as_f64()? {
-                None => builder.append_null(),
-                Some(val) => builder.append_value(val),
+            let idx = (band_index - 1) as usize;
+            match (raster.band_nodata(idx), raster.band_data_type(idx)) {
+                (Some(bytes), Some(dt)) => {
+                    let val = nodata_bytes_to_f64(bytes, &dt)
+                        .map_err(datafusion_common::DataFusionError::from)?;
```
paleolimbot: Can you make sure that `nodata_bytes_to_f64()` errors for int64 and uint64 types where this conversion is lossy? (At least one version I reviewed does but we have some duplication and it's a good time to make sure)

rust/sedona-raster-functions/src/rs_convexhull.rs
```diff
@@ -107,8 +107,18 @@ impl SedonaScalarKernel for RsConvexHull {
...
+            "Raster has no spatial dimensions; cannot determine width".into(),
+        ));
+    };
+    let Some(height) = raster.height() else {
+        return Err(DataFusionError::Execution(
+            "Raster has no spatial dimensions; cannot determine height".into(),
+        ));
+    };
+    let width = width as i64;
+    let height = height as i64;
```
paleolimbot: Should this be checked in the executor before it reaches the function? Or should `width()` return a `Result<>` that propagates this error so it doesn't have to be copy/pasted?

paleolimbot (commented):
I got partway through this before realizing it was the wrong PR 😬  (comments may be helpful anyway)

rust/sedona-raster-functions/src/rs_band_accessors.rs
james-willis: done in the other PR

rust/sedona-raster-functions/src/rs_convexhull.rs
james-willis: done in the other PR

docs/reference/sql/rs_bandtodim.qmd
```diff
@@ -0,0 +1,64 @@
...
+data layout.
+
+This is useful for converting GeoTIFF-style multi-band rasters into a single
+N-dimensional chunk for export to formats like Zarr.
+
+## Examples
+
+```
+-- Collapse 3 RGB bands [y, x] into 1 band [band=3, y, x]
+SELECT RS_BandToDim(raster, 'band');
```
paleolimbot: These aren't labelled as `sql` so they won't run. We should ensure these examples are runnable so they don't go stale...can you use a `RS_DimToBand(RS_BandToDim(RS_Example())` or do you need another fixture?

docs/reference/sql/rs_dimtoband.qmd
```diff
@@ -0,0 +1,63 @@
...
+condition usually signals a typo'd dimension name.
+
+The spatial dimensions (`x_dim` and `y_dim`) cannot be promoted.
+
+The inverse operation is [RS_BandToDim](rs_bandtodim.qmd).
+
+## Examples
+
+```
+-- Promote wavelength into separate bands for band math
```
paleolimbot: This one is also missing a `sql` qualifier.

rust/sedona-raster-functions/src/rs_dim_band.rs
```diff
@@ -0,0 +1,749 @@
...
+    fn invoke_batch(
+        &self,
+        arg_types: &[SedonaType],
+        args: &[ColumnarValue],
+    ) -> Result<ColumnarValue> {
+        let executor = RasterExecutor::new(arg_types, args);
+
+        let dim_name_array = args[1].clone().cast_to(&DataType::Utf8, None)?;
+        let dim_name_array = dim_name_array.into_array(executor.num_iterations())?;
+        let dim_name_array = as_string_array(&dim_name_array)?;
```
paleolimbot: Casting into a Utf8View array is probably more efficient for the scalar case

rust/sedona-raster-functions/src/rs_dimensions.rs
```diff
@@ -0,0 +1,1012 @@
...
+    }
+    let band0 = raster.band(0).map_err(|e| arrow_datafusion_err!(e))?;
+    let value = extractor(band0.as_ref());
+    for i in 1..raster.num_bands() {
+        if let Ok(band) = raster.band(i) {
+            let other = extractor(band.as_ref());
+            if other != value {
+                return exec_err!(
+                    "{func_name}: bands have different {property_name} — specify a band index"
+                );
```
paleolimbot, line 63: Do you have a (potential) outdb_uri you could pass on here to help a user debug which value went awry?

rust/sedona-raster-functions/src/rs_dimensions.rs
```diff
@@ -0,0 +1,1012 @@
...
+            match raster_opt {
+                None => {
+                    builder.append_null();
+                    Ok(())
+                }
+                Some(raster) => {
+                    if band_index < 1 || band_index > raster.num_bands() as i32 {
+                        builder.append_null();
+                        return Ok(());
+                    }
```
paleolimbot: Is the silent discarding of invalid specified indices intentional/desirable?

rust/sedona-raster-functions/src/rs_dimensions.rs
```diff
@@ -0,0 +1,1012 @@
...
+        matcher.match_args(args)
+    }
+
+    fn invoke_batch(
+        &self,
+        arg_types: &[SedonaType],
+        args: &[ColumnarValue],
+    ) -> Result<ColumnarValue> {
+        let executor = RasterExecutor::new(arg_types, args);
+        let mut list_builder = ListBuilder::new(StringBuilder::new());
```
paleolimbot: Returning a stringview might be helpful here (the din names are often short and/or repeated)

rust/sedona-raster-functions/src/rs_dimensions.rs
```diff
@@ -0,0 +1,1012 @@
...
+    }
+
+    fn invoke_batch(
+        &self,
+        arg_types: &[SedonaType],
+        args: &[ColumnarValue],
+    ) -> Result<ColumnarValue> {
+        let executor = RasterExecutor::new(arg_types, args);
+        let dim_name_array = args[1].clone().cast_to(&DataType::Utf8, None)?;
+        let dim_name_array = dim_name_array.into_array(executor.num_iterations())?;
```
paleolimbot: It may be worth casting to a string view here for a more efficient scalar case

rust/sedona-raster-functions/src/rs_dimensions.rs
```diff
@@ -0,0 +1,1012 @@
...
+    use sedona_schema::datatypes::RASTER;
+    use sedona_schema::raster::BandDataType;
+    use sedona_testing::rasters::generate_test_rasters;
+    use sedona_testing::testers::ScalarUdfTester;
+
+    /// Build a single-row 2D raster StructArray.
+    fn build_2d_raster(width: u64, height: u64) -> StructArray {
+        let mut builder = RasterBuilder::new(1);
+        builder
+            .start_raster_2d(
```
paleolimbot: Can any of these use your utilities to be more compact?

rust/sedona-raster-functions/src/rs_dimensions.rs
```diff
@@ -0,0 +1,1012 @@
...
+        let tester = ScalarUdfTester::new(udf, vec![RASTER]);
+        tester.assert_return_type(DataType::Int32);
+
+        let rasters = build_2d_raster(4, 5);
+        let result = tester.invoke_array(Arc::new(rasters)).unwrap();
+        let arr = result
+            .as_any()
+            .downcast_ref::<Int32Array>()
+            .expect("Expected Int32Array");
+        assert_eq!(arr.value(0), 2);
```
paleolimbot: You may have to tweak the tester a little, but for the cases that are testing behaviour that doesn't depend on scalar/arrayness, these are better as

```rust
let result = tester.invoke_scalar(build_s2_raster(4, 5)).unwrap();
tester.assert_scalar_result(result, 2);
```

...which is sufficiently compact that you can fit more than one in a test function.

rust/sedona-raster-functions/src/rs_dimensions.rs
```diff
@@ -0,0 +1,1012 @@
...
+            .downcast_ref::<ListArray>()
+            .expect("Expected ListArray");
+        let values = list_arr.value(0);
+        let str_arr = values
+            .as_any()
+            .downcast_ref::<StringArray>()
+            .expect("Expected StringArray");
+        assert_eq!(str_arr.len(), 2);
+        assert_eq!(str_arr.value(0), "y");
+        assert_eq!(str_arr.value(1), "x");
```
paleolimbot: Again, you might have to tweak the tester, but this could be

```rust
tester.assert_scalar_result(result, ["y", "x"])` // maybe json!["y", "x"] or vec!["y", "x"] would be easier
```

rust/sedona-raster-functions/src/rs_slice.rs
```diff
@@ -0,0 +1,865 @@
...
+        let mut builder = RasterBuilder::new(1);
+        let transform = [0.0, 1.0, 0.0, 0.0, 0.0, -1.0];
+        builder
+            .start_raster_nd(
+                &transform,
+                &["x", "y"],
+                &[width as i64, height as i64],
+                None,
+            )
+            .unwrap();
```
paleolimbot: Can this use a new helper?

paleolimbot (commented):
Great! I think this can merge soon, although I would feel better about them if it was easier to read the tests (I'm not sure I would be able to spot a place where an LLM added an expectation that locked in the behaviour of a bug, which I find they often do). Tests in Python would probably be easier to read (and you could compare your answer against the same operation in numpy or xarray) but you'd need to extend the tester I just added.

For this to merge I'd suggest cleaning up the verbosity of your expectations where you can reasonably do so, marking the functions as EXPERIMENTAL in the documentation (so uses know the behaviour can change without notice), and ensuring the examples can actually run. We can add integration tests with these operations against an actual implementation after the release.

rust/sedona-raster-functions/src/rs_dimensions.rs
james-willis: done

rust/sedona-raster-functions/src/rs_slice.rs
james-willis: done

docs/reference/sql/rs_dimtoband.qmd
james-willis: done

james-willis:
python lacks an nd array fixture right now. don't want to build it at this moment. 

but I reduced the verbosity of the rust tests.

rust/sedona-raster-functions/src/rs_dimensions.rs
james-willis: done

rust/sedona-raster-functions/src/rs_dim_band.rs
james-willis: done

rust/sedona-raster-functions/src/rs_dimensions.rs
james-willis: done

rust/sedona-raster-functions/src/rs_dimensions.rs
james-willis: I got pretty close to this. marking done

rust/sedona-raster-functions/src/rs_dimensions.rs
james-willis: cleaned up verbosity here

rust/sedona-raster-functions/src/rs_dimensions.rs
james-willis: done

rust/sedona-raster-functions/src/rs_dimensions.rs
james-willis: I think that is idiomatic in data processing/sql based systems: One bad row should not fail the query. If the result being null is bad users can deal with that downstream

I found a couple cases that do error that I am going to make return null instead

python/sedonadb/python/sedonadb/raster.py
```diff
@@ -81,61 +151,102 @@ def lazy(
         Args:
             uri: The URI of the external data source (e.g., file path or cloud URL).
                 This URI is not validated unless the pixels are read.
-            shape: The shape of the raster as (height, width). Must have exactly
-                two dimensions.
+            shape: The shape of the raster. The trailing two axes are the spatial
+                ``(y, x)`` pair; any leading axes are extra (e.g. time) dimensions.
```
paleolimbot:
```suggestion
                `(y, x)` pair; any leading axes are extra (e.g. time) dimensions.
```

python/sedonadb/python/sedonadb/raster.py
```diff
@@ -81,61 +151,102 @@ def lazy(
...
             dtype: The pixel data type (e.g., 'uint8', 'float32', 'int16').
             format: The format of the external data (e.g., 'tif', or 'zarr'). If None,
                 the format will be inferred when the data is accessed.
             crs: The coordinate reference system. Can be any value accepted by
                 geoarrow.types.type_spec (e.g., string, pyproj.CRS).
+            dim_names: Names of the dimensions. Defaults to ``["y", "x"]`` for a 2-D
```
paleolimbot:
```suggestion
            dim_names: Names of the dimensions. Defaults to `["y", "x"]` for a 2-D
```

python/sedonadb/python/sedonadb/raster.py
```diff
@@ -81,61 +151,102 @@ def lazy(
...
+        array: "np.ndarray",
+        *,
+        dim_names: Optional[Iterable[str]] = None,
+        crs: Any = None,
+        nodata: Any = None,
+        transform: Optional[Iterable[float]] = None,
+    ) -> "Raster":
+        """Create an in-database raster from a NumPy array, holding pixels inline.
+
+        The mirror of :meth:`lazy`: identical dimension/CRS conventions, but the
```
paleolimbot: I am not actually sure how to properly cross-reference using our current Python docs setup, but I don't think ReST works.

```suggestion
        The mirror of `lazy`: identical dimension/CRS conventions, but the
```

python/sedonadb/python/sedonadb/raster.py
```diff
@@ -81,61 +151,102 @@ def lazy(
...
+            array: The pixel data. The trailing two axes are the spatial ``(y, x)``
+                pair; any leading axes are extra (e.g. time) dimensions. The dtype
+                must be a supported raster type (uint8, int16, float32, ...).
+            dim_names: Names of the dimensions. Defaults to ``["y", "x"]`` for a 2-D
+                array; required when the array has more than two dimensions.
+            crs: The coordinate reference system (any value accepted by
+                geoarrow.types.type_spec).
+            nodata: Optional nodata sentinel, packed in the array's dtype.
+            transform: Optional GDAL-order geotransform
+                ``[origin_x, scale_x, skew_x, origin_y, skew_y, scale_y]``;
```
paleolimbot: I know you're working on this, but ndarray + bbox is an easier way to specify this

```suggestion
                `[origin_x, scale_x, skew_x, origin_y, skew_y, scale_y]`;
```

python/sedonadb/python/sedonadb/raster.py
```diff
@@ -81,61 +151,102 @@ def lazy(
...
+
+    @staticmethod
+    def from_numpy(
+        array: "np.ndarray",
+        *,
+        dim_names: Optional[Iterable[str]] = None,
+        crs: Any = None,
+        nodata: Any = None,
+        transform: Optional[Iterable[float]] = None,
+    ) -> "Raster":
```
paleolimbot, line 204: Perhaps leave a loud comment in the docstring that this copies the entire raster. I tried a bit of work on true zero copy and there is some pyarrow behaviour that's tricky to work around...better suited for follow-up https://github.com/paleolimbot/sedona-db/commit/41adadd18210caf06783ee4b9e9bab6560c98456

python/sedonadb/tests/functions/test_raster_nd.py
```diff
@@ -0,0 +1,175 @@
...
+def test_num_dimensions(r2d, r3d):
+    assert _eval("RS_NumDimensions(r1)", r2d) == 2
+    assert _eval("RS_NumDimensions(r1)", r3d) == 3
+
+
+def test_num_dimensions_with_band(r3d):
+    assert _eval("RS_NumDimensions(r1, 1)", r3d) == 3
+    # A null or out-of-range band index yields NULL; it does not default to band 1.
+    assert _eval("RS_NumDimensions(r1, CAST(NULL AS INT))", r3d) is None
+    assert _eval("RS_NumDimensions(r1, 99)", r3d) is None
```
paleolimbot, line 85: Thanks for these!

paleolimbot (approved):
Thank you!

A few ReST formattings that should be markdown but I think this is great and has all that is needed to iterate on these + improve them easily if we run into an issue.

python/sedonadb/python/sedonadb/raster.py
james-willis: done

python/sedonadb/python/sedonadb/raster.py
james-willis: done

python/sedonadb/tests/functions/test_raster_nd.py
james-willis: im working to move as many as possible into python. encountered some blockers & bugs. good testing I suppose!

python/sedonadb/python/sedonadb/raster.py
james-willis: yeah these were done real quick to get the tests into python for now. didnt give much thought yet

python/sedonadb/python/sedonadb/raster.py
james-willis: done

paleolimbot (approved):
Thank you!

PR #756, opened 2026-04-10
chore(python/sedonadb): Add geography integration test framework

paleolimbot:
This PR adds a framework for geography integration tests. This specifically adds one for each category (categories based on the BigQuery categories here: https://docs.cloud.google.com/bigquery/docs/reference/standard-sql/geography_functions ), and, critically, adds a `BigQuery` engine to test against (because PostGIS only supports a subset of the functions we implement now and bigquery uses the same underlying library so the results in theory should match more closely).

For BigQuery, the engine uses application default credentials (i.e., `gcloud login`) and caches the responses by query in a YAML file. This is a bit crude but is needed because the latency of a single BigQuery roundtrip is very slow. We may need something that scales better at some point but this should get us through the basic suite of "is all of this plugged in and can we add regression tests".


python/sedonadb/python/sedonadb/testing.py
```diff
@@ -18,10 +18,11 @@
 import os
 import warnings
 from pathlib import Path
-from typing import TYPE_CHECKING, List, Tuple, Any
+from typing import TYPE_CHECKING, Any, List, Tuple
 
 import geoarrow.pyarrow as ga
 import pyarrow as pa
+import yaml
```
zhangfengcdt: Do you want to lazily import this in ArrowSQLCache?

zhangfengcdt (approved):
LGTM

Just minor suggestions of importing and we may want to add "__init__.py" to test module as well.

PR #766, opened 2026-04-16
feat(sedona-spatial-join): Fix for the customized join provider that may not take effective

pwrliang:
We have provided a method (https://github.com/apache/sedona-db/pull/722) `with_spatial_join_provider` to overwrite the `DefaultSpatialJoinProvider` with a customized one by calling `with_spatial_join_provider` of`SpatialJoinExec`.
```rust
    pub fn with_spatial_join_provider(
        mut self,
        join_provider: Arc<dyn SpatialJoinProvider>,
    ) -> Self {
        self.join_provider = join_provider;
        self
    }
```

However, there are many methods in `SpatialJoinExec`, such as `swap_inputs`, that return a modified `SpatialJoinExec` by calling `try_new_internal`, which binds `join_provider` to `DefaultSpatialJoinProvider` again. We need to call `with_spatial_join_provider` after the returning of a `SpatialJoinExec` object to overwrite it.

### DIFF
--- a/rust/sedona-spatial-join/src/exec.rs
+++ b/rust/sedona-spatial-join/src/exec.rs
@@ -226,7 +226,8 @@ impl SpatialJoinExec {
             &self.join_type.swap(),
             swapped_projection,
             self.seed,
-        )?;
+        )?
+        .with_spatial_join_provider(Arc::clone(&self.join_provider));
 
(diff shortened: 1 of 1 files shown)


pwrliang:
Do you prefer this modification or add an extra parameter `join_provider` to the constructors of `SpatialJoinExec`? @paleolimbot

paleolimbot (approved):
Thank you!

PR #767, opened 2026-04-16
 feat(c/sedona-libgpuspatial): Upgrade RMM to 26.04.00

pwrliang:
This PR upgrades RAPIDS RMM to 26.04.00, which addresses an upstream bug (https://github.com/rapidsai/rmm/issues/2303), so we don't have to patch it. Additionally, the PR reads environment variables to adjust the log level and switch on/off profiling mode.


paleolimbot (approved):
Thank you!

The clippy errors aren't related to this change, but do you mind fixing them while you're here? (`rustup update` should get you the freshly released version causing the issue)

PR #777, opened 2026-04-21
feat(rust/sedona-functions): Aded ST_LineSubstring function

sapienza88:


paleolimbot (commented):
Thanks for opening!

- This needs `cargo fmt` to (and all the problems found with `cargo clippy` to be fixed).
- This will need tests in Rust to verify the basic behaviour of the function. You can use one of the other functions in this directory as a template for the tests you'll need (basically ensure that all the branches of the implementation are covered)
- The new function needs to be documented in `docs/reference/sql`
- The new function needs Python integration tests, which is where we check the behaviour against PostGIS.

I think there are links for some of those in the initial ticket you filed but let me know if you get lost!

sapienza88:
@paleolimbot I added some tests and the doc for this function, not sure I understand: "This needs cargo fmt to (and all the problems found with cargo clippy to be fixed)."

python/sedonadb/tests/functions/test_linesubstring.py
```diff
@@ -0,0 +1,61 @@
...
+        if isinstance(eng, PostGIS):
+            
+            expected = expected.replace("Z(", "Z (")
+            expected = expected.replace("M(", "M (")
+            expected = expected.replace("ZM(", "ZM (")
+
+    eng.assert_query_result(
+        f"SELECT ST_AsText(ST_LineSubstring({geom_or_null(geom)}, {start}, {end}))",
+        expected,
+    )
```
paleolimbot: You should be able to just feed the WKT directly into the `expected` slot, which should help with the difference between the WKT output:

```suggestion
        eng.assert_query_result(
        f"SELECT ST_LineSubstring({geom_or_null(geom)}, {start}, {end})",
        expected,
    )

```

paleolimbot (commented):
`cargo fmt` is something you can run from the terminal that normalizes the formatting of the rust code. It is also run automatically by `pre-commit run --all-files`. There's a CI check to make sure the pre-commit command runs cleanly, so if that check fails you'll have to run that locally to get it to pass.

python/sedonadb/tests/functions/test_linesubstring.py
```diff
@@ -0,0 +1,46 @@
...
+
+import pytest
+from sedonadb.testing import geom_or_null, PostGIS, SedonaDB
+
+
+@pytest.mark.parametrize("eng", [SedonaDB, PostGIS])
+@pytest.mark.parametrize(
+    ("geom", "start", "end", "expected"),
+    [
+        (None, 0.0, 1.0, None),
```
paleolimbot, line 29:
```suggestion
        (None, 0.0, 1.0, None),
        ("LINESTRING EMPTY", None, 1.0, None),
        ("LINESTRING EMPTY, 0.0, None, None),
```

rust/sedona-functions/src/st_linesubstring.rs
```diff
@@ -0,0 +1,164 @@
...
+
+        let end_frac: Option<f64> = match &args[2].cast_to(&DataType::Float64, None)? {
+            ColumnarValue::Scalar(ScalarValue::Float64(e)) => *e,
+            _ => None,
+        };
+        fn interpolate<C: CoordTrait<T = f64>>(p1: C, p2: C, fraction: f64) -> (f64, f64) {
+            let x = p1.x() + (p2.x() - p1.x()) * fraction;
+            let y = p1.y() + (p2.y() - p1.y()) * fraction;
+            (x, y)
+        }
```
paleolimbot: You have a test for Z that's failing because this doesn't handle anything except XY coordinates. I think it should be relatively easy to modify this to work with any dimension...maybe:

```rust
        fn interpolate<C: CoordTrait<T = f64>>(p1: C, p2: C, fraction: f64, dim: Dimensions, buf: &mut impl Write) ->  Result<(), SedonaGeometryError> {
            for i in 0..dim.size() {
                let v = p1.nth_unchecked(i) + (p2.nth_unchecked(i) - p1.nth_unchecked(i)) * fraction;
                buf.write_all(v.to_le_bytes())?;
            }
            Ok(())
        }
```

rust/sedona-functions/src/st_linesubstring.rs
```diff
@@ -0,0 +1,164 @@
...
+                            let fraction = if segment_len > 0.0 {
+                                (start_dist - d1) / segment_len
+                            } else {
+                                0.0
+                            };
+                            new_coords.push(interpolate(p1, p2, fraction));
+                        }
+
+                        if d1 > start_dist && d1 < end_dist {
+                            new_coords.push((p1.x(), p1.y()));
```
paleolimbot:
```suggestion
                            write_wkb_coord_trait(p1, &mut coords)?;
```

rust/sedona-functions/src/st_linesubstring.rs
```diff
@@ -0,0 +1,164 @@
...
+                        let p2 = line.coord(i + 1).unwrap();
+
+                        if start_dist >= d1 && start_dist <= d2 {
+                            let segment_len = d2 - d1;
+                            let fraction = if segment_len > 0.0 {
+                                (start_dist - d1) / segment_len
+                            } else {
+                                0.0
+                            };
+                            new_coords.push(interpolate(p1, p2, fraction));
```
paleolimbot:
```suggestion
                            new_coords.push(interpolate(p1, p2, fraction, line.dim(), &mut coords)?);
```

rust/sedona-functions/src/st_linesubstring.rs
```diff
@@ -0,0 +1,164 @@
...
+                let type_id: u32 = 2;
+                builder.write_all(&type_id.to_le_bytes())?;
+
+                let num_points = new_coords.len() as u32;
+                builder.write_all(&num_points.to_le_bytes())?;
+
+                for (x, y) in new_coords {
+                    builder.write_all(&x.to_le_bytes())?;
+                    builder.write_all(&y.to_le_bytes())?;
+                }
```
paleolimbot: You can use `write_wkb_linestring_header(&mut builder, line.dim(), num_output_coords)?` for this (there are other utility functions in sedona_geometry::wkb_factory that might be useful as well. I've added suggestions above about how to accumulate the coordinates as bytes directly in your implementation (which also simplifies supporting Z and M here).

paleolimbot (commented):
Sorry for being slow to review this!

I left a few suggestions about how to support Z and M right out of the gate (since I saw that test was failing and it's a nice feature to be able to support).

The clippy error you should be able to fix with `cargo clippy --fix` (it's an unused import).

rust/sedona-functions/src/st_linesubstring.rs
sapienza88: I’m a little rusty when it comes to Rust, but I’ll try to use some AI tools to help make this mergeable.

python/sedonadb/tests/functions/test_linesubstring.py
paleolimbot: Can you add these cases to ensure proper null semantics?

python/sedonadb/tests/functions/test_linesubstring.py
```diff
@@ -0,0 +1,45 @@
...
+        ("LINESTRING (0 0, 10 0)", 0.2, 0.8, "LINESTRING (2 0, 8 0)"),
+        ("LINESTRING (0 0, 10 10)", 0.3, 0.6, "LINESTRING (3 3, 6 6)"),
+        ("LINESTRING (0 0, 10 10)", 0.5, 0.5, "POINT (5 5)"),
+        ("LINESTRING (0 0, 10 0, 10 10)", 0.25, 0.75, "LINESTRING (5 0, 10 0, 10 5)"),
+        (
+            "LINESTRING Z (0 0 0, 10 10 10)",
+            0.5,
+            0.8,
+            "LINESTRING Z (5 5 5, 8 8 8)",
+        ),
```
paleolimbot, line 105: Can you add:

- Z, M, and ZM cases to ensure  dimensions are propagated correctly? You will want to use different numbers for the coordinates to ensure the dimensions are not mixed up (e.g., `LINESTRING Z (0 10 20, 10 20 30)`)
- Cases that cover different values of start/end (for the same 2D linestring)

rust/sedona-functions/src/st_linesubstring.rs
```diff
@@ -0,0 +1,205 @@
...
+                            final_wkb.extend_from_slice(&point_count.to_le_bytes());
+                            final_wkb.extend_from_slice(&wkb_body);
+                        }
+                        builder.append_value(final_wkb);
+                    } else {
+                        builder.append_null();
+                    }
+                } else {
+                    builder.append_null();
+                }
```
paleolimbot: Should non-linestrings really return null here or should they error? (You can check by adding a non-linestring case to the Python integration tests where they also run against PostGIS)

Cargo.toml
```diff
@@ -92,6 +92,7 @@ datafusion-datasource-parquet = { version = "52.5.0" }
 datafusion-execution = { version = "52.5.0", default-features = false }
 datafusion-expr = { version = "52.5.0" }
 datafusion-ffi = {  version = "52.5.0" }
+datafusion-functions-nested = { version = "52.5.0" }
```
paleolimbot: Was it necessary to add this?

Cargo.lock
paleolimbot, line 1: I am not sure these updates are needed but I'm happy to push that commit to fix this when we get that close to merging 🙂

rust/sedona-functions/src/st_linesubstring.rs
```diff
@@ -0,0 +1,205 @@
...
+                            if wkb_body.len() >= coord_bytes {
+                                final_wkb.extend_from_slice(&wkb_body[..coord_bytes]);
+                            }
+                        } else {
+                            // LINESTRING Result (Fixes Z-coordinate drop)
+                            let l_type: u32 = if dim == Dimensions::Xyz { 1002 } else { 2 };
+                            final_wkb.extend_from_slice(&l_type.to_le_bytes());
+                            final_wkb.extend_from_slice(&point_count.to_le_bytes());
+                            final_wkb.extend_from_slice(&wkb_body);
+                        }
```
paleolimbot: We have utilities in `sedona_geometry::wkb_factory` for this (e.g., write the linestring header or write the point header). This should propagate the dimensions of the input (this section special-cases Z but this could be M or ZM).

rust/sedona-functions/src/st_linesubstring.rs
```diff
@@ -0,0 +1,205 @@
...
+            } else {
+                builder.append_null();
+            }
+
+            Ok(())
+        })?;
+
+        executor.finish(Arc::new(builder.finish()))
+    }
+}
```
paleolimbot, line 136: Before this merges it will also need a `mod test {}` like those in the other sedona-functions implementations. These include enough pure Rust test coverage to cover the main functionality including Scalar and Array inputs in execution and any errors that should occur. Basically, enough to cover the branches of the implementation.

rust/sedona-functions/src/st_linesubstring.rs
```diff
@@ -0,0 +1,205 @@
...
+
+        let start_frac: Option<f64> = match &args[1].cast_to(&DataType::Float64, None)? {
+            ColumnarValue::Scalar(ScalarValue::Float64(s)) => *s,
+            _ => None,
+        };
+
+        let end_frac: Option<f64> = match &args[2].cast_to(&DataType::Float64, None)? {
+            ColumnarValue::Scalar(ScalarValue::Float64(e)) => *e,
+            _ => None,
+        };
```
paleolimbot: This covers the case where these are constants but doesn't cover the case where you have Array input here. Typically we use something like `let start_frac_arrayref = args[1].cast_to(&DataType::Float64, None)?.to_array(executor.num_iterations());` and get an iterator over the float64s. I think the implementation for `st_pointn()` has an example you can draw from.

sapienza88: Done!

rust/sedona-functions/src/st_linesubstring.rs
sapienza88: done!

python/sedonadb/tests/functions/test_linesubstring.py
sapienza88: done!

Cargo.toml
sapienza88: for some reason I remember that was needed

rust/sedona-functions/src/st_linesubstring.rs
sapienza88: for now null should be fine?

rust/sedona-functions/src/st_linesubstring.rs
sapienza88: done! pls check all and merge whenever possible. Thanks.

docs/reference/sql/st_linesubstring.qmd
```diff
@@ -0,0 +1,37 @@
...
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+---
+title: ST_LineSubstring
+description: Returns a linestring being a substring of the input one starting and ending at the given fractions of total 2d length. Second and third arguments are double precision values between 0 and 1.
```
paleolimbot:
```suggestion
description: Returns a linestring being a substring of the input one starting and ending at the given fractions of total 2d length.
```

docs/reference/sql/st_linesubstring.qmd
```diff
@@ -0,0 +1,37 @@
...
+kernels:
+  - returns: geometry
+    args:
+      - geometry: geometry
+      - start_fraction: double
+      - end_fraction: double
```
paleolimbot:
```suggestion
    args:
      - geometry
      - name: start_fraction
        type: double
        description: The fraction from which the returned line will start (between 0 and 1)
      - name: end_fraction
        type: double
        description: The fraction from which the returned line will end (between 0 and 1)
```

python/sedonadb/tests/functions/test_linesubstring.py
```diff
@@ -0,0 +1,63 @@
...
+
+@pytest.mark.parametrize("eng", [SedonaDB, PostGIS])
+@pytest.mark.parametrize(
+    ("geom", "start", "end", "expected"),
+    [
+        (None, 0.0, 1.0, None),
+        ("LINESTRING(0 0, 10 10)", None, 1.0, None),  # Start fraction is NULL -> Output is NULL
+        ("LINESTRING(0 0, 10 10)", 0.0, None, None),  # End fraction is NULL -> Output is NULL
+        (None, None, None, None),
+        ("LINESTRING (0 0, 10 0)", 0.2, 0.8, "LINESTRING (2 0, 8 0)"),
```
paleolimbot, line 47: Can you add a case for LINESTRING EMPTY input?

rust/sedona-functions/src/st_linesubstring.rs
```diff
@@ -0,0 +1,296 @@
...
+        ]).unwrap();
+
+        let actual_array = match tester.invoke(vec![
+            ColumnarValue::Array(geoms_input),
+            ColumnarValue::Array(starts_input),
+            ColumnarValue::Array(ends_input)
+        ]).unwrap() {
+            ColumnarValue::Array(arr) => arr,
+            _ => panic!("Expected array block context output"),
+        };
```
paleolimbot: I think this works:

```suggestion
        let actual_array = tester.invoke_arrays(vec![geoms_input, starts_input, ends_input]).unwrap();
```

rust/sedona-functions/src/st_linesubstring.rs
```diff
@@ -0,0 +1,296 @@
...
+        let actual_point = tester.invoke_scalar_scalar_scalar("LINESTRING(0 0, 10 10)", 0.5, 0.5).unwrap();
+        let expected_point = tester.invoke_wkb_scalar(Some("POINT(5 5)")).unwrap();
+        assert_scalar_equal(&actual_point, &expected_point);
+
+
+        let geoms_input = tester.invoke_wkb_array(vec![
+            Some("LINESTRING(0 0, 10 10)"),
+            None,
+            Some("LINESTRING(0 0, 10 10)")
+        ]).unwrap();
```
paleolimbot: I think this may have been what you meant (I'm pretty sure our testing function is called `create_array()` here):

```suggestion
        let geoms_input = create_array(&[
            Some("LINESTRING(0 0, 10 10)"),
            None,
            Some("LINESTRING(0 0, 10 10)")
        ]);
```

paleolimbot (commented):
Thank you for the edits!

I pushed an update to fix the merge conflict since it was preventing CI from running...this is getting close!

rust/sedona-functions/src/st_linesubstring.rs
sapienza88: ok

rust/sedona-functions/src/st_linesubstring.rs
sapienza88: it cannot find create_array in the scope...

python/sedonadb/tests/functions/test_linesubstring.py
sapienza88: done

rust/sedona-functions/src/st_linesubstring.rs
paleolimbot: You'll have to import it from `sedona_testing::create`!

sapienza88:
@paleolimbot not sure why Git complains about a Cargo.lock conflicting, I removed it from my local code...

python/sedonadb/tests/functions/test_linesubstring.py
```diff
@@ -0,0 +1,122 @@
...
+
+import pytest
+from sedonadb.testing import PostGIS, SedonaDB, geom_or_null, val_or_null
+
+
+@pytest.mark.parametrize("eng", [SedonaDB, PostGIS])
+@pytest.mark.parametrize(
+    ("geom", "start", "end", "expected"),
+    [
+        ("LINESTRING EMPTY", 0.0, 1.0, None),
```
paleolimbot, line 26: I added a set of multi-segment cases here that surfaced an error where the endpoint was duplicated in some cases if the proportion landed exactly on a vertex! I also ran these against PostGIS and had to update a few semantics (e.g., EMPTY output is always NULL here)

rust/sedona-functions/src/st_linesubstring.rs
```diff
@@ -0,0 +1,341 @@
...
+
+        executor.execute_wkb_void(|maybe_wkb| {
+            let (Some(wkb), Some(s_f), Some(e_f)) = (
+                maybe_wkb,
+                start_floats_iter.next().unwrap(),
+                end_floats_iter.next().unwrap(),
+            ) else {
+                builder.append_null();
+                return Ok(());
+            };
```
paleolimbot, line 102: I moved these into an iterator + early return for any null, which helps reduce nesting in the implementation

rust/sedona-functions/src/st_linesubstring.rs
```diff
@@ -0,0 +1,341 @@
...
+
+            if !(0.0..=1.0).contains(&e_f) {
+                return exec_err!("end_fraction must be between 0.0 and 1.0 (got {e_f}");
+            }
+
+            if e_f < s_f {
+                return exec_err!(
+                    "end_fraction must be greater than start_fraction (got {e_f} < {s_f})"
+                );
+            }
```
paleolimbot, line 120: PostGIS seems to error for all of these cases. We can relax this at some point if there's a good reason to.

rust/sedona-functions/src/st_linesubstring.rs
```diff
@@ -0,0 +1,341 @@
...
+        let end_floats = end_array
+            .as_any()
+            .downcast_ref::<arrow_array::Float64Array>()
+            .unwrap();
+
+        let mut start_floats_iter = start_floats.iter();
+        let mut end_floats_iter = end_floats.iter();
+
+        let mut wkb_body = Vec::new();
+        let mut cumulative_distances = Vec::new();
```
paleolimbot, line 92: I moved these outside the loop so that the scratch space isn't reallocated for every element (cleared + reused).

rust/sedona-functions/src/st_linesubstring.rs
```diff
@@ -0,0 +1,341 @@
...
+}
+
+fn invoke_scalar(
+    line: &impl LineStringTrait<T = f64>,
+    s_f: f64,
+    e_f: f64,
+    wkb_body: &mut Vec<u8>,
+    cumulative_distances: &mut Vec<f64>,
+    builder: &mut BinaryBuilder,
+) -> Result<(), SedonaGeometryError> {
```
paleolimbot, line 145: I moved this into a standalone function

rust/sedona-functions/src/st_linesubstring.rs
```diff
@@ -0,0 +1,341 @@
...
+
+    let first_segment = cumulative_distances
+        .partition_point(|&d| d <= start_dist)
+        .saturating_sub(1)
+        .min(num_segments - 1);
+
+    let last_segment = cumulative_distances
+        .partition_point(|&d| d < end_dist)
+        .saturating_sub(1)
+        .min(num_segments - 1);
```
paleolimbot, line 185: I added this to limit the subset of the linestring that is considered in the loop (the previous version looped through all vertices, even if only a tiny portion of the linestring is actually needed).

paleolimbot (commented):
@sapienza88 I fixed the merge and did a bit of polishing on top of your implementation to help get this over the line! I've annotated the high level changes...I'll see if Copilot has anything to add here to the bits I added and then we'll merge!

rust/sedona-functions/src/st_linesubstring.rs
sapienza88: Looks good to me!

rust/sedona-functions/src/st_linesubstring.rs
sapienza88: Here whats for?

Issue #772, opened 2026-04-19
python/sedonadb: No mechanism to pass layer parameter to read_pyogrio


paleolimbot:
When reading from a gpkg or other pylori-accessible datasource with SedonaDB, we don't get the ability to either list multiple layers or query anything except the default:

```python
import sedona.db

sd = sedona.db.connect()
sd.options.interactive = True

gpkg = "/Volumes/data/usgs-3dhp/data/3dhp_all_CONUS_20250313_GPKG.gpkg"
sd.read_pyogrio(gpkg, layer="hydro_3dhp_all_waterbody")
#> TypeError: SedonaContext.read_pyogrio() got an unexpected keyword argument 'layer'
sd.read_pyogrio(gpkg, options={"layer": "hydro_3dhp_all_waterbody"})
#> UserWarning: More than one layer found in '3dhp_all_CONUS_20250313_GPKG.gpkg': 'hydro_3dhp_all_catchment' (default), 'hydro_3dhp_all_drainagearea', 'hydro_3dhp_all_flowline', 'hydro_3dhp_all_landscape', 'hydro_3dhp_all_network', 'hydro_3dhp_all_rc_team', 'hydro_3dhp_all_waterbody'. Specify layer parameter to avoid this warning.

```

pratheekrebala:
Proposed a PR to address this in #778

paleolimbot:
Closed in #778!

PR #778, opened 2026-04-22
feat(python/sedonadb): Support specifying layer name and accessing sub-paths inside archives when working with pyogrio sources.

pratheekrebala:
## Summary

- Enables passing pyogrio keyword arguments (e.g., `layer`, `where`, `sql`, `max_features`) through `read_pyogrio(options=...)` by switching from the internal `pyogrio.raw.ogr_open_arrow()` to the public `pyogrio.raw.open_arrow()` API, which accepts these as keyword arguments directly including specifying the `layer=` flag to handle multi-layer files (#772)

- Adds a `path_suffix` option to append a subpath to the resolved GDAL source, enabling reading of formats like FileGeoDatabase (`.gdb`) stored inside `.zip` archives (e.g., `options={"path_suffix": 'subdir/test.gdb')`). This is equivalent to running `ogrinfo /vsizip/test.zip/subdir/test.gdb`.

### Motivation

Previously, `read_pyogrio()` had no way to pass options through to pyogrio, making it impossible to select a specific layer in a multi-layer file or use GDAL driver-specific open options. For example, reading a named layer from a GeoPackage now works:

```python
con.read_pyogrio("file.gpkg", options={"layer": "my_layer"})
```

For formats like FileGeoDatabase distributed inside .zip archives, GDAL expects a path of the form `/vsizip/archive.zip/data.gdb`. Because the file listing infrastructure resolves real filesystem paths before the Python reader runs, passing a full `/vsizip/...` path directly fails. The path_suffix option bridges this gap:

```python
con.read_pyogrio("path/to/file.zip", options={"path_suffix": "data.gdb"})
```

### DIFF
--- a/python/sedonadb/tests/io/test_pyogrio.py
+++ b/python/sedonadb/tests/io/test_pyogrio.py
@@ -18,6 +18,7 @@
 import io
 import tempfile
 import warnings
+import zipfile
 from pathlib import Path
 
 import geoarrow.pyarrow as ga
@@ -143,6 +144,27 @@ def test_read_ogr_layer_selection(con):
         # Reading with the correct layer name should work
         geopandas.testing.assert_geodataframe_equal(
             con.read_pyogrio(gpkg_path, options={"layer": "my_layer"}).to_pandas(),
+            gdf,
+        )
+
+
+def test_read_ogr_path_suffix(con):
+    series = geopandas.GeoSeries.from_xy([0, 1], [1, 2], crs="EPSG:3857")
+    gdf = geopandas.GeoDataFrame({"val": ["a", "b"], "wkb_geometry": series})
+    gdf = gdf.set_geometry(gdf["wkb_geometry"])
+
+    with tempfile.TemporaryDirectory() as td:
+        gpkg_path = f"{td}/data.gpkg"
+        gdf.to_file(gpkg_path)
+
+        zip_path = f"{td}/archive.zip"
+        with zipfile.ZipFile(zip_path, "w") as zf:
+            zf.write(gpkg_path, "nested/data.gpkg")
+
+        geopandas.testing.assert_geodataframe_equal(
+            con.read_pyogrio(
+                zip_path, options={"path_suffix": "nested/data.gpkg"}
+            ).to_pandas(),
             gdf,
         )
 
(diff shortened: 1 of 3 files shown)


python/sedonadb/tests/io/test_pyogrio.py
```diff
@@ -131,6 +132,43 @@ def test_read_ogr_filter(con):
...
+        geopandas.testing.assert_geodataframe_equal(
+            con.read_pyogrio(gpkg_path, options={"layer": "my_layer"}).to_pandas(),
+            gdf,
+        )
+
+
+def test_read_ogr_path_suffix(con):
+    series = geopandas.GeoSeries.from_xy([0, 1], [1, 2], crs="EPSG:3857")
+    gdf = geopandas.GeoDataFrame({"val": ["a", "b"], "wkb_geometry": series})
+    gdf = gdf.set_geometry(gdf["wkb_geometry"])
```
paleolimbot:
```suggestion
    gdf = geopandas.GeoDataFrame({"val": ["a", "b"], "geom": series})
    gdf = gdf.set_geometry(gdf["geom"])
```

python/sedonadb/tests/io/test_pyogrio.py
```diff
@@ -131,6 +132,43 @@ def test_read_ogr_filter(con):
         )
 
 
+def test_read_ogr_layer_selection(con):
+    series = geopandas.GeoSeries.from_xy([0, 1], [1, 2], crs="EPSG:3857")
+    gdf = geopandas.GeoDataFrame({"val": ["a", "b"], "wkb_geometry": series})
+    gdf = gdf.set_geometry(gdf["wkb_geometry"])
```
paleolimbot:
```suggestion
    gdf = geopandas.GeoDataFrame({"val": ["a", "b"], "geom": series})
    gdf = gdf.set_geometry(gdf["geom"])
```

paleolimbot (commented):
Thank you!

In CI I see:

```python
        # column comparison
>       assert_index_equal(
            left.columns, right.columns, exact=check_column_type, obj="GeoDataFrame.columns"
        )
E       AssertionError: GeoDataFrame.columns are different
E       
E       GeoDataFrame.columns values are different (50.0 %)
E       [left]:  Index(['val', 'geom'], dtype='str')
E       [right]: Index(['val', 'wkb_geometry'], dtype='str')

```

and left a few suggestions for how to fix the test. I'm never quite sure what the name of a geometry column will be when I write or read a GeoDataFrame (everybody chooses something different for the "no geometry column name" case).

paleolimbot (approved):
Thank you!

PR #787, opened 2026-04-26
feat(sedona-raster-gdal) add GDAL foundation library for supporting GDAL-based RS functions

Kontinuation:
## Summary

This PR is the starting point of implementing complex RS_* functions requiring GDAL library.

- add the new `sedona-raster-gdal` crate and wire it into the workspace as the root GDAL raster foundation
- introduce the core GDAL conversion and provider layers for in-db, out-db, and mixed rasters, plus GDAL session options
- expose `RasterExecutor` for later GDAL-backed raster functions while intentionally leaving user-facing `RS_*` functions and registrations to follow-up PRs

## Notes
- this PR is the root of the planned stacked raster PR series


rust/sedona-raster-gdal/src/gdal_common.rs
```diff
@@ -0,0 +1,752 @@
...
+        },
+    )
+    .map_err(convert_gdal_err)
+}
+
+/// Normalize out-db raster URLs to GDAL VSI paths.
+///
+/// Supported translations:
+/// - `s3://bucket/key` -> `/vsis3/bucket/key`
+/// - `s3a://bucket/key` -> `/vsis3/bucket/key`
```
paleolimbot, line 316: No need to do it here, but `{rest}.zip` -> `/vsizip/{rest}` is also useful.

rust/sedona-raster-gdal/src/gdal_dataset_provider.rs
```diff
@@ -0,0 +1,993 @@
...
+const DEFAULT_GDAL_VRT_CACHE_SIZE: usize = 32;
+
+pub(crate) fn configure_thread_local_options(
+    gdal: &Gdal,
+    config_options: Option<&ConfigOptions>,
+) -> Result<()> {
+    // Set frequently requested GDAL config options as thread-local options to eliminate the
+    // need for acquiring configs from global config or environment variable, which is very
+    // likely to result in heavy contention in multi-threaded environments.
+    let cpl_debug_enabled = config_options
```
paleolimbot, line 98: It is probably a good pattern to always do this within some context like `with_thread_local_options()` that takes care to reset the values on exit, particularly when borrowing the shared object from something like rasterio which may be doing the same thing. During execution we'll often have our own threads from tokio but I think constant folding happens on the main Python thread. We can follow up with this if it's difficult to implement here.

rust/sedona-raster-gdal/src/gdal_dataset_provider.rs
```diff
@@ -0,0 +1,993 @@
...
+    Ok(())
+}
+
+/// Get or create the thread-local `GDALDatasetCache`.
+pub(crate) fn thread_local_cache() -> Result<Rc<GDALDatasetCache>> {
+    TL_GDAL_DATASET_CACHE.with(|cell| {
+        let mut opt = cell.borrow_mut();
+        if let Some(rc) = opt.as_ref() {
+            Ok(Rc::clone(rc))
+        } else {
```
paleolimbot, line 129: I hope there is a future where we can move some of these caches to be session scoped (this feels like it could potentially be large in some cases).

rust/sedona-raster-gdal/src/gdal_common.rs
```diff
@@ -0,0 +1,752 @@
...
+                );
+            }
+            Ok(bytes[0] as i8 as f64)
+        }
+        BandDataType::UInt16 => read_le_f64!(u16, 2),
+        BandDataType::Int16 => read_le_f64!(i16, 2),
+        BandDataType::UInt32 => read_le_f64!(u32, 4),
+        BandDataType::Int32 => read_le_f64!(i32, 4),
+        BandDataType::UInt64 => read_le_f64!(u64, 8),
+        BandDataType::Int64 => read_le_f64!(i64, 8),
```
paleolimbot: Maybe remove these two since they're lossy and you handle/test them separately via the dedicated gdal setters?

rust/sedona-raster-gdal/src/lib.rs
```diff
@@ -0,0 +1,37 @@
...
+//!
+//! - in-db raster to GDAL MEM dataset conversion
+//! - out-db and mixed raster to GDAL VRT dataset conversion
+//! - GDAL datatype and nodata conversion helpers
+//! - path normalization for GDAL VSI-backed raster sources
+
+#[allow(dead_code)]
+mod gdal_common;
+#[allow(dead_code)]
+mod gdal_dataset_provider;
```
paleolimbot, line 33: Do you still need these after the pub use? If you do perhaps add a comment with a link to a PR or issue that will remove it (or just make it pub).

```suggestion
mod gdal_common;
mod gdal_dataset_provider;
```

rust/sedona-raster-gdal/src/utils.rs
```diff
@@ -0,0 +1,167 @@
...
+use crate::gdal_common::{gdal_to_band_data_type, nodata_f64_to_bytes};
+
+/// Load a raster from any GDAL-openable path as an in-db raster `StructArray`.
+///
+/// The `path` can be a regular file path, a `/vsimem/` memory path,
+/// a `/vsicurl/` URL, or any other GDAL virtual filesystem path.
+pub fn load_as_indb_raster(gdal: &Gdal, path: &str) -> Result<StructArray> {
+    // Open dataset from path
+    let dataset = gdal
+        .open_ex_with_options(
```
paleolimbot: Should we file a follow-up issue to add tests for this?

paleolimbot (approved):
Thank you! The conversion to/from GDAL Dataset is huge. I'm happy to file follow-ups for any of my comments here.

cc @james-willis

rust/sedona-raster-gdal/src/utils.rs
Kontinuation: I was intended to move it to a dedicated patch. I'll remove it from this patch.

rust/sedona-raster-gdal/src/gdal_common.rs
Kontinuation: I have patched these arms to return Err to inform the developer to handle these types specially.

rust/sedona-raster-gdal/src/lib.rs
Kontinuation: Filed https://github.com/apache/sedona-db/issues/804 and added comments

paleolimbot (approved):
Thank you!

PR #788, opened 2026-04-26
fix(ci): Fix Windows build path limits and unify rust-cache configuration

pwrliang:
The Windows CI pipeline is currently failing on a forked repo (https://github.com/wherobots/sedona-db-gpu-tester/actions/runs/24955442999/job/73072769657) during the compilation of the sedonadb package, specifically when building the aws-lc-sys Rust dependency.

The root cause is the default Windows MAX_PATH limitation (260 characters). Because Cargo and CMake create deeply nested build directories, the absolute path for target artifacts exceeds this limit. This causes the compiler (GCC/MSYS2) to silently fail to create directory trees, resulting in a fatal file IO error when it tries to write the dependency (.d) files.

Changes

- Dynamic Target Directory: Added a Configure Target Directory step that sets a dynamically shortened SEDONADB_TARGET_DIR on Windows (D:/a/t) to safely bypass the compiler's 260-character limit. Mac and Linux continue to use the standard repository path.

- Dynamic Cache Workspace: Extracted the cache mapping into a new CACHE_WORKSPACE environment variable. This allows the Swatinem/rust-cache@v2 action to correctly locate and cache the build artifacts across all platforms, preserving the caching optimization the original workflow intended

### DIFF
--- a/.github/workflows/r.yml
+++ b/.github/workflows/r.yml
@@ -83,6 +83,13 @@ jobs:
           rustup toolchain install stable-x86_64-pc-windows-gnu --no-self-update
           rustup default stable-x86_64-pc-windows-gnu
 
+      - name: Enable Long Paths support (Windows)
+        if: matrix.config.os == 'windows-latest'
+        shell: pwsh
+        run: |
+          git config --system core.longpaths true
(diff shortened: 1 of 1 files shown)


paleolimbot (approved):
This is a super annoying limitation...thank you for fixing!

PR #790, opened 2026-04-26
ST_HausdorffDistance function

sapienza88:


zhangfengcdt:
Thanks for the contribution! Could you please clarify the use cases you want to support by adding this st function? Also,

1. The UDF needs to be registered in rust/sedona-functions/src/register.rs inside register_scalar_udfs — otherwise it's not callable from SQL.
2. GEOS-backed kernels in this project live in the c/sedona-geos crate (see c/sedona-geos/src/distance.rs) and use GeosExecutor. Could you move the implementation there and drop the new geos dep from rust/sedona-functions/Cargo.toml?

Also, please add the Apache license header to any new files.

sapienza88:
@zhangfengcdt thanks for the comment, I want to first ask if g1.hausdorff_distance(&g2) would work?, is hausdorff_distance already defined?

zhangfengcdt:
> @zhangfengcdt thanks for the comment, I want to first ask if g1.hausdorff_distance(&g2) would work?, is hausdorff_distance already defined?

Yes, geos already support it - https://raw.githubusercontent.com/georust/geos/master/src/geometry.rs

You might want to check how this one works and follow the same pattern:
c/sedona-geos/src/distance.rs

sapienza88:
@zhangfengcdt yet I cannot see the defining function, could you point out to the exact logic defining the hausdorff_distance?

sapienza88:
@zhangfengcdt why the tests are in python while the implementation in rust?

sapienza88:
@paleolimbot this function should implemented in c/sedona-geos/src or rust/sedona-functions/src?

docs/reference/sql/st_hausdorffdistance.qmd
```diff
@@ -0,0 +1,38 @@
...
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+---
+title: ST_HausdorffDistance
+description: Returns the Hausdorff distance between two geometries. This is a measure of how similar two geometries are.
```
paleolimbot, line 19:
```suggestion
description: Returns the Hausdorff distance between two geometries.
```

docs/reference/sql/st_hausdorffdistance.qmd
```diff
@@ -0,0 +1,38 @@
...
+kernels:
+  - returns: double
+    args:
+      - geom1: geometry
+      - geom2: geometry
+  - returns: double
+    args:
+      - geom1: geometry
+      - geom2: geometry
+      - density_frac: double
```
paleolimbot, line 29: We have funny hand-rolled syntax for these, but this should work:

```suggestion
  - returns: double
    args:
      - geometry
      - geometry
  - returns: double
    args:
      - geometry
      - geometry
      - name: density_frac
        type: float64
        description: Densification to apply before calculation
```

python/sedonadb/tests/functions/st_hausdorffdistance.py
```diff
@@ -0,0 +1,49 @@
...
+import pytest
+from sedonadb.testing import geom_or_null, PostGIS, SedonaDB
+
+
+@pytest.mark.parametrize("eng", [SedonaDB, PostGIS])
+@pytest.mark.parametrize(
+    ("geom1", "geom2", "density_frac", "expected"),
+    [
+        (None, None, None, None),
+        ("POINT (0 0)", None, None, None),
```
paleolimbot, line 27: Can you add `(None, "POINT (0 1)", None, ...)`, `("POINT (0 1)", "POINT (0 1)", None)`, and checks for EMTPY behaviour (empty linestring + non empty linestring and reversed)

python/sedonadb/tests/functions/st_hausdorffdistance.py
```diff
@@ -0,0 +1,49 @@
...
+    ],
+)
+def test_st_hausdorff_distance(eng, geom1, geom2, density_frac, expected):
+    eng = eng.create_or_skip()
+
+    if density_frac is None:
+        query = f"SELECT ST_HausdorffDistance({geom_or_null(geom1)}, {geom_or_null(geom2)})"
+    else:
+        query = f"SELECT ST_HausdorffDistance({geom_or_null(geom1)}, {geom_or_null(geom2)}, {density_frac})"
+
```
paleolimbot, line 44: Can you separate the tests for the version with and without `density_frac` (i.e., `def test_st_hasdorff_distance_density_frac()`? This will let us separate the "null" from "not specified" case.

python/sedonadb/tests/functions/st_hausdorffdistance.py
paleolimbot, line 1: This can be added next to wherever `st_distance` is tested (I'm not sure it needs its own file)

c/sedona-geos/src/st_hausdorffdistance.rs
```diff
@@ -0,0 +1,86 @@
...
+
+                    builder.append_value(dist);
+                }
+                _ => builder.append_null(),
+            }
+            Ok(())
+        })?;
+
+        executor.finish(Arc::new(builder.finish()))
+    }
```
paleolimbot, line 85: Can you add Rust tests (you can use the tests for st_distance as a template). The rust tests check for basic functionality and ensure that scalar and array arguments work everywhere.

paleolimbot (commented):

PR #799, opened 2026-04-28
refactor(rust/sedona-spatial-join): Support wraparound rectangles in EvaluatedGeometryArray

### LINKED ISSUE #782: rust/sedona-spatial-join: Add proper support for wraparound rectangles in geography join
After https://github.com/apache/sedona-db/pull/775 we have a geography join! To get things working right away I didn't do anything special for wraparound rectangles in the `EvaluatedGeometryArray` except set the longitude extent to the full longitude range (-180 to 180) if they show up.

The index itself is fairly well equipped to deal with this...it deduplicates build indices on the way out of the index such that a query that intersects both the east and west halves of the wraparound box shows up once in the result. We might want to pawn off that deduplication to the refiners since it might not apply to anything except a geography join, or conversely, we might want to do a better job doing wraparound bounding on geometries (there is a mechanism to do this in sedona-geometry or arrow-rs, I forget where exactly I put it).

One place that the rectangles from the EvaluatedGeometryArray are explicitly used is to divvy up geometries into partitions for the out of core join. This function would have to be made wraparound-aware. We would also have to make sure that when inserting into the index that a wraparound box is inserted twice.

paleolimbot:
This PR adds support for wraparound bounds in the `EvaluatedGeometryArray`. It introduces a new struct, `Bounds2D`, which are just f32 bounds with wraparound support. The main consequence of this was removing the bounding box from the `GeometrySummary` and reusing the bounds from the evaluated array when constructing the `GeoStatistics`. This solves the issue of geographies having possibly bogus `GeoStatistics`, although does introduce the possibility that we trigger some not-yet implemented behaviour in the out-of core join (that may require further eliminating Rect usage in the kdb and rtree partitioners).

Closes #782.


c/sedona-libgpuspatial/src/lib.rs
```diff
@@ -113,9 +112,10 @@ mod sys {
         /// Inserts a batch of bounding boxes into the index.
         /// Each rectangle is represented as a `Rect<f32>` with minimum and maximum x and y coordinates.
         /// This method accumulates these rectangles until `finish_building` is called to finalize the index.
-        /// The method can be called multiple times to insert data in batches before finalizing.
-        pub fn push_build(&mut self, rects: &[Rect<f32>]) -> Result<()> {
-            // Re-interpreting Rect<f32> as flat f32 array (xmin, ymin, xmax, ymax)
+        /// The method can be called multiple times to insert data in batches before finalizing. The values
+        /// in rects are ordered (xmin, ymin, xmax, ymax).
+        pub fn push_build(&mut self, rects: &[(f32, f32, f32, f32)]) -> Result<()> {
+            // Re-interpreting rects as a flat f32 array (xmin, ymin, xmax, ymax)
```
paleolimbot, line 118: cc @pwrliang to make sure I didn't muck this up...the evaluated geometry array is no longer using Rect so I updated this to just use f32s. I don't think the GPU tests actually run here but this does build.

rust/sedona-spatial-join-geography/src/join_provider.rs
```diff
@@ -177,28 +176,16 @@ fn try_new_evaluated_array_impl(
...
-                if min_x > max_x {
-                    min_x = -180.0;
-                    max_x = 180.0;
-                }
-
-                let (min_x, min_y, max_x, max_y) = f64_box_to_f32(min_x, min_y, max_x, max_y);
-                let rect = Rect::new(coord!(x: min_x, y: min_y), coord!(x: max_x, y: max_y));
-                rect_vec.push(Some(rect));
+            if let Some((min_x, min_y, max_x, max_y)) = maybe_bounds {
+                rect_vec.push(Bounds2D::new((min_x, max_x), (min_y, max_y)));
```
paleolimbot, line 180: This is the main point of the PR...eliminate the expansion of potentially very small feature bounds into the entire width of the earth.

rust/sedona-spatial-join-geography/tests/spatial_join_integration.rs
paleolimbot, line 1: These tests were copied by via LLM from the sedona-spatial-join crate, taking the random geometry parameters from the Python tests. I needed this to debug a deduplication issue but it is good to have regardless.

rust/sedona-spatial-join-gpu/src/index/gpu_spatial_index_builder.rs
```diff
@@ -212,11 +208,12 @@ impl SpatialIndexBuilder for GPUSpatialIndexBuilder {
...
-            for (idx, rect_opt) in rects.iter().enumerate() {
-                if let Some(rect) = rect_opt {
-                    native_rects.push(*rect);
-                } else {
+            for (idx, rect) in rects.iter().enumerate() {
+                if rect.is_empty() {
                     native_rects.push(empty_rect);
+                } else {
+                    let (x, y) = rect.clone().into_inner();
+                    native_rects.push((x.0, y.0, x.1, y.1));
```
paleolimbot: cc @pwrliang...did I reinterpret this correctly?

rust/sedona-spatial-join/src/index/default_spatial_index_builder.rs
```diff
@@ -142,25 +164,41 @@ impl DefaultSpatialIndexBuilder {
...
+                    let data_idx = rtree_builder.add(x.0, y.0, x.1, y.1);
+                    batch_pos_vec[data_idx as usize] = (batch_idx as i32, idx as i32);
+                }
+
+                if !right.is_empty() {
+                    let (x, y) = right.into_inner();
+                    let data_idx = rtree_builder.add(x.0, y.0, x.1, y.1);
+                    batch_pos_vec[data_idx as usize] = (batch_idx as i32, idx as i32);
+                }
             }
```
paleolimbot, line 216: This is the main change in the spatial index: we can now insert two rectangles into the index instead of one.

rust/sedona-spatial-join/src/index/default_spatial_index.rs
```diff
@@ -538,9 +564,21 @@ impl SpatialIndex for DefaultSpatialIndex {
...
+
+                // Dedup by position: when a geometry spans the antimeridian, it may be indexed
+                // as two separate boxes with different data_idx values that map to the same position.
+                let mut seen_positions: std::collections::HashSet<(i32, i32)> =
+                    std::collections::HashSet::new();
+                candidates.retain(|data_idx| {
+                    let pos = self.inner.data_id_to_batch_pos[*data_idx as usize];
+                    seen_positions.insert(pos)
+                });
+            }
```
paleolimbot: ...and when we query it we have to deduplicate the results. The deduplication happened already before this PR but the DWithin tests exposed that it didn't handle the multiple source rectangles properly.

We may want to solve this performance issue before merging this PR (a `HashSet` is almost certainly slowing us down here).

rust/sedona-spatial-join/src/utils/bounds.rs
```diff
@@ -0,0 +1,277 @@
...
+
+/// A float32 bounding box with wraparound support
+///
+/// This struct is conceptually similar to the Rect<f32> but explicitly supports
+/// wraparound x intervals to ensure raw xmin and xmax values are not misused.
+#[derive(Debug, Clone, PartialEq)]
+pub struct Bounds2D {
+    x: (f32, f32),
+    y: (f32, f32),
+}
```
paleolimbot, line 32: This is the main utility supporting the EvaluatedGeometryArray. I played with making the `IntervalTrait` generic but it was a bit of a pain and I was keen to keep this scoped.

rust/sedona-spatial-join/src/partitioning/stream_repartitioner.rs
```diff
@@ -601,26 +600,27 @@ pub(crate) fn assign_rows(
...
-            for rect_opt in batch.geom_array.rects() {
-                let partition = match rect_opt {
-                    Some(rect) => partitioner.partition(&geo_rect_to_bbox(rect))?,
-                    None => SpatialPartition::None,
+            for rect in batch.geom_array.rects() {
+                let partition = if rect.is_empty() {
+                    SpatialPartition::None
+                } else {
+                    partitioner.partition(&BoundingBox::xy(rect.x(), rect.y()))?
                 };
```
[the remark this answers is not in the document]
paleolimbot: This is valid...if a partition ends up with wraparound bounds, the out of core join will currently error. I opened https://github.com/apache/sedona-db/issues/800 to track.

zhangfengcdt:
Can we also add tests for regions around the poles, hemisphere-spanning polygons, or full-globe geographies?

c/sedona-libgpuspatial/src/lib.rs
pwrliang: The changes look good to me. I also pulled the branch and ran the CI on it https://github.com/wherobots/sedona-db-gpu-tester/actions/runs/25140655727/job/73694467245

paleolimbot: Thanks!

zhangfengcdt (approved):
LGTM

PR #802, opened 2026-04-30
feat(c/sedona-geos): Adds ST_Normalize(geometry) in c/sedona-geos using GEOS

oglego:
## Description
This PR introduces the `ST_Normalize` function, leveraging the **GEOS** library. `ST_Normalize` returns a geometry in its normalized (canonical) form, ensuring consistent representation across different processing steps.

## Key Changes
* **GEOS Integration**: Implemented the core normalization logic using GEOS.
* **Rust Unit Tests**
* **Python Integration Tests**
* **PostGIS Compatibility**: Specific integration tests added to document and verify behavior regarding **M** and **ZM** coordinates, matching PostGIS expectations.

## Tracking
This work is part of the following development tracks:
* **Epic:** #174 (ST function coverage)
* **Feature:** #224 (Implement functions using GEOS library)

### DIFF
--- a/python/sedonadb/tests/functions/test_functions.py
+++ b/python/sedonadb/tests/functions/test_functions.py
@@ -2309,6 +2309,89 @@ def test_st_length(eng, geom, expected):
     ("geom", "expected"),
     [
         (None, None),
+        ("POINT EMPTY", "POINT EMPTY"),
+        ("LINESTRING EMPTY", "LINESTRING EMPTY"),
+        ("POLYGON EMPTY", "POLYGON EMPTY"),
+        ("MULTIPOINT EMPTY", "MULTIPOINT EMPTY"),
+        ("MULTILINESTRING EMPTY", "MULTILINESTRING EMPTY"),
+        ("MULTIPOLYGON EMPTY", "MULTIPOLYGON EMPTY"),
+        ("GEOMETRYCOLLECTION EMPTY", "GEOMETRYCOLLECTION EMPTY"),
+        ("POINT (0 0)", "POINT (0 0)"),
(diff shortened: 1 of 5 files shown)


python/sedonadb/tests/functions/test_functions.py
```diff
@@ -2304,6 +2304,89 @@ def test_st_length(eng, geom, expected):
...
+        expected = expected.replace(r"M(", r"M (")
+        expected = expected.replace(r"Z(", r"Z (")
+
+    if isinstance(eng, SedonaDB) and expected is not None:
+        expected = expected.replace(", ", ",")
+        expected = expected.replace(" (", "(")
+
+    eng.assert_query_result(
+        f"SELECT ST_AsText(ST_Normalize({geom_or_null(geom)}))", expected
+    )
```
paleolimbot, line 2387: This *should* work:

```suggestion
    eng.assert_query_result(
        f"SELECT ST_Normalize({geom_or_null(geom)})", expected
    )
```

paleolimbot (approved):
Thank you!

It's worth seeing if the slightly more compact version of the test works (in theory you should be able to put WKT in the `expected` slot but there are some quirks).

python/sedonadb/tests/functions/test_functions.py
oglego: Thank you for reviewing this, I really appreciate it!  I tried the compact version, but it led to formatting-only failures in the Python harness, for example one of the diffs I ran into is:

E                   Got:

paleolimbot: Got it! I will open a follow up to see if we can improve this at some point since at least in theory the testing module is supposed to deal with that 😬

paleolimbot (commented):
Thank you!

PR #805, opened 2026-05-01
feat(rust/sedona-geoparquet): Add GeoParquet 2.0/Parquet-native geometry and geography support to Parquet writer

paleolimbot:
Stacked on https://github.com/apache/sedona-db/pull/797 because this needs Arrow 57.1.

This PR adds support for writing Parquet GEOMETRY and GEOGRAPHY, including statistics for both. We use the escape hatch added in the arrow-rs repo that lets us override the statistics accumulator, which lets us accumulate geography statistics using s2geography.

```python
from pyarrow import parquet
import sedona.db

sd = sedona.db.connect()

sd.funcs.table.sd_random_geometry("Point", bounds=[170, 10, 190, 30]).to_view(
    "pts", overwrite=True
)

sd.sql(
    "SELECT ST_SetSRID(ST_GeogFromWKB(ST_AsBinary(geometry)), 4326) as geog FROM pts"
).to_parquet("geog.parquet", geoparquet_version="2.0")

# Wraparound statistics!
f = parquet.ParquetFile("geog.parquet")
f.metadata.row_group(0).column(0).geo_statistics
# <pyarrow._parquet.GeoStatistics object at 0x12b6ace00>
#   geospatial_types: [1]
#   xmin: 170.02695197909054, xmax: -170.0081463408811
#   ymin: 10.002287497049661, ymax: 29.998259204130765
#   zmin: None, zmax: None
#   mmin: None, mmax: None
```


rust/sedona-geoparquet/src/writer.rs
```diff
@@ -532,8 +574,183 @@ fn append_float_bbox(
...
+    let crs_component = crs
+        .as_ref()
+        .map(|crs| format!(r#""crs":{}"#, crs.to_json()));
+
+    let edges_component = match edges {
+        Edges::Planar => None,
+        // This is where we apply the workaround relative to our usual
+        // serialize_edges_and_crs().
+        Edges::Spherical => Some(r#""algorithm":"spherical""#),
+    };
```
paleolimbot: I filed https://github.com/apache/arrow-rs/issues/9929 for this one

rust/sedona-geoparquet/src/writer.rs
```diff
@@ -77,40 +79,67 @@ pub fn create_geoparquet_writer_physical_plan(
     }
 
     // If there is no geometry, just use the inner implementation
-    let mut output_geometry_column_indices = conf.output_schema().geometry_column_indices()?;
-    if output_geometry_column_indices.is_empty() {
+    let input_geometry_column_indices = conf.output_schema().geometry_column_indices()?;
+    if input_geometry_column_indices.is_empty() {
```
paleolimbot, line 83: There is a lot of churn in this function, which prepares a plan before writing. Basically, we now *always* do a "projection" (that just modifies metadata) because we either have to strip the geoarrow.wkb type for GeoParquet 1.0 and 1.1 (Parquet with the geospatial feature would write that as Geometry and we want plain byte array with no logical type), canonicalize the CRS to write PROJJSON for sure in the logical type (GeoParquet 2.0/no GeoParquet), and/or create some invalid GeoArrow metadata to work around a parquet Rust bug (see above). Finally, sometimes we don't write the key/value metadata (no GeoParquet).

rust/sedona-geoparquet/src/statistics_accumulator.rs
```diff
@@ -0,0 +1,169 @@
...
+        init_geo_stats_accumulator_factory(Arc::new(Self))?;
+        Ok(())
+    }
+}
+
+impl GeoStatsAccumulatorFactory for SedonaGeoStatsAccumulatorFactory {
+    fn new_accumulator(&self, descr: &ColumnDescPtr) -> Box<dyn GeoStatsAccumulator> {
+        if let Some(LogicalType::Geometry { .. }) = descr.logical_type_ref() {
+            return Box::new(ParquetGeoStatsAccumulator::default());
+        }
```
paleolimbot, line 43: To get Geography statistics written we have to do some plumbing to get s2geography's rectangle bounder called from deep within the depths of the Parquet crate. This is the mechanism I added in the PR that enabled Geometry statistics to be written...you can initialize an "accumlator factory" on startup that dishes out dynamic stats accumulators. It's a little unfortunate we need an s2geography dependency here but the fact that we can write stats at all is very cool.

python/sedonadb/tests/io/test_parquet.py
```diff
@@ -483,6 +483,104 @@ def test_write_geoparquet_geography(con, geoarrow_data):
...
+        assert file_kv_metadata is None or b"geo" not in file_kv_metadata
+
+        file.metadata.schema.column(2).logical_type.to_json() == '{"Type": "Geography"}'
+
+        # We should only have stats if s2geography is enabled
+        geo_stats = file.metadata.row_group(0).column(2).geo_statistics
+        if "s2geography" not in sedonadb.__features__:
+            assert geo_stats is None
+        else:
+            assert geo_stats is not None
```
paleolimbot, line 587: Geography statistics + Geography logical type!

python/sedonadb/tests/io/test_parquet.py
```diff
@@ -483,6 +483,104 @@ def test_write_geoparquet_geography(con, geoarrow_data):
...
+        assert table_roundtrip == table
+
+        # Check for metadata and logical type
+        file = parquet.ParquetFile(tmp_parquet)
+        file_kv_metadata = file.metadata.metadata
+        assert b"geo" in file_kv_metadata
+        geo_metadata = json.loads(file_kv_metadata[b"geo"])
+        assert geo_metadata["version"] == "2.0.0"
+
+        file.metadata.schema.column(2).logical_type.to_json() == '{"Type": "Geometry"}'
```
paleolimbot: GeoParquet 2.0!

zhangfengcdt:
I am looking at this code sample
```
sd.sql(
    "SELECT ST_SetSRID(ST_GeogFromWKB(ST_AsBinary(geometry)), 4326) as geog FROM pts"
).to_parquet("geog.parquet", geoparquet_version="2.0")
```

Is there any reason we cannot just do the following?

df.write.option("geoparquet_version", "2.0").parquet("path/to/destination")


In general, close to what spark sql syntax is doing ...

zhangfengcdt:
> > Is there any reason we cannot just do the following?
> 
> `.to_parquet()` is the idiom in Pandas, GeoPandas, DuckDB, and Ibis. Personally I like it because autocomplete works when working interactively (parameter documentation is highlighted as you type).
> 
> Feel free to open an issue about adding a standalone Python package implementing a Spark compatibility layer like DuckDB's to see if there is interest!

Sure, I will create a ticket on that.

zhangfengcdt (approved):

PR #806, opened 2026-05-01
feat(rust/sedona-expr): Add pruning capability for geography type

paleolimbot:
Stacked on https://github.com/apache/sedona-db/pull/805, since without the ability to write the geography type with statistics there is no meaningful way to test this!


james-willis (approved):

Maxxen:
Is it actually spec-compliant to assume that the row-group bounding boxes contain the coverage (e.g. edge interpolated bounds), and not just the pairwise min/max vertex values?

Maxxen:
Awesome, thanks for filling me in! (maybe we need an extra sentence or something in the spec about this as it wasn't really obvious to me)

Issue #791, opened 2026-04-26
python/sedonadb: Add Pandas/GeoPandas-style DataFrame API


jiayuasu:
> **Status (2026-06-09):** In progress — P1 and most of P2 have shipped via a series of stacked PRs. P3 (`Series` + `Scalar`) is the next milestone. See [§ 9 — Shipped status](#9-shipped-status-2026-06-09) at the bottom for the merged-PR breakdown.

## 0. Scope

Review feedback challenged the GeoPandas compatibility layer: most of SedonaDB's perf comes from rethinking workflows in the relational model, and the locked non-goals (no `Index`, no `.apply`, no `.iloc`) make a pure import-change migration impossible anyway. This issue keeps the **relational pandas surface** and defers the **GeoPandas-specific facade** (`.geometry`, `.crs`, `.to_crs`, `.sjoin`, geometry methods on `Series`) to a separate, demand-gated effort.

Cookbook entries that show how to rewrite GeoPandas workflows into the relational model run **alongside** the API build — each entry doubles as a mini design doc for the methods it exercises.

The in-flight prior art is the R bindings: PR #468 added `r/sedonadb/R/expression.R` (454 LoC) + `r/sedonadb/src/rust/src/expression.rs` (8.5 KB). The Python expression layer mirrors that pattern.

## 1. Motivation

SedonaDB's Python package today is SQL-driven: users get a lazy `DataFrame` from `sd.sql("SELECT ...")`, `sd.read_parquet(...)`, or `sd.create_data_frame(...)`, and the only way to transform it is to write another SQL string. The target audience — Python data scientists who use pandas — expects a different surface entirely: a `DataFrame` they can index with `df["col"]`, a `Series` with `.mean()` / `.str.lower()` / `.fillna()`, a `groupby().agg(...)` chain, and `df.merge(other, on=...)`.

**Goal:** make the relational surface of the SedonaDB Python interface as close to pandas as is feasible on top of a SQL-backed lazy engine. Where pandas idioms force a real semantic conflict with the engine, follow pandas at the surface and document the deviation. Where they would gut the engine, refuse, and offer a clear alternative.

Explicit non-goals (locked):

- No row labels / `Index` / `.loc` / `.iloc` / `.reset_index` / `.set_index`.
- No `.apply(func, axis=1)` — would force per-row Python over a SQL plan.
- **No GeoPandas facade in this scope.** Tracked separately; gated on demand signal.

## 2. Current state

Relevant files:

[...]

PR #807, opened 2026-05-02
feat(python/sedonadb): add Expr foundation

jiayuasu:
Adds the foundation of a Python expression layer that wraps DataFusion's logical `Expr` through PyO3, mirroring the pattern used in the R bindings (`r/sedonadb/src/rust/src/expression.rs`).

This is the first of four small stacked PRs that together implement Phase P1 of #791. This one ships only the foundation:

- `sedonadb.expr.Expr` — column-expression wrapper with `alias`, `cast`, `is_null`, `is_not_null`, `isin`, `negate`.
- `sedonadb.expr.col(name, qualifier=None)` — reference a column by name (with an optional table qualifier for joins).

Operator overloading and DataFrame integration land in follow-up PRs. `Expr` objects are pure syntax: they are not bound to a DataFrame at construction time, and column-validity errors surface only when an `Expr` is consumed by a DataFrame method.

The pre-existing `sedonadb.expr.literal.lit` (returning `Literal`, the lazy parameterized-query helper) is unchanged by this PR and is intentionally kept off `sedonadb.expr`'s package-level surface.

## Test plan
- [x] Unit tests in `tests/expr/test_expression.py` covering construction, qualified columns, alias, cast, null checks, `isin`, `negate`, chaining, and the `__init__` type guard.
- [x] No regressions in the existing `tests/test_dataframe.py`.
- [x] CI green.

### DIFF
--- a/python/sedonadb/python/sedonadb/expr/__init__.py
+++ b/python/sedonadb/python/sedonadb/expr/__init__.py
@@ -14,3 +14,8 @@
 # KIND, either express or implied.  See the License for the
 # specific language governing permissions and limitations
 # under the License.
+
+from sedonadb.expr.expression import Expr, col, lit
+from sedonadb.expr.literal import Literal
+
+__all__ = ["Expr", "Literal", "col", "lit"]
--- a/python/sedonadb/python/sedonadb/expr/__init__.py
+++ b/python/sedonadb/python/sedonadb/expr/__init__.py
@@ -15,7 +15,7 @@
 # specific language governing permissions and limitations
 # under the License.
 
-from sedonadb.expr.expression import Expr, col, lit
-from sedonadb.expr.literal import Literal
+from sedonadb.expr.expression import Expr, col
+from sedonadb.expr.literal import Literal, lit
 
 __all__ = ["Expr", "Literal", "col", "lit"]
--- a/python/sedonadb/python/sedonadb/expr/__init__.py
+++ b/python/sedonadb/python/sedonadb/expr/__init__.py
@@ -16,6 +16,11 @@
 # under the License.
 
 from sedonadb.expr.expression import Expr, col
-from sedonadb.expr.literal import Literal, lit
+from sedonadb.expr.literal import Literal
 
-__all__ = ["Expr", "Literal", "col", "lit"]
+# Note: `lit` is intentionally not re-exported here. It returns a `Literal`
+# (the lazy parameterized-query wrapper), not an `Expr`. Users who need it
+# import it directly from `sedonadb.expr.literal`, matching its role as a
+# SQL-parameter helper rather than part of the Expr construction surface.
+
+__all__ = ["Expr", "Literal", "col"]
--- a/python/sedonadb/python/sedonadb/expr/expression.py
+++ b/python/sedonadb/python/sedonadb/expr/expression.py
@@ -60,8 +60,13 @@ class Expr:
         return Expr(self._impl.cast(target))
 
     def is_null(self) -> "Expr":
-        """Return a boolean expression that is true where this expression is
-        null (matches both SQL NULL and floating-point NaN)."""
+        """Return a boolean expression that is true where this expression
+        is SQL NULL.
+
+        Note that floating-point NaN is *not* matched by `is_null` — the
+        SQL `IS NULL` predicate only matches NULL. A pandas-style
+        NaN-aware helper is planned on the future `Series` type.
+        """
         return Expr(self._impl.is_null())
 
     def is_not_null(self) -> "Expr":
--- a/python/sedonadb/python/sedonadb/expr/expression.py
+++ b/python/sedonadb/python/sedonadb/expr/expression.py
@@ -1,0 +1,111 @@
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+from typing import Any, Iterable
+
+from sedonadb._lib import expr_col as _expr_col
+from sedonadb._lib import expr_lit as _expr_lit
+from sedonadb.expr.literal import Literal
+
+
+class Expr:
+    """A column expression.
+
+    `Expr` represents a logical expression that will be evaluated against a
+    `DataFrame` when the frame is executed. Expressions are pure syntax — they
+    do not carry data and are not bound to a particular frame at construction
+    time. Errors such as referring to a column that does not exist surface only
+    when the expression is consumed (for example, by `DataFrame.select()` or
+    `DataFrame.filter()`).
+
+    Construct an `Expr` with `col(name)` or `lit(value)`.
+    """
+
+    __slots__ = ("_impl",)
+
+    def __init__(self, impl):
+        # impl is the underlying _lib.InternalExpr handle. Users normally
+        # do not construct Expr directly; use col() / lit() instead.
+        self._impl = impl
+
+    def __repr__(self) -> str:
+        return f"Expr({self._impl!r})"
+
+    def alias(self, name: str) -> "Expr":
+        """Return a copy of the expression with a new output name."""
+        return Expr(self._impl.alias(name))
+
+    def cast(self, target) -> "Expr":
+        """Cast the expression to the given Arrow type.
+
+        `target` must be an object exposing the Arrow C schema interface
+        (e.g. `pyarrow.int64()`, `pyarrow.string()`, a `pyarrow.Field`, or any
+        object with `__arrow_c_schema__`). Casting to Arrow extension types is
+        not supported.
+        """
+        return Expr(self._impl.cast(target))
+
+    def is_null(self) -> "Expr":
+        """Return a boolean expression that is true where this expression is
+        null (matches both SQL NULL and floating-point NaN)."""
+        return Expr(self._impl.is_null())
+
+    def is_not_null(self) -> "Expr":
+        """Return a boolean expression that is true where this expression is
+        not null."""
+        return Expr(self._impl.is_not_null())
+
+    def isin(self, values: Iterable[Any]) -> "Expr":
+        """Return a boolean expression that is true where this expression
+        equals any of the given values."""
+        coerced = [_to_expr(v) for v in values]
+        return Expr(self._impl.isin([e._impl for e in coerced], False))
+
+    def negate(self) -> "Expr":
+        """Return the arithmetic negation of this expression."""
+        return Expr(self._impl.negate())
+
+
+def col(name: str) -> Expr:
+    """Reference a column by name.
+
+    Examples:
+        >>> from sedonadb.expr import col
+        >>> col("x").alias("y")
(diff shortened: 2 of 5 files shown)


python/sedonadb/src/expr.rs
```diff
@@ -0,0 +1,203 @@
...
+    /// Steps:
+    ///
+    /// 1. Pull an `arrow_schema::Field` out of `target` via the Arrow
+    ///    PyCapsule schema interface (any object exposing
+    ///    `__arrow_c_schema__` works — `pyarrow.DataType`,
+    ///    `pyarrow.Field`, etc.).
+    /// 2. Reject Arrow extension types up front. SedonaDB's spatial types
+    ///    are extension types over WKB; users who reach for `cast` on
+    ///    those almost certainly want a different operation, so we surface
+    ///    a clear error rather than silently dropping the extension.
```
paleolimbot, line 114: 👍  (I added support for these in DataFusion 53 that we can transform using an optimizer rule to a scalar function call). We can support these before DataFusion 53 here if we need to by inserting the scalar function call directly (I may give this a go in the near future since it's maybe useful for geometry/geography).

python/sedonadb/src/expr.rs
```diff
@@ -0,0 +1,203 @@
...
+}
+
+/// Construct an unqualified column reference: `Expr::Column("x")`.
+///
+/// Qualified columns (e.g. `t.x`) are not exposed yet; the Python
+/// `col()` helper takes only a single name. When we add joins and
+/// multi-table references we can grow this to accept an optional
+/// table qualifier, matching the R side's `column(name, qualifier)`.
+#[pyfunction]
+pub fn expr_col(name: &str) -> PyExpr {
```
paleolimbot: If it's easy you can probably just add a qualifier argument here while you're at it

python/sedonadb/tests/expr/test_expression.py
```diff
@@ -0,0 +1,147 @@
...
+import pytest
+
+from sedonadb.expr import Expr, col, lit
+
+
+def test_col_returns_expr():
+    e = col("x")
+    assert isinstance(e, Expr)
+    assert e._impl.variant_name() == "Column"
+    assert "x" in repr(e)
```
paleolimbot: Testing the exact repr output is probably easy to do and is a slightly better test

python/sedonadb/python/sedonadb/expr/expression.py
```diff
@@ -0,0 +1,116 @@
...
+    return Expr(_expr_col(name))
+
+
+def lit(value: Any) -> Expr:
+    """Wrap a Python value as a literal expression.
+
+    Accepts the same value types as `sedonadb.expr.literal.lit`, including
+    Python scalars, pyarrow arrays/scalars, and Shapely geometries. Returns an
+    `Expr` suitable for composition with column expressions.
+    """
```
paleolimbot: FWIW there is already a `lit()` function that returns `Literal`. Literal is intentionally lazy so (the python wrappers around) different functions can interpret then differently if they want to (e.g., RS_Intersects() doesn't actually need to convert a whole rasterio object to an Arrow scalar, which is expensive, to compute a correct result). Purely theoretical, not currently used, and can change, but that's why it's like that 🙂 .

Probably you can just fold this logic into `_to_expr()` and leave the existing `lit()` (which powers parameterized queries).

python/sedonadb/python/sedonadb/expr/expression.py
```diff
@@ -0,0 +1,116 @@
...
+
+    Construct an `Expr` with `col(name)` or `lit(value)`.
+    """
+
+    __slots__ = ("_impl",)
+
+    def __init__(self, impl):
+        # impl is the underlying _lib.InternalExpr handle. Users normally
+        # do not construct Expr directly; use col() / lit() instead.
+        self._impl = impl
```
paleolimbot, line 57: An `isinstance()` check here would be good so that this errors if used incorrectly

python/sedonadb/python/sedonadb/expr/expression.py
```diff
@@ -0,0 +1,116 @@
...
+
+    def __repr__(self) -> str:
+        return f"Expr({self._impl!r})"
+
+    def alias(self, name: str) -> "Expr":
+        """Return a copy of the expression with a new output name."""
+        return Expr(self._impl.alias(name))
```
paleolimbot, line 74: I forget if we have CI checks for this, but we parameter docs and examples for most functions in the Python APIs

paleolimbot (commented):
Very cool! I know you're still working here so feel free to ignore anything that was still in the works.

python/sedonadb/src/expr.rs
jiayuasu: Thanks for the heads-up. Leaving extension-type cast as a clear error in this PR. Happy to revisit with the DF53 optimizer rule once that's available — geometry/geography is the obvious payoff.

python/sedonadb/src/expr.rs
jiayuasu: Done in 0d1a4abe — `expr_col(name, qualifier=None)` on the Rust side and `col(name, qualifier=None)` on the Python side. Test in `test_col_with_qualifier`.

python/sedonadb/tests/expr/test_expression.py
jiayuasu: Reverted in 0d1a4abe — exact repr substrings are back (`x AS y`, `CAST(x AS Int32)`, etc.), with `variant_name()` checks kept alongside as structural anchors. Added a module-level comment explaining the policy so the next person doesn't re-loosen them.

python/sedonadb/python/sedonadb/expr/expression.py
jiayuasu: Got it, that makes sense — thanks for the context. Folded into 0d1a4abe: dropped the public `lit() -> Expr` and moved the Python-value-to-Expr coercion into the private `_to_expr()` helper. The existing `sedonadb.expr.literal.lit` returning `Literal` is untouched and stays re-exported from the package.

python/sedonadb/python/sedonadb/expr/expression.py
jiayuasu: Added in 0d1a4abe — `Expr.__init__` now `isinstance`-checks its argument against `_lib.InternalExpr` and raises `TypeError` with a message pointing at `col()`. Covered by `test_expr_init_rejects_wrong_type`.

python/sedonadb/python/sedonadb/expr/expression.py
jiayuasu: Beefed up in 0d1a4abe. Public `Expr` methods and `col()` now have `Args:` and `Examples:` docstring blocks matching the style of `dataframe.py:head/limit/count`.

python/sedonadb/python/sedonadb/expr/__init__.py
```diff
@@ -14,3 +14,8 @@
 # KIND, either express or implied.  See the License for the
 # specific language governing permissions and limitations
 # under the License.
+
+from sedonadb.expr.expression import Expr, col
+from sedonadb.expr.literal import Literal, lit
+
+__all__ = ["Expr", "Literal", "col", "lit"]
```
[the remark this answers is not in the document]
jiayuasu: Done in f633481f — `lit` is removed from `sedonadb.expr.__init__.py`'s re-exports. It's a `Literal` constructor (the lazy parameterized-query helper), not an `Expr` factory, so surfacing it on the same package as `col()` was misleading. Users continue to `from sedonadb.expr.literal import lit`, matching the pre-PR layout. PR description updated to drop the obsolete bullet.

python/sedonadb/python/sedonadb/expr/expression.py
```diff
@@ -0,0 +1,186 @@
...
+
+    `Expr` represents a logical expression that will be evaluated against a
+    `DataFrame` when the frame is executed. Expressions are pure syntax — they
+    do not carry data and are not bound to a particular frame at construction
+    time. Errors such as referring to a column that does not exist surface only
+    when the expression is consumed (for example, by `DataFrame.select()` or
+    `DataFrame.filter()`).
+
+    Construct an `Expr` with `col(name)`. Plain Python values composed with an
+    `Expr` via operators are coerced to literal expressions automatically.
```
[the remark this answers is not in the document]
jiayuasu: Good catch — fixed in f633481f. The `Expr` docstring now describes only what this PR ships (coercion happens inside methods like `isin`, not via operators) and calls out that operator overloading arrives in a follow-up.

python/sedonadb/tests/expr/test_expression.py
```diff
@@ -0,0 +1,133 @@
...
+
+
+def test_isin_python_scalars():
+    # Plain Python scalars are coerced to literal expressions automatically.
+    e = col("x").isin([1, 2, 3])
+    assert e._impl.variant_name() == "InList"
+    rep = repr(e)
+    assert "IN" in rep
+    assert "Int64(1)" in rep
+    assert "Int64(3)" in rep
```
paleolimbot: A few more of these that should just be checking the repr. (LLMs love to write tests that will almost definitely pass, particularly if they are testing their own implementation 🙂 )

python/sedonadb/tests/expr/test_expression.py
```diff
@@ -0,0 +1,133 @@
...
+
+
+def test_isin_with_expr_values():
+    # Mixed Expr + scalar input — Exprs pass through, scalars are coerced.
+    e = col("x").isin([col("a"), 2])
+    assert e._impl.variant_name() == "InList"
+    rep = repr(e)
+    assert "a" in rep
+    assert "Int64(2)" in rep
```
paleolimbot: This one too

python/sedonadb/tests/expr/test_expression.py
```diff
@@ -0,0 +1,133 @@
...
+
+
+def test_negate():
+    e = col("x").negate()
+    assert e._impl.variant_name() == "Negative"
+    assert "(- x)" in repr(e)
```
paleolimbot: also this one

python/sedonadb/tests/expr/test_expression.py
```diff
@@ -0,0 +1,133 @@
...
+
+
+def test_chain_alias_after_predicate():
+    e = col("x").is_null().alias("missing")
+    assert e._impl.variant_name() == "Alias"
+    assert "missing" in repr(e)
+    assert "IS NULL" in repr(e)
```
paleolimbot: and this one

python/sedonadb/tests/expr/test_expression.py
```diff
@@ -0,0 +1,133 @@
...
+
+
+def test_expr_is_not_bound_to_dataframe():
+    # Constructing an Expr referring to a non-existent column does not error.
+    # Errors surface only at DataFrame consumption.
+    e = col("nonexistent_column_xyz")
+    assert "nonexistent_column_xyz" in repr(e)
```
paleolimbot: and this one

paleolimbot (approved):

python/sedonadb/tests/expr/test_expression.py
jiayuasu: Tightened to exact `repr() == ...` in f89149b5.

python/sedonadb/tests/expr/test_expression.py
jiayuasu: Tightened to exact `repr() == ...` in f89149b5.

python/sedonadb/tests/expr/test_expression.py
jiayuasu: Tightened to exact `repr() == ...` in f89149b5.

python/sedonadb/tests/expr/test_expression.py
jiayuasu: Tightened to exact `repr() == ...` in f89149b5.

python/sedonadb/tests/expr/test_expression.py
jiayuasu: Tightened to exact `repr() == ...` in f89149b5.

PR #811, opened 2026-05-04
feat(rust/sedona-raster-gdal,c/sedona-gdal): Add in-db GDAL raster loading functions

Kontinuation:
## Summary
- add `append_as_indb_raster()` and `dataset_to_indb_raster` to materialize any GDAL dataset as in-db arrow array representation of SedonaDB rasters.
- add focused loader tests covering single-band, multi-band, `UInt64`, and `Int64` GeoTIFF inputs

## Stack
- follows PR https://github.com/apache/sedona-db/pull/787, which has already merged into `main`
- foundation-only PR with no user-facing SQL function. `dataset_to_indb_raster` will be mainly used by functions such as `RS_FromGeoTIFF` or `RS_FromGDAL`. It will also be used by tests for various RS functions for loading in-db raster from test raster files.


rust/sedona-raster-gdal/src/utils.rs
```diff
@@ -0,0 +1,429 @@
...
+        let raster_array = with_gdal(|gdal| load_as_indb_raster(gdal, &path_str)).unwrap();
+        let raster_struct = RasterStructArray::new(&raster_array);
+        let raster = raster_struct.get(0).unwrap();
+        let band1 = raster.bands().band(1).unwrap();
+        let band2 = raster.bands().band(2).unwrap();
+
+        assert_eq!(raster.bands().len(), 2);
+        assert_eq!(band1.metadata().storage_type().unwrap(), StorageType::InDb);
+        assert_eq!(band1.metadata().data_type().unwrap(), BandDataType::UInt8);
+        assert_eq!(band1.metadata().nodata_value().unwrap(), [255u8]);
```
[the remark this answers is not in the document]
Kontinuation: Fixed. I aligned the GeoTIFF fixture with its dataset-level nodata behavior by setting both bands to 255, and added a separate MEM-dataset test that verifies distinct per-band nodata preservation through `dataset_to_indb_raster()`.

rust/sedona-raster-gdal/src/utils.rs
```diff
@@ -0,0 +1,458 @@
...
+            outdb_url: None,
+            outdb_band_id: None,
+        };
+
+        builder
+            .start_band(band_metadata)
+            .map_err(|e| exec_datafusion_err!("Failed to start band: {}", e))?;
+
+        let band_data = read_band_data(dataset, band_idx, width, height, band_data_type)?;
+        builder.band_data_writer().append_value(&band_data);
```
paleolimbot: If I'm reading this correctly, this will heap allocate + load a potentially large `Vec<u8>`, copy it into another allocation, and then discard it.

Can we do something like this?

```rust
let mut writer = builder.band_data_writer();
visit_band_data(dataset, band_idx, width, height, band_data_type, |item| {
    writer.write_unsafe(item);
})?;
```

(or `chunk` instead of `item` if the per-item overhead for that is too high)

paleolimbot (approved):
Thank you!

rust/sedona-raster-gdal/src/utils.rs
Kontinuation: I have added another method `read_as_bytes` for retrieving the band data as `Vec<u8>`, and it could be wrapped into an arrow buffer and directly appended to BinaryViewArray without additional allocation.

paleolimbot (approved):
Thank you!

PR #812, opened 2026-05-04
feat(raster-gdal): add parse_outdb_source helper for the GDAL format driver

james-willis:
## Summary

Adds `parse_outdb_source(uri) -> (Cow<str>, u32)` in
`rust/sedona-raster-gdal/src/source_uri.rs`. It is a **GDAL-format-driver-internal helper**: when an outdb band's `outdb_format` dispatches to GDAL, the loader calls this to pick the source band out of `outdb_uri`. Two recognised forms:

- **SedonaDB convention (GDAL driver only)** — `<uri>#band=N` → `(uri without fragment, N)`.
  `rsplit_once('#')` is used so a trailing `#band=N` wins over an earlier `#anchor` in the URI.
- **GDAL native subdataset URI** — `HDF5:\"x.h5\":/var`, `NETCDF:\"file.nc\":var`, `GTIFF_DIR:N:foo.tif`, …  → `(uri verbatim, 1)`. We don't try to interpret the GDAL grammar; we just pass it through.
- **Plain URI / no fragment / non-`band=` fragment** → `(uri, 1)`.
- **Malformed `#band=` fragment** (non-numeric, zero, negative, > `u32::MAX`) — returns an `Execution` error. The user explicitly asked for a band; we refuse to silently fall back to band 1.

Returns `Cow::Borrowed` for both common cases (no allocation when the input has no fragment, and only the trimmed prefix is borrowed when a `#band=N` is present).

## Scope

This is **not** a SedonaDB-wide URI convention. The schema treats `outdb_uri` as opaque, and `RS_BandPath` and other format-agnostic surfaces continue to do so. The `#band=N` shape only has meaning inside the GDAL format driver — other format drivers (Zarr, MDArray, …) handle their own URIs however they want, and unrelated callers like `RS_BandPath` are untouched.

## Why encode the source band index inside `outdb_uri` (for GDAL)?

#787 introduced an `outdb_band_id: u32` column alongside `outdb_url`. The follow-up work on N-D rasters needs the GDAL driver to also support GDAL native subdataset URIs (HDF5 groups, multi-page GeoTIFFs, NetCDF variables). Those URIs already carry the source-selector inline, so a typed sibling column would be redundant *and* couldn't represent them. Inside the GDAL driver, folding both forms (`#band=N` for plain rasters, native subdataset syntax for everything else) into one parser keeps the driver's URI handling in one place.

## Purely additive

No existing call sites are rewired. The first caller will appear in a follow-up PR that ports the GDAL loader to read the source band selector from `outdb_uri` instead of a separate `outdb_band_id` column.

## Test plan

- [x] `cargo test -p sedona-raster-gdal` — 43 unit tests + 1 doctest pass
- [x] `cargo build -p sedona-schema -p sedona-raster -p sedona-raster-gdal -p sedona-raster-functions -p sedona-testing`
- [x] `cargo test  -p sedona-schema -p sedona-raster -p sedona-raster-gdal -p sedona-raster-functions -p sedona-testing` — 315 unit tests + 4 doctests, 0 failures
- [x] `cargo clippy -p sedona-schema -p sedona-raster -p sedona-raster-gdal -p sedona-raster-functions -p sedona-testing --all-targets -- -D warnings` — clean
- [x] `cargo fmt --all --check` — clean
- [x] `pre-commit run --files rust/sedona-raster-gdal/src/source_uri.rs rust/sedona-raster-gdal/src/lib.rs` — clean

The unit tests in `source_uri::tests` cover: default band 1; explicit `band=N`; malformed `#band=` (zero, negative, overflow, non-numeric) returning a clear Execution error; URL query strings preserved before fragment; local paths; HDF5/NETCDF/GTIFF_DIR subdataset passthrough; trailing `#band=N` winning over earlier `#anchor`; empty URI; `Cow::Borrowed` invariants.

### DIFF
--- a/rust/sedona-raster-gdal/src/lib.rs
+++ b/rust/sedona-raster-gdal/src/lib.rs
@@ -31,6 +31,8 @@ mod gdal_common;
 // Temporary until https://github.com/apache/sedona-db/issues/804 is resolved.
 #[allow(dead_code)]
 mod gdal_dataset_provider;
+#[cfg(test)]
+mod source_uri;
 
 // Re-export main dataset conversion functions
 pub use gdal_common::{
--- a/rust/sedona-raster-gdal/src/source_uri.rs
+++ b/rust/sedona-raster-gdal/src/source_uri.rs
@@ -1,0 +1,223 @@
+// Licensed to the Apache Software Foundation (ASF) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership.  The ASF licenses this file
+// to you under the Apache License, Version 2.0 (the
+// "License"); you may not use this file except in compliance
+// with the License.  You may obtain a copy of the License at
+//
+//   http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+//! GDAL-format-driver-internal parser for out-db raster source URIs.
+//!
+//! When a band's `outdb_format` dispatches to the GDAL driver, the loader
+//! uses this helper to extract a 1-based source band index from `outdb_uri`
+//! via the SedonaDB convention `<uri>#band=N`. The convention is private to
+//! the GDAL driver — the schema and format-agnostic surfaces (e.g.
+//! `RS_BandPath`) treat `outdb_uri` as opaque. Other format drivers handle
+//! their own URIs however they like.
+
+use datafusion_common::{error::Result, exec_err};
+
+/// Parse a SedonaDB out-db source URI into the GDAL-side URI and 1-based
+/// source band index.
+///
+/// Behaviour:
+///
+/// - `<uri>#band=N` where `N` parses as a `u32` in `1..=u32::MAX`: strips
+///   the fragment and returns `(<uri>, N)`.
+/// - `<uri>#band=...` with a value that is not a positive `u32` (zero,
+///   negative, non-numeric, empty, or overflowing `u32`): returns an
+///   `Execution` error. The user explicitly asked for a band; we refuse to
+///   silently substitute a default.
+/// - GDAL-native subdataset URIs (e.g. `HDF5:"x.h5":/var`,
+///   `NETCDF:"x.nc":var`, `GTIFF_DIR:1:multi.tif`) and any URI whose
+///   fragment is not `band=...`: pass through verbatim with default band
+///   index 1.
+/// - Plain URIs without any fragment: pass through verbatim with default
+///   band index 1.
+pub(crate) fn parse_outdb_source(uri: &str) -> Result<(String, u32)> {
+    // rsplit lets a trailing `#band=N` win over any earlier `#anchor` in the
+    // URI — useful for users who append the SedonaDB convention to a URI
+    // that already carries a fragment.
+    if let Some((prefix, fragment)) = uri.rsplit_once('#') {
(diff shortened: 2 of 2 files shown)


rust/sedona-raster-gdal/src/lib.rs
```diff
@@ -31,6 +31,8 @@ mod gdal_common;
 // Temporary until https://github.com/apache/sedona-db/issues/804 is resolved.
 #[allow(dead_code)]
 mod gdal_dataset_provider;
+#[allow(dead_code)]
+mod source_uri;
```
paleolimbot:
```suggestion
#[cfg(test)]
mod source_uri;
```

I very much dislike merging allow dead code because it is very easy to accumulate it (at one point I removed hundreds of lines from sedona-spatial-join that ended up being unused and unnoticed). LLMs love adding it because it makes clippy errors go away. DataFusion no longer allows it (you must use `expect(dead_code)` if you absolutely must).

rust/sedona-raster-gdal/src/source_uri.rs
```diff
@@ -0,0 +1,237 @@
...
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    fn parse_ok(uri: &str) -> (String, u32) {
+        let (s, n) = parse_outdb_source(uri).expect("expected Ok");
+        (s.into_owned(), n)
+    }
```
paleolimbot: These will give more informative backtraces in the failures if you just call `unwrap()` in the test

rust/sedona-raster-gdal/src/source_uri.rs
```diff
@@ -0,0 +1,237 @@
...
+/// - GDAL-native subdataset URIs (e.g. `HDF5:"x.h5":/var`,
+///   `NETCDF:"x.nc":var`, `GTIFF_DIR:1:multi.tif`) and any URI whose
+///   fragment is not `band=...`: pass through verbatim with default band
+///   index 1.
+/// - Plain URIs without any fragment: pass through verbatim with default
+///   band index 1.
+///
+/// On success the returned URI is always borrowed from the input; this
+/// function never allocates on the happy path.
+pub(crate) fn parse_outdb_source(uri: &str) -> Result<(Cow<'_, str>, u32)> {
```
paleolimbot: Unless you think there will be a performance issue here, it's probably easier to just return a String so that the caller doesn't have to deal with lifetimes.

paleolimbot (approved):
A few optional suggestions...thank you!

rust/sedona-raster-gdal/src/lib.rs
james-willis: done, I'll also remove the cfg(test) in the next PR once this is used.

rust/sedona-raster-gdal/src/source_uri.rs
james-willis: done. Probably doesnt even wind up saving an allocation once we pass into GDAL.

rust/sedona-raster-gdal/src/source_uri.rs
james-willis: done

rust/sedona-raster-gdal/src/source_uri.rs
```diff
@@ -0,0 +1,215 @@
...
+    }
+
+    fn parse_err_msg(uri: &str) -> String {
+        parse_outdb_source(uri).unwrap_err().to_string()
+    }
+
+    #[test]
+    fn no_fragment_defaults_to_band_one() {
+        assert_eq!(
+            parse_ok("s3://bucket/file.tif"),
```
paleolimbot: nit: 

```suggestion
            parse_outdb_source("s3://bucket/file.tif").unwrap(),
```

I think your test function is now redundant (the unwrap being within the `#[test]` is what generates the better failure)

paleolimbot (approved):
Optional nit (feel free to merge whenever after an approve when you're happy with it 🙂 )

PR #813, opened 2026-05-05
[WIP] feat(raster): view machinery for non-identity band views

james-willis:
## Summary

Integrates the view-machinery layer into the raster type. Bands can be constructed and read with non-identity `view` entries — slices, broadcasts, axis permutations, and reverse-iteration. The byte-access surface is a single zero-copy accessor, `nd_buffer()`, which exposes the source buffer plus the visible region's shape/strides/offset; consumers that need flat row-major bytes borrow them via `NdBuffer::as_contiguous()`, which errors on strided layouts rather than allocating.

Builds on two pieces that already landed on `main`: the `ViewEntries` view-spec module (`validate` / `compose` / `visible_shape` / `is_identity`, #934) and the i64 `source_shape` column (#927). This PR wires them into band construction and reads.

## What's in this PR

Three layers, plus the GDAL bridge:

**Trait surface — zero-copy byte access** (`rust/sedona-raster/src/traits.rs`)
- `NdBuffer::is_contiguous()` — pure function of `(shape, strides, data_type)`. C-order packed-stride check, innermost-first, **offset-agnostic** so an outer-axis slice from a nonzero offset still counts as contiguous. Broadcast (stride 0), reverse (negative stride), permuted, gapped, and inner-strided layouts are not contiguous; a zero-extent axis is trivially contiguous.
- `NdBuffer::as_contiguous() -> Result<&[u8]>` — borrows the visible bytes zero-copy when contiguous, else errors directing the caller to `RS_EnsureContiguous` (#899). Never allocates.
- Unit tests for the contiguity classifier.

**Builder** (`rust/sedona-raster/src/builder.rs`)
- `start_band_with_view()` API (args bundled in `StartBandWithViewArgs`).
- `with_view()` API (args bundled in `WithViewArgs`) — compose a new view over an existing band.
- Tests cover the construction and Arrow-serialization paths: identity-view equivalence with `start_band`, 0-D rejection, view-step validation, the `view` column being NULL for identity bands, and Arrow IPC round-trip.

**Reader** (`rust/sedona-raster/src/array.rs`)
- `RasterRefImpl::band` resolves the band row, decodes the data type, and composes view -> byte strides + byte offset (`compose_band_layout` / `check_band_buffer_bounds`) with checked arithmetic; malformed views and arithmetic overflows route through `sedona_internal_datafusion_err!` so they surface with the standard "SedonaDB internal error" framing.
- `nd_buffer()` returns the source buffer + shape/strides/offset (zero-allocation strided access) and is the sole byte accessor. OutDb bands return an error from `nd_buffer()` (backend resolvers are tracked separately).
- Reader-level byte-layout tests: identity and outer-axis-slice contiguity (zero-copy borrow), strided `nd_buffer` strides across dtypes, broadcast and negative-step rejection, and permutation+slice composition.

**GDAL bridge** (`rust/sedona-raster-gdal/src/{gdal_common,gdal_dataset_provider}.rs`)
- `raster_ref_to_gdal_mem` borrows each band via `nd_buffer()?.as_contiguous()?`. `is_2d()` already requires an identity view, so the borrow always succeeds — no owned-bytes copy is threaded through the dataset provider.


rust/sedona-schema/src/raster.rs
```diff
@@ -115,8 +115,11 @@ impl RasterSchema {
     /// describing how the visible axis maps onto the band's source shape.
     /// The field is nullable: a null view denotes the identity view
     /// `[(i, 0, 1, source_shape[i]) for i in 0..ndim]` and is the canonical
-    /// representation for any band whose data has not been sliced. See
-    /// `RasterSchema` doc for full semantics.
+    /// representation for any band whose data has not been sliced. An
+    /// empty (non-null, zero-length) list is read equivalently to NULL
+    /// for forward compatibility with producers that emit one form versus
+    /// the other; writers should prefer NULL. See `RasterSchema` doc for
+    /// full semantics.
```
paleolimbot: Can we just say it has to be NULL since we just invented this?

rust/sedona-raster/src/traits.rs
```diff
@@ -667,6 +673,188 @@ pub trait BandRef {
...
+}
+
+/// True iff `view` is the canonical identity over a C-order source buffer:
+/// every visible axis `k` is a no-op (no permutation, no slice offset, no
+/// stride/reverse, full coverage of the corresponding source axis).
+///
+/// Identity views let the reader borrow the underlying `data` column
+/// directly through `BandRef::contiguous_data()` and `BandRef::data()`
+/// with no allocation or copy. Non-identity views must materialize a
+/// row-major copy on first byte access.
```
```suggestion
/// with no allocation or copy. Non-identity views must materialize a
/// row-major copy on first byte access when using these functions
/// (use `nd_buffer()` to guarantee that a copy is never produced).
```

rust/sedona-raster/src/traits.rs
```diff
@@ -667,6 +673,188 @@ pub trait BandRef {
...
+///
+/// Uses checked arithmetic at every step. Returns the composed view; the
+/// caller must `validate_view` it against the source_shape before use.
+pub(crate) fn compose_view(
+    input_view: &[ViewEntry],
+    next_view: &[ViewEntry],
+) -> Result<Vec<ViewEntry>, ArrowError> {
+    let mut out = Vec::with_capacity(next_view.len());
+    for (k, next) in next_view.iter().enumerate() {
+        if next.source_axis < 0 || (next.source_axis as usize) >= input_view.len() {
```
paleolimbot: Should `source_axis` be a `usize`? I'm wary of using `as` here given that this is sensitive territory (will panic on out of bounds) although it is probably fine.

rust/sedona-raster/src/traits.rs
```diff
@@ -667,6 +673,188 @@ pub trait BandRef {
...
+        seen[sa] = true;
+
+        if v.steps < 0 {
+            return Err(ArrowError::InvalidArgumentError(format!(
+                "view[{k}].steps = {} must be >= 0",
+                v.steps
+            )));
+        }
+        if v.steps > 0 {
+            let s = source_shape[sa] as i64;
```
paleolimbot: Should source_shape be an array of `i64`?

rust/sedona-raster/src/traits.rs
```diff
@@ -667,6 +673,188 @@ pub trait BandRef {
...
+/// broadcasting. Callers should treat the returned shape as authoritative for
+/// the visible region; `source_shape` is only meaningful in conjunction with
+/// the per-entry `source_axis`.
+///
+/// `validate_view` guarantees `steps >= 0`, so the `as u64` cast is lossless
+/// when the input has already been validated. Callers that haven't validated
+/// yet should still call `validate_view` first.
+pub(crate) fn visible_shape_from_view(view: &[ViewEntry]) -> Vec<u64> {
+    view.iter().map(|v| v.steps as u64).collect()
+}
```
paleolimbot: These should probably be all `i64s` unless there is a compelling reason to use the unsigned version to minimize the number of times we have to `as` between signed and unsigned types.

rust/sedona-raster-gdal/src/gdal_dataset_provider.rs
```diff
@@ -69,6 +69,11 @@ pub(crate) struct RasterDataset<'a> {
     _gdal_mem_source: Option<Rc<Dataset>>,
     /// External datasets referenced by the VRT; kept alive for the lifetime of this struct.
     _gdal_outdb_sources: Vec<Rc<Dataset>>,
+    /// Reader-allocated band bytes that GDAL pointers in the MEM dataset may
+    /// reference (i.e. bytes returned by `BandRef::contiguous_data()` as
+    /// `Cow::Owned`, moved here without a copy). Kept alive for as long as
+    /// the MEM dataset that holds the pointers.
+    _owned_band_bytes: Vec<Vec<u8>>,
```
paleolimbot: I am not sure about this one...I think it would be better for now to reject non-identity views with an error along the lines of `Use RS_ForceContinguous()` (or something). Alternatively, any scalar function's "invoke" body can be in charge of managing a temporary buffer that can be used for this purpose (that may be reused between iterations instead of reallocating).

rust/sedona-raster-gdal/src/gdal_common.rs
```diff
@@ -238,8 +253,39 @@ pub unsafe fn raster_ref_to_gdal_mem<R: RasterRef + ?Sized>(
...
+        // For Cow::Borrowed the pointer is into the StructArray (caller keeps
+        // it alive). For Cow::Owned we move the Vec into `owned_band_bytes`
+        // — no extra copy of the reader's materialization — and point GDAL
+        // at it; the Vec is kept alive alongside the returned Dataset.
+        let data_ptr: *const u8 = match band_data {
+            Cow::Borrowed(b) => b.as_ptr(),
+            Cow::Owned(v) => {
+                // SAFETY (Vec<Vec<u8>> pointer stability):
+                //
+                // GDAL holds a raw pointer into the inner `Vec<u8>`'s heap
```
paleolimbot: I think you should just reject these. The step where we resolve all of the InDB and OutDB rasters into contiguous ranges that GDAL can operate on is special and this is unlikely to be the way we want to end up doing it.

rust/sedona-raster/src/array.rs
```diff
@@ -44,14 +48,26 @@ struct BandRefImpl<'a> {
...
+    byte_offset: i64,
+    /// True iff this view is the identity over a C-order source buffer —
+    /// `contiguous_data()` can then borrow `data` directly. Relies on
+    /// `validate_view` having enforced `view.len() == source_shape.len()`;
+    /// otherwise a "shorter identity" could be wrongly accepted here.
+    is_identity_view: bool,
+    /// Lazy row-major copy of the visible bytes, materialized on the first
+    /// `data()` call against a non-identity view. Identity views borrow
+    /// directly from the Arrow column and never touch this cell.
+    materialized: std::cell::OnceCell<Vec<u8>>,
```
paleolimbot: I don't think you want this: force the `data()` function to accept a `&mut Vec` and force all users of `data()` to manage their scratch space (or operate only on contiguous views). Unless I'm misunderstanding this pattern, this will make it easy to proliferate allocations that are difficult to track and may be large.

rust/sedona-raster/src/array.rs
```diff
@@ -86,14 +102,32 @@ impl<'a> BandRef for BandRefImpl<'a> {
...
+            return &[];
+        }
+        // Identity-view InDb: the column bytes ARE the row-major visible
+        // buffer. Borrow directly from Arrow — no allocation, no copy.
+        if self.is_identity_view {
+            return self.data_array.value(self.band_row);
+        }
+        // Non-identity view: walk strides and cache the row-major copy in
+        // `materialized` so repeat `.data()` calls reuse the same buffer.
+        // The walk is infallible by construction — `RasterRef::band()` runs
```
paleolimbot: You may have to `panic!()` here for now and add another function that accepts externally managed scratch space.

rust/sedona-raster/src/array.rs
```diff
@@ -134,29 +168,206 @@ impl<'a> BandRef for BandRefImpl<'a> {
         }
         // shape and strides are owned by NdBuffer (see its doc comment).
         // Cloning here is cheap — both vecs are O(ndim), a handful of values.
+        // Cast offset i64 -> u64: safe because RasterRefImpl::band asserts
+        // byte_offset >= 0 before storing.
```
paleolimbot: Probably getting rid of the u64s will help here

rust/sedona-raster/src/array.rs
```diff
@@ -134,29 +168,206 @@ impl<'a> BandRef for BandRefImpl<'a> {
...
             buffer: self.data_array.value(self.band_row),
             shape: self.visible_shape.clone(),
             strides: self.byte_strides.clone(),
-            offset: self.byte_offset,
+            offset: self.byte_offset as u64,
             data_type: self.data_type,
         })
     }
 
     fn contiguous_data(&self) -> Result<Cow<'_, [u8]>, ArrowError> {
```
paleolimbot: I am not sure about this one anymore, either. You almost always want externally managed scratch space because the implementations are almost always looping over multiple rasters.

rust/sedona-raster/src/array.rs
```diff
@@ -134,29 +168,206 @@ impl<'a> BandRef for BandRefImpl<'a> {
...
+/// product can't wrap and silently pass the subsequent bound checks.
+/// Returns an error rather than panicking so the caller decides how to
+/// surface the failure.
+fn materialize_strided(
+    buffer: &[u8],
+    visible_shape: &[u64],
+    byte_strides: &[i64],
+    byte_offset: i64,
+    dtype_size: usize,
+) -> Vec<u8> {
```
paleolimbot: The easiest way to let the caller manage the allocation would be to accept a `&mut Vec<u8>` (which may or may not be empty, having possibly used it to materialize a previous band) and have it return a `&[u8]` (the slice of the provided scratch that was written to).

Another interface worth exposing is a visitor-based one (i.e., `visit_strided(..., |element_bytes| { ... })`. Or possibly a templated version where you `visit_strided<32>(..., |element_f32| { ... })`. You can write the version that appends to a buffer in terms of the visitor interface (the compiler should inline the lambda).

rust/sedona-raster/src/array.rs
```diff
@@ -134,29 +168,206 @@ impl<'a> BandRef for BandRefImpl<'a> {
...
+    let inner_steps = visible_shape[inner] as usize;
+    let inner_stride = byte_strides[inner];
+    // `dtype_size` is always >= 1 (every BandDataType has a positive
+    // byte size), so it doesn't need its own guard.
+    let row_bytes_contiguous = inner_stride == dtype_size as i64 && inner_steps > 0;
+
+    // Precompute a small index vector for outer axes (everything except
+    // the innermost). For 1D this is empty and we run a single pass.
+    let mut outer_idx = vec![0u64; ndim.saturating_sub(1)];
+    loop {
```
paleolimbot: This is potentially a bottleneck and would benefit from a dedicated benchmark. Once you have this, you can check to see if skipping the zero improves performance (you can use some of the unsafe slice/vec functions to resize without zeroing, and to do raw copying of memory regions). I am guessing that the looped calculations involving `start..start + len` are slow but compilers are very smart these days and we need a benchmark to be sure.

rust/sedona-raster/src/array.rs
```diff
@@ -242,36 +457,127 @@ impl<'a> RasterRef for RasterRefImpl<'a> {
...
+        let visible_shape = visible_shape_from_view(&view_entries);
 
         let dtype_size = data_type.byte_size() as i64;
-        let mut byte_strides = vec![0i64; source_shape.len()];
-        byte_strides[source_shape.len() - 1] = dtype_size;
+
+        // Internal helper: arithmetic overflow during stride composition on a
+        // validated band is corruption beyond what validate_view catches, so
+        // route it through the standard internal-error framing.
+        let overflow_err = |msg: &str| {
```
paleolimbot: Should validate_view catch this?

rust/sedona-raster/src/array.rs
```diff
@@ -242,36 +457,127 @@ impl<'a> RasterRef for RasterRefImpl<'a> {
...
-        // path, which is not yet implemented. Surface it loudly here rather
-        // than silently rejecting the band, so callers see the standardised
-        // SedonaDB-internal-error framing.
-        if !self.band_view_list.is_null(band_row) {
+        // Read view entries. Identity view is encoded as either a NULL row
+        // or an empty (non-null, zero-length) list — both synthesise the
+        // canonical identity from `source_shape`. A non-empty row is decoded
+        // from the `view` column. Treating empty-non-null and NULL the same
+        // makes the reader tolerant of producers that emit one form versus
+        // the other; the alternative is a misleading "view length 0 must
```
paleolimbot: This is mixing a few levels of abstraction in the `band()` accessor. Can you split it up into smaller named methods with smaller scope to make it clearer what `band()` is doing?

rust/sedona-raster/src/array.rs
```diff
@@ -829,11 +1192,137 @@ mod tests {
...
+                        step: 1,
+                        steps: 2,
+                    },
+                    ViewEntry {
+                        source_axis: 1,
+                        start: 0,
+                        step: 1,
+                        steps: 3,
+                    },
+                ],
```
paleolimbot: I may be misunderstanding the `ViewEntry` system here, but is there a reason why we encode `source_axis` instead of force the source axis to be the same length as the shape that it's viewing?

rust/sedona-raster/src/builder.rs
```diff
@@ -374,6 +377,184 @@ impl RasterBuilder {
...
+    /// Each `ViewEntry` describes one *visible* axis in `dim_names` order:
+    /// `(source_axis, start, step, steps)`. Validates that:
+    /// - `dim_names`, `source_shape`, and `view` have equal length.
+    /// - Across `view`, `source_axis` values form a permutation of
+    ///   `0..ndim` (no axis duplicated, none missing).
+    /// - For each entry with `steps > 0`: `start` and (when `step != 0`)
+    ///   `start + (steps - 1) * step` are in `[0, source_shape[source_axis])`.
+    /// - `steps >= 0`.
+    #[allow(clippy::too_many_arguments)]
+    pub(crate) fn start_band_with_view(
```
paleolimbot: Clippy has a good point here. Define `struct StartBandWithViewArgs<'a> { ... }` with all `pub` fields, which will force callers to write `start_band_with_view(StartBandWithViewArgs { <named params> })`.

You could also expose more than one function or have a BandBuilder of some kind to separate these things into more than one call.

rust/sedona-raster/src/builder.rs
```diff
@@ -374,6 +377,184 @@ impl RasterBuilder {
...
+    /// - **OutDb input** → output is OutDb. The data column stays empty;
+    ///   the input's `outdb_uri` and `outdb_format` are inherited (unless
+    ///   overridden via the explicit `outdb_uri` / `outdb_format` args).
+    ///   The composed view lives alongside the same external pointer —
+    ///   loading is deferred to whoever reads the visible bytes.
+    ///
+    /// Identity-input shortcut: when `input` carries identity view, the
+    /// composed view equals `view` verbatim.
+    #[allow(clippy::too_many_arguments)]
+    pub fn with_view(
```
paleolimbot: Here as well (you can probably use a common Args struct)

rust/sedona-raster/src/builder.rs
```diff
@@ -1717,222 +1902,1410 @@ mod tests {
...
         let mut builder = RasterBuilder::new(1);
         let transform = [0.0, 1.0, 0.0, 0.0, 0.0, -1.0];
         builder
-            .start_raster_nd(&transform, &["x", "y"], &[3, 2], None)
+            .start_raster_nd(&transform, &["x", "y"], &[5, 4], None)
             .unwrap();
+
+        let view = [
+            ViewEntry {
+                source_axis: 0,
```
paleolimbot: A helper struct to manage `Vec<ViewEntry>` would make it more likely that other people will write reasonable tests. Or a macro (`view_entries![0:4, 1:5]`; use Python slice syntax since that's what you'd type in numpy to do this).

rust/sedona-raster/src/builder.rs
```diff
@@ -1717,222 +1902,1410 @@ mod tests {
...
+        builder
+            .start_band_with_view(
+                None,
+                &["y", "x"],
+                &[3, 3], // 3x3 source
+                &view,
+                BandDataType::Float32,
+                None,
+                None,
+                None,
```
paleolimbot: A BandBuilder may help you quite a bit here since there are a lot of `None` that a default constructor could fill in

rust/sedona-raster/src/builder.rs
```diff
@@ -1717,222 +1902,1410 @@ mod tests {
...
+
+    #[test]
+    fn test_nd_buffer_permutation_and_slice_combined() {
+        // 2D source [Y=4, X=3]. View permutes (visible order [X, Y]) and
+        // slices Y from 1, step 2, steps 2. Expected:
+        //   visible_shape = [3, 2]
+        //   byte_strides  = [step_X * stride_X_src, step_Y * stride_Y_src]
+        //                 = [1 * 1, 2 * 3] = [1, 6]
+        //   byte_offset   = start_X * stride_X_src + start_Y * stride_Y_src
+        //                 = 0 * 1 + 1 * 3 = 3
```
paleolimbot: These kinds of tests you probably don't want in this file. This file is about serializing stuff into Arrow and it's tricky to see what's going on...I'm not sure what the perfect abstraction is here but it should probably live in `src/views.rs` with tests and some pub functions that can be benchmarked.

Proabably `struct ViewEntries { inner: Vec<ViewEntry> }` with associated methods would make this a bit cleaner than the `&[ViewEntry]` + free functions.

paleolimbot (commented):
This is very cool!

I think I got there by the end in the inline comments, but I think a `struct ViewEntries` with associated functions in `src/view_entries.rs` where the non serialize-to-arrow and deserialize-from-arrow logic behind this lives will make this clearer.

The main issue I have with the implementation here is that it relies on hidden `Vec<u8>`s that aren't reusable. To manage memory properly we are going to have to register some scratch spaces with the session's memory pool and reuse them between iterations of a loop. Hiding them within various structures for convenience is going to make it tricky to implement a correct pattern going forward.

james-willis:
Somewhat offtopic but are you going to have the same concern with lazy outdb loading as well? (Outdb lazy loading + zarr loader is going to be higher on my priority list next week than this PR)

> The main issue I have with the implementation here is that it relies on hidden Vec<u8>s that aren't reusable. To manage memory properly we are going to have to register some scratch spaces with the session's memory pool and reuse them between iterations of a loop. Hiding them within various structures for convenience is going to make it tricky to implement a correct pattern going forward.

rust/sedona-schema/src/raster.rs
james-willis: done

rust/sedona-raster/src/traits.rs
james-willis: only if we want to change the arrow type. for now I dont want to touch that.

rust/sedona-raster/src/traits.rs
james-willis: I cleaned out some of the u64s. These are the ones I left in the raster ecosystem:

  1. BandRef::width() -> u64 / height() -> u64 (and the RasterMetadata shim equivalents) — 2D-compat accessors. Changing these is API-breaking across sedona-functions, sedona-raster-gdal, callers in tests, etc.
  2. BandRef::raw_source_shape() -> &[u64] and dim_size() -> Option<u64> — same reason; they sit on the public trait.
  3. source_shape, spatial_shape Arrow columns type → could be done as a follow up.

rust/sedona-raster/src/array.rs
james-willis: addressed in another comment

rust/sedona-raster-gdal/src/gdal_common.rs
james-willis: maybe we use the same strategy from the OutDB loader PR and have some RS_EnsureContiguous function.

rust/sedona-raster/src/array.rs
james-willis: I'm going to tweak the whole PR to remove anything that needs a net-new buffer or scratch vector to do its work. instead we will do something similar to what we did with outdb: `RS_EnsureContiguous` will create a new raster with contiguous bands.

rust/sedona-raster-gdal/src/gdal_dataset_provider.rs
james-willis: yeah this EnsureContiguous direction will be the way we go. I stripped out data & contiguous_data from the API to match with this change.

for now these will just error

rust/sedona-raster/src/array.rs
james-willis: the EnsureContiguous is the approach I decided on. the NdBuffer will give a sedona internal error when as_contiguous is called on a non-contiguous buffer.

rust/sedona-raster/src/array.rs
james-willis: removed.

rust/sedona-raster/src/array.rs
james-willis: I just delete this method instead - deferred to when we do the implementation of RS_EnsureContiguous.

rust/sedona-raster/src/array.rs
james-willis: delete this method - comment is moot for now

rust/sedona-raster/src/builder.rs
james-willis: I reduce the tests here. they should now be just builder-integration specific. will confirm before resolving this comment

rust/sedona-raster/src/traits.rs
james-willis: moot with the remove of these transparent copies

rust/sedona-raster/src/traits.rs
james-willis: added a guard before the cast. 

source_axis is a i64 to consolidate all these fields to that one type.

rust/sedona-raster/src/array.rs
james-willis: we do force them to be the same length. The source_axis field is essential to support permutation.

The ordering tells us how to permute the view. the source_axis tells us where to get the data from.

rust/sedona-raster/src/builder.rs
james-willis: done

rust/sedona-raster/src/builder.rs
james-willis: done

rust/sedona-raster/src/builder.rs
james-willis: done

rust/sedona-raster/src/builder.rs
james-willis: I am generally unhappy with the band builder interfaces; I agree I am making them substantially worse. Not sure how you feel about deferring this but I started putting items for band builder improvements in this ticket:

https://github.com/apache/sedona-db/issues/896

rust/sedona-raster/src/array.rs
james-willis: validate_view does not know the dtype_byte_size so it cannot the dtype-scaled product of source strides. Only compose_byte_strides computes that so we put this check here.

james-willis:
subsumed by #1113

PR #816, opened 2026-05-05
chore(ci): Port distance and predicate geography tests from s2geography and run them against BigQuery and PostGIS where possible

paleolimbot:
This PR adds tests for predicates and distance functions. These tests were scraped from s2geography but here they also run against bigquery (where possible) and postgis (where possible). This PR is a pretty good summary of our geography implementation and the functions we support (as well as our behaviour differs from PostGIS or BigQuery). I know this is a lot of lines but I promise it's almost all tests 🙂 

I opened up some follow-ups for tests that probably should pass but don't yet as well as some functions that need implementing upstream in s2geography to complete parity with BigQuery.

- https://github.com/apache/sedona-db/issues/817
- https://github.com/apache/sedona-db/issues/819
- https://github.com/apache/sedona-db/issues/820
- https://github.com/apache/sedona-db/issues/822


c/sedona-s2geography/src/kernels.rs
```diff
@@ -46,7 +53,202 @@ pub fn s2_scalar_kernels() -> Result<Vec<(String, ScalarKernelRef)>> {
...
+    // in s2geography.
+
+    // Binary (geography, geography) -> Boolean
+    let binary_bool_fns = [
+        "st_contains",
+        "st_disjoint",
+        "st_equals",
+        "st_intersects",
+        "st_within",
+    ];
```
paleolimbot, line 69: This was a code change...I added kernels to match things like `st_intersects(geography, NULL)`. This is mostly for parameter binding...params are inserted into a plan as a Null type and the plan is validated before it's replace with something of the correct type.

c/sedona-s2geography/s2geography
paleolimbot, line 1: I fixed a few issues in s2geography and updated the submodule.

python/sedonadb/tests/geography/test_geog_measures.py
```diff
@@ -16,38 +16,283 @@
...
+    ],
+)
+def test_st_length(eng, geog, expected):
+    eng = eng.create_or_skip()
+    # Use reduced precision for PostGIS because it is calculating true ellipsoidal length
+    # and not the spherical approximation
+    if eng.name() == "postgis":
+        eps = 1e-2
+    else:
+        eps = 1e-15
```
james-willis: should we encode this into the assert_query_results method somehow? maybe an is_spheroidal flag?

python/sedonadb/tests/geography/test_geog_distance.py
```diff
@@ -0,0 +1,1065 @@
...
+        pytest.param(
+            "POINT (0 0)", "POINT (0 1)", 111195.10117748393, id="point_distance_point"
+        ),
+        pytest.param(
+            "POINT (0 0)",
+            "POINT (360 90)",
+            10007559.105973553,
+            id="point_distance_wraparound_lng",
+        ),
+        # Point x linestring (point on linestring)
```
james-willis, line 48: should we consider doing some more "difficult" point on linestring, specifically a line that is not axis aligned?

python/sedonadb/tests/geography/test_geog_distance.py
```diff
@@ -0,0 +1,1065 @@
...
+            id="linestring_distance_polygon_through",
+        ),
+        # Polygon x linestring (linestring crosses through, neither vertex inside)
+        pytest.param(
+            "POLYGON ((0 0, 2 0, 0 2, 0 0))",
+            "LINESTRING (-1 0.5, 3 0.5)",
+            0.0,
+            id="polygon_distance_linestring_through",
+        ),
+        # Linestring x polygon (linestring fully outside)
```
james-willis, line 146: instead of transposing these cases, should we build that into the harness/fixture? then we could also automatically test:
* both cases individually
* equality between both cases

python/sedonadb/tests/geography/test_geog_distance.py
```diff
@@ -0,0 +1,1065 @@
...
+        pytest.param(
+            "POINT Z (-1 0 10)",
+            "POLYGON Z ((0 0 12, 2 0 12, 0 2 12, 0 0 12))",
+            333585.3035324518,
+            id="point_z_max_distance_polygon_outside",
+        ),
+    ],
+)
+def test_st_max_distance_zm(geom1, geom2, expected):
+    eng = SedonaDB.create_or_skip()
```
james-willis: should we always parameterize for consistency?

python/sedonadb/tests/geography/test_geog_overlay.py
```diff
@@ -0,0 +1,605 @@
...
+    [
+        # Point + Point: different
+        pytest.param(
+            "POINT (0 0)", "POINT (0 1)", "POINT (nan nan)", id="point_different"
+        ),
+        # Multipoint + Point: disjoint
+        pytest.param(
+            "MULTIPOINT ((0 0), (1 1))",
+            "POINT (2 2)",
+            "POINT (nan nan)",
```
james-willis, line 229: is this really the right thing to return? I would think POINT EMPTY is more correct.

python/sedonadb/tests/geography/test_geog_overlay.py
```diff
@@ -0,0 +1,605 @@
...
+        # Polygon + Polygon: same
+        pytest.param(
+            "POLYGON ((0 0, 10 0, 10 10, 0 10, 0 0))",
+            "POLYGON ((0 0, 10 0, 10 10, 0 10, 0 0))",
+            "POLYGON ((0 0, 10 0, 10 10, 0 10, 0 0))",
+            id="polygon_same",
+        ),
+    ],
+)
+def test_st_union(eng, geom1, geom2, expected):
```
james-willis, line 518: overlapping polygon and overlapping at terminus linestring could be good cases.

python/sedonadb/tests/geography/test_geog_overlay.py
```diff
@@ -0,0 +1,605 @@
...
+        pytest.param(
+            "POLYGON ((0 0, 5 0, 5 5, 0 5, 0 0))",
+            "POLYGON ((10 10, 15 10, 15 15, 10 15, 10 10))",
+            "MULTIPOLYGON (((0 0, 5 0, 5 5, 0 5, 0 0)), "
+            "((10 10, 15 10, 15 15, 10 15, 10 10)))",
+            id="polygon_disjoint",
+        ),
+    ],
+)
+def test_st_symdifference(eng, geom1, geom2, expected):
```
james-willis, line 648: this could use some partially overlapping cases too

python/sedonadb/tests/geography/test_geog_predicates.py
```diff
@@ -49,3 +236,395 @@ def test_st_intersects(eng, geom1, geom2, expected):
...
+            "POINT EMPTY", "POINT EMPTY", True, id="empty_point_equals_empty_point"
+        ),
+        pytest.param(
+            "POINT EMPTY",
+            "LINESTRING EMPTY",
+            True,
+            id="empty_point_equals_empty_linestring",
+        ),
+        # Fast path for identical values
+        pytest.param(
```
james-willis, line 510: should there be a case for wrap around?

python/sedonadb/tests/geography/test_geog_transformations.py
```diff
@@ -37,3 +37,839 @@ def test_st_centroid(eng, geom, expected):
...
+            "LINESTRING EMPTY", "GEOMETRYCOLLECTION EMPTY", id="linestring_empty"
+        ),
+        pytest.param("POLYGON EMPTY", "GEOMETRYCOLLECTION EMPTY", id="polygon_empty"),
+        # Points
+        pytest.param("POINT (0 1)", "POINT (0 1)", id="point"),
+        pytest.param("MULTIPOINT ((0 0), (0 1))", "POINT (0 0.5)", id="multipoint"),
+        # Linestrings
+        pytest.param("LINESTRING (0 0, 0 1)", "POINT (0 0.5)", id="linestring"),
+        pytest.param(
+            "LINESTRING (0 0, 0 1, 0 5)", "POINT (0 2.5)", id="linestring_two_segments"
```
james-willis, line 60: should there be a case where the midpoint isnt colinear?

python/sedonadb/tests/geography/test_geog_transformations.py
```diff
@@ -37,3 +37,839 @@ def test_st_centroid(eng, geom, expected):
...
+            53452519026.41781,
+            id="linestring_positive_distance",
+        ),
+        # Polygon with positive distance: expands the polygon
+        pytest.param(
+            "POLYGON ((0 0, 1 0, 1 1, 0 1, 0 0))",
+            100000.0,
+            88052039626.29015,
+            id="polygon_positive_distance",
+        ),
```
james-willis, line 484: should we do a triangle? rectangle? concave geometry?

python/sedonadb/tests/geography/test_geog_transformations.py
```diff
@@ -37,3 +37,839 @@ def test_st_centroid(eng, geom, expected):
...
+        # Polygon: single ring with snapping
+        pytest.param(
+            "POLYGON ((0.001 0.001, 10.001 0.001, 10.001 10.001, "
+            "0.001 10.001, 0.001 0.001))",
+            1.0,
+            "POLYGON ((0 0, 10 0, 10 10, 0 10, 0 0))",
+            id="polygon_snap",
+        ),
+    ],
```
james-willis, line 666: what about polygon where two coordinates get rounded to the same value? also smae case but it degenerates the polygon

james-willis (approved):
some ideas/nitpicks. up to you

zhangfengcdt (approved):
Nitpick: The only thing we might consider is to see if documents for these functions have been updated to accept both geometry and geography, for example, STPoints, should say "Native implementation to count all the points of a geometry or a geography".

Otherwise, LGTM!

prantogg (approved):
LGTM!

python/sedonadb/tests/geography/test_geog_measures.py
paleolimbot: I added `eng.geography_numeric_epsilon()` and I think I managed to remove all of these

python/sedonadb/tests/geography/test_geog_distance.py
paleolimbot: That's a good point...I added a polar case. These are coming from C++ where I'm absolutely sure nothing planar is happening but here I suppose that's not a given.

python/sedonadb/tests/geography/test_geog_overlay.py
paleolimbot: This is a bug in the testing framework/geoarrow-c. I annotated this like is done elsewhere in the tests

python/sedonadb/tests/geography/test_geog_overlay.py
paleolimbot: I added these for all the overlays (a few had to go in the bug bin of follow-ups)

python/sedonadb/tests/geography/test_geog_predicates.py
paleolimbot: I added one for wraparound for the predicates (the equals one had to go in the bug overflow bin to fix in s2geography)

python/sedonadb/tests/geography/test_geog_transformations.py
paleolimbot: I added a polar case here!

python/sedonadb/tests/geography/test_geog_transformations.py
paleolimbot: These are mostly just checking that it's wired up to the correct buffer-er...as further cases come up we can add to this list.

python/sedonadb/tests/geography/test_geog_transformations.py
paleolimbot: I have a case for this but it doesn't pass yet 😬  https://github.com/apache/sedona-db/issues/822

python/sedonadb/tests/geography/test_geog_distance.py
paleolimbot: I skipped this one for now to avoid more testing infrastructure. The distance cases in C++ have a bit more structure (e.g., checking that dwithin on the exact size works...)

PR #818, opened 2026-05-06
feat(raster-gdal): port indb-raster loader utilities to canonical N-D schema

james-willis:
## Summary

Reintroduces `append_as_indb_raster` and `dataset_to_indb_raster` on top of the canonical N-D raster schema. These helpers were deleted on #749 (`jw/nd-raster-type`) because the version on `main` (added in #811) targets the legacy `BandMetadata` / `StorageType` / `band_data_writer` API that the N-D port removes. This PR ports the helpers and their full test suite to the new builder + reader API so the deletion in #749 isn't permanent.

- Loader rewritten against `start_raster_2d` / `start_band_2d` and `band_data_writer().append_value` — bands are emitted as canonical `["y", "x"]` 2-D bands with the identity view.
- Test suite (single-band GeoTIFF, `UInt64` / `Int64` / `UInt16` nodata round-trips, multi-band GeoTIFF, per-band MEM nodata, multi-raster append) ported to the new `RasterRef` / `BandRef` accessors and `is_indb()`.

## Stacking

> ⚠️ **Blocked on #749** — this branch sits on top of `jw/nd-raster-type`. The first 4 commits in the diff are #749's; only the final commit (`feat(raster-gdal): port indb-raster loader utilities to canonical N-D schema`) is unique to this PR. Review will be cleanest after #749 merges and this PR rebases onto `main`.

## Test plan

- [x] `cargo build -p sedona-raster-gdal`
- [x] `cargo test -p sedona-raster-gdal` (51 passed, including 7 new `utils::tests::*`)
- [x] `cargo clippy -p sedona-raster-gdal --all-targets -- -D warnings`
- [x] `cargo fmt --all --check`


james-willis:
Folded into #749. The loader () has been restored verbatim and the GDAL backend now uses the compatibility shim to stay close to main's API surface, so a separate port is no longer needed.

PR #831, opened 2026-05-11
feat(rust/sedona-raster-gdal): add RS_FromPath

Kontinuation:
## Summary
- add `RS_FromPath` for constructing out-db rasters from GDAL-openable paths
- wire `sedona-raster-gdal` and `SedonaContext` minimally for this standalone child PR
- keep other GDAL-backed raster functions out of scope for follow-up PRs

## Testing
- `cargo test -p sedona-raster-gdal rs_from_path --no-default-features`
- `cargo clippy -p sedona-raster-gdal --no-default-features -- -D warnings`
- `cargo test -p sedona-raster-gdal -p sedona --lib`
- `cargo clippy -p sedona-raster-gdal -p sedona --lib -- -D warnings`


rust/sedona-raster-gdal/src/rs_frompath.rs
```diff
@@ -0,0 +1,216 @@
...
+use crate::gdal_dataset_provider::configure_thread_local_options;
+use crate::utils::append_as_outdb_raster;
+
+pub fn rs_from_path_udf() -> SedonaScalarUDF {
+    SedonaScalarUDF::new(
+        "rs_frompath",
+        vec![Arc::new(RsFromPath)],
+        Volatility::Volatile,
+    )
+}
```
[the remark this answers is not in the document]
Kontinuation: Kept rs_frompath intentionally. RS_FromPath lowercases to rs_frompath, which matches the existing raster SQL naming convention such as RS_BandPath -> rs_bandpath. I also renamed the Rust helper to rs_frompath_udf and added a test to lock in the registered name.

rust/sedona-raster-gdal/src/utils.rs
```diff
@@ -109,6 +104,68 @@ pub fn append_as_indb_raster(dataset: &Dataset, builder: &mut RasterBuilder) ->
...
+            },
+        )
+        .map_err(|e| {
+            exec_datafusion_err!(
+                "Failed to open raster file '{}'(GDAL path '{}'): {}",
+                path,
+                gdal_path,
+                e
+            )
+        })?;
```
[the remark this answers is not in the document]
Kontinuation: Fixed in 2756654f. Updated the error message to include the missing space before the GDAL path segment.

rust/sedona/src/context.rs
```diff
@@ -221,6 +221,10 @@ impl SedonaContext {
             Arc::new(RandomGeometryFunction::default()),
         );
 
+        for udf in sedona_raster_gdal::all_gdal_udfs() {
+            out.ctx.register_udf(udf.into());
+        }
```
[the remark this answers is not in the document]
Kontinuation: Leaving this as-is intentionally. Sedona-DB uses a trusted-query security model similar to PostGIS, where file and URL-backed access is available to users who can execute SQL. We do not plan to add an extra opt-in gate for RS_FromPath in this PR.

rust/sedona-raster-gdal/src/gdal_common.rs
```diff
@@ -313,6 +288,54 @@ pub fn nodata_bytes_to_f64(nodata_bytes: Option<&[u8]>, band_type: &BandDataType
...
+            .no_data_value_u64()
+            .map(|nodata| nodata.to_le_bytes().to_vec()),
+        BandDataType::Int64 => band
+            .no_data_value_i64()
+            .map(|nodata| nodata.to_le_bytes().to_vec()),
+        _ => band
+            .no_data_value()
+            .map(|nodata| nodata_f64_to_bytes(nodata, &band_type)),
+    })
+}
```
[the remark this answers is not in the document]
Kontinuation: Leaving this as-is intentionally. band_nodata_to_bytes is meant to stay self-contained over a GDAL band, and reading the band type from the band object is trivial here. We do not want to push that extraction responsibility onto callers for a negligible optimization.

rust/sedona-raster-gdal/src/rs_frompath.rs
```diff
@@ -0,0 +1,216 @@
...
+            let len = path_array.len();
+            let mut builder = RasterBuilder::new(len);
+            for i in 0..len {
+                if path_array.is_null(i) {
+                    builder.append_null()?;
+                } else {
+                    let path = path_array.value(i);
+                    append_as_outdb_raster(gdal, path, &mut builder)?;
+                }
+            }
```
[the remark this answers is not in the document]
Kontinuation: Fixed in 2756654f. Added unit tests for null propagation and invalid-path errors in rs_frompath.rs so the expected semantics are now locked in.

jiayuasu:
@james-willis please review too

rust/sedona-raster-gdal/src/rs_frompath.rs
```diff
@@ -0,0 +1,273 @@
...
```
paleolimbot: This works, although our iteration elsewhere is generally `for item_opt in path_array { ... }`

rust/sedona-raster-gdal/src/rs_frompath.rs
```diff
@@ -0,0 +1,273 @@
...
+
+            let result: Arc<dyn Array> = Arc::new(builder.finish()?);
+
+            match &args[0] {
+                ColumnarValue::Scalar(_) => {
+                    let scalar = datafusion_common::ScalarValue::try_from_array(&result, 0)?;
+                    Ok(ColumnarValue::Scalar(scalar))
+                }
+                ColumnarValue::Array(_) => Ok(ColumnarValue::Array(result)),
+            }
```
paleolimbot:
```suggestion
            executor.finish(result)
```

rust/sedona-raster-gdal/src/rs_frompath.rs
```diff
@@ -0,0 +1,273 @@
...
+        _arg_types: &[SedonaType],
+        args: &[ColumnarValue],
+        _return_type: &SedonaType,
+        _num_rows: usize,
+        config_options: Option<&ConfigOptions>,
+    ) -> Result<ColumnarValue> {
+        with_gdal(|gdal| {
+            configure_thread_local_options(gdal, config_options)?;
+
+            let paths = args[0].cast_to(&DataType::Utf8, None)?.into_array(1)?;
```
paleolimbot:
```suggestion
            let executor = WkbBytesExecutor::new(arg_types, args);
            let paths = args[0].cast_to(&DataType::Utf8, None)?.into_array(executor.num_iterations())?;
```

The rest of this loop suggests this can handle both scalars and columns but I think as written this will fail for anything except a scalar or one row table

rust/sedona-raster-gdal/src/rs_frompath.rs
```diff
@@ -0,0 +1,273 @@
...
+        assert_eq!(rs_frompath_udf().name(), "rs_frompath");
+    }
+
+    fn assert_raster_dimensions(
+        result: &ColumnarValue,
+        expected_len: usize,
+        width: u64,
+        height: u64,
+    ) {
+        match result {
```
paleolimbot, line 133: Not your problem for this PR, but we really need support for rasters in the `ScalarUdfTester` so that these tests can be more compact.

In the meantime this could be moved to `sedona_testing::rasters` which has a few asserters along these lines.

rust/sedona-raster-gdal/src/utils.rs
```diff
@@ -109,6 +104,68 @@ pub fn append_as_indb_raster(dataset: &Dataset, builder: &mut RasterBuilder) ->
...
+pub fn append_as_outdb_raster(gdal: &Gdal, path: &str, builder: &mut RasterBuilder) -> Result<()> {
+    let gdal_path = normalize_outdb_source_path(path);
+    let dataset = gdal
+        .open_ex_with_options(
+            &gdal_path,
+            DatasetOptions {
+                open_flags: GDAL_OF_RASTER | GDAL_OF_READONLY,
+                ..Default::default()
+            },
+        )
```
paleolimbot, line 117: I am guessing this does blocking IO. Should we make any attempt to mitigate this or implement `RS_FromPath` as an async UDF to get the auto batch size adjustment from DataFusion for these types of functions?

rust/sedona-raster-gdal/Cargo.toml
```diff
@@ -31,13 +31,18 @@ rust-version.workspace = true
 result_large_err = "allow"
 
 [dependencies]
+arrow = { workspace = true }
```
paleolimbot: Can we use a targeted subcrate for whatever uses this import?

rust/sedona/src/context.rs
```diff
@@ -232,6 +221,10 @@ impl SedonaContext {
             Arc::new(RandomGeometryFunction::default()),
         );
 
+        for udf in sedona_raster_gdal::all_gdal_udfs() {
+            out.ctx.register_udf(udf.into());
```
paleolimbot: Can this use the same pattern as the other crates that provide scalar functions? (e.g., `out.register_scalar_kernels(sedona_raster_gdal::register::scalar_kernels().into_iter())?;`). I suppose if we make it an async UDF we'll need the pattern you have here.

paleolimbot (commented):

zhangfengcdt (approved):
We may want to align the registration the same way as other UDFs via scalar_kernels() from a register module inside sedona-raster-gdal as @paleolimbot already points out, otherwise, LTGM!

Kontinuation:
I tried to implement `RS_FromPath` as a DataFusion async UDF but was blocked by a DataFusion bug: https://github.com/apache/datafusion/issues/22662. I have submitted a patch for fixing this: https://github.com/apache/datafusion/pull/22663.

I'd like to leave `RS_FromPath` as synchronous UDF for now until the fix got accepted and released by the upstream. The async version of `RS_FromPath` can be found in https://github.com/Kontinuation/sedona-db/tree/pr-c-rs-frompath-async. We extended `SedonaScalarUDF` to support async function and made `RS_FromPath` async by dispatching GDAL calls to tokio worker threads.

rust/sedona-raster-gdal/src/rs_frompath.rs
Kontinuation: Switched to `for path_opt in path_array`

rust/sedona-raster-gdal/Cargo.toml
Kontinuation: Depend on `arrow-array`, `arrow-buffer` and `arrow-schema`.

rust/sedona/src/context.rs
Kontinuation: Now we have `sedona_raster_gdal::register::default_function_set`, so this loop was replaced by `out.register_function_set(sedona_raster_gdal::register::default_function_set());`.

rust/sedona-raster-gdal/src/rs_frompath.rs
Kontinuation: Switched to use `WkbBytesExecutor`. It looks a bit strange: we use the executor for purposes other than iterating over the input data, but it still make the code shorter.

paleolimbot (approved):
> I tried to implement RS_FromPath as a DataFusion async UDF but was blocked by a DataFusion bug:

Thank you for trying and for fixing that!

> I'd like to leave RS_FromPath as synchronous UDF for now until the fix got accepted and released by the upstream.

Sounds good to me. If an async context is truly helpful we can store an `Arc<Runtime>` in the UDF and instantiate/register from the Cli/Python where we have access to it, but no need to do that here.

PR #832, opened 2026-05-11
feat(python/sedonadb): add DataFrame.select

jiayuasu:
Wire the `Expr` layer into the existing lazy `DataFrame` so users can project columns without writing SQL strings.

This is the third small PR in the Phase P1 series of #791, building on #807 (Expr foundation) and #823 (Expr operators).

## What's new

```python
from sedonadb.expr import col

df.select("x", "y")                                # bare column names
df.select(col("x"), (col("y") + 1).alias("y1"))    # Expr objects
df.select("x", (col("y") * 2).alias("y2"))         # mix
```

- Strings are converted to column references via `col(name)` internally; the same plan is produced as the all-Expr form.
- Empty argument list → `ValueError`. Non-str/non-Expr argument → `TypeError`. Unknown column → DataFusion plan-build error (same behavior locked in the foundation PR).

## Implementation

`InternalDataFrame::select` (Rust) is a thin wrapper that unwraps `Vec<PyExpr>` to `Vec<Expr>` and calls DataFusion's `DataFrame::select` directly. No new query-engine code.

## Test plan

- 11 tests in `tests/expr/test_dataframe_select.py` cover string projection, Expr projection, mixed args, arithmetic Expr, alias output, literal-coercion via operators, lazy return, and the three error paths.
- Assertions are exact `column_names` / `to_pylist()` — no substring matching — so any change in projection semantics fails loudly.
- All 11 pass locally; no regressions in existing `test_dataframe.py`.

`DataFrame.filter` / `.where` come in the next small PR.

### DIFF
--- a/python/sedonadb/python/sedonadb/dataframe.py
+++ b/python/sedonadb/python/sedonadb/dataframe.py
@@ -85,6 +85,50 @@ class DataFrame:
         """
         return self.limit(n)
 
+    def select(self, *exprs) -> "DataFrame":
+        """Project a set of columns or expressions.
+
+        Returns a new lazy `DataFrame` whose columns are exactly the
+        projection. Column-name strings are converted to column references
+        via `sedonadb.expr.col` internally, so `df.select("x", "y")` and
+        `df.select(col("x"), col("y"))` produce the same plan.
+
+        Args:
+            *exprs: One or more arguments. Each argument is either a column
+                name (`str`) or a `sedonadb.expr.Expr`. At least one argument
+                is required.
+
+        Examples:
+
+            >>> from sedonadb.expr import col
+            >>> sd = sedona.db.connect()
+            >>> df = sd.sql("SELECT 1 AS a, 2 AS b")
+            >>> df.select("a", (col("b") + 1).alias("b_plus_1")).show()
+            ┌───────┬──────────┐
+            │   a   │ b_plus_1 │
+            │ int64 │   int64  │
+            ╞═══════╪══════════╡
+            │     1 │        3 │
+            └───────┴──────────┘
+        """
+        from sedonadb.expr import Expr
+        from sedonadb.expr import col as _col
+
+        if not exprs:
+            raise ValueError("select() requires at least one column or expression")
+
+        coerced = []
+        for e in exprs:
+            if isinstance(e, Expr):
+                coerced.append(e._impl)
+            elif isinstance(e, str):
+                coerced.append(_col(e)._impl)
+            else:
+                raise TypeError(
+                    f"select() expects str or Expr arguments, got {type(e).__name__}"
+                )
+        return DataFrame(self._ctx, self._impl.select(coerced), self._options)
+
     def limit(self, n: Optional[int], /, *, offset: int = 0) -> "DataFrame":
         """Limit result to n rows starting at offset
 
--- a/python/sedonadb/python/sedonadb/dataframe.py
+++ b/python/sedonadb/python/sedonadb/dataframe.py
@@ -85,7 +85,7 @@ class DataFrame:
         """
         return self.limit(n)
 
-    def select(self, *exprs) -> "DataFrame":
+    def select(self, *exprs: "Expr | str") -> "DataFrame":
         """Project a set of columns or expressions.
 
         Returns a new lazy `DataFrame` whose columns are exactly the
@@ -105,10 +105,10 @@ class DataFrame:
             >>> df = sd.sql("SELECT 1 AS a, 2 AS b")
             >>> df.select("a", (col("b") + 1).alias("b_plus_1")).show()
             ┌───────┬──────────┐
-            │   a   │ b_plus_1 │
-            │ int64 │   int64  │
+            │   a   ┆ b_plus_1 │
+            │ int64 ┆   int64  │
             ╞═══════╪══════════╡
-            │     1 │        3 │
+            │     1 ┆        3 │
             └───────┴──────────┘
         """
         from sedonadb.expr import Expr
--- a/python/sedonadb/python/sedonadb/dataframe.py
+++ b/python/sedonadb/python/sedonadb/dataframe.py
@@ -26,6 +26,9 @@ if TYPE_CHECKING:
     import pandas
     import pyarrow
 
+    from sedonadb.expr import Expr
+    from sedonadb.expr import Literal as _SedonaLit
+
 
 class DataFrame:
     """Representation of a (lazy) collection of columns
@@ -85,22 +88,24 @@ class DataFrame:
         """
         return self.limit(n)
 
-    def select(self, *exprs: "Expr | str") -> "DataFrame":
+    def select(self, *exprs: "Expr | str | _SedonaLit") -> "DataFrame":
         """Project a set of columns or expressions.
 
         Returns a new lazy `DataFrame` whose columns are exactly the
         projection. Column-name strings are converted to column references
         via `sedonadb.expr.col` internally, so `df.select("x", "y")` and
-        `df.select(col("x"), col("y"))` produce the same plan.
+        `df.select(col("x"), col("y"))` produce the same plan. Literals
+        produced by `sedonadb.expr.lit()` are also accepted; use
+        `lit(value).alias(name)` to give the literal column a name.
 
         Args:
             *exprs: One or more arguments. Each argument is either a column
-                name (`str`) or a `sedonadb.expr.Expr`. At least one argument
-                is required.
+                name (`str`), a `sedonadb.expr.Expr`, or a `sedonadb.expr.Literal`.
+                At least one argument is required.
 
         Examples:
 
-            >>> from sedonadb.expr import col
+            >>> from sedonadb.expr import col, lit
             >>> sd = sedona.db.connect()
             >>> df = sd.sql("SELECT 1 AS a, 2 AS b")
             >>> df.select("a", (col("b") + 1).alias("b_plus_1")).show()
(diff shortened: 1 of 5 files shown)


python/sedonadb/python/sedonadb/dataframe.py
```diff
@@ -85,6 +85,50 @@ def head(self, n: int = 5) -> "DataFrame":
...
+        coerced = []
+        for e in exprs:
+            if isinstance(e, Expr):
+                coerced.append(e._impl)
+            elif isinstance(e, str):
+                coerced.append(_col(e)._impl)
+            else:
+                raise TypeError(
+                    f"select() expects str or Expr arguments, got {type(e).__name__}"
+                )
```
paleolimbot, line 141: Literals are missing here: we can check `isinstance(e, Literal)` (so that literals wrapped in `lit()` work).

I think it will also be useful to ensure that the error that occurs when `str` is an invalid column lists the valid columns (DataFusion will do this for you in SQL and the error we get might already include it but also might not when calling via the DataFrame API).

python/sedonadb/tests/expr/test_dataframe_select.py
```diff
@@ -0,0 +1,106 @@
...
+
+import pytest
+
+import sedonadb
+from sedonadb.expr import col
+
+
+@pytest.fixture
+def sd():
+    return sedonadb.connect()
```
paleolimbot: We already have this (the fixture is called `con`, because it predates inventing of `sd = sedona.db.connect()` 🙂 )

python/sedonadb/tests/expr/test_dataframe_select.py
```diff
@@ -0,0 +1,106 @@
...
+    out = df_xy.select((col("x") * 0 + 7).alias("seven")).to_arrow_table()
+    assert out.column("seven").to_pylist() == [7, 7, 7, 7]
+
+
+def test_select_returns_lazy_dataframe(df_xy):
+    out = df_xy.select("x")
+    # Plan should be lazy until materialization.
+    assert hasattr(out, "to_arrow_table")
+
+
```
paleolimbot: I think you can skip this one

```suggestion
```

python/sedonadb/tests/expr/test_dataframe_select.py
```diff
@@ -0,0 +1,106 @@
...
+def test_select_mix_strings_and_exprs(df_xy):
+    out = df_xy.select("x", (col("y") * 2).alias("y2")).to_arrow_table()
+    assert out.column_names == ["x", "y2"]
+    assert out.column("y2").to_pylist() == [20, 40, 60, 80]
+
+
+def test_select_literal_via_isin_path(df_xy):
+    # No public lit() -> Expr in this PR; literals come in via operator
+    # coercion (and inside isin). This test exercises a literal-on-the-LHS
+    # arithmetic operation, where the int is auto-coerced to Expr.
```
paleolimbot: I think there's no problem exposing `lit()` and accepting it here?

python/sedonadb/tests/expr/test_dataframe_select.py
```diff
@@ -0,0 +1,106 @@
...
+@pytest.fixture
+def sd():
+    return sedonadb.connect()
+
+
+@pytest.fixture
+def df_xy(sd):
+    return sd.sql(
+        "SELECT * FROM (VALUES (1, 10), (2, 20), (3, 30), (4, 40)) AS t(x, y)"
+    )
```
paleolimbot: I know it seems repetitive, but it's worth the one extra line to keep all of the test context within the `test_` function for the day you have to debug a random failure:

```python
df = sd.create_data_frame(pd.DataFrame({"x": [1, 2, 3]})
```

python/sedonadb/tests/expr/test_dataframe_select.py
```diff
@@ -0,0 +1,106 @@
...
+
+
+@pytest.fixture
+def sd():
+    return sedonadb.connect()
+
+
+@pytest.fixture
+def df_xy(sd):
+    return sd.sql(
```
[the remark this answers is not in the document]
jiayuasu: Done in 098393c1 — fixtures now consume the shared `con` fixture from `tests/conftest.py` and the local `sd` fixture is gone.

python/sedonadb/tests/expr/test_dataframe_select.py
```diff
@@ -0,0 +1,106 @@
...
+def test_select_bad_arg_type_raises(df_xy):
+    with pytest.raises(TypeError, match="str or Expr"):
+        df_xy.select(123)
+
+
+def test_select_unknown_column_errors_at_plan_build(df_xy):
+    # DataFusion validates column references at plan-build time. The Expr
+    # itself is unbound (col("nonexistent") alone is fine), but selecting
+    # it against a frame that doesn't have that column fails immediately.
+    with pytest.raises(Exception, match="nonexistent"):
```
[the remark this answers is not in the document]
jiayuasu: Done in 098393c1 — `pytest.raises` now catches `sedonadb._lib.SedonaError` specifically.

python/sedonadb/tests/expr/test_dataframe_select.py
```diff
@@ -0,0 +1,106 @@
...
+    out = df_xy.select("x", (col("y") * 2).alias("y2")).to_arrow_table()
+    assert out.column_names == ["x", "y2"]
+    assert out.column("y2").to_pylist() == [20, 40, 60, 80]
+
+
+def test_select_literal_via_isin_path(df_xy):
+    # No public lit() -> Expr in this PR; literals come in via operator
+    # coercion (and inside isin). This test exercises a literal-on-the-LHS
+    # arithmetic operation, where the int is auto-coerced to Expr.
+    out = df_xy.select((col("x") * 0 + 7).alias("seven")).to_arrow_table()
```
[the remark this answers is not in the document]
jiayuasu: Renamed to `test_select_literal_via_operator_coercion` and rewrote the comment in 098393c1 to describe what's actually being exercised (right-hand scalar coercion through `_to_expr`), not the `isin` path the old name implied.

python/sedonadb/python/sedonadb/dataframe.py
```diff
@@ -85,6 +85,50 @@ def head(self, n: int = 5) -> "DataFrame":
         """
         return self.limit(n)
 
+    def select(self, *exprs) -> "DataFrame":
+        """Project a set of columns or expressions.
+
```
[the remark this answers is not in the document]
jiayuasu: Done in 098393c1 — `*exprs: "Expr | str"` (string annotation so the forward reference works without restructuring imports). This also fixes the docs-and-deploy CI failure, which was griffe rejecting the untyped vararg under strict mode.

python/sedonadb/tests/expr/test_dataframe_select.py
```diff
@@ -0,0 +1,106 @@
...
+def df_xy(sd):
+    return sd.sql(
+        "SELECT * FROM (VALUES (1, 10), (2, 20), (3, 30), (4, 40)) AS t(x, y)"
+    )
+
+
+def test_select_by_string(df_xy):
+    out = df_xy.select("x").to_arrow_table()
+    assert out.column_names == ["x"]
+    assert out.column("x").to_pylist() == [1, 2, 3, 4]
```
paleolimbot: Unless there is a good reason not to, use `pd.testing.assert_dataframe_equal()` for these (the failures are generally very informative). GeoPandas has a similar API for the GeoDataFrame.

python/sedonadb/python/sedonadb/dataframe.py
jiayuasu: Done in 98bf0bd0 — `select()` now accepts `Literal` and routes it through the same `_to_expr` path operators use. On the error message, I checked what DataFusion already emits via plan-build: `No field named nonexistent. Valid fields are ...`. Strengthened the test to lock that contract (asserts both `Valid fields` and the actual column names appear). No Python-side rewrap needed.

python/sedonadb/tests/expr/test_dataframe_select.py
jiayuasu: Switched to the shared `con` fixture in 098393c1 (and now in 98bf0bd0 each test inlines its own `df` from `con.create_data_frame` per your other comment).

python/sedonadb/tests/expr/test_dataframe_select.py
jiayuasu: Dropped in 98bf0bd0. Replaced with `test_select_with_aliased_lit` / `test_select_with_unaliased_lit` that exercise `lit()` directly now that it's exposed.

python/sedonadb/tests/expr/test_dataframe_select.py
jiayuasu: Done in 98bf0bd0. `sedonadb.expr.lit` is re-exported at package level, `select()` accepts `Literal`, and `Literal.alias(name)` promotes the literal to an `Expr` so `lit(7).alias("seven")` works ergonomically.

python/sedonadb/tests/expr/test_dataframe_select.py
jiayuasu: Inlined in 98bf0bd0 — each test now builds its own `df` via `con.create_data_frame(pd.DataFrame({...}))`. The shared `con` fixture from `tests/conftest.py` stays per your other comment.

python/sedonadb/tests/expr/test_dataframe_select.py
jiayuasu: Switched in 98bf0bd0 — every output assertion now uses `pd.testing.assert_frame_equal`.

paleolimbot (approved):
Thank you!

PR #833, opened 2026-05-11
feat(rust/sedona-raster-gdal): add RS_MetaData

Kontinuation:
## Summary
- add `RS_MetaData` backed by the GDAL dataset provider
- expose raster dimensions, transform, srid, band count, and GDAL block size metadata
- keep the change limited to the incremental metadata surface on top of the current GDAL raster work

## Dependency
- Depends on #831 (`RS_FromPath`)
- This branch currently includes the parent `RS_FromPath` commit because GitHub upstream PRs cannot target a fork-only base branch.
- Please review/merge this after #831, or compare this PR commit-by-commit with that dependency in mind.

## Testing
- `cargo clippy -p sedona-raster-gdal -- -D warnings`
- `cargo test -p sedona-gdal`

### DIFF
--- a/docs/reference/sql/rs_metadata.qmd
+++ b/docs/reference/sql/rs_metadata.qmd
@@ -1,0 +1,53 @@
+---
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+title: RS_MetaData
+description: Returns raster metadata as a struct.
+kernels:
+  - returns: struct
+    args: [raster]
+---
+
+## Description
+
+`RS_MetaData()` returns a struct containing:
+
+- `upperLeftX`, `upperLeftY`: origin of the raster geotransform
+- `gridWidth`, `gridHeight`: raster dimensions in pixels
+- `scaleX`, `scaleY`: pixel scale
+- `skewX`, `skewY`: geotransform skew terms
+- `srid`: raster SRID if available, otherwise `0`
+- `numSampleDimensions`: band count
+- `tileWidth`, `tileHeight`: GDAL block size derived tile dimensions
+
+For rasters with no bands, `tileWidth` and `tileHeight` are `0`.
+
+## Examples
+
+```sql
+SELECT RS_MetaData(
+  RS_FromPath('../../../submodules/sedona-testing/data/raster/test4.tiff')
+);
+```
+
+```sql
+SELECT meta."gridWidth", meta."gridHeight"
+FROM (
+  SELECT RS_MetaData(RS_FromPath('../../../submodules/sedona-testing/data/raster/test4.tiff')) AS meta
+);
+```
--- a/docs/reference/sql/rs_metadata.qmd
+++ b/docs/reference/sql/rs_metadata.qmd
@@ -49,5 +49,5 @@ SELECT RS_MetaData(
 SELECT meta."gridWidth", meta."gridHeight"
 FROM (
   SELECT RS_MetaData(RS_FromPath('../../../submodules/sedona-testing/data/raster/test4.tiff')) AS meta
-);
+) q;
 ```
--- a/docs/reference/sql/rs_metadata.qmd
+++ b/docs/reference/sql/rs_metadata.qmd
@@ -40,14 +40,14 @@ For rasters with no bands, `tileWidth` and `tileHeight` are `0`.
 ## Examples
 
 ```sql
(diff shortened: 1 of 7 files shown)


rust/sedona-raster-gdal/src/rs_metadata.rs
```diff
@@ -0,0 +1,440 @@
...
+
+use crate::gdal_common::with_gdal;
+use crate::gdal_dataset_provider::{configure_thread_local_options, thread_local_provider};
+
+pub fn rs_metadata_udf() -> SedonaScalarUDF {
+    SedonaScalarUDF::new(
+        "rs_metadata",
+        vec![Arc::new(RsMetaData {})],
+        Volatility::Immutable,
+    )
```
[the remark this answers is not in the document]
Kontinuation: We intentionally kept `Volatility::Immutable` here. Sedona assumes external raster sources remain stable for the duration of a query; under that system assumption, `RS_MetaData` is deterministic for a given query input. If a user mutates an underlying out-db raster during query execution, that violates the same external-data stability assumption the system already relies on, and they have to accept the consequences. So we are treating this as an explicit design choice rather than relaxing the volatility classification.

docs/reference/sql/rs_metadata.qmd
```diff
@@ -0,0 +1,53 @@
...
+
+For rasters with no bands, `tileWidth` and `tileHeight` are `0`.
+
+## Examples
+
+```sql
+SELECT RS_MetaData(
+  RS_FromPath('../../../submodules/sedona-testing/data/raster/test4.tiff')
+);
+```
```
paleolimbot, line 44: Can we use the `http://` url here so that this example is runnable by a user? Or enable `RS_Example("fixture name")`?

rust/sedona-raster-gdal/src/rs_metadata.rs
```diff
@@ -0,0 +1,458 @@
...
+}
+
+fn metadata_struct_fields() -> Fields {
+    Fields::from(vec![
+        Field::new("upperLeftX", DataType::Float64, true),
+        Field::new("upperLeftY", DataType::Float64, true),
+        Field::new("gridWidth", DataType::Int64, true),
```
paleolimbot, line 50: I get that we probably have to do this for compatibility with Sedona Spark, but it's a bit of a bummer these aren't `snake_case`.

rust/sedona-raster-gdal/src/rs_metadata.rs
```diff
@@ -0,0 +1,458 @@
...
+        let mut grid_height_builder = Int64Builder::with_capacity(capacity);
+        let mut scale_x_builder = Float64Builder::with_capacity(capacity);
+        let mut scale_y_builder = Float64Builder::with_capacity(capacity);
+        let mut skew_x_builder = Float64Builder::with_capacity(capacity);
+        let mut skew_y_builder = Float64Builder::with_capacity(capacity);
+        let mut srid_builder = Int32Builder::with_capacity(capacity);
+        let mut num_bands_builder = UInt32Builder::with_capacity(capacity);
+        let mut tile_width_builder = UInt64Builder::with_capacity(capacity);
+        let mut tile_height_builder = UInt64Builder::with_capacity(capacity);
+        let mut struct_validity = Vec::with_capacity(capacity);
```
paleolimbot: Can you use a `NullsBuffer` here?

rust/sedona-raster-gdal/src/rs_metadata.rs
```diff
@@ -0,0 +1,458 @@
...
+                                Ok(Some(crs_ref)) => crs_ref
+                                    .srid()
+                                    .ok()
+                                    .flatten()
+                                    .and_then(|s| i32::try_from(s).ok())
+                                    .unwrap_or(0),
+                                _ => 0i32,
+                            },
+                        };
+                        srid_builder.append_value(srid);
```
paleolimbot, line 157: I get how this is needed for compatibility, but we should arguably change the SedonaSpark version to return a string here so that we avoid this being lossy for un-sridable tiffs.

rust/sedona-raster-gdal/src/rs_metadata.rs
```diff
@@ -0,0 +1,458 @@
...
+            .as_struct()
+            .clone()
+    }
+
+    fn assert_scalar_result(result: ScalarValue) -> Arc<StructArray> {
+        match result {
+            ScalarValue::Struct(struct_array) => struct_array,
+            other => panic!("Expected struct scalar result, got {other:?}"),
+        }
+    }
```
paleolimbot, line 326: For analyze agg we have a JSON based struct approach that could be copied or moved to sedona-testing and reused (but your tests here are quite readable and I'm not sure it's an improvement other than less helper code).

https://github.com/apache/sedona-db/blob/f64ed45532a23f5065f9e92e0a8695a01991cfa6/rust/sedona-functions/src/st_analyze_agg.rs#L519-L540

paleolimbot (approved):
Thank you!

It would be great to get the example to be user runnable (whatever we do here will also help make realistic examples for other functions, too). Other than that these are just nits.

rust/sedona-raster-gdal/src/rs_metadata.rs
Kontinuation: Updated. Use NullBuffer instead of vec.

docs/reference/sql/rs_metadata.qmd
Kontinuation: Replaced this example with 2 examples: one using `RS_Example` and another using `RS_FromPath('https://download.osgeo.org/geotiff/samples/gdal_eg/cea.tif')`

paleolimbot (approved):
Thank you!

PR #835, opened 2026-05-12
feat(python/sedonadb): add DataFrame.filter / DataFrame.where

jiayuasu:
Wire the `Expr` layer into the existing lazy `DataFrame` so users can filter rows without writing SQL predicates.

This is the fourth and final small PR completing Phase P1 of #791, building on #807 (Expr foundation), #823 (operators), and #832 (`DataFrame.select`).

## What's new

```python
from sedonadb.expr import col

df.filter(col("x") > 0)
df.filter(col("x") > 0, col("y") < 10)              # multiple preds → AND
df.filter((col("x") > 0) & col("y").is_not_null())  # explicit & also fine
df.filter(col("x") > 1).filter(col("x") < 4)        # chained: two filter nodes
```

Multiple predicates are AND-combined into a single composed `Expr` via DataFusion's `conjunction` helper, giving the optimizer one conjunction to reason about rather than N stacked filter nodes. Chained `.filter().filter()` calls produce two filter nodes in the plan but the same result.

## Decision worth flagging — rejecting `Literal` 🛎️ cc @paleolimbot

`filter()` rejects both `str` and `Literal` arguments at the Python boundary.

- **Strings** are not interpreted as SQL predicates — that's a separate `query()`-style feature, not this PR's surface.
- **`Literal`** is rejected because `filter(lit(True))` (or any `filter(lit(value))`) is almost always a typo: DataFusion would silently accept `WHERE 7` as truthy, producing a filter that looks present in the source but is actually a no-op. If a user genuinely wants a constant predicate they can write `col("flag") == lit(True)`, which forces the intent.

This is more restrictive than `select`, which does accept `Literal`. The asymmetry is intentional: a literal as a projection is meaningful (a constant column), but a literal as a filter is almost never what you want. Happy to revisit if you'd rather have `filter(lit(...))` accepted and the resulting no-op surface as an optimizer-time warning instead.

## Implementation

Rust: `InternalDataFrame::filter(Vec<PyExpr>)` — empty list errors at the Python boundary; multi-element lists are folded via `datafusion_expr::utils::conjunction`, matching the pattern used elsewhere in the workspace (`sedona-pointcloud`, `sedona-query-planner`). Step-by-step comments explain the conjunction choice.

Python: `DataFrame.filter(*exprs: "Expr") -> DataFrame` with explicit `isinstance` checks rejecting `str`, `Literal`, and other types.

## Test plan

12 tests in `tests/expr/test_dataframe_filter.py`:

- **Positive**: simple predicate, multi-AND, explicit `&`, `|`, `~`, `isin`, chained `filter().filter()` matches multi-AND.
- **Lazy**: filter returns a `DataFrame` without materializing.
- **Errors**: empty filter → `ValueError`; string arg → `TypeError`; `Literal` arg → `TypeError` with the actionable suggestion; unknown column → `SedonaError` whose message includes the valid field names.

Each test builds its own input `df` inline from the smallest `pd.DataFrame({...})` that exercises the predicate. Existing select/expression/literal tests all still pass. `ruff format` and doctests both clean.

## What's next

This completes the issue-tracker Phase P1 (Expr layer + select/filter on DataFrame). Phase P2 in the [design doc](https://github.com/apache/sedona-db/issues/791) covers `Series` / `Scalar`, `groupby`, `merge`, the curated `sd.st` module, and the cookbook track.

### DIFF
--- a/python/sedonadb/python/sedonadb/dataframe.py
+++ b/python/sedonadb/python/sedonadb/dataframe.py
@@ -198,8 +198,6 @@ class DataFrame:
             self._options,
         )
 
-    where = filter
-
     def limit(self, n: Optional[int], /, *, offset: int = 0) -> "DataFrame":
         """Limit result to n rows starting at offset
 
--- a/python/sedonadb/python/sedonadb/dataframe.py
+++ b/python/sedonadb/python/sedonadb/dataframe.py
@@ -141,6 +141,65 @@ class DataFrame:
                 )
         return DataFrame(self._ctx, self._impl.select(coerced), self._options)
 
+    def filter(self, *exprs: "Expr") -> "DataFrame":
+        """Filter rows by one or more boolean expressions.
+
+        Multiple expressions are combined with logical AND, so
+        `df.filter(a, b)` is equivalent to `df.filter(a & b)` and to
+        `df.filter(a).filter(b)` (the planner sees one conjunction in
+        the first two forms and two filter nodes in the third).
+
+        Only `Expr` arguments are accepted. Strings are not interpreted
+        as SQL predicates (that is a separate feature). Bare `Literal`
+        values are also rejected — `filter(lit(True))` is almost
+        certainly a typo; if you really mean a constant predicate, wrap
+        a column expression like `col("flag") == lit(True)`.
+
+        Args:
+            *exprs: One or more boolean `sedonadb.expr.Expr` predicates.
+                At least one argument is required.
+
+        Examples:
+
+            >>> from sedonadb.expr import col
+            >>> sd = sedona.db.connect()
+            >>> df = sd.sql("SELECT * FROM (VALUES (1), (2), (3), (4)) AS t(x)")
+            >>> df.filter(col("x") > 2).show()
+            ┌───────┐
+            │   x   │
+            │ int64 │
+            ╞═══════╡
+            │     3 │
+            ├╌╌╌╌╌╌╌┤
+            │     4 │
+            └───────┘
+        """
+        from sedonadb.expr import Expr, Literal
+
+        if not exprs:
+            raise ValueError("filter() requires at least one predicate")
+
+        for e in exprs:
+            if isinstance(e, Literal):
+                raise TypeError(
+                    "filter() does not accept Literal arguments. "
+                    "filter(lit(value)) is almost always a typo; if you really "
+                    "want a constant predicate, wrap a column expression like "
+                    "col('flag') == lit(value)."
+                )
+            if not isinstance(e, Expr):
+                raise TypeError(
+                    f"filter() expects Expr arguments, got {type(e).__name__}"
+                )
+
+        return DataFrame(
+            self._ctx,
+            self._impl.filter([e._impl for e in exprs]),
+            self._options,
+        )
+
+    where = filter
+
     def limit(self, n: Optional[int], /, *, offset: int = 0) -> "DataFrame":
         """Limit result to n rows starting at offset
 
--- a/python/sedonadb/src/dataframe.rs
+++ b/python/sedonadb/src/dataframe.rs
@@ -123,6 +123,34 @@ impl InternalDataFrame {
         Ok(InternalDataFrame::new(inner, self.runtime.clone()))
     }
 
+    /// Filter rows by one or more boolean expressions, producing a new
+    /// lazy `DataFrame`.
+    ///
+    /// The Python side guarantees `exprs` is non-empty and that every
+    /// element is an `Expr` (no strings, no Literals — those rejections
+    /// happen at the Python boundary so the error message can point at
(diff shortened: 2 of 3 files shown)


python/sedonadb/python/sedonadb/dataframe.py
```diff
@@ -141,6 +141,65 @@ def select(self, *exprs: "Expr | str | _SedonaLit") -> "DataFrame":
...
+                    f"filter() expects Expr arguments, got {type(e).__name__}"
+                )
+
+        return DataFrame(
+            self._ctx,
+            self._impl.filter([e._impl for e in exprs]),
+            self._options,
+        )
+
+    where = filter
```
paleolimbot: Can we not add an alias for filter here? We can specifically add a pyspark compatibility later at a different time if there is interest.

```suggestion
```

python/sedonadb/src/dataframe.rs
```diff
@@ -123,6 +123,34 @@ impl InternalDataFrame {
...
+        if exprs.is_empty() {
+            return Err(PySedonaError::SedonaPython(
+                "filter() requires at least one predicate".to_string(),
+            ));
+        }
+        let mut iter = exprs.into_iter().map(|e| e.inner);
+        let mut combined = iter.next().unwrap();
+        for next in iter {
+            combined = combined.and(next);
+        }
```
paleolimbot: There's a helper for this in DataFusion (I think it's called `conjunction!()`; there may be one for both physical and logical expressions and you'll need the logical one).

python/sedonadb/tests/expr/test_dataframe_filter.py
```diff
@@ -0,0 +1,138 @@
...
+import pytest
+
+from sedonadb._lib import SedonaError
+from sedonadb.expr import col, lit
+
+
+def _xy_df(con):
+    return con.create_data_frame(
+        pd.DataFrame({"x": [1, 2, 3, 4], "y": [10, 20, 30, 40]})
+    )
```
paleolimbot: Can you inline this into each test? Probably `pd.DataFrame({"x": [1, 2, 3]})` would be sufficient.

python/sedonadb/python/sedonadb/dataframe.py
jiayuasu: Done in 92afb9b1 — `where = filter` is gone and `test_where_alias_produces_same_output` along with it. PR description updated to drop the where mentions.

python/sedonadb/src/dataframe.rs
jiayuasu: Switched in 92afb9b1 — now using `datafusion_expr::utils::conjunction`, matching the pattern in `sedona-pointcloud` and `sedona-query-planner`. `unwrap`/`expect` is safe because the empty case is already rejected at the Python boundary.

python/sedonadb/tests/expr/test_dataframe_filter.py
jiayuasu: Done in 92afb9b1 — `_xy_df` helper removed; each test builds its own `df` inline from the smallest `pd.DataFrame({...})` sufficient for the assertion. Most tests use `{"x": [1, 2, 3]}`; tests that need a second column use `{"x": ..., "y": ...}`.

python/sedonadb/src/dataframe.rs
```diff
@@ -123,6 +123,32 @@ impl InternalDataFrame {
...
+            return Err(PySedonaError::SedonaPython(
+                "filter() requires at least one predicate".to_string(),
+            ));
+        }
+        // `conjunction` returns None only for an empty iterator, which we
+        // already rejected above; unwrap is safe.
+        let combined = datafusion_expr::utils::conjunction(exprs.into_iter().map(|e| e.inner))
+            .expect("non-empty predicates checked above");
+        let inner = self.inner.clone().filter(combined)?;
+        Ok(InternalDataFrame::new(inner, self.runtime.clone()))
```
paleolimbot: Optional but possibly cleaner:

```suggestion
        if let Some(combined) = conjunction(exprs.into_iter().map(|e| e.inner)) {
            let inner = self.inner.clone().filter(combined)?;
            Ok(InternalDataFrame::new(inner, self.runtime.clone()))
        } else {
            Err(PySedonaError::SedonaPython(
                "filter() requires at least one predicate".to_string(),
            ));
        }
```

paleolimbot (approved):
Thank you!

python/sedonadb/src/dataframe.rs
jiayuasu: Applied in 06fdfc7a — `filter()` now wraps the work in `if let Some(combined) = conjunction(...)` with the empty-list branch in `else`, dropping the `expect()`. (Removed the trailing `;` on the `Err(...)` line so the else arm types as `Result`.)

PR #843, opened 2026-05-14
fix(sedona-geoparquet): cache metadata in inner ParquetOpener

prantogg:
## Summary

PR #646 ([#250](https://github.com/apache/sedona-db/issues/250)) wired the file metadata cache through the GeoParquet wrapper's own `DFParquetMetadata` calls but not through the inner `ParquetOpener`'s `ParquetFileReaderFactory`. With the inner factory left at `DefaultParquetFileReaderFactory`, SQL queries against a registered view kept paying full `metadata_load_time` on every execution even with `datafusion.runtime.metadata_cache_limit` configured.

Install `CachedParquetFileReaderFactory` on the inner `ParquetSource` in `GeoParquetFormat::create_physical_plan`, mirroring DataFusion's stock `ParquetFormat::create_physical_plan`.

## Test plan

- New unit test asserts the inner `ParquetSource` has a custom factory after `create_physical_plan` — fails on `main`, passes here
- `cargo test -p sedona-geoparquet --lib` → 61 passed

## Follow-up (out of scope)

`GeoParquetFileSource::try_from_file_source` resets `metadata_cache` to `None` and relies on each caller to re-set it manually — easy to forget when adding a new caller. The factory added here doesn't have this footgun (it rides on `inner` via `ParquetSource::clone`). Worth a small refactor in a separate PR.

### DIFF
--- a/rust/sedona-geoparquet/src/format.rs
+++ b/rust/sedona-geoparquet/src/format.rs
@@ -41,6 +41,7 @@ use datafusion_catalog::{memory::DataSourceExec, Session};
 use datafusion_common::tree_node::{Transformed, TreeNode, TreeNodeRecursion};
 use datafusion_common::{plan_err, GetExt, Result, Statistics};
 use datafusion_datasource_parquet::metadata::DFParquetMetadata;
+use datafusion_datasource_parquet::{CachedParquetFileReaderFactory, ParquetFileReaderFactory};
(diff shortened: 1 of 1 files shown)


paleolimbot (approved):
These kinds of things are a pain to track down...remind me to bake you a cake the next time we're in the same place 🍰. Thank you!

PR #844, opened 2026-05-14
feat(rust/sedona-functions,c/sedona-geos): Implement geography kernels for structural transformations and accessors

### LINKED ISSUE #819: rust/sedona-functions: Add geography support for mechanical transformations
Most mechanical extractors (e., number of geometries, is empty) that can support geography already do via the `is_geometry_or_geography()` matcher. Transformers (that output the same type they input) aren't quite supported yet in many cases. A few that can be implemented trivially:

- st_dump
- st_force_dim functions
- st_flipcoordinates
- st_geometryn
- st_interiorringn
- st_reverse
- st_start_point
- st_pointN

Some of the GEOS functions are structural (could possibly be reimplemented in sedona-functions for speed but could also be left as is and given a geography matcher)

- st_numpoints (maybe this is repeated from  sedona-functions? or could be easily ported)
- st_numinteriorrings (could probably be easily ported to sedona-functions)
- st_nrings (could probably be easily ported to sedona-functions)
- st_isring
- st_boundary

paleolimbot:
This PR adds geography matchers for a whack of functions whose output is the same for geometry and geography:

- ST_FlipCoordinates
- ST_Force2D, ST_Force3D, ST_Force3DM, ST_Force4D
- ST_GeometryN
- ST_InteriorRingN
- ST_Reverse
- ST_PointN
- ST_Boundary
- ST_LineMerge
- ST_Normalize
- ST_NRings
- ST_NumInteriorRings
- ST_NumPoints
- ST_Points

There's no implementation changes here, only updates to the matchers so that we get geography -> geography and geometry -> geometry.

Closes #819

### DIFF
--- a/c/sedona-geos/src/st_normalize.rs
+++ b/c/sedona-geos/src/st_normalize.rs
@@ -96,9 +96,7 @@ mod tests {
     use geos::{Geom, Geometry};
     use rstest::rstest;
     use sedona_expr::scalar_udf::SedonaScalarUDF;
-    use sedona_schema::datatypes::{
-        WKB_GEOGRAPHY, WKB_GEOGRAPHY_ITEM_CRS, WKB_GEOMETRY, WKB_GEOMETRY_ITEM_CRS,
-    };
+    use sedona_schema::datatypes::{WKB_GEOGRAPHY_ITEM_CRS, WKB_GEOMETRY_ITEM_CRS};
     use sedona_testing::testers::ScalarUdfTester;
 
     use super::*;
@@ -143,7 +141,7 @@ mod tests {
                 Some("POLYGON ((0 0, 0 1, 1 1, 1 0, 0 0))"),
                 Some("MULTILINESTRING ((3 3, 4 4), (1 1, 2 2))"),
             ],
-            &WKB_GEOMETRY,
+            &sedona_type,
         );
         sedona_testing::compare::assert_array_equal(&batch_result, &expected);
     }
--- a/c/sedona-geos/src/st_normalize.rs
+++ b/c/sedona-geos/src/st_normalize.rs
@@ -96,7 +96,9 @@ mod tests {
     use geos::{Geom, Geometry};
     use rstest::rstest;
     use sedona_expr::scalar_udf::SedonaScalarUDF;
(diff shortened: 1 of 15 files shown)


c/sedona-geos/src/st_normalize.rs
```diff
@@ -34,17 +34,24 @@ use crate::geos_to_wkb::write_geos_geometry;
 
 /// ST_Normalize() implementation using the geos crate
 pub fn st_normalize_impl() -> Vec<ScalarKernelRef> {
```
james-willis, line 36: is there some risk to calling normalize on geogs and swapping the winding order? should we provide normalize on a geog?

python/sedonadb/tests/geography/test_geog_mechanical_transforms.py
```diff
@@ -0,0 +1,695 @@
...
+            id="linestring_m",
+        ),
+        pytest.param(
+            "LINESTRING ZM (0 0 1 4, 1 1 2 5, 2 2 3 6)",
+            "LINESTRING ZM (2 2 3 6, 1 1 2 5, 0 0 1 4)",
+            id="linestring_zm",
+        ),
+    ],
+)
+def test_st_reverse(eng, geog, expected):
```
james-willis, line 320: should there be a test here about very large geographies and their interaction with the winding order and how we determine what the inside of the polygon is?

Not sure how we handle determining the inside of a geog polygon in sedona db.

james-willis (commented):
Just the two concerns about polygon interior/winding order behavior.

python/sedonadb/tests/geography/test_geog_mechanical_transforms.py
paleolimbot: That's a good point...I added a test to verify the behaviour of ST_Normalize() with respect to winding order. I don't think the winding order is updated by ST_Normalize (just the point at which the ring starts).

FWIW, we don't use the winding order to determine containment for Geography (we use ring position and we do a zero copy ring reverse using a similar mechanism to the views you've been working on if incorrectly wound). When we get to a place where we can assert a winding order by a dedicated type or using statistics, we'll probably speed things up a bit.

james-willis (approved):

PR #845, opened 2026-05-15
feat(python/sedonadb): normalize geometry WKT expected values in harness

### LINKED ISSUE #815: python/sedonadb: Investigate formatting issues when using WKT as an expected value

E                   Got:

oglego:
## Summary

Fixes #815 - formatting-only failures in Python tests that use WKT as an expected value by normalizing geometry WKT in the Python test harness before comparison.

## Problem

Some Python tests compare query results against expected WKT strings using `assert_query_result()`. When the query returns a geometry value, the Python harness converts that geometry to WKT before comparing tuple results.

The issue is that equivalent WKT can be formatted differently depending on which writer produced it. In practice, this showed up as whitespace-only differences, such as whether spaces appear after commas:

Expected:
`POLYGON Z ((0 0 5,0 1 5,1 1 5,1 0 5,0 0 5))`

Got:
`POLYGON Z ((0 0 5, 0 1 5, 1 1 5, 1 0 5, 0 0 5))`

These failures are not about the geometry result itself, they come from string formatting differences.

## What changed

- Updated the Python test harness to canonicalize WKT only for geometry-typed result columns during tuple-based comparisons
- Left non-geometry string columns on exact string comparison
- Added a regression test that verifies this distinction:
  - geometry results with spacing-only WKT differences compare equal
  - plain string results with the same spacing differences still fail

## Why this approach

This keeps geometry tests focused on geometry semantics rather than renderer-specific text formatting.

The important detail is scope: we do not normalize all strings globally. Only columns that are actually typed as geometry are canonicalized. That avoids masking real regressions in plain-text outputs while still fixing the formatting-only failures that motivated this change.

## Testing

- `pytest -q python/sedonadb/tests/test_testing.py`
- `pytest -q python/sedonadb/tests/functions/test_functions.py`

### DIFF
--- a/python/sedonadb/python/sedonadb/testing.py
+++ b/python/sedonadb/python/sedonadb/testing.py
@@ -63,6 +63,31 @@ def _env_no_skip():
     return env_no_skip in ("true", "1")
 
 
+def _normalize_geometry_wkt_columns(rows, geometry_indexes):
+    """Canonicalize WKT only for geometry-typed tuple values."""
+    import shapely
+
+    geometry_indexes = set(geometry_indexes)
+    normalized_rows = []
+
+    for row in rows:
+        normalized_row = list(row)
+
+        for i in geometry_indexes:
+            value = normalized_row[i]
+            if not isinstance(value, str):
+                continue
+
+            try:
+                normalized_row[i] = shapely.to_wkt(shapely.from_wkt(value))
+            except (ValueError, TypeError, shapely.errors.GEOSException):
+                pass
+
+        normalized_rows.append(tuple(normalized_row))
+
+    return normalized_rows
+
+
 class DBEngine:
     """Engine-agnostic catalog and SQL engine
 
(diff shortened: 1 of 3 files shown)


paleolimbot (commented):

Before looking at this closely I want to double check that our existing WKT normalization is working (you've clearly hit a case where it isn't!):

https://github.com/apache/sedona-db/blob/4daf73bbde764f300cdf7d52e4610f3e91232db4/python/sedonadb/python/sedonadb/testing.py#L204-L222

This is using geoarrow-c's writer, which we did at the time because shapely didn't yet support M values. In theory all geometry and geography should be getting normalized by the time it gets to `assert_result()`. There is an unfortunate bug in geoarrow-c where POINT EMPTY is formatted as `POINT (nan nan)`.

oglego:
Thank you for looking into this!  If I need to make any modifications or need to close out the PR just let me know, thanks again!

oglego:
Fixed by #1016

PR #848, opened 2026-05-15
feat(rust/sedona-functions,c/sedona-proj): Support geography in CRS/SRID functions

paleolimbot:
This PR adds support for Geography in functions that work with CRSes. There was limited support for Geography with item-crs but it had been excluded from ST_Transform and ST_SetSRID and the srid-setting shortcuts in ST_Point and friends.

This PR also validates target CRSes so that it's harder to specify an invalid crs in ST_SetSRID or ST_Transform. This isn't perfect...CRSes that contain an elevation aren't going to be validated here but there are workarounds (specify a 2D crs, which will leave the Z values intact, and restore the 3D crs on the trip back to geometry).

I'm not sure exaclty what the use case is for item crs with geography but it was easy enough to support here so I went for it. Supporting NAD83 is probably useful because I think the average radius over north america is slightly more accurate (but still not a replacement for true geodescic edge support).

```python
import pyproj
import sedona.db

sd = sedona.db.connect()

# Defaults to ogc:crs84
sd.sql("SELECT ST_GeogFromWKT('POINT (0 1)')").schema
# SedonaSchema with 1 field:
#   st_geogfromwkt(Utf8("POINT (0 1)")): geography<Wkb(ogc:crs84, Spherical)>

# Can be other stuff like NAD27
sd.sql("SELECT ST_GeogFromWKT('POINT (0 1)', 'EPSG:4267')").schema
# SedonaSchema with 1 field:
#   st_geogfromwkt(Utf8("POINT (0 1)"),Utf8("EPSG:4267")): geography<Wkb(epsg:4267, Spherical)>

# Works for any GeographicCRS if PROJJSON (even on Mars!)
mars_crs = pyproj.CRS('IAU:49900')

sd.sql("SELECT ST_GeogFromWKT('POINT (0 1)', $1)", params=(mars_crs, )).schema
# SedonaSchema with 1 field:
#   st_geogfromwkt(Utf8("POINT (0 1)"),$1): geography<Wkb(iau:49900, Spherical)>

# Works for item/crs case if you are searching for ways to make your life difficult
# Only works with a few hard-coded auth:code combinations for now
sd.sql("""
    SELECT ST_GeogFromWKT('POINT (0 1)', crs) AS geog
    FROM (VALUES 
        ('EPSG:4326'),
        ('EPSG:4267'),
        ('EPSG:4269')
    ) AS t(crs)
""").show()
# ┌────────────────────────────────────┐
# │                geog                │
# │               struct               │
# ╞════════════════════════════════════╡
# │ {item: POINT(0 1), crs: OGC:CRS84} │
# ├╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌┤
# │ {item: POINT(0 1), crs: EPSG:4267} │
# ├╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌┤
# │ {item: POINT(0 1), crs: EPSG:4269} │
# └────────────────────────────────────┘
```

### DIFF
--- a/c/sedona-proj/src/st_transform.rs
+++ b/c/sedona-proj/src/st_transform.rs
@@ -131,6 +131,8 @@ impl SedonaScalarKernel for STTransform {
             WKB_MIN_PROBABLE_BYTES * executor.num_iterations(),
         );
 
+        let (return_item_type, _) = parse_item_crs_arg_type(return_type)?;
+
         // Optimize the easy case, where we have exactly one transformation and there are no
         // null or missing CRSes to contend with.
         let from_index = inputs.len() - 2;
@@ -144,7 +146,7 @@ impl SedonaScalarKernel for STTransform {
                 // Validate the target CRS is valid for the output type (e.g., geography
                 // requires a geographic CRS)
                 let to_crs_opt: Crs = Some(to_crs.clone());
-                validate_crs_for_type(&to_crs_opt, return_type)?;
+                validate_crs_for_type(&to_crs_opt, &return_item_type)?;
 
                 with_global_proj_engine(|engine| {
                     let crs_transform = engine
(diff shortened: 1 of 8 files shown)


zhangfengcdt (approved):
LGTM

PR #858, opened 2026-05-19
feat(raster-zarr): sedona-raster-zarr crate + sedona-zarr plugin

james-willis:
## Summary

Adds Zarr support to SedonaDB as an opt-in Python plugin (`sedonadb-zarr`). Each chunk position in the Zarr group becomes one row of a single-column raster table, similar to how each TIFF becomes one row.

```python
import sedonadb
import sedonadb_zarr

con = sedonadb.connect()
con.read_format(sedonadb_zarr.ZarrFormatSpec(), "file:///path/to/foo.zarr").show()
```

The main `sedonadb` package has zero Zarr code or dependencies — applications that don't `import sedonadb_zarr` pay no build or runtime cost. This is the plugin pattern future format crates (`sedonadb-cog`, `sedonadb-icechunk`, …) will follow.

We built on the `zarrs` 0.23 crate rather than GDAL MultiDim because two GDAL gaps are independently disqualifying:

- Zarr v3 sharding is unreadable in GDAL ≤ 3.12.
- vlen-utf8 string coordinate variables are unreadable through GDAL MultiDim (climate/datacube Zarr uses these for band names).

## What's in this PR

**New `sedona-raster-zarr` crate.** A streaming `ZarrChunkReader: RecordBatchReader` that walks the group's chunk grid lazily, emitting one raster row per chunk position with one band per array. Per-batch memory is `O(rows_per_batch × per-row-metadata)`, not `O(total_chunks × chunk_size)`. Each row carries an `outdb_uri` chunk anchor; `data` stays empty until the async byte resolver lands.

**New `sedonadb-zarr` Python plugin package.** A PyO3 shim around `sedona-raster-zarr`: `PyZarrChunkReader` implements `__arrow_c_stream__`; `ZarrFormatSpec(ExternalFormatSpec)` consumes it from Python. Plugin Cargo.toml has no host-crate dependency — the plugin and host talk through the Python `ExternalFormatSpec` contract and Arrow C Stream FFI.

**New host machinery in `sedonadb` / `sedona-datasource`.**

- `SedonaContext.read_format(spec, paths)` Python API for plugin-defined formats. Parallels `read_parquet` / `read_pyogrio` in shape.
- `ExternalFormatSpec.list_single_object` property (default `False`) on the Python base class and the matching Rust trait method.
- `SingleObjectExternalTable` Rust `TableProvider` that bypasses DataFusion's `ListingTable` for directory-shape formats — a Zarr group is a directory, so the listing layer would return zero matching objects at a `.zarr` prefix.
- `external_table` dispatcher that picks between `SingleObjectExternalTable` and `ListingTable` based on the spec's `list_single_object`.

**Loader feature surface.**

- GeoZarr metadata: `proj:wkt2` / `proj:projjson` / `proj:epsg` for CRS; `spatial:transform` / `spatial:dims` for affine + spatial axes.
- Reads dimension names from either Zarr v3's first-class `dimension_names` field or the xarray-on-Zarr `_ARRAY_DIMENSIONS` attribute (supports xarray-produced v2 Zarrs).
- `arrays = [...]` option for an explicit subset; default auto-skips 1-D coord variables. Explicit 1-D names error at parse time.
- Strict georef validation: CRS-without-`spatial:transform` errors; both-absent falls back to identity + `log::warn!`.
- Group-constraint validation: shared chunk grid / chunk shape / dim_names across the group.
- Dtype mapping: `bool`, `u8`/`i8`, `u16`/`i16`, `u32`/`i32`, `u64`/`i64`, `f32`/`f64` — each exercised by a parameterized smoke test.
- Local filesystem stores only (`file://` URIs or bare paths).

## Out of scope (separate follow-ups)

- **SQL UDTF (`sd_read_zarr`).** Split out to draft PR #866 to keep this PR focused on the loader + plugin pattern.
- **URL-as-table (`SELECT * FROM 'foo.zarr'`).** Needs filesystem-catalog customisation in DataFusion's resolver.
- **R-side `sedonadbzarr` package.** Mirrors the Python plugin pattern for R sedonadb users.
- **Async OutDb byte resolver + `RS_EnsureLoaded` UDF.** Materialises the chunk-anchor URIs into actual pixel bytes; until then every metadata-only kernel (`count(*)`, `RS_Envelope`, `RS_Width`, …) works against the anchor-only rows.
- **Cloud storage backends** (`s3://`, `gs://`, `az://`, `https://`).
- **Chunk-grid partitioning** for intra-group parallelism.
- **Lifting `GeoTransform` from `c/sedona-gdal` into `sedona-raster`** so all raster backends share the same affine-transform type.


rust/sedona-raster-zarr/src/loader.rs
```diff
@@ -0,0 +1,513 @@
...
+    let spatial_dims_names: Vec<&str> = spatial_dim_indices
+        .iter()
+        .map(|&i| array_infos[0].dim_names[i].as_str())
+        .collect();
+    let chunk_spatial_shape: Vec<i64> = spatial_dim_indices
+        .iter()
+        .map(|&i| array_infos[0].chunk_shape[i] as i64)
+        .collect();
+
+    let group_transform = geo.transform.unwrap_or([0.0, 1.0, 0.0, 0.0, 0.0, -1.0]);
```
james-willis: Currently the type implementation does not allow null geotransforms so we need to provide this potentially misleading value here when geozarr metadata is not present.

Havent considered what to do about this.

james-willis: added some validation and warning to help users who dont have full geozarr metadata.

rust/sedona-raster-zarr/Cargo.toml
```diff
@@ -0,0 +1,53 @@
...
+sedona-schema = { workspace = true }
+serde_json = { workspace = true }
+zarrs = { workspace = true, features = ["filesystem", "gzip", "zstd", "crc32c", "sharding", "transpose"] }
+zarrs_filesystem = { workspace = true }
+
+# `blosc` is gated off Windows: c-blosc (statically linked) bundles its own
+# `pthread_create` / `pthread_cond_*` symbols, which conflict with rtools45's
+# `libpthread.a` during the MinGW link of the R `sedonadb.dll`. Non-Windows
+# targets get the full blosc-compressed Zarr reading capability.
+[target.'cfg(not(target_os = "windows"))'.dependencies]
```
paleolimbot: Can you use a more targeted cfg so that Windows Python users can still do this? I think the triplet for mingw is windows-gnu or something

rust/sedona-raster-zarr/src/geozarr.rs
```diff
@@ -0,0 +1,228 @@
...
+#[derive(Debug, Clone, PartialEq)]
+pub struct GroupGeoMetadata {
+    /// CRS string in PROJ or WKT format (whichever the group declared).
+    /// `None` if no `proj:wkt2` / `proj:projjson` / `proj:epsg` attribute
+    /// is present on the group.
+    pub crs: Option<String>,
+    /// Affine transform in GDAL GeoTransform order:
+    /// `[origin_x, scale_x, skew_x, origin_y, skew_y, scale_y]`. `None`
+    /// when no `spatial:transform` attribute is present.
+    pub transform: Option<[f64; 6]>,
```
paleolimbot, line 40: I think at one point we had an object for this to help encapsulate what you can do with this? If not, we should add one and use it.

rust/sedona-raster-zarr/src/udtf.rs
```diff
@@ -0,0 +1,553 @@
...
+///     - `indb`: `true` (default) materializes chunk bytes into the
+///       Arrow `data` column eagerly; `false` emits chunk-anchor URIs
+///       only and defers byte resolution to the OutDb loader.
+///     - `rows_per_batch`: chunks per `RecordBatch` (default 1024)
+///     - `num_partitions`: scan partitions (default 1; > 1 currently errors)
+///     - `arrays`: optional list of array names to read. Default reads
+///       every multi-dimensional array (1-D coord variables are
+///       auto-skipped); an explicit list reads exactly those arrays.
+#[derive(Debug, Default)]
+pub struct ZarrReadFunction {}
```
paleolimbot: Rather than a UDTF in the sedona crate (which make zarr everybody's problem), how about:

- The ability to resolve a `Box<dyn RecordBatchReader>` based on these JSON options is exposed as a function in sedona-raster-zarr
- Use the pre-existing `ExternalFormatSpec` to write the FileFormat/TableProvider for you:

https://github.com/apache/sedona-db/blob/c2613699056b67c001a7ef6899e8568f91bc6596/python/sedonadb/python/sedonadb/datasource.py#L24-L34

- Build a separate Python package `sedonadb-zarr` that exposes your `Box<dyn RecordBatchReader>` to Python and wire it together there. You can add to the ExternalFormatSpec if you need more features.

Either way the UDTF in the sedona-raster-zarr crate is a better fit (make sure it is feature flagged here under `zarr` at the very very least). We can still build it into another Python package using the DataFusion FFI but it's more work and in this case you don't really need any bells and whistles of the DataFusion UDTF because there are not a lot of total rows.

Should we default to reading outdb since this will almost certainly be the future behaviour?

It may be worth leaving off the default row count (you may well want to have this be adaptive based on the byte size you need to load)

rust/sedona/Cargo.toml
```diff
@@ -50,6 +50,14 @@ s2geography = ["dep:sedona-s2geography", "dep:sedona-spatial-join-geography", "s
...
 tokio = { workspace = true, features = ["rt-multi-thread"] }
 rstest = { workspace = true }
+sedona-raster = { workspace = true }
+zarrs = { workspace = true, features = ["filesystem", "gzip", "zstd", "crc32c", "sharding", "transpose"] }
+zarrs_filesystem = { workspace = true }
+
+# See sedona-raster-zarr/Cargo.toml — `blosc` is gated off Windows to avoid
+# a libwinpthreads/c-blosc symbol collision in the R `sedonadb.dll` link.
+[target.'cfg(not(target_os = "windows"))'.dev-dependencies]
+zarrs = { workspace = true, features = ["blosc"] }
```
paleolimbot: I think if you move these to `python/sedona-zarr` we can avoid making this crate take on every single piece of functionality we'll need here

rust/sedona-raster-zarr/src/loader.rs
```diff
@@ -0,0 +1,674 @@
...
+            &array_infos[0].chunk_shape,
+            &spatial_dim_indices,
+        );
+        let crs_str = geo.crs.as_deref();
+        builder.start_raster_nd(
+            &row_transform,
+            &spatial_dims_names,
+            &chunk_spatial_shape,
+            crs_str,
+        )?;
```
paleolimbot: This step involves getting all chunks of a Zarr into memory before emitting them. Is this always a sufficiently small number of chunks that this is safe/the behaviour we want? (We could also have a zarr chunk iterator that can be pulled from)

rust/sedona-raster-zarr/src/loader.rs
```diff
@@ -0,0 +1,674 @@
...
+                info.data_type,
+                nodata_ref,
+                Some(anchor.as_str()),
+                Some("zarr"),
+            )?;
+            match mode {
+                Mode::InDb => {
+                    let bytes = retrieve_chunk_bytes(&info.array, &chunk_indices)?;
+                    builder.band_data_writer().append_value(&bytes);
+                }
```
paleolimbot: This part you definitely don't want in a forward synchronous loop that does serial blocking IO. If you're building this as a TableProvider or UDTF, you can have this be `async` so the runtime is provided for you and you can `await` or implement the lower level RecordBatchStream to handle the polling yourself. In any case you probably want

```rust
struct ZarrChunkIterator {}

impl RecordBatchStream for ZarrChunkIterator {}

// or

impl RecordBatchReader for ZarrChunkIterator {}
```

As written it's a bit of a footgun because this type of loop can't be cancelled after a user realizes they've tried to load a few GB by accident

rust/sedona-raster-zarr/src/loader.rs
```diff
@@ -0,0 +1,674 @@
...
+        }
+    }
+
+    builder.finish()
+}
+
+/// Collect per-array metadata from open zarrs `Array` handles.
+///
+/// Sorts arrays by path so band ordering across rows is deterministic
+/// (zarrs's underlying store listing order is implementation-defined —
```
paleolimbot, line 373: This seems like something you might want to leave to DataFusion (`ORDER_BY RS_BandPath(rast)`). You can have this as a testing utility too if that's the use case.

python/sedonadb/tests/test_funcs.py
```diff
@@ -24,3 +24,35 @@ def test_random_geometry(con):
 
     # Ensure the output is reproducible
     assert df.to_arrow_table() == df.to_arrow_table()
+
+
+def test_read_zarr(con, tmp_path):
+    # Skip cleanly if the optional `zarr` Python lib isn't installed —
+    # the binding is exercised by the Rust-side integration tests; this
+    # test only verifies the Python wrapper threads arguments through.
+    import pytest
```
paleolimbot: This can be a top-level import

python/sedonadb/tests/test_funcs.py
```diff
@@ -24,3 +24,35 @@ def test_random_geometry(con):
...
+
+    zarr = pytest.importorskip("zarr")
+    np = pytest.importorskip("numpy")
```
paleolimbot: This can be a top-level import (zarr can stay importorskip)

python/sedonadb/tests/test_funcs.py
```diff
@@ -24,3 +24,35 @@ def test_random_geometry(con):
...
+    arr[:] = np.array([[10, 11], [20, 21]], dtype=np.uint8)
+
+    # Default-mode (InDb) read — every chunk materialized into the
+    # Arrow `data` column.
+    df = con.funcs.table.sd_read_zarr(f"file://{tmp_path}")
+    assert df.count() == 2
+
+    # Options thread through: rows_per_batch slices the output.
+    df = con.funcs.table.sd_read_zarr(f"file://{tmp_path}", rows_per_batch=1)
+    assert df.count() == 2
```
paleolimbot: Make sure you also materialize these via `to_arrow_table()` or `to_pandas()`. We don't have a great way to inspect these but `arrow_tab["raster"][0].as_py()` should get you a Python `dict` you can do some basic asserts on.

paleolimbot (commented):
This is very cool!

The main comment I have is the pattern for iterating through the chunks at the start...if we have to have it be one big forward loop (i.e., we're sure there are never enough chunks that we'll meaninfully need to inspect only a few at a time), we at least need to have the byte loading (1) be async or (2) pullable in tiny chunks.

I would also really like this to not be built in to the main `sedonadb` Python package. I'm certainly willing to let it go a tiny bit for progress' sake but we're hitting the point where we've got too much built in by default (pointcloud and GPU are on my radar too for how we have to separate the build). If you have it in you to work through exposing this in `python/sedonadb-zarr` as a simple object that implements `__arrow_c_stream__` it would help me out because (1) then I don't have to do it later and (2) it will help iron out some of the missing functionality that we'll need to separate other stuff like pointclouds, which I would otherwise have to do later. An obvious hiccup in that approach is that `RecordBatchReader` is not async (so you'll just have to serve batches bounded by some byte transfer size to start).

rust/sedona-raster-zarr/src/loader.rs
james-willis: Yeah this is quite bad. This exists as a temporary way to make it possible to load a zarr while we wait for all the outdb machinery. 

My intention is to replace this with the `RS_EnsureLoaded` call once all the outdb loading + zarr integration work lands so `load_eager` effectively means load outdb then inject a Project operator with a `RS_EnsureLoaded` call.

rust/sedona-raster-zarr/src/geozarr.rs
james-willis: yeah the ND refactor lays them out in the arrow as a single array. we can add methods to the trait that indexes into the array for this.

ETA: I see my response is not totally responsive here. Ill look into if there exists such an object.

python/sedonadb/tests/test_funcs.py
james-willis: done

python/sedonadb/tests/test_funcs.py
james-willis: done

rust/sedona-raster-zarr/src/loader.rs
james-willis: That wont work here. We are just ordering the arrays (ie the bands) lexically within the raster so that band(n) will refer to the same zarr variable in every raster record.

Its something that:
1) will tend to have low cardinality (n << 100 in almost all cases, typically 1-5)
2) does not have access to a datafusion session to query.

rust/sedona-raster-zarr/src/geozarr.rs
james-willis: I filed this as a follow up in my notes. The best way to make this change would involve plumbing such an object through to the trait surface so I dont want to take that on in this PR because it would involve touching the core trait which doesnt seem in scope to me. 

`GeoTransform` + `GeoTransformEx::apply/invert` already exist in `sedona-gdal` but I don't want to bring a dependency on gdal in here. So that would probably warrant a second refactor of crates I don't want to touch right now

rust/sedona-raster-zarr/src/loader.rs
james-willis: I decided to just rip all pixel access out now and instead wait for outdb + RS_EnsureLoaded to land. at that point we will use that stack to make load_eager=true work. for now we will just error out when load_eager is true.

This PR will not allow pixel retrieval from zarrs

rust/sedona-raster-zarr/Cargo.toml
james-willis: done

rust/sedona-raster-zarr/src/udtf.rs
james-willis: > Should we default to reading outdb since this will almost certainly be the future behaviour?

Done

> It may be worth leaving off the default row count 

Done

Looking at the rest.

rust/sedona-raster-zarr/src/loader.rs
james-willis: this is somewhat mitigated by getting rid of the indb path as described in other comments. the metadata for all chunks being sync work will be resolved by the migration to ExternalFormatSpec

rust/sedona-raster-zarr/src/udtf.rs
james-willis: ok I implemented against the ExternalFormatSpec

rust/sedona/Cargo.toml
james-willis: I have a pluggable approach. should work with the outdb loader as well.

python/sedonadb/tests/test_funcs.py
james-willis: done

python/sedonadb-zarr/python/sedonadb_zarr/__init__.py
```diff
@@ -0,0 +1,127 @@
...
+        if uri is None:
+            raise ValueError(
+                "ZarrFormatSpec: could not resolve a URL from the source object"
+            )
+        if self._options.get("load_eager"):
+            raise NotImplementedError(
+                "ZarrFormatSpec: load_eager=True is not yet supported. "
+                "Pixel-byte materialisation lands with the async RS_EnsureLoaded "
+                "resolver."
+            )
```
paleolimbot: We could also just not provide this option yet?

python/sedonadb-zarr/src/lib.rs
```diff
@@ -0,0 +1,126 @@
...
+/// has been spoofed.
+#[pyfunction]
+fn zarr_udtf_capsule(py: Python<'_>) -> PyResult<Bound<'_, PyCapsule>> {
+    let udtf: Arc<dyn TableFunctionImpl> = Arc::new(ZarrReadFunction::default());
+    let payload = UdtfCapsule::new(udtf);
+    // CString::new requires owned bytes; reuse the host-side constant
+    // so the name (and its `.vN` ABI tag) lives in exactly one place.
+    let name = CString::new(UDTF_CAPSULE_NAME.to_bytes()).unwrap();
+    PyCapsule::new_with_destructor(py, payload, Some(name), |_v: UdtfCapsule, _ctx| {})
+}
```
paleolimbot: I think you can use the built-in pattern for UDTF capsules in datafusion-python:

https://github.com/apache/datafusion-python/blob/dac9ec6230dba8717d8a0d27de19a141600486b1/examples/datafusion-ffi-example/src/table_function.rs#L30-L66

We already support this in SedonaDB Python for scalar_udf and table provider if I remember correctly.

python/sedonadb-zarr/Cargo.toml
```diff
@@ -0,0 +1,44 @@
...
+datafusion = { workspace = true }
+pyo3 = { version = "0.25.1" }
+sedona-raster-zarr = { workspace = true }
+# The sedonadb Python extension; its rustc crate name is `_lib` per
+# the `[lib].name` override in `python/sedonadb/Cargo.toml`. We import
+# it in `src/lib.rs` via the leading `::_lib` syntax to disambiguate
+# from our own crate's `_lib` pymodule. Inlined here (not via
+# `workspace = true`) to avoid a self-referential workspace dep that
+# trips `cargo metadata` in cibuildwheel's isolated build env.
+sedonadb = { path = "../sedonadb", default-features = false }
```
paleolimbot: This part we need to remove (we can't take on a dependency to another Python package's private crate in another Python package to safely separate the builds). I believe the UDTF FFI suggestion I made above will eliminate the need for this one. If that doesn't work, eliminating the UDTF is probably better than this (i.e., just do the externalformatspec).

python/sedonadb-zarr/pyproject.toml
```diff
@@ -0,0 +1,49 @@
...
+requires-python = ">=3.9"
+classifiers = [
+    "Programming Language :: Rust",
+    "Programming Language :: Python :: Implementation :: CPython",
+    "Programming Language :: Python :: Implementation :: PyPy",
+]
+dynamic = ["version"]
+dependencies = [
+    "sedonadb",
+]
```
paleolimbot, line 35: Don't do this if it causes problems in CI, but this will need to be `sedonadb>=0.4.0` because it relies on newly added features.

python/sedonadb/src/context.rs
```diff
@@ -200,4 +200,60 @@ impl InternalContext {
...
+    /// [`UDTF_CAPSULE_NAME`]: crate::plugin::UDTF_CAPSULE_NAME
+    /// [`UDTF_CAPSULE_MAGIC`]: crate::plugin::UDTF_CAPSULE_MAGIC
+    pub fn register_udtf_capsule(
+        &self,
+        name: &str,
+        capsule: &Bound<'_, pyo3::types::PyCapsule>,
+    ) -> Result<(), PySedonaError> {
+        use crate::plugin::{UdtfCapsule, UDTF_CAPSULE_MAGIC, UDTF_CAPSULE_NAME};
+
+        let actual = capsule
```
paleolimbot: For this side you can use the DataFusion Python pattern:

https://github.com/apache/datafusion-python/blob/dac9ec6230dba8717d8a0d27de19a141600486b1/crates/core/src/udtf.rs#L50-L91

python/sedonadb/src/datasource.rs
```diff
@@ -143,7 +153,28 @@ impl PyExternalFormat {
...
+            list_single_object,
+            py_spec,
+        })
+    }
+}
+
+/// Read the `list_single_object` attribute on a Python spec, defaulting
+/// to `false` if the attribute is missing (older spec implementations
+/// that predate the directory-format path).
+fn read_list_single_object<'py>(
```
paleolimbot: You can also just update the pyogrio one (it's the only one so far)

python/sedonadb/tests/test_no_zarr_by_default.py
```diff
@@ -0,0 +1,43 @@
...
+# under the License.
+
+"""Architectural regression test: `sedonadb` alone has no Zarr.
+
+This locks in the plugin separation. A fresh sedonadb connection,
+created without importing `sedonadb_zarr`, must not know about the
+`sd_read_zarr` UDTF. If a future change accidentally re-bundles Zarr
+into the main package — for example by adding a convenience import or
+re-attaching the registration to `SedonaContext::new_from_context` —
+this test fails and forces a conscious decision.
```
paleolimbot: I appreciate the comprehensiveness but you can probably remove this test

python/sedonadb/Cargo.toml
```diff
@@ -23,10 +23,35 @@ edition = "2021"
...
+# like `RecordBatch` are monomorphised per-cdylib, so the drop site
+# runs the consumer's allocator regardless of who allocated, and
+# mimalloc-free on a libc-malloc pointer segfaults.
+#
+# !! WARNING: opting in with `--features mimalloc` re-introduces the
+# segfault for any user who imports a plugin alongside this wheel.
+# This isn't a perf knob — it's a "I'm shipping a closed binary that
+# will never load a plugin cdylib" knob. The plugin-safe path is to
+# link mimalloc *dynamically* (one shared `libmimalloc.so` for the
+# whole process), which is a packaging change, not a feature flag.
```
paleolimbot: Hopefully this goes away with the DataFusionFFI approach. If not, we should remove the table function completely because mimalloc is more important (mimalloc is what makes returning `Vec<>`s in a loop much faster than with a system allocator).

rust/sedona-datasource/src/provider.rs
```diff
@@ -110,3 +149,153 @@ impl ReadOptions<'_> for RecordBatchReaderTableOptions {
...
+
+/// [`TableProvider`] that treats each input URI as one opaque object.
+///
+/// Built when [`ExternalFormatSpec::list_single_object`] is `true`. The
+/// listing layer is bypassed entirely: each URI is synthesised into a
+/// single-element [`PartitionedFile`] whose `object_meta.location` is
+/// the URL's path within the [`ObjectStoreUrl`]. The format's
+/// [`ExternalFileFormat::file_source`] then drives the same scan path
+/// used by [`ListingTable`] — projections, filter pushdown, and
+/// streaming behaviour are identical.
```
paleolimbot: This is very cool, although I wonder if `SELECT * FROM 'file://path/to/something.zarr'` will work (not really an issue if it doesn't, and we can customize the filesystem catalog if we really want it to work)

rust/sedona-raster-zarr/src/dtype.rs
```diff
@@ -0,0 +1,66 @@
...
+
+/// Map a Zarr `DataType` to a SedonaDB `BandDataType`.
+///
+/// Bool maps to UInt8 losslessly (Zarr packs bools to one byte each, matching
+/// our representation). Variable-length, complex, and extended-precision
+/// types error with `NotYetImplemented` — they have no numeric counterpart
+/// in `BandDataType` today.
+pub fn zarr_to_band_data_type(dt: &ZarrDataType) -> Result<BandDataType, ArrowError> {
+    if dt.is::<BoolDataType>() {
+        Ok(BandDataType::UInt8)
```
paleolimbot, line 40: Not a blocker, but you can parameterize your smoke test with numpy dytpes to sanity check these

rust/sedona-raster-zarr/src/format_spec.rs
```diff
@@ -0,0 +1,135 @@
...
+        args: &OpenReaderArgs,
+    ) -> Result<Box<dyn RecordBatchReader + Send>> {
+        if self.load_eager {
+            return Err(DataFusionError::Plan(
+                "ZarrFormatSpec: load_eager = true is not yet supported. \
+                 Pixel-byte materialisation will be wired up when the async \
+                 RS_EnsureLoaded resolver lands."
+                    .into(),
+            ));
+        }
```
paleolimbot: Probably just leave out this option for now? Depending on what we do for ensuring loading, we may never pass this into the scan

rust/sedona-raster-zarr/src/loader.rs
```diff
@@ -0,0 +1,814 @@
...
+/// types — those don't have a `BandDataType` counterpart anyway, so the
+/// dtype check in `collect_array_infos` rejects them upstream.
+///
+/// This is the only pixel-byte read primitive in the crate. The loader
+/// itself never calls it today — it always emits OutDb anchors — but
+/// the async `RS_EnsureLoaded` resolver (follow-up PR) will. Kept here
+/// rather than dropped because (a) it's tested directly below and (b)
+/// the resolver lives in this crate and wants a sync chunk-fetch
+/// helper as a starting point.
+#[allow(dead_code)]
```
paleolimbot: Instead of allowing dead code here you can do `#[cfg(test)]` (if this is tested) or copy/paste into a follow-up issue (if it isn't)

rust/sedona-raster-zarr/src/source_uri.rs
```diff
@@ -0,0 +1,285 @@
...
+        array_path,
+        chunk_indices,
+    })
+}
+
+/// Normalize a user-supplied group URI into a local filesystem path.
+///
+/// Only `file://` and bare-path URIs are supported. Cloud schemes
+/// (`s3://`, `gs://`, `az://`, `https://`) error with a clear message.
+pub fn group_uri_to_filesystem_path(uri: &str) -> Result<std::path::PathBuf, ArrowError> {
```
paleolimbot, line 145: One of the things we may need to figure out how to do is get the object store properly across the FFI boundary (which we're currently doing with URIs, but there are issues with bridging the authentication between the store in sedonadb and sedonadb-zarr). Not a problem for this PR.

rust/sedona-raster-zarr/src/udtf.rs
```diff
@@ -0,0 +1,614 @@
...
+use sedona_schema::datatypes::SedonaType;
+use serde::{Deserialize, Serialize};
+
+use crate::loader::ZarrChunkReader;
+
+/// Register the `sd_read_zarr` UDTF on a `SessionContext`. Called from
+/// the `sedonadb-zarr` Python package's `register(con)` entry point.
+pub fn register(ctx: &SessionContext) {
+    ctx.register_udtf("sd_read_zarr", Arc::new(ZarrReadFunction::default()));
+}
```
paleolimbot: You probably want to avoid this line because the `SessionContext` is a heavy object that forces a heavy dependency (`datafusion`). Otherwise you can get away with the individual sub-crates as dependencies (compiles much faster).

rust/sedona-raster-zarr/Cargo.toml
```diff
@@ -0,0 +1,62 @@
...
+zarrs = { workspace = true, features = ["filesystem", "gzip", "zstd", "crc32c", "sharding", "transpose"] }
+zarrs_filesystem = { workspace = true }
+
+# `blosc` is gated off the Windows MinGW (gnu) toolchain only: c-blosc
+# (statically linked) bundles its own `pthread_create` / `pthread_cond_*`
+# symbols, which conflict with rtools45's `libpthread.a` during the MinGW
+# link of the R `sedonadb.dll`. The MSVC toolchain (Python wheels, etc.)
+# uses a different pthread implementation and is fine.
+[target.'cfg(not(all(target_os = "windows", target_env = "gnu")))'.dependencies]
+zarrs = { workspace = true, features = ["blosc"] }
```
paleolimbot: Do you still need this since R is not participating in the plugin system here?

rust/sedona-raster-zarr/Cargo.toml
```diff
@@ -0,0 +1,62 @@
...
+rust-version.workspace = true
+
+[lints.clippy]
+result_large_err = "allow"
+
+[dependencies]
+arrow-array = { workspace = true }
+arrow-schema = { workspace = true }
+async-trait = { workspace = true }
+datafusion = { workspace = true, features = ["sql"] }
```
paleolimbot: If you can avoid this one in `dependencies` (dev-dependencies is ok), the crate will compile faster and be smaller.

paleolimbot (commented):
Thank you for working on this, and especially thank you for going with me on the plugin concept.

I think we can solve the linking issues with by dropping the table function. This is a trivial table function (in the sense that it is not slow in the outdb case and never particularly large) and we can make a system like the ExternalFormatSpec (i.e., just routing a RecordBatchReader, which already can be FFIed) for this if the table function is a truly important way to interact with this. I also left some comments that will let the table function hang around without causing a linking issue (I'm pretty sure).

rust/sedona-raster-zarr/src/loader.rs
james-willis: sorry this slipped by my eye. fixed

rust/sedona-raster-zarr/src/udtf.rs
james-willis: done - this was dead code from an early approach I was testing anyway.

rust/sedona-raster-zarr/Cargo.toml
james-willis: done - easy once I got rid of SessionContext

rust/sedona-raster-zarr/Cargo.toml
james-willis: Should we backlog a task for r zarr support?

python/sedonadb-zarr/python/sedonadb_zarr/__init__.py
james-willis: fair, done

python/sedonadb/tests/test_no_zarr_by_default.py
james-willis: done

rust/sedona-raster-zarr/src/format_spec.rs
james-willis: done

.github/workflows/packaging.yml
```diff
@@ -159,7 +159,11 @@ jobs:
 
       - name: Install dev SedonaDB Python
         env:
-          MATURIN_PEP517_ARGS: "--features s2geography,pyo3/extension-module"
+          # `extension-module` is sedonadb's own feature gating the
+          # `#[pymodule] fn _lib` symbol; required for the wheel to load
+          # as a Python module. Plugin crates linking the rlib leave it
+          # off to avoid `PyInit__lib` duplication.
+          MATURIN_PEP517_ARGS: "--features s2geography,pyo3/extension-module,extension-module"
```
paleolimbot: Can we remove this?

.github/workflows/python.yml
```diff
@@ -117,8 +117,18 @@ jobs:
       - name: Install
         run: |
           # Keep this export in sync with the export in dev/release/verify-release-candidate.sh
-          export MATURIN_PEP517_ARGS="--features s2geography"
+          # `extension-module` is sedonadb's own feature gating the
+          # `#[pymodule] fn _lib` symbol — required for the wheel to
+          # load as a Python module. Plugin crates linking the rlib
+          # leave it off to avoid `PyInit__lib` duplication.
+          export MATURIN_PEP517_ARGS="--features s2geography,extension-module"
```
paleolimbot: Do you still need extension-module here?

ci/scripts/wheels-build-linux.sh
```diff
@@ -48,7 +48,7 @@ BEFORE_ALL_MANYLINUX="yum install -y curl zip unzip tar clang perl"
 # add quite a bit of complexity but could save time if we build wheels for linux frequently.
 # The native and Rust builds are cached on each image such that compile work is effectively
 # cached between Python versions (just not between invocations of this script).
-export CIBW_ENVIRONMENT_LINUX="VCPKG_ROOT=/vcpkg VCPKG_REF=$VCPKG_REF VCPKG_DEFAULT_TRIPLET=$VCPKG_DEFAULT_TRIPLET CMAKE_TOOLCHAIN_FILE=/vcpkg/scripts/buildsystems/vcpkg.cmake PKG_CONFIG_PATH=/vcpkg/installed/$VCPKG_DEFAULT_TRIPLET/lib/pkgconfig LD_LIBRARY_PATH=/vcpkg/installed/$VCPKG_DEFAULT_TRIPLET/lib MATURIN_PEP517_ARGS='--features s2geography,pyo3/extension-module'"
+export CIBW_ENVIRONMENT_LINUX="VCPKG_ROOT=/vcpkg VCPKG_REF=$VCPKG_REF VCPKG_DEFAULT_TRIPLET=$VCPKG_DEFAULT_TRIPLET CMAKE_TOOLCHAIN_FILE=/vcpkg/scripts/buildsystems/vcpkg.cmake PKG_CONFIG_PATH=/vcpkg/installed/$VCPKG_DEFAULT_TRIPLET/lib/pkgconfig LD_LIBRARY_PATH=/vcpkg/installed/$VCPKG_DEFAULT_TRIPLET/lib MATURIN_PEP517_ARGS='--features s2geography,pyo3/extension-module,extension-module'"
```
paleolimbot: Can you remove extension-module here?

ci/scripts/wheels-build-macos.sh
```diff
@@ -47,7 +47,7 @@ source ./wheels-bootstrap-vcpkg.sh
 export CIBW_REPAIR_WHEEL_COMMAND_MACOS="DYLD_LIBRARY_PATH=$VCPKG_INSTALL_NAME_DIR delocate-listdeps {wheel} && DYLD_LIBRARY_PATH=$VCPKG_INSTALL_NAME_DIR delocate-wheel --require-archs {delocate_archs} -w {dest_dir} {wheel}"
 
 # Pass on environment variables specifically for the build
-export CIBW_ENVIRONMENT_MACOS="$CIBW_ENVIRONMENT_MACOS _PYTHON_HOST_PLATFORM=macosx-12.0-${SEDONADB_MACOS_ARCH} MACOSX_DEPLOYMENT_TARGET=12.0 CMAKE_TOOLCHAIN_FILE=${CMAKE_TOOLCHAIN_FILE} MATURIN_PEP517_ARGS='--features s2geography,pyo3/extension-module'"
+export CIBW_ENVIRONMENT_MACOS="$CIBW_ENVIRONMENT_MACOS _PYTHON_HOST_PLATFORM=macosx-12.0-${SEDONADB_MACOS_ARCH} MACOSX_DEPLOYMENT_TARGET=12.0 CMAKE_TOOLCHAIN_FILE=${CMAKE_TOOLCHAIN_FILE} MATURIN_PEP517_ARGS='--features s2geography,pyo3/extension-module,extension-module'"
```
paleolimbot: Can you remove extension-module here?

ci/scripts/wheels-build-windows.ps1
```diff
@@ -73,7 +73,7 @@ $env:PATH += ";$scriptDirectory\windows"
 $env:GEOS_LIB_DIR = "$vcpkgLibDirectory"
 $env:GEOS_VERSION = "3.13.0"
 
-$env:MATURIN_PEP517_ARGS='--features s2geography,pyo3/extension-module'
+$env:MATURIN_PEP517_ARGS='--features s2geography,pyo3/extension-module,extension-module'
```
paleolimbot: Can you remove extension-module here?

python/sedonadb-zarr/python/sedonadb_zarr/__init__.py
```diff
@@ -0,0 +1,117 @@
...
+
+>>> import sedonadb
+>>> import sedonadb_zarr
+>>> con = sedonadb.connect()
+>>> sedonadb_zarr.register(con)
+>>> con.sql("SELECT count(*) FROM sd_read_zarr('file:///path/to/foo.zarr')").show()  # doctest: +SKIP
+
+The plugin is opt-in: SedonaDB itself does not bundle Zarr support, so
+applications that don't import ``sedonadb_zarr`` pay no zarr build or
+runtime cost.
```
paleolimbot: This is written using ReST cross-linking but mkdocs is building this using Markdown

python/sedonadb-zarr/python/sedonadb_zarr/__init__.py
```diff
@@ -0,0 +1,117 @@
...
+    registers the UDTF without error.
+
+    Parameters
+    ----------
+    con
+        A ``sedonadb`` ``Context`` (the object returned by
+        ``sedonadb.connect()``). Internally, this function hands a
+        :class:`ZarrTableFunction` instance to sedonadb's
+        ``register_udtf_capsule``, which uses ``datafusion-ffi``'s
+        ``FFI_TableFunction`` to bridge the cdylib boundary.
```
paleolimbot: Here too

python/sedonadb-zarr/Cargo.toml
```diff
@@ -0,0 +1,43 @@
...
+crate-type = ["cdylib"]
+
+[dependencies]
+arrow-array = { workspace = true }
+# UDTF handoff to sedonadb is via `datafusion-ffi`'s `FFI_TableFunction`
+# (the `datafusion-python` pattern). The plugin and host don't share
+# any Rust types directly — concrete `RecordBatch` traffic flows
+# through Arrow C with release callbacks owned by the producing
+# cdylib, so plugin and host can use different allocators.
+datafusion-catalog = { workspace = true }
```
paleolimbot: You should be able to remove these comments since this is now back to a normal Python with maturin crate

python/sedonadb-zarr/pyproject.toml
```diff
@@ -0,0 +1,49 @@
...
+    "numpy",
+]
+
+[tool.maturin]
+features = ["pyo3/extension-module"]
+# `_zarr_lib` rather than `_lib` so the generated PyInit symbol
+# doesn't collide with sedonadb's `_lib` under cargo workspace
+# feature unification. The Python wrapper imports from this path.
+module-name = "sedonadb_zarr._zarr_lib"
+python-source = "python"
```
paleolimbot: This should be able to go back to a normal line without a comment

python/sedonadb-zarr/src/lib.rs
james-willis: done

python/sedonadb-zarr/Cargo.toml
james-willis: done. was able to keep the UDTF

python/sedonadb-zarr/pyproject.toml
james-willis: done- testing now.

rust/sedona-raster-zarr/src/source_uri.rs
james-willis: 🏈  I think this is something I will have to worry about later!

python/sedonadb/src/context.rs
james-willis: done

rust/sedona-raster-zarr/Cargo.toml
james-willis: I have r support for the plugin in my notes. Will eventually file against the repo

james-willis: removed for now

python/sedonadb/src/context.rs
```diff
@@ -200,4 +212,39 @@ impl InternalContext {
...
+
+        let session_capsule = crate::plugin::create_session_capsule(py, &self.udtf_task_provider)?;
+        let returned = spec
+            .getattr(crate::plugin::UDTF_ATTR)?
+            .call1((session_capsule,))?;
+        let returned = returned.downcast::<PyCapsule>().map_err(|e| {
+            PySedonaError::SedonaPython(format!(
+                "plugin's {} must return a PyCapsule, got {e}",
+                crate::plugin::UDTF_ATTR
+            ))
```
paleolimbot: Might be nice to have this follow `import_ffi_scalar_udf()` and use any of the utilities in that file if it's possible to do so

python/sedonadb-zarr/python/sedonadb_zarr/__init__.py
```diff
@@ -0,0 +1,117 @@
...
+
+    Use with ``con.read_format(spec, uri)``:
+
+    >>> con.read_format(ZarrFormatSpec(), "file:///path/to/foo.zarr")  # doctest: +SKIP
+
+    Supported options (via :meth:`with_options`):
+
+    - ``arrays`` (``list[str]`` or JSON-string) — explicit subset of
+      group arrays to read.
+    """
```
paleolimbot, line 51: This is also ReST and not markdown

python/sedonadb-zarr/src/lib.rs
```diff
@@ -0,0 +1,177 @@
...
+/// Codec capsule name the host passes to us via the `session` argument.
+const CODEC_CAPSULE_NAME: &std::ffi::CStr = c"datafusion_logical_extension_codec";
+
+/// Zarr UDTF surface. The Python plugin instantiates one of these and
+/// hands the instance to sedonadb's `con._impl.register_udtf` (Python
+/// wrapper). Sedonadb then calls `__datafusion_table_function__(session)`
+/// on it to obtain the FFI struct.
+#[pyclass(name = "ZarrTableFunction", module = "sedonadb_zarr")]
+#[derive(Default, Debug, Clone)]
+pub struct ZarrTableFunction;
```
paleolimbot: This table function is causing a lot of code and isn't necessary to access this from Python. Are you sure it needs to exist or can you push it to a follow up and evaluate the priority later? We don't have a `read_parquet()` either.

python/sedonadb-zarr/src/lib.rs
```diff
@@ -0,0 +1,177 @@
...
+    }
+}
+
+// Named `_zarr_lib` (not `_lib`) so the generated `PyInit__zarr_lib`
+// symbol doesn't collide with sedonadb's `PyInit__lib` when cargo's
+// workspace feature unification (e.g. `cargo build --all-features`)
+// brings sedonadb's pymodule into our cdylib's link.
+#[pymodule]
+fn _zarr_lib(m: &Bound<'_, PyModule>) -> PyResult<()> {
+    m.add_class::<ZarrTableFunction>()?;
```
paleolimbot: This can go back to `_lib`?

rust/sedona-datasource/src/provider.rs
james-willis: it doesn't — the URL-table path goes through ListingTable, which has the same directory-vs-file mismatch the read_format path hit before. Same gap applies to pyogrio.

We can file a followup bug. It would be good to make this work but also we dont want to assume a zarr ends in `.zarr` (they often dont)

rust/sedona-raster-zarr/src/dtype.rs
james-willis: done

python/sedonadb/Cargo.toml
james-willis: datafusion ffi solved this one! plugin + mimalloc is working.

python/sedonadb-zarr/src/lib.rs
james-willis: It doesnt need to exist, but I think its a much better interface.

 I think it is fair to split it out into a separate PR that is lower priority. Thatll give us an opportunity to evaluate its integration with the plugin model better and maybe from someone with more subject matter expertise

rust/sedona/Cargo.toml
```diff
@@ -50,6 +50,7 @@ s2geography = ["dep:sedona-s2geography", "dep:sedona-spatial-join-geography", "s
 tempfile = { workspace = true }
 tokio = { workspace = true, features = ["rt-multi-thread"] }
 rstest = { workspace = true }
+sedona-raster = { workspace = true }
```
paleolimbot: Is this still needed?

```suggestion
```

paleolimbot (approved):
Thank you! Looking forward to this!

james-willis:
sorry dewey, I havent finished dealing with your last round of comments yet!

python/sedonadb/src/datasource.rs
james-willis: we have the Zarr one now as well.

While pyogrio only supports single file formats, zarr (and maybe future file formats) support dirs so I think this is the right place to do this change.

or maybe im missing your point. Are you saying something about having a default behavior?

rust/sedona/Cargo.toml
james-willis: removed

james-willis:
@paleolimbot now I am happy with it. review/approve/merge at your discretion.
 
 
 ETA: I think the test failure is a flakey test in GDAL

python/sedonadb/python/sedonadb/context.py
```diff
@@ -333,6 +346,50 @@ def read_pyogrio(
             self.options,
         )
 
+    def read_format(
+        self,
+        spec: "ExternalFormatSpec",
+        table_paths: Union[str, Path, Iterable[str]],
+        check_extension: bool = False,
+    ) -> DataFrame:
+        """Read one or more paths using a Python-defined `ExternalFormatSpec`.
```
paleolimbot, line 355: This is great!

paleolimbot (approved):
Thank you!

python/sedonadb/python/sedonadb/context.py
james-willis: I almost forgot to create it! Would have made this feature challenging to use 😆

PR #859, opened 2026-05-19
feat(python/sedonadb): add DataFrame.sort with composable SortExpr

jiayuasu:
Redesigned per @paleolimbot's [review on the earlier draft](https://github.com/apache/sedona-db/pull/859) — replacing pandas-style `sort_values(by=, ascending=)` with the composable API pattern from DataFusion-python / DuckDB / Ibis.

Continues Phase P2 of #791.

## API

```python
from sedonadb.expr import col, sort_expr

# Common cases — Expr.asc()/.desc() return SortExpr
df.sort("x")                          # str auto-promotes to ascending
df.sort(col("x").desc())
df.sort(col("x"), col("y").desc())    # varargs, multi-key
df.sort(col("x") + col("y"))          # arbitrary Expr key

# Full control via the factory
df.sort(sort_expr(col("x"), asc=False, nulls_first=True))
```

`DataFrame.sort` accepts each key as `str | Expr | SortExpr`. Strings and bare `Expr` auto-promote to ascending keys with nulls last; `Expr.asc()` / `Expr.desc()` cover the common direction switch; `sedonadb.expr.sort_expr(expr, asc=..., nulls_first=...)` exposes the full DataFusion `SortExpr` knobs.

## What's gone vs. the earlier draft

- `sort_values()` — removed entirely. Per the discussion: no pandas-compat duplicate at this layer; that can come later in a dedicated GeoPandas-compat surface.
- `by=` / `ascending=` keyword args — replaced by varargs (formats better with Ruff and matches the three reference interfaces).

## Null placement

`Expr.asc()` and `Expr.desc()` default to `nulls_first=False` (nulls last) regardless of direction — matches the previous draft's pandas-style default and what the rest of the SedonaDB Python API ships. SQL-style nulls-first-on-descending is available via `sort_expr(col("x"), asc=False, nulls_first=True)`.

## Implementation

| File | Change |
|---|---|
| `python/sedonadb/src/expr.rs` | New `PySortExpr` PyO3 class wrapping `datafusion_expr::SortExpr`. `Expr.asc(nulls_first)` / `.desc(nulls_first)` methods. `expr_sort_expr(expr, asc, nulls_first)` factory. |
| `python/sedonadb/src/dataframe.rs` | `InternalDataFrame::sort(Vec<PySortExpr>)`; old `sort_by_keys` gone. |
| `python/sedonadb/src/lib.rs` | Register `PySortExpr` + `expr_sort_expr`. |
| `python/sedonadb/python/sedonadb/expr/expression.py` | Python `SortExpr` class, `sort_expr()` factory, `Expr.asc/.desc` methods. |
| `python/sedonadb/python/sedonadb/expr/__init__.py` | Re-export `SortExpr`, `sort_expr`. |
| `python/sedonadb/python/sedonadb/dataframe.py` | `DataFrame.sort(*keys)`. Module-level `SortExpr` import (no lazy imports, per the policy locked on #852). |

## Tests

- `tests/expr/test_dataframe_sort.py` — 14 tests covering string/Expr/SortExpr keys, mixed-direction multi-key, computed-Expr keys, nulls-last default in both directions, `sort_expr(nulls_first=True)`, lazy return type (`isinstance(out, DataFrame)`), and the empty / bad-type / unknown-column error paths.
- `tests/expr/test_expression.py` — 9 new tests pinning the exact `repr()` of `SortExpr` for asc / desc / both null placements / the factory / both type-guard rejections.

Local: 97 unit + 29 doctests + `ruff format` + `ruff check` all clean.

### DIFF
--- a/python/sedonadb/python/sedonadb/dataframe.py
+++ b/python/sedonadb/python/sedonadb/dataframe.py
@@ -19,7 +19,7 @@ import io
 from pathlib import Path
 from typing import TYPE_CHECKING, Any, Dict, Iterable, List, Literal, Optional, Union
 
-from sedonadb.expr import Expr
+from sedonadb.expr import Expr, SortExpr
 from sedonadb.expr import Literal as _SedonaLit
 from sedonadb.expr import col as _col
 from sedonadb.expr.expression import _to_expr
@@ -250,6 +250,73 @@ class DataFrame:
             self._impl.filter([e._impl for e in exprs]),
             self._options,
         )
+
+    def sort(self, *keys: Union[str, Expr, SortExpr]) -> "DataFrame":
+        """Sort rows by one or more keys.
+
+        Each argument is either a column name (`str`), an `Expr`, or a
+        `SortExpr` (built via `Expr.asc()` / `Expr.desc()` or
+        `sedonadb.expr.sort_expr(...)`). Strings and bare `Expr`s
+        auto-promote to ascending sort keys with nulls placed last; for
+        descending order or null-first placement, use the explicit
+        `SortExpr` forms.
+
+        Null placement defaults to "nulls last" for both ascending and
+        descending sorts (overriding DataFusion's SQL-style nulls-first-
+        on-descending). Override via `sort_expr(expr, asc=..., nulls_first=...)`
+        for the SQL behavior.
+
+        Args:
+            *keys: One or more sort keys, in order of priority. At least
+                one is required.
+
+        Examples:
+
+            >>> from sedonadb.expr import col
+            >>> sd = sedona.db.connect()
+            >>> df = sd.sql("SELECT * FROM (VALUES (3), (1), (2)) AS t(x)")
+            >>> df.sort("x").show()
+            ┌───────┐
+            │   x   │
+            │ int64 │
+            ╞═══════╡
+            │     1 │
+            ├╌╌╌╌╌╌╌┤
+            │     2 │
+            ├╌╌╌╌╌╌╌┤
+            │     3 │
+            └───────┘
+            >>> df.sort(col("x").desc()).show()
+            ┌───────┐
+            │   x   │
+            │ int64 │
+            ╞═══════╡
+            │     3 │
+            ├╌╌╌╌╌╌╌┤
+            │     2 │
+            ├╌╌╌╌╌╌╌┤
+            │     1 │
+            └───────┘
+        """
+        if not keys:
+            raise ValueError("sort() requires at least one sort key")
+
+        coerced: List = []
+        for k in keys:
+            if isinstance(k, SortExpr):
+                coerced.append(k._impl)
+            elif isinstance(k, Expr):
+                # Default direction is ascending, nulls last.
+                coerced.append(k.asc()._impl)
+            elif isinstance(k, str):
+                coerced.append(_col(k).asc()._impl)
+            else:
+                raise TypeError(
+                    f"sort() expects str, Expr, or SortExpr arguments, "
+                    f"got {type(k).__name__}"
+                )
+
+        return DataFrame(self._ctx, self._impl.sort(coerced), self._options)
 
     def limit(self, n: Optional[int], /, *, offset: int = 0) -> "DataFrame":
         """Limit result to n rows starting at offset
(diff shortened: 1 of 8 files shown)


paleolimbot (commented):
Since Pandas invented `sort_values()` there have been more elegant/composable ways to handle this that have evolved.

For our purposes, I think we should:

- Expose all the bells and whistles of DataFusion's `SortExpr` via `sedonadb.expr.sort_expr(expr, <options like asc/dsc/nulls>)`
- Add methods to `Expr` (`.asc(nulls)`, `.desc(nulls)`) that return `SortExpr`
- Accept `str`, `Expr`, or `SortExpr` in sort_values
- Name it either `sort()` (DataFusion python, DuckDB) or `order_by()` (Ibis)
- Accept multiple arguments instead of a list (auto formats with Ruff into multiple lines better, all three newer interfaces do this)

Some more elegant interfaces for reference:

- https://datafusion.apache.org/python/autoapi/datafusion/dataframe/index.html
- https://ibis-project.org/reference/expression-tables#ibis.expr.types.relations.Table.order_by
- https://duckdb.org/docs/current/clients/python/relational_api#sort

jiayuasu:
Should we keep both `sort_values` and `sort` to maintain some compatibility with Pandas?

jiayuasu:
@paleolimbot — redesigned per your review. Force-pushed `d88b91f6` over the earlier `sort_values` commit. Branch name is stale (kept it to preserve this PR thread); the PR title and description are updated.

All five of your asks landed:

1. **`sedonadb.expr.sort_expr(expr, asc=True, nulls_first=False)`** — full-knob factory.
2. **`Expr.asc(nulls_first=False)` / `Expr.desc(nulls_first=False)`** — common-case sugar, return `SortExpr`.
3. **`DataFrame.sort` accepts `str | Expr | SortExpr`** — strings and bare `Expr` auto-promote to ascending keys.
4. **Named `sort()`** (matching DataFusion-python and DuckDB) rather than `order_by()`.
5. **Varargs** — `df.sort(col("x"), col("y").desc())`, no list.

`sort_values()` is gone — no pandas-compat duplicate at this layer, per the thread.

The `Expr.asc/.desc` shortcuts default to `nulls_first=False` (nulls last) in both directions, matching what the rest of the SedonaDB Python API already ships. SQL-style nulls-first-on-descending is reachable via `sort_expr(col("x"), asc=False, nulls_first=True)`.

python/sedonadb/python/sedonadb/expr/expression.py
```diff
@@ -18,10 +18,12 @@
 from typing import Any, Iterable, Optional
 
 from sedonadb._lib import InternalExpr as _InternalExpr
+from sedonadb._lib import InternalSortExpr as _InternalSortExpr
 from sedonadb._lib import expr_binary as _expr_binary
 from sedonadb._lib import expr_col as _expr_col
 from sedonadb._lib import expr_lit as _expr_lit
 from sedonadb._lib import expr_not as _expr_not
+from sedonadb._lib import expr_sort_expr as _expr_sort_expr
```
paleolimbot: Can you consense these into a single import? Ruff's "format imports" might do this for you nicely.

python/sedonadb/python/sedonadb/dataframe.py
```diff
@@ -251,6 +251,73 @@ def filter(self, *exprs: Expr) -> "DataFrame":
...
+            elif isinstance(k, Expr):
+                # Default direction is ascending, nulls last.
+                coerced.append(k.asc()._impl)
+            elif isinstance(k, str):
+                coerced.append(_col(k).asc()._impl)
+            else:
+                raise TypeError(
+                    f"sort() expects str, Expr, or SortExpr arguments, "
+                    f"got {type(k).__name__}"
+                )
```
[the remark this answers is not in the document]
paleolimbot: I think you're fine here 🙂

paleolimbot (approved):
Thank you!

python/sedonadb/python/sedonadb/expr/expression.py
jiayuasu: Combined in d17eccc8 — single `from sedonadb._lib import (..., ...)` block. Didn't enable a workspace-wide isort rule in this PR; happy to do it separately if you'd like.

PR #865, opened 2026-05-21
Fix ST_AsBinary return type for WkbView input

### LINKED ISSUE #864: ST_AsBinary returns BinaryView for WkbView / GeoParquet input
`ST_AsBinary` currently returns Arrow `BinaryView` when the input geometry is stored as `WkbView`, while it returns Arrow `Binary` for `Wkb`.

That makes the SQL-visible return type depend on internal storage representation. This shows up with GeoParquet inputs because a valid GeoParquet file can be read with its WKB geometry column backed by Arrow `BinaryView`, producing `SedonaType::WkbView`.

Expected behavior:

`ST_AsBinary` / `ST_AsWKB` should consistently return Arrow `Binary`.

Actual behavior:

For `WkbView` input, `ST_AsBinary` returns Arrow `BinaryView` to avoid a cast.

Impact:

Downstream consumers expecting normal binary output need to work around this locally, for example by wrapping `ST_AsBinary(...)` in `sd_simplifystorage(...)` or another explicit cast.

Suggested fix:

Keep the existing clone fast path for plain `Wkb` input, but cast `WkbView` and other non-`Wkb` inputs to Arrow `Binary` in `invoke_batch()`.

A regression test can use a small valid GeoParquet file with WKB geometry stored as Arrow `BinaryView`; before the fix, the test fails because `ST_AsBinary` returns `BinaryView` instead of `Binary`.

camden-lowrance:
## Summary

This updates `ST_AsBinary` / `ST_AsWKB` so the return type is consistently Arrow `Binary`, including when the input geometry is stored as `WkbView`.

The previous implementation returned Arrow `BinaryView` for `WkbView` input to avoid a cast. That exposed an internal storage detail through the SQL-visible return type and required downstream callers to add their own simplification/cast workaround.

This change keeps the no-copy clone path for plain `Wkb` input and casts non-`Wkb` inputs to Arrow `Binary`.

Fixes #864.

## Changes

- Remove the special-case `WkbView -> BinaryView` return type.
- Preserve the clone fast path for `SedonaType::Wkb`.
- Cast `WkbView` output to Arrow `Binary`.
- Update existing `WkbView` UDF tests to expect `Binary`.
- Add a regression test that writes a valid GeoParquet file with WKB geometry stored as Arrow `BinaryView`, reads it as `WkbView`, and verifies `ST_AsBinary` returns Arrow `Binary`.

## Testing

- `pixi run cargo fmt --all -- --check`
- `pixi run test-st-asbinary-rust`
- `pixi run test-st-asbinary-geoparquet`

I also verified the new GeoParquet regression test catches the previous behavior. With the old `WkbView -> BinaryView` special case restored, this command fails:

```shell
pixi run test-st-asbinary-geoparquet
```

Failure excerpt:

```text
thread 'context::tests::st_asbinary_geoparquet_returns_binary' panicked at rust/sedona/src/context.rs:825:9:
assertion `left == right` failed
  left: BinaryView
 right: Binary
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5327,6 +5327,7 @@ dependencies = [
  "geo-types",
  "object_store",
  "parking_lot",
+ "parquet",
  "regex",
  "rstest",
  "sedona-common",
--- a/rust/sedona-functions/src/st_asbinary.rs
+++ b/rust/sedona-functions/src/st_asbinary.rs
@@ -42,13 +42,6 @@ struct STAsBinary {}
 
 impl SedonaScalarKernel for STAsBinary {
     fn return_type(&self, args: &[SedonaType]) -> Result<Option<SedonaType>> {
-        // If we have WkbView input, return BinaryView to avoid a cast
-        if args.len() == 1 {
-            if let SedonaType::WkbView(_, _) = args[0] {
-                return Ok(Some(SedonaType::Arrow(DataType::BinaryView)));
-            }
-        }
-
         let matcher = ArgMatcher::new(
             vec![ArgMatcher::is_geometry_or_geography()],
             SedonaType::Arrow(DataType::Binary),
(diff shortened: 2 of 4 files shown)


camden-lowrance:
Closing this since the issue discussion clarified that ST_AsBinary() preserving view storage is intentional. I will follow up with a docs-focused PR for the sd_simplifystorage(ST_AsBinary(...)) pattern instead.

PR #867, opened 2026-05-22
feat(c/sedona-s2geography,rust/sedona-functions): Add ST_Segmentize, ST_TessellateGeom, ST_TessellateGeog

paleolimbot:
This PR wires up the existing support for ST_Segmentize, ST_TessellateGeom, and ST_TessellateGeog provided by SedonaDB, including documentation and tests from the Python side. I also added a `ST_ToGeography()` and `ST_ToGeometry()` (Snowflake-inspried) that do a metadata-only (non-validating) cast between the two. Validation functions don't currently exist in s2geography (when they do we can consider whether it's a reasonable default to validate Geography on the way in).

While I was here I added a native `ST_Segmentize()` for geometry, since it was missing.

Screenshot 2026-05-26 at 3 48 31 PM

Screenshot 2026-05-26 at 3 50 05 PM


docs/reference/sql/st_segmentize.qmd
```diff
@@ -0,0 +1,42 @@
...
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+title: ST_Segmentize
+description: Densifies a geometry by adding intermediate points along segments that exceed a maximum length.
+kernels:
+  - returns: geometry
+    args:
+    - geometry
```
james-willis, line 24: geog -> geog is supported too, right?

docs/reference/sql/st_tessellategeom.qmd
```diff
@@ -0,0 +1,60 @@
...
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+title: ST_TessellateGeom
+description: Tessellates a geography into a geometry using spherical interpolation.
```
james-willis: I find the current wording here a bit difficult to parse. Same comment for the Geog function and maybe the longer descriptions as well.


```suggestion
description: Converts a geography to geometry, densifying edges to approximate geodesic segments as planar lines.
```

python/sedonadb/tests/functions/test_segmentize.py
```diff
@@ -0,0 +1,237 @@
...
+
+@pytest.mark.parametrize("eng", [SedonaDB, PostGIS])
+@pytest.mark.parametrize(
+    ("geom", "max_segment_length", "expected"),
+    [
+        # Nulls
+        pytest.param("POINT (0 0)", None, None, id="null_length"),
+        # Empties
+        pytest.param("POINT EMPTY", 1e9, "POINT (nan nan)", id="empty_point"),
+        pytest.param(
```
james-willis, line 30: what happens if max length  <= 0?

python/sedonadb/tests/geography/test_tessellate.py
```diff
@@ -0,0 +1,641 @@
...
+        # GeometryCollections without segmentation (large max segment)
+        pytest.param(
+            "GEOMETRYCOLLECTION (POINT (0 1), LINESTRING (0 1, 1 2))",
+            1e9,
+            "GEOMETRYCOLLECTION (POINT (0 1), LINESTRING (0 1, 1 2))",
+            id="geometrycollection_large_seg",
+        ),
+    ],
+)
+def test_st_segmentize_no_split(eng, geog, max_segment_length, expected):
```
james-willis, line 127: should this be in the segmentize tests file?

python/sedonadb/tests/geography/test_tessellate.py
```diff
@@ -0,0 +1,641 @@
...
+            "LINESTRING (-10 45, 10 45, 30 45)",
+            10000.0,
+            "LINESTRING (-10 45, -5.019332 45.328489, 0 45.438549, "
+            "5.019332 45.328489, 10 45, 14.980668 45.328489, 20 45.438549, "
+            "25.019332 45.328489, 30 45)",
+            id="linestring_tessellate_multiseg",
+        ),
+    ],
+)
+def test_st_tessellategeom_linestring_split(eng, geog, tolerance, expected):
```
james-willis, line 548: tolerance = 0 would be a good case to test

python/sedonadb/tests/geography/test_tessellate.py
```diff
@@ -0,0 +1,641 @@
...
+        pytest.param("POINT (0 1)", "POINT (0 1)", id="point"),
+    ],
+)
+def test_st_togeometry(eng, geog, expected):
+    eng = eng.create_or_skip()
+    eng.assert_query_result(f"SELECT ST_ToGeometry({geog_or_null(geog)})", expected)
+
+
+@pytest.mark.parametrize("eng", [SedonaDB])
+@pytest.mark.parametrize(
```
james-willis, line 669: can we test CRS preservation?

rust/sedona-functions/src/st_togeomgeog.rs
```diff
@@ -0,0 +1,270 @@
...
+    fn return_type(&self, args: &[SedonaType]) -> Result<Option<SedonaType>> {
+        // First check if the matcher matches
+        let matched = self.matcher.match_args(args)?;
+        if matched.is_none() {
+            return Ok(None);
+        }
+
+        // Return the input type with the target edges. NULL returns the target edge
+        // type with no CRS for the purposes of parameter binding.
+        let input_type = &args[0];
```
james-willis, line 82: should we reject Geom -> Geog when the CRS is not something that converts properly to a geog?

rust/sedona-functions/src/st_segmentize.rs
```diff
@@ -0,0 +1,708 @@
...
+/// Calculate the number of segments needed for a single edge
+fn calc_num_segments<C: CoordTrait<T = f64>>(c1: &C, c2: &C, max_segment_length: f64) -> usize {
+    let dx = c2.x() - c1.x();
+    let dy = c2.y() - c1.y();
+    let distance = (dx * dx + dy * dy).sqrt();
+
+    if distance <= max_segment_length {
+        1
+    } else {
+        (distance / max_segment_length).ceil() as usize
```
james-willis, line 257: Should we worry about overflow when max_segment_length is really low?

Should we use try_into instead of a raw cast? Then we can deal with the overflow.

james-willis (commented):
I did not review the submodule changes

python/sedonadb/tests/functions/test_segmentize.py
paleolimbot: I added tests here (it's an error)

python/sedonadb/tests/geography/test_tessellate.py
paleolimbot: For better or worse, all the geography stuff is tested under the tests/geography package. In both it's a little tricky to find what's tested where (there's a lot to test!)

rust/sedona-functions/src/st_togeomgeog.rs
paleolimbot: Good catch, I added this check!

rust/sedona-functions/src/st_segmentize.rs
paleolimbot: Good catch...for geography I had added a hard-coded maximum of 65536 added segments per edge and I added/tested that here as well (one level up)

rust/sedona-functions/src/st_segmentize.rs
```diff
@@ -0,0 +1,749 @@
...
+    c2: &C,
+    num_segments: usize,
+) -> Result<(), SedonaGeometryError> {
+    let x1 = c1.x();
+    let y1 = c1.y();
+    // SAFETY: Only called when dims.size() == 3, so index 2 is valid
+    let d1 = unsafe { c1.nth_unchecked(2) };
+    let x2 = c2.x();
+    let y2 = c2.y();
+    let d2 = unsafe { c2.nth_unchecked(2) };
```
james-willis:
```suggestion
    let d2 = c2.nth(2).unwrap();
```

rust/sedona-functions/src/st_segmentize.rs
```diff
@@ -0,0 +1,749 @@
...
+fn write_interpolated_3d<C: CoordTrait<T = f64>>(
+    writer: &mut impl Write,
+    c1: &C,
+    c2: &C,
+    num_segments: usize,
+) -> Result<(), SedonaGeometryError> {
+    let x1 = c1.x();
+    let y1 = c1.y();
+    // SAFETY: Only called when dims.size() == 3, so index 2 is valid
+    let d1 = unsafe { c1.nth_unchecked(2) };
```
james-willis: Unless there a good perf reason we should probably avoid unsafe here

```suggestion
    let d1 = c1.nth(2).unwrap();
```

PR #871, opened 2026-05-23
feat(python/sedonadb): add DataFrame.drop

jiayuasu:
Continues Phase P2 of #791 with `DataFrame.drop` — the smallest of the remaining P2 ops.

## API

```python
df.drop("a")
df.drop("a", "b", "c")
```

- **Varargs** of column names (no `columns=` kwarg). Same pattern as `sort()` from #859 — matches DataFusion-python / Ibis / Polars, avoids the pandas-style keyword.
- **Strings only**. `Expr` arguments are rejected at the Python boundary; "drop a computed expression" has no meaning at the schema level.

## Unknown-column behavior

Worth calling out — DataFusion's `DataFrame::drop_columns` is **permissive**: it silently no-ops on names that aren't in the schema. That hides typos. To match user expectations (and pandas' `KeyError` behavior), this PR validates the column names Python-side and raises a `KeyError` listing the available columns when one is missing. The exact message format is pinned by `test_drop_unknown_column_raises_keyerror`.

This is more restrictive than `select`/`filter` (where DataFusion validates at plan-build time and includes "Valid fields are X, Y" in the error). The asymmetry is forced by DataFusion's silent-no-op default on `drop_columns`; the workaround is one Python-side schema lookup per call, which is cheap.

## Implementation

| File | Change |
|---|---|
| `python/sedonadb/src/dataframe.rs` | New `InternalDataFrame::drop_columns(Vec<String>)`. Materializes a `Vec<&str>` and calls DataFusion's `DataFrame::drop_columns`. Step-by-step comments. |
| `python/sedonadb/python/sedonadb/dataframe.py` | `DataFrame.drop(*cols: str)`. Validates non-empty, str-only, and known columns. |

## Test plan

9 tests in `tests/expr/test_dataframe_drop.py`:

- **Positive**: single-column, multi-column, column-order preservation.
- **Errors**: empty args → `ValueError`; non-str arg → `TypeError`; `Expr` arg → `TypeError`; `columns=` kwarg → Python's unexpected-keyword `TypeError`; unknown column → `KeyError` with exact pinned message listing available columns.

Local: 9 unit + 19 doctests + `ruff format` + `ruff check` all clean.

### DIFF
--- a/python/sedonadb/python/sedonadb/dataframe.py
+++ b/python/sedonadb/python/sedonadb/dataframe.py
@@ -318,6 +318,51 @@ class DataFrame:
 
         return DataFrame(self._ctx, self._impl.sort(coerced), self._options)
 
+    def drop(self, *cols: str) -> "DataFrame":
+        """Drop the named columns.
+
+        Returns a new lazy `DataFrame` with each named column removed.
+        Only column-name strings are accepted; expression arguments are
+        rejected because "drop a computed expression" has no meaning at
+        the schema level. Unknown column names raise a `SedonaError` at
+        plan-build time, with the list of valid field names included in
+        the message.
+
+        Args:
+            *cols: One or more column names to drop. At least one is
+                required.
+
+        Examples:
+
+            >>> sd = sedona.db.connect()
+            >>> df = sd.sql("SELECT 1 AS a, 2 AS b, 3 AS c")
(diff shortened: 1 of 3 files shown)


paleolimbot (approved):
Thank you!

PR #872, opened 2026-05-24
chore(docker): Add Dockerfile for building GPU-enabled sedonadb image

pwrliang:
This PR provides a Dockerfile to build a Sedona-DB image with GPU support. When a container is created with the image, users can use JupyterLab to use sedona-db.

Here's the example:
```
docker build -f docker/sedonadb-gpu.dockerfile --build-arg CMAKE_CUDA_ARCHITECTURES=“86;89” -t sedonadb-gpu . # Build image (taking about 20mins)
docker run -it --rm --gpus all sedonadb-gpu # Start a container with the image
```

This PR also updates the docs on using the image.

### DIFF
--- a/docker/sedonadb-gpu.dockerfile
+++ b/docker/sedonadb-gpu.dockerfile
@@ -64,7 +64,7 @@ ENV CMAKE_TOOLCHAIN_FILE=$VCPKG_ROOT/scripts/buildsystems/vcpkg.cmake
 ENV PKG_CONFIG_PATH=$VCPKG_ROOT/installed/x64-linux/lib/pkgconfig
 ENV CC=clang-18
 ENV CXX=clang++-18
-ARG CMAKE_CUDA_ARCHITECTURES=native
+ARG CMAKE_CUDA_ARCHITECTURES="86"
 ENV CMAKE_CUDA_ARCHITECTURES=${CMAKE_CUDA_ARCHITECTURES}
 ENV LD_LIBRARY_PATH=/usr/local/cuda/lib64/stubs:${LD_LIBRARY_PATH}
 
@@ -76,7 +76,7 @@ ENV PATH="/opt/venv/bin:$PATH"
 ENV MATURIN_PEP517_ARGS="--features s2geography,gpu"
 
 # Install maturin AND JupyterLab inside the virtual environment
-RUN pip3 install maturin jupyterlab pyproj
+RUN pip3 install maturin jupyterlab pyproj geoarrow-pyarrow pandas
 
 # Install the project. You may add '-e' to enable the editable mode (uses pyproject -> maturin under the hood) for debug/development purposes
 RUN pip3 install "python/sedonadb" -vv
--- a/docker/sedonadb-gpu.dockerfile
+++ b/docker/sedonadb-gpu.dockerfile
@@ -77,16 +77,26 @@ ENV MATURIN_PEP517_ARGS="--features s2geography,gpu"
 # Install maturin AND JupyterLab inside the virtual environment
 RUN pip3 install maturin jupyterlab pyproj
 
-# Install the project in editable mode (uses pyproject -> maturin under the hood)
-RUN pip3 install -e "python/sedonadb" -vv
+# Install the project. You may add '-e' to enable the editable mode (uses pyproject -> maturin under the hood) for debug/development purposes
+RUN pip3 install "python/sedonadb" -vv
 
 # Clean up the library path so the container uses the real driver at runtime
 ENV LD_LIBRARY_PATH=/usr/local/cuda/lib64
-
 ENV NVIDIA_DRIVER_CAPABILITIES=compute,utility,graphics
+
+# --- SECURITY CONFIGURATION ---
+# Create a non-root user
+RUN useradd -ms /bin/bash jupyteruser
+
+# Change ownership of the workspace and virtual environment to the new user
+RUN chown -R jupyteruser:jupyteruser /workspace /opt/venv
+
+# Switch to the non-root user
+USER jupyteruser
+# ------------------------------
 
 # Expose the default JupyterLab port
 EXPOSE 8888
 
-# Launch JupyterLab by default when the container starts
-CMD ["jupyter", "lab", "--ip=0.0.0.0", "--port=8888", "--no-browser", "--allow-root"]
+# Launch JupyterLab
+CMD ["jupyter", "lab", "--ip=0.0.0.0", "--port=8888", "--no-browser"]
--- a/docker/sedonadb-gpu.dockerfile
+++ b/docker/sedonadb-gpu.dockerfile
@@ -64,7 +64,8 @@ ENV CMAKE_TOOLCHAIN_FILE=$VCPKG_ROOT/scripts/buildsystems/vcpkg.cmake
 ENV PKG_CONFIG_PATH=$VCPKG_ROOT/installed/x64-linux/lib/pkgconfig
 ENV CC=clang-18
 ENV CXX=clang++-18
-ENV CMAKE_CUDA_ARCHITECTURES="86"
+ARG CMAKE_CUDA_ARCHITECTURES=native
+ENV CMAKE_CUDA_ARCHITECTURES=${CMAKE_CUDA_ARCHITECTURES}
 ENV LD_LIBRARY_PATH=/usr/local/cuda/lib64/stubs:${LD_LIBRARY_PATH}
 
 # Create and activate virtual environment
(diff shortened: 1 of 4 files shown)


pwrliang:
@jiayuasu We also need to consider which GPU models are supported in the image. We can pass a list of Compute Capability by referring to this: https://developer.nvidia.com/cuda/gpus

The example above will build an image with compute capability 8.6 and 8.9, which should cover many mainstream desktop and data-center GPU models. We may add two more: 8.0 and 9.0, which cover A100 and H100. These are data-center-grade GPUs for AI training, and I think users are less likely to use these models for spatial data processing tasks. Adding one more compute capability to the list will increase the image build time by approximately 5 minutes, but will not impact performance.

docs/gpu-acceleration.md
```diff
@@ -72,18 +72,23 @@ When `gpu.fallback_to_cpu = true` (default), unsupported predicates fall back to
 When `gpu.fallback_to_cpu = false`, unsupported predicates fail the query.
 
 ## Install from Source with the GPU Feature
+**Build from source**
 
-If you build the Python package from source, enable GPU at build time and configure the CUDA
+If you build the Python package from source, enable GPU at build time, and configure the CUDA
 environment before running `MATURIN_PEP517_ARGS="--features gpu" pip install`.
```
paleolimbot: Does this need to be `MATURIN_PEP517_ARGS="--features='gpu,s2geography,pyo3/extension-module' pip install`? (I've been doing this for s2geography with the extension-module thing in our wheel build but that might not be needed in all cases)

docs/gpu-acceleration.md
```diff
@@ -97,6 +102,13 @@ ctx = sedonadb.connect()
 ctx.sql("SET gpu.enable = true")
```
paleolimbot: This should probably be `ctx.sql("SET gpu.enable = true").execute()`

docs/gpu-acceleration.md
```diff
@@ -148,9 +160,10 @@ ctx.sql("SET datafusion.execution.batch_size = 100000")
 
 ```python
 !nvidia-smi
+!pip install huggingface_hub ipywidgets rasterio pyogrio
```
paleolimbot: Maybe put the `pip install` in a markdown chunk (polite for folks running the notebook so their environment isn't clobbered). This will also clean up the install output from the notebook

docs/gpu-acceleration.md
```diff
@@ -211,16 +285,24 @@ import sedonadb
 
 ctx = sedonadb.connect()
 ctx.options.memory_limit = "unlimited"
+ctx.sql("SET datafusion.execution.batch_size = 100000")
```
paleolimbot:
```suggestion
ctx.sql("SET datafusion.execution.batch_size = 100000").execute()
```

docs/gpu-acceleration.md
paleolimbot:
```suggestion
ctx.sql("""
   CREATE EXTERNAL TABLE trip
   STORED AS PARQUET
   LOCATION 'hf-data/v0.1.0/sf1/trip/'
""").execute()
```

docs/gpu-acceleration.md
```diff
@@ -241,6 +323,7 @@ from tqdm.notebook import tqdm
 from IPython.display import display, HTML
 import ipywidgets as widgets
 
+
```
paleolimbot: This is a slightly confusing way to document this...did you want to add this to `benchmarks/` somewhere (anywhere really...not a lot of structure in that section of the repo yet). Then possibly add examples of the highest value queries here as normal cells).

docker/sedonadb-gpu.dockerfile
```diff
@@ -0,0 +1,92 @@
...
+ENV PATH="/opt/venv/bin:$PATH"
+
+# Pass features to maturin when pip builds the package via PEP517
+ENV MATURIN_PEP517_ARGS="--features s2geography,gpu"
+
+# Install maturin AND JupyterLab inside the virtual environment
+RUN pip3 install maturin jupyterlab pyproj
+
+# Install the project in editable mode (uses pyproject -> maturin under the hood)
+RUN pip3 install -e "python/sedonadb" -vv
```
paleolimbot: You might want to drop the `-e` (you can leave a note that for development this is needed/helpful).

docker/sedonadb-gpu.dockerfile
```diff
@@ -0,0 +1,92 @@
...
+# Clean up the library path so the container uses the real driver at runtime
+ENV LD_LIBRARY_PATH=/usr/local/cuda/lib64
+
+ENV NVIDIA_DRIVER_CAPABILITIES=compute,utility,graphics
+
+# Expose the default JupyterLab port
+EXPOSE 8888
+
+# Launch JupyterLab by default when the container starts
+CMD ["jupyter", "lab", "--ip=0.0.0.0", "--port=8888", "--no-browser", "--allow-root"]
```
paleolimbot: I'm not an expert in security for this kind of thing but is this best practice for spinning up a remote AWS host?

paleolimbot (approved):
@jiayuasu should also take a look here but this looks good to me!

When there are instructions for how to use this on a remote host I'm happy to give this a try, too.

docs/gpu-acceleration.md
pwrliang: I believe executing subsequent SQL statements will make the setting take effect.

docs/gpu-acceleration.md
pwrliang: I agree with you. I have split this huge cell into multiple steps. This `.md` does not serve as a formal benchmark document but for the readers to have a sense of the effectivness of the GPU acceleration.

docs/gpu-acceleration.md
pwrliang: Thanks for the suggestion. I have enabled the suggested features in case users need them.

docker/sedonadb-gpu.dockerfile
pwrliang: I have created a separate user for the notebook, which makes it safer to use. For the network part, I think it's fine. The users can choose whether to expose the port to the public when creating the container. Anyone who wants to use the notebook also needs a token/password to access.

docs/gpu-acceleration.md
paleolimbot: They might not ever be executed if you don't call `.execute()` or print the data frame (and you're not in interactive mode)

docs/gpu-acceleration.md
paleolimbot: The install output seems to still be in the output

pwrliang: I have removed them now

pwrliang:
This PR may still be changing. I'm waiting for @jiayuasu  to finalize the Dockerfile and upload the image to Docker Hub.

jiayuasu:
Thanks for adding this. I think building from source is reasonable for this PR since we do not yet have a released SedonaDB GPU wheel.

I can handle the Docker Hub tag naming when we publish the image under `apache/sedona`, so this PR does not need to settle that part. My current plan is to keep SedonaDB tags separate from the main Sedona Docker image tags: use `sedonadb-latest` for source-built images for now, and later use tags like `sedonadb-0.4.0` once we have a released SedonaDB GPU wheel.

For this PR, I would suggest focusing on the following release-readiness details:

1. Fix the `CMAKE_CUDA_ARCHITECTURES` build arg. The docs show `--build-arg CMAKE_CUDA_ARCHITECTURES="86;89"`, but the Dockerfile currently hard-codes `ENV CMAKE_CUDA_ARCHITECTURES="86"`, so the build arg is ignored. I would add:

```dockerfile
ARG CMAKE_CUDA_ARCHITECTURES=86
ENV CMAKE_CUDA_ARCHITECTURES=${CMAKE_CUDA_ARCHITECTURES}
```

2. Add a Dockerfile-specific `.dockerignore`. Since the Dockerfile does `COPY . /workspace`, the build context can accidentally include `.git`, local build outputs, downloaded datasets, caches, notebook artifacts, etc. The main Sedona Docker image has a Dockerfile-specific ignore file for this reason, and I think we should do the same here before publishing.

3. Please document that this image is source-built and not tied to a released SedonaDB wheel yet. That will avoid confusion with the regular versioned Sedona Docker images.

4. The Jupyter run command should probably include port mapping, e.g. `docker run -it --rm --gpus all -p 8888:8888 sedonadb-gpu`, otherwise users may start the container but not be able to access JupyterLab from the host.

Overall, I am fine with this starting as a source-built image, as long as the docs make that clear and the build is reproducible enough for Docker Hub.

PR #873, opened 2026-05-24
Expose simplified Arrow stream export

### LINKED ISSUE #869: Expose simplified Arrow storage for Python stream export
Some Python callers consume SedonaDB results through the Arrow stream interface instead of collecting a full table with `to_pandas()` / `to_arrow_table()`. This is useful for large results where callers need to process record batches incrementally.

Today, the public stream path preserves Arrow storage types. That is efficient, but it means callers may receive newer storage such as `BinaryView` or `Utf8View`. Callers that need simpler Arrow storage can work around this in SQL with `sd_simplifystorage(...)`, but this requires them to know which columns need simplification.

SedonaDB already has an internal path that can simplify all columns while exporting a stream:

```python
self._impl.to_stream(self._ctx, simplify=True)
```

It would be useful to expose this capability through the Python API so streaming callers can request simplified Arrow storage at the export boundary, without adding column-specific SQL casts or `sd_simplifystorage(...)` calls.

Possible API shape, naming TBD:

```python
reader = df.to_record_batch_reader(simplify=True)
```

or another public streaming method that maps to the existing simplified stream behavior.

Related context: #864 and #868.

camden-lowrance:
## Summary
- add `DataFrame.to_record_batch_reader(simplify=False)` for public Arrow streaming export
- map `simplify=True` to the existing simplified stream path
- cover default storage preservation and `Utf8View`/`BinaryView` simplification

Fixes #869.

## Tests
- `python -m pytest python/sedonadb/tests/test_dataframe.py`

### DIFF
--- a/python/sedonadb/python/sedonadb/dataframe.py
+++ b/python/sedonadb/python/sedonadb/dataframe.py
@@ -28,7 +28,7 @@ from sedonadb.utility import sedona  # noqa: F401
 if TYPE_CHECKING:
     import geopandas
     import pandas
-    import pyarrow
+    import pyarrow as pa
 
 
 class DataFrame:
@@ -457,9 +457,7 @@ class DataFrame:
             requested_schema=requested_schema
         )
 
-    def to_record_batch_reader(
-        self, *, simplify: bool = False
-    ) -> "pyarrow.RecordBatchReader":
+    def to_arrow_reader(self, *, simplify: bool = False) -> "pa.RecordBatchReader":
         """Execute and stream results as a PyArrow RecordBatchReader
 
         Executes the logical plan represented by this object and returns a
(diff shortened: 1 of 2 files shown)


python/sedonadb/python/sedonadb/dataframe.py
```diff
@@ -457,6 +457,39 @@ def __arrow_c_stream__(self, requested_schema: Any = None):
             requested_schema=requested_schema
         )
 
+    def to_record_batch_reader(
+        self, *, simplify: bool = False
+    ) -> "pyarrow.RecordBatchReader":
```
paleolimbot: - Can we name this `to_arrow_reader()` (this is what DuckDB calls the same method)
- Can we expose `.arrow(simplify=...)` either in place of or in addition to this? This is what DuckDB does for this concept (return a stream reader without a pyarrow dependency)

We haven't set up type checking but I think you may need at the start of this file

```python
if TYPE_CHECKING:
    import pyarrow as pa
```

...and then `-> "pa.RecordBatchReader"`. (I'm not an expert on the typing, though)

paleolimbot (commented):
Thank you! A slight comment on the naming but in general this look good!

paleolimbot (approved):
Thank you!

PR #878, opened 2026-05-26
feat(python/sedonadb): add DataFrame.rename

jiayuasu:
Continues Phase P2 of #791 with `DataFrame.rename`. Same shape as `drop` from #871 — small schema op, strings only, varargs-style (well, dict-style) API instead of pandas' `columns=` keyword.

## API

```python
df.rename({"a": "x"})
df.rename({"a": "x", "b": "y"})
```

- **Single `dict[str, str]` arg.** No `columns=` kwarg.
- **Direction is `{old: new}`** — matches Polars and the inner shape of `pandas.rename(columns={...})`. Not Ibis's kwarg-flipped style.
- **Strings only** for both keys and values; no `Expr`.
- **Multi-rename in one call**, applied as a single plan transformation.

## Validation

All Python-side, locked with exact-message tests where the message is the feature being verified:

- Empty dict → `ValueError`.
- Non-dict arg / non-str key or value → `TypeError`.
- Unknown old-name → `KeyError` listing available columns. Forced by DataFusion's `with_column_renamed` being permissive (same trap as `drop_columns`).
- Final-state collisions (`{"a": "z", "b": "z"}` or `{"a": "b"}` when `b` already exists) → `ValueError("duplicate column names")`.

## Swap-pair behavior (worth flagging)

`{"a": "b", "b": "a"}` has a unique final schema `[b, a]` so the Python-side collision check passes. But DataFusion applies renames sequentially, and the intermediate state after `a→b` collides with the original `b`. The error surfaces as `SedonaError` from plan-build with message `Projections require unique expression names...`.

I considered building a rename-graph and detecting cycles Python-side, but the DataFusion error is correct, clear-enough, and trying to reorder Python-side would either reimplement DataFusion's check or miss edge cases. Locked in `test_rename_swap_pair_raises_at_plan_build`. Users wanting a swap route through a temporary name explicitly.

## Tests

13 in `tests/expr/test_dataframe_rename.py`:

- **Positive**: single, multi, column-order preservation.
- **Errors**: empty dict; non-dict; non-str key; non-str value; `columns=` kwarg; unknown old-name (exact `KeyError` message); rename-onto-existing collision; new-name-to-new-name collision; sequential-application swap.

Local: 13 unit + 20 doctests + `ruff format` + `ruff check` all clean.

### DIFF
--- a/python/sedonadb/python/sedonadb/dataframe.py
+++ b/python/sedonadb/python/sedonadb/dataframe.py
@@ -363,6 +363,73 @@ class DataFrame:
 
         return DataFrame(self._ctx, self._impl.drop_columns(list(cols)), self._options)
 
+    def rename(self, mapping: Dict[str, str]) -> "DataFrame":
+        """Rename one or more columns.
+
+        Takes a single dict mapping old names to new names. Multiple
+        renames in one call are applied as a unit; the resulting set of
+        column names must be unique. Unknown old-names raise a
+        `KeyError` listing the available columns (DataFusion's
+        `with_column_renamed` is permissive and silently no-ops on an
+        unknown name, which would hide typos; we validate Python-side).
+
+        Args:
+            mapping: A `dict[str, str]` mapping current column names to
+                new names. Direction is `{old: new}` — matching Polars
+                and the inner shape of pandas' `rename(columns={...})`.
+
+        Examples:
+
+            >>> sd = sedona.db.connect()
+            >>> df = sd.sql("SELECT 1 AS a, 2 AS b")
+            >>> df.rename({"a": "x"}).show()
+            ┌───────┬───────┐
+            │   x   ┆   b   │
+            │ int64 ┆ int64 │
+            ╞═══════╪═══════╡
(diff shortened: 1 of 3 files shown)


paleolimbot (commented):
My preference would be to skip this one for now...it can be replicated with a one liner as a workaround (`df.select(*[col(k).alias(v) for k, v in mapping.items()])`) and there is more important APIs to surface like grouping, aggregation, join, and UDFs.

I also very much dislike the Pandas rename syntax (`rename(a="b")` is easier to type for those of us still typing this stuff)

jiayuasu:
> My preference would be to skip this one for now...it can be replicated with a one liner as a workaround (`df.select(*[col(k).alias(v) for k, v in mapping.items()])`) and there is more important APIs to surface like grouping, aggregation, join, and UDFs.
> 
> I also very much dislike the Pandas rename syntax (`rename(a="b")` is easier to type for those of us still typing this stuff)

Fine by me. Moving on to the next operator then.

PR #881, opened 2026-05-27
fix(rust/sedona-geo): support array distances in ST_Buffer

### LINKED ISSUE #856: rust/sedoa-geo: buffer with non-scalar buffer argument errors with invalid distance
The primary kernel for Buffer rejects non-scalar arguments, which we need to fix:

https://github.com/apache/sedona-db/blob/4daf73bbde764f300cdf7d52e4610f3e91232db4/rust/sedona-geo/src/st_buffer.rs#L71-L80

camden-lowrance:
## Summary
- support scalar or array distance arguments in ST_Buffer through the shared scalar/array argument pattern
- return null when either geometry or distance is null
- add regression coverage for array distances

## Tests
- cargo fmt --check
- cargo test -p sedona-geo st_buffer

Fixes #856

### DIFF
--- a/rust/sedona-geo/src/st_buffer.rs
+++ b/rust/sedona-geo/src/st_buffer.rs
@@ -19,7 +19,7 @@ use std::sync::Arc;
 
 use arrow_array::builder::BinaryBuilder;
 use arrow_schema::DataType;
-use datafusion_common::{error::Result, exec_err, DataFusionError};
+use datafusion_common::{cast::as_float64_array, error::Result, DataFusionError};
 use datafusion_expr::ColumnarValue;
 use geo::algorithm::buffer::{Buffer, BufferStyle};
 use geo_types::Polygon;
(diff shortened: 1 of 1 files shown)


rust/sedona-geo/src/st_buffer.rs
```diff
@@ -65,36 +65,52 @@ impl SedonaScalarKernel for STBuffer {
...
-                    invoke_scalar(&wkb, params, &mut builder)?;
-                    builder.append_value([]);
-                }
-                _ => builder.append_null(),
+
+        let arg1 = args[1].cast_to(&DataType::Float64, None)?;
+        match &arg1 {
+            // Extract the constant scalar value before looping over the input geometries
+            ColumnarValue::Scalar(scalar_arg) => {
+                let params = if scalar_arg.is_null() {
```
paleolimbot: We should document this better, but the way we've converged to do this for other functions (where the other argument is not geometry) is to do

```rust
let arg1 = args[1].cast_to(&DataType::Float64, None)?.to_array(executor.num_iterations())?;
let mut arg1_iter = arg1.iter();
```

...and then inside the loop

```rust
executor.execute_wkb_void(|maybe_wkb| {
    
    match (maybe_wkb, arg1_iter.next().unwrap()) {
        // ...
    }
}
```

...which unifies the scalar and array path. We've never been able to measure a performance regression as a result of this but I'm sure for some functions it matters.

paleolimbot (commented):

rust/sedona-geo/src/st_buffer.rs
camden-lowrance: Updated to use that unified scalar/array pattern. Would it be useful for me to open a follow-up issue to document or factor this pattern?

paleolimbot (approved):
Thank you!

> Would it be useful for me to open a follow-up issue to document or factor this pattern?

Yes! We've tried a number of patterns and it would be good to have a document in the contributors guide (or linked to from the contributor guide) that specifically addresses the style guide for adding scalar functions.

PR #882, opened 2026-05-27
feat(python/sedonadb): add aggregate expression methods on Expr

jiayuasu:
First step toward grouping/aggregation, per the reprioritization recorded on #791 (aggregation/join/UDFs ahead of the remaining small schema ops).

## What's new

Aggregate builder methods on `Expr`:

```python
col("x").sum()      # Expr(sum(x))
col("x").count()    # Expr(count(x))
col("x").mean()     # Expr(avg(x))
col("x").min()      # Expr(min(x))
col("x").max()      # Expr(max(x))
```

- **Methods on `Expr`** (matching Polars / PySpark) rather than free functions — consistent with the operator-method pattern already on `Expr`.
- **`mean`** maps to DataFusion's `avg` aggregate but is named for the pandas/Polars vocabulary.
- Each wraps the corresponding helper from `datafusion::functions_aggregate::expr_fn`.

## Scope

This PR ships **only the expression builders**. The resulting `Expr` is an aggregate AST node — valid only inside an aggregation context (`DataFrame.agg(...)` / `group_by().agg()`), which come in follow-up PRs. Shipping the builders in isolation mirrors how the original `Expr` foundation (#807) landed before any DataFrame integration existed; tests are repr-level.

Planned follow-ups:
1. `df.agg(*exprs)` — global (ungrouped) aggregation.
2. `df.group_by(*keys).agg(*exprs)` — the `GroupedDataFrame` layer.

## Test plan

6 tests in `tests/expr/test_expression.py`:
- Parametrized exact-`repr()` check for all five aggregates.
- Aggregate over a compound expression (`(col("x") + col("y")).sum()` → `Expr(sum(x + y))`).

Local: 54 expression tests + 15 doctests + `ruff format` + `ruff check` all clean.

### DIFF
--- a/python/sedonadb/python/sedonadb/expr/expression.py
+++ b/python/sedonadb/python/sedonadb/expr/expression.py
@@ -157,6 +157,69 @@ class Expr:
         """
         return Expr(self._impl.negate())
 
+    # Aggregate expressions ------------------------------------------------
+    #
+    # These build aggregate nodes (SUM, COUNT, AVG, MIN, MAX). They are
+    # meaningful only inside an aggregation context — `DataFrame.agg(...)`
+    # or `DataFrame.group_by(...).agg(...)` — which land in follow-up PRs.
+    # As standalone expressions they are valid AST nodes but cannot be
+    # evaluated by `select` / `filter`.
+
+    def sum(self) -> "Expr":
+        """Aggregate this expression with SUM.
+
+        Examples:
+
+            >>> from sedonadb.expr import col
+            >>> col("x").sum()
+            Expr(sum(x))
+        """
+        return Expr(self._impl.sum())
+
+    def count(self) -> "Expr":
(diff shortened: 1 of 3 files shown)


jiayuasu:
Closing in favor of the approach in #885 — exposing scalar and aggregate UDFs through a registry-walking `sd.funcs.<name>(args)` dispatch rather than hand-rolling per-function methods on `Expr`. That mechanism:

- avoids growing the Rust bindings by one method per aggregate,
- automatically surfaces plugin and Python-registered UDFs that a hard-coded list would miss,
- uses one path for scalar (spatial `st_*`) and aggregate functions instead of two parallel mechanisms.

Once #885 lands I'll build `DataFrame.agg(...)` and `DataFrame.group_by(...).agg(...)` on top, with call sites like `sd.funcs.sum(col("x"))`. The repr-level tests from this PR will move to that follow-up.

PR #883, opened 2026-05-27
feat(docs): Add scalar function style guide

camden-lowrance:
## Summary
- add a focused scalar SQL function implementation-conventions section to the contributor guide
- document the preferred scalar-or-array argument pattern using executor helpers
- call out expected ScalarUdfTester coverage for scalar, array, and null handling

## Verification
- git diff --check


paleolimbot (approved):
Thank you!

PR #884, opened 2026-05-27
fix(rust/sedona-spatial-join): Improve spatial join evaluated batch memory accounting

camden-lowrance:
## Summary
- avoid double-counting geometry Arrow buffers when evaluated spatial join geometry reuses an input batch column
- split evaluated geometry memory accounting into storage size plus evaluation overhead
- add regression tests for reused-column and computed-geometry cases

## Verification
- cargo test -p sedona-spatial-join evaluated_batch::tests --lib
- cargo test -p sedona-spatial-join --lib *(302 passed; 1 ignored)*

Closes #745

### DIFF
--- a/rust/sedona-spatial-join/src/evaluated_batch.rs
+++ b/rust/sedona-spatial-join/src/evaluated_batch.rs
@@ -36,13 +36,23 @@ pub struct EvaluatedBatch {
 
 impl EvaluatedBatch {
     pub fn in_mem_size(&self) -> Result<usize> {
-        // NOTE: sometimes `geom_array` will reuse the memory of `batch`, especially when
-        // the expression for evaluating the geometry is a simple column reference. In this case,
-        // the in_mem_size will be overestimated. It is a conservative estimation so there's no risk
-        // of running out of memory because of underestimation.
         let record_batch_size = get_record_batch_memory_size(&self.batch)?;
-        let geom_array_size = self.geom_array.in_mem_size()?;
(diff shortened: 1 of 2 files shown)


rust/sedona-spatial-join/src/evaluated_batch.rs
```diff
@@ -56,3 +66,77 @@ impl EvaluatedBatch {
...
+    use std::sync::Arc;
+
+    fn point_wkb(x: f64, y: f64) -> Vec<u8> {
+        let mut out = Vec::with_capacity(21);
+        out.push(1);
+        out.extend_from_slice(&1_u32.to_le_bytes());
+        out.extend_from_slice(&x.to_le_bytes());
+        out.extend_from_slice(&y.to_le_bytes());
+        out
+    }
```
paleolimbot: Can we use one of the existing helpers we have for this? Maybe

https://github.com/apache/sedona-db/blob/6fe541cade2fcd7132feffb17091f83ac1411777/rust/sedona-testing/src/create.rs#L147-L160

paleolimbot (approved):
Thank you! One optional nit:

rust/sedona-spatial-join/src/evaluated_batch.rs
camden-lowrance: I used `create_array`, let me know if you prefer `make_wkb`

paleolimbot (approved):
Thank you!

PR #886, opened 2026-05-28
feat(rust): Lazy Raster Loading Support for Zarr and GDAL

james-willis:
## Summary 
This PR adds support for OutDB raster loading via the `RS_EnsureLoaded` UDF. When a function that needs the bytes of a raster available is called, an optimizer rule injects RS_EnsureLoaded into the expression tree. This _should also leverage Common Subexpression Elimination from Datafusion to avoid repeated loads. 

This PR supercedes #849. This is the new approach based on the review comments I received from Dewey on that PR.

## Copy reduction
Returning `arrow_buffer::Buffer` (not `Vec<u8>` or `Bytes`) and  constructing the output `BinaryViewArray` directly from collected `Buffer`s — bypassing `BinaryViewBuilder`'s internal block copy — gives **one** copy total (storage → loader-internal `Buffer`) on the byte path, down from three with a naive `Vec<u8>` interface. 

## Zarr Registration
I punted on this but we will need it in the near future. The standard datafusion FFI doesnt help us here because these OutDBLoaders are not part of the datafusion interface. the Agent presented two solutions:

  A. Build a C-ABI wrapper for AsyncByteLoader following the datafusion-ffi pattern (#[repr(C)] struct, extern "C" function pointers, abi_stable types). ~200-400 LoC of FFI scaffolding + safety review. Future-proof but heavyweight, and we'd own the ABI compatibility story going forward.

  B. PyCapsule with raw Arc<dyn AsyncByteLoader> transit. Non-standard but works when both wheels link the same sedona-raster version. PyArrow has analogous patterns (PyCapsule carrying ArrowSchema* etc., where the C struct happens to be standardized — ours wouldn't be). Document the version-coupling constraint.

## Merge coordination (with #888)

#888 migrates `sedona-raster-zarr` to all-async and establishes the `open_storage_from_uri(uri, store)` store-resolution contract. Whichever of #886/#888 lands second must reconcile:

- [ ] If #888 has merged: thread an `Arc<dyn ObjectStore>` (from the session `ObjectStoreRegistry`) into `OutDbLoadRequest` so the Zarr loader reads through the host's credentialed store instead of re-deriving creds from env (Dewey review comment on `OutDbLoadRequest`).
- [ ] If #888 has merged: move the Zarr `AsyncByteLoader` onto the all-async storage path and drop the sync `FilesystemStore` / `spawn_blocking` shim in the zarr loader.


rust/sedona-raster-zarr/src/raster_loader.rs
```diff
@@ -0,0 +1,282 @@
...
+                    sedona_internal_datafusion_err!(
+                        "Zarr OutDb band metadata claims {:?} but array {} is {:?}",
+                        expected_dtype,
+                        array_path,
+                        file_dtype
+                    ),
+                )));
+            }
+
+            let bytes = retrieve_chunk_bytes(&array, &anchor.chunk_indices)?;
```
paleolimbot: While this is nicely happening on another thread so as to not block what's happening elsewhere, doing it all at once makes it non-cancelable. If it's possible to break this into chunks of work that are independently `await`able it would make it easier for tokio to schedule them effectively. In the absence of that, perhaps put a limit on either the number of bytes or the number of seconds and loop in such a way that a user won't get a hanging process.

rust/sedona-raster-zarr/src/raster_loader.rs
```diff
@@ -0,0 +1,282 @@
...
+impl AsyncByteLoader for ZarrLoader {
+    async fn load(&self, req: &OutDbLoadRequest<'_>) -> Result<Buffer, ArrowError> {
+        // Take owned copies for the spawn_blocking closure.
+        let uri = req.uri.to_string();
+        let expected_dtype = req.data_type;
+
+        let buffer = tokio::task::spawn_blocking(move || -> Result<Buffer, ArrowError> {
+            let anchor = parse_chunk_anchor(&uri)?;
+            let fs_path = group_uri_to_filesystem_path(&anchor.store_uri)?;
+            let store = FilesystemStore::new(&fs_path).map_err(|e| {
```
paleolimbot: Does the zarr crate have an async API you can use to avoid a spawn here? I was expecting this to look something like

```rust
let read_tasks = open_zarr(uri).await?;
futures::stream::iter(read_tasks)
    .map(|task| {
        async move {
            task.doit().await
        }
    })
    .boxed()
    .buffered(max_concurrency)
    .try_collect()
    .await?;
```

...but maybe that's what's happening one level higher than this. As long as the data transfer size is bounded (a few MB per chance to evaluate whether this should continue or not) the current approach is OK too.

rust/sedona-raster-gdal/src/outdb_loader.rs
```diff
@@ -0,0 +1,309 @@
...
+                        band_num,
+                        file_dtype
+                    ));
+                }
+
+                // `read_as_bytes` allocates a fresh Vec internally.
+                // `Buffer::from_vec` wraps it without copying.
+                let bytes = band
+                    .read_as_bytes((0, 0), (width, height), (width, height), None)
+                    .map_err(convert_gdal_err)?;
```
paleolimbot: Possibly worth sanity checking the number of bytes that should be loaded since this will be non-cancelable. I forget if there's a way to loop here to be able to insert a check + bail if there's a reason to stop.

rust/sedona-raster-gdal/src/raster_loader.rs
```diff
@@ -0,0 +1,309 @@
...
+                "GDAL OutDb loader requires dim_names=[\"y\", \"x\"]; got {:?}",
+                req.dim_names
+            )));
+        }
+        let height = usize::try_from(req.source_shape[0]).map_err(|_| {
+            ArrowError::InvalidArgumentError(format!(
+                "GDAL OutDb source_shape[0]={} exceeds usize::MAX",
+                req.source_shape[0]
+            ))
+        })?;
```
paleolimbot, line 157: Is this loading the source shape or the target shape? (I think Kristin added a getter for an extract that considers a target shape as long as it is contiguous)

rust/sedona-raster/src/raster_loader.rs
```diff
@@ -0,0 +1,255 @@
...
+    /// Anchor URI from the band's `outdb_uri` column. Bare paths and
+    /// scheme'd URIs both allowed; backend is responsible for parsing.
+    pub uri: &'a str,
+    /// Per-axis names parallel to `source_shape`. Backends use this to
+    /// map their native axis order onto the band's.
+    pub dim_names: &'a [&'a str],
+    /// Raw source shape in `dim_names` order. The loader returns a
+    /// `Buffer` whose length equals `Π source_shape × data_type.byte_size()`
+    /// bytes, encoding pixels in C-order over `dim_names`.
+    pub source_shape: &'a [u64],
```
paleolimbot: Shouldn't this see the view as well? In some cases the IO will be the same (but we can discard a lot of pixels right away); in other cases we can avoid IO because the retile will map to a small subset of a contiguous range. (Apologies if I misunderstood what source_shape is here)

rust/sedona-raster/src/raster_loader.rs
```diff
@@ -0,0 +1,255 @@
...
+/// `bytes::Bytes`) hand them off zero-copy, and so the dispatcher can
+/// build the output `BinaryViewArray` directly from collected Buffers
+/// without an extra copy through a `BinaryViewBuilder` block buffer.
+#[async_trait::async_trait]
+pub trait AsyncByteLoader: Send + Sync {
+    /// Fetch the band's bytes. The returned `Buffer` must contain exactly
+    /// `Π source_shape × data_type.byte_size()` bytes in C-order over
+    /// `dim_names`. Errors propagate to the caller of `RS_EnsureLoaded`.
+    async fn load(&self, req: &OutDbLoadRequest<'_>) -> Result<Buffer, ArrowError>;
+}
```
paleolimbot, line 145: I am not sure exactly what the parameter should be here, but this should possibly also pass an `Arc<dyn ObjectStore>`. The loader might not be able to make use of this (i.e., it might use the URI and hope/pray that the authentication ENV vars specified auth identically) but it would be great to centralize IO around this (we need this for non-raster datasources as well, in particular we need to FFI it to get true auth unification across a plugin boundary).

rust/sedona/src/context.rs
```diff
@@ -225,8 +234,49 @@ impl SedonaContext {
...
         };
 
+        // Register the RS_EnsureLoaded async UDF, bound to this session's
+        // OutDb loader registry. Compiled-in backends (`sedona-raster-gdal`
+        // under the `gdal` feature) and plugin crates (`sedona-raster-zarr`,
+        // …) register their loaders into `out.outdb_registry` either from
+        // this constructor or via `SedonaContext::register_outdb_loader`
+        // after construction; the UDF instance below closes over the same
+        // Arc, so registrations are immediately visible at query time.
+        let ensure_loaded_udf = {
```
paleolimbot: It is a little funny, but you could also put this in the `ConfigOptions`. I put a `CrsProvider` there in a similar capacity. The `ConfigOptions` are useful because they are available to scalar function exec steps (although this is a totally reasonable approach as well if we only ever use this in one function).

paleolimbot (commented):
I didn't get through the analyzer bits today but thought I'd leave the loader part and I'll take a look at the analyzer tomorrow. In general this looks great!

rust/sedona-raster-zarr/src/raster_loader.rs
james-willis: the Cloud ones will. AFAIK file doesnt. I can look for something to avoid forking those two cases but no promises.

rust/sedona-raster-gdal/src/raster_loader.rs
james-willis: source_shape is the name of the field in the raster trait. it represents the shape of:

a. in the indentity view case, the data
b. in the non-identity view case, the data that backs the view.

it is named as such to be clear of the difference between the visible bytes vs the backing bytes.

So in this case `source_shape` is the target shape.

rust/sedona-raster/src/raster_loader.rs
james-willis: I'm not sure. I havent thought about designing the system to allow "pushing" views into loaders. This PR does not support that and I'm not sure when we would be able to take advantage of such a thing.

Generally a cog tile or zarr chunk its retrieved atomically so we can't benefit from "pushing" the view in these cases. so I dont think theres a way to reduce io with such support.

Im not sure when you want to eagerly slice/crop vs when you want to perform that lazily. 

Do you think crops on indb rasters should be eager in general? if so we could probably solve this with operator fusion + a rule in the crop operators impl that always eager crops the indb raster

rust/sedona-raster/src/raster_loader.rs
james-willis: Probably a good idea. GDAL wouldnt use but native loaders (ie zarr) will. 

I think probably this can find a good home in the zarr cloud PR. 

I'll sketch that up tonight or tomorrow in parallel with this PR in case we want to land that first. Some of the changes in that might make this PR simpler to review.

rust/sedona-raster-gdal/src/outdb_loader.rs
james-willis: There's a way to loop: added that so we get cancellability

rust/sedona-raster-zarr/src/raster_loader.rs
james-willis: Ok I did some more research. There is a way to use zarrs with obs store to get a local fs that works asynchronously.

My plan (wrote it down in my notes for that work) is that when I do that work to integrate zarr cloud support I will remove all of this synchronous code.

rust/sedona-raster-zarr/src/raster_loader.rs
james-willis: This will all go away when we migrate to the async/obstore-based implementation when we add cloud support to the zarr crate.

rust/sedona-raster-zarr/src/raster_loader.rs
paleolimbot: Not a big deal, just wondering if there was already an async way to split this up (totally fine if there isn't, although best if you can find a way to do it chunks like the GDAL loader does)

rust/sedona/src/context.rs
james-willis: ill do it. Never know where we might want to load rasters from I suppose. There might be some complex cases where the function needs to internally manage loading (ie conditional loading for some reason?)

rust/sedona-raster-zarr/src/raster_loader.rs
james-willis: rough(!) draft of migrating sedona-zarr to be all async: https://github.com/apache/sedona-db/pull/888

rust/sedona-raster-zarr/src/raster_loader.rs
james-willis: rough(!) draft of migrating sedona-zarr to be all async: https://github.com/apache/sedona-db/pull/888

rust/sedona-raster/src/raster_loader.rs
james-willis: rough(!) draft of migrating sedona-zarr to be all async: https://github.com/apache/sedona-db/pull/888

depending which of these PRs we land first ill think about how to best pipe the obstore into the byte load request.

rust/sedona-raster/src/raster_loader.rs
paleolimbot: Yes, I think that if a loader is asked to load a 3x3 chunk of a 1e6 by 1e6 raster, the loader should return 9 pixels. It would be up to the caller to choose whether retiling or cropping is appropriate (and up to the loader to decide whether or not to cache 1e6 by 1e6 pixels in anticipation of reading another slice of the same chunk).

It's OK not to implement crop pushdown but if the information isn't there it can never happen.

rust/sedona-raster/src/outdb_loader.rs
```diff
@@ -0,0 +1,415 @@
...
+/// public `SedonaContext::register_outdb_loader` API.
+#[derive(Default)]
+pub struct OutDbLoaderRegistry {
+    loaders: HashMap<String, Arc<dyn AsyncByteLoader>>,
+}
+
+impl std::fmt::Debug for OutDbLoaderRegistry {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        // Show registered format keys but hide the opaque loader impls
+        // (which don't carry a meaningful Debug surface).
```
paleolimbot: Probably better to just add `Debug` (`AsyncByteLoader: Send + Sync + Debug`) and `#[derive(Debug)]` here.

rust/sedona/src/context.rs
```diff
@@ -225,8 +234,66 @@ impl SedonaContext {
...
+        // Register the analyzer rule that wraps raster args of
+        // `needs_bytes` UDFs with `RS_EnsureLoaded`. The rule holds an
+        // Arc clone of the same UDF so the wraps it injects route
+        // through the same async dispatcher (and thus the same
+        // per-session registry) as any explicit `RS_EnsureLoaded(...)`
+        // already written by the user.
+        out.ctx
+            .state_ref()
+            .write()
+            .add_analyzer_rule(Arc::new(EnsureLoadedAnalyzerRule::new(ensure_loaded_udf)));
```
paleolimbot: I think you don't need this because logical optimizer rules have access to the `function_registry()`. (If this has to stay an analyzer rule and analyzer rules don't share that context this is a good solution)

rust/sedona/src/context.rs
```diff
@@ -225,8 +234,66 @@ impl SedonaContext {
...
+        // mandatory dep on `sedona`, but libgdal itself is dlopen'd
+        // lazily by `sedona-gdal` (workspace-default-features = false),
+        // so this registration is safe on systems without libgdal —
+        // the loader's `load()` call will surface a clean "libgdal not
+        // found" error when first invoked, but registration and import
+        // succeed regardless.
+        out.register_outdb_loader(
+            sedona_raster_gdal::GDAL_FORMAT,
+            Arc::new(sedona_raster_gdal::GdalLoader::new()),
+        );
```
paleolimbot: If this becomes problematic we could gate it behind the `[gdal]` feature and just set this up automatically in the Python package when creating the context, but I think lazy GDAL is OK to include.

rust/sedona/src/rs_ensure_loaded.rs
```diff
@@ -0,0 +1,681 @@
...
+
+    fn signature(&self) -> &Signature {
+        &self.signature
+    }
+
+    fn return_type(&self, arg_types: &[DataType]) -> Result<DataType> {
+        // Output shape mirrors input — the schema is preserved; only the
+        // `data` column's contents change. We require exactly one Struct
+        // argument shaped like a raster.
+        if arg_types.len() != 1 {
```
paleolimbot: I think you need `return_field()` here (or the extension `"sedona.raster"` from the raster type will get dropped).

rust/sedona-raster-functions/src/rs_ensure_loaded.rs
paleolimbot, line 1: Should this live in the sedona-raster-funcs crate?

rust/sedona/src/rs_ensure_loaded.rs
```diff
@@ -0,0 +1,681 @@
...
+                let data_type = band.data_type();
+                let nodata: Option<Vec<u8>> = band.nodata().map(|b| b.to_vec());
+                let outdb_uri: Option<String> = band.outdb_uri().map(|s| s.to_string());
+                let outdb_format: Option<String> = band.outdb_format().map(|s| s.to_string());
+                // For InDb bands, copy bytes into an owned Buffer.
+                // `Buffer::from_vec` is zero-copy ownership transfer of
+                // the Vec; the per-row clone of `band.data()` itself is
+                // the one InDb copy we accept for sequential simplicity.
+                // Parallel fan-out + zero-copy borrowing of the input
+                // column is a follow-up optimisation.
```
paleolimbot: If it isn't already, this should have a tracking issue + link here because it is a particularly egregious copy that should be fixed before anybody uses this seriously

rust/sedona/src/ensure_loaded_analyzer.rs
```diff
@@ -0,0 +1,477 @@
...
+
+/// Analyzer rule that wraps raster arguments of `needs_bytes` UDFs with
+/// `RS_EnsureLoaded`. Holds an `Arc<ScalarUDF>` clone of the registered
+/// `RS_EnsureLoaded` UDF so wraps it injects reference the same async
+/// dispatcher (and thus the same per-session loader registry) as any
+/// explicit `RS_EnsureLoaded(...)` already written by the user.
+#[derive(Debug)]
+pub struct EnsureLoadedAnalyzerRule {
+    ensure_loaded_udf: Arc<ScalarUDF>,
+}
```
paleolimbot: Unless I'm missing something I think this is probably better as a logical optimizer rule (you can put it in sedona-query-planner with the other logical optimizer rule we have there).

rust/sedona-expr/src/scalar_udf.rs
```diff
@@ -71,6 +71,13 @@ pub struct SedonaScalarUDF {
     signature: Signature,
     kernels: Vec<ScalarKernelRef>,
     aliases: Vec<String>,
+    /// Class-level metadata: does this UDF read raster pixel bytes from
+    /// any of its inputs? The optimiser/analyzer rule for `RS_EnsureLoaded`
+    /// uses this to decide whether to wrap raster arguments. Default
+    /// `false`; set via [`SedonaScalarUDF::with_needs_bytes`] at
+    /// construction time for UDFs whose kernels call
+    /// `BandRef::nd_buffer()` / `BandRef::contiguous_data()`.
+    needs_bytes: bool,
```
paleolimbot: Can you make this more generic (like: `HashMap<String, String>`)? This would need to be added to the FFI in sedona-extension and a single bool is tricky to future proof. Alternatively, just keep a list of the functions that need pixels in the optimizer rule.

rust/sedona/src/ensure_loaded_analyzer.rs
```diff
@@ -0,0 +1,477 @@
...
+        // inside-out: the inner call's raster arg is wrapped first, then
+        // the outer call's (now-wrapped) arg is rewritten if it itself
+        // is a needs_bytes target.
+        plan.transform_up(|p| {
+            // Only single-input nodes (Projection, Filter, Aggregate,
+            // Window) carry expressions whose args are evaluated against
+            // a single input schema. Multi-input nodes (Join) and leaf
+            // nodes (TableScan, EmptyRelation) are passed through —
+            // joins with needs_bytes UDFs in the condition are a niche
+            // case to handle in a follow-up.
```
paleolimbot: If this becomes a logical optimizer rule it will need to be able to run twice on the same plan, which might make the inside out logic need guarding. (Maybe you'd have to check the expression for any RS_EnsureLoaded call and skip transforming it if so).

paleolimbot (commented):
A few notes, but the details here are all great!

rust/sedona-raster/src/outdb_loader.rs
james-willis: done

rust/sedona/src/rs_ensure_loaded.rs
james-willis: done. good catch. added a test as well

rust/sedona/src/rs_ensure_loaded.rs
james-willis: done. added something with a bit wider scope: leverage the BinaryViewArray to give zero copy functionality in the raster band builder APIs & all call sites.

rust/sedona/src/ensure_loaded_analyzer.rs
james-willis: done

rust/sedona-expr/src/scalar_udf.rs
james-willis: done

rust/sedona/src/ensure_loaded_analyzer.rs
james-willis: done

rust/sedona-raster-functions/src/rs_ensure_loaded.rs
james-willis: done

rust/sedona/src/context.rs
james-willis: done

rust/sedona-raster/src/raster_loader.rs
james-willis: I created https://github.com/apache/sedona-db/issues/897 to track this.

I'm still not really sure this kind of feature belongs in the loader and I feel it adds a large amount of complexity to each loader implementation, even if it doesnt support view-pruning.

I think we can do this later if we decide to do it; adding fields and tweaking/shimming existing identity based implementations is a 2 way door.

rust/sedona-raster/src/raster_loader.rs
james-willis: I added a note to the PR descriptions of this and 888 to track this tweak based on the order things land

rust/sedona-expr/src/scalar_udf.rs
```diff
@@ -191,17 +216,50 @@ impl SedonaScalarUDF {
...
+        self.with_metadata(NEEDS_PIXELS_METADATA_KEY, "true")
+    }
+
+    /// Returns whether this UDF reads raster pixel bytes from its inputs
+    /// (i.e. carries [`NEEDS_PIXELS_METADATA_KEY`] = `"true"`).
+    pub fn needs_bytes(&self) -> bool {
+        self.metadata
+            .get(NEEDS_PIXELS_METADATA_KEY)
+            .map(String::as_str)
+            == Some("true")
```
paleolimbot: I don't love having raster function (or any geo-specific) stuff live here. `with_needs_bytes` is sufficiently compact that you can probably just inline it to the functions that need it. The named constant can probably live in sedona-raster-functions and you can duplicate it if that creates weird cross-crate interactions?

rust/sedona/src/context.rs
```diff
@@ -287,6 +345,40 @@ impl SedonaContext {
         Ok(())
     }
 
+    /// Register an async OutDb byte loader under a `format` key.
+    ///
+    /// `format` matches the band-level `outdb_format` column value
+    /// (`"gdal"`, `"zarr"`, …). At query time the `RS_EnsureLoaded` UDF
```
paleolimbot: Should this format be more like `"tif"`, `"zarr"`, etc.? As in, should the byte loader get registered by file format or backend? (I had assumed we'd register by file format). You could also have the `trait` have a `pub fn supported_formats() -> Vec<&str>` and just `register_outdb_loader(loader)`.

rust/sedona/src/context.rs
```diff
@@ -287,6 +345,40 @@ impl SedonaContext {
...
+    /// looks up the loader for each OutDb band's format and dispatches
+    /// the byte fetch through it.
+    ///
+    /// Used by both compiled-in backends (`sedona-raster-gdal::register`
+    /// called from `new_from_context` under `#[cfg(feature = "gdal")]`)
+    /// and out-of-tree plugins (`sedona-raster-zarr::register(&ctx)`
+    /// from user code after construction). Later registrations under the
+    /// same `format` key overwrite earlier ones.
+    pub fn register_outdb_loader(
```
paleolimbot: nit:  Should this be an `array_loader` or `raster_loader`? OutDb is a bit of a historical name.

rust/sedona/src/context.rs
```diff
@@ -649,6 +741,52 @@ mod tests {
...
 
+    #[tokio::test]
+    async fn outdb_registry_has_gdal_at_bootstrap_and_accepts_plugin_registration() {
+        use arrow_buffer::Buffer;
+        use sedona_raster::outdb_loader::{AsyncByteLoader, OutDbLoadRequest};
+
+        let ctx = SedonaContext::new();
+        // GDAL is always registered at bootstrap (compiled-in backend).
+        // Plugin backends like Zarr add themselves later via
+        // `register_outdb_loader`.
```
paleolimbot: nit: I know you've been using the word "plugin" but "extension" or "registered at runtime" is closer to existing language (e.g., sedona-extension).

rust/sedona-raster-zarr/src/source_uri.rs
```diff
@@ -80,10 +76,6 @@ pub fn build_chunk_anchor(store_uri: &str, array_path: &str, chunk_indices: &[u6
 ///
 /// Strict: rejects URIs that don't carry both `array=` and `chunk=`
 /// fragment parameters with valid values.
-///
-/// `#[cfg(test)]`: pairs with [`ChunkAnchor`]; resurrected when the
-/// async byte resolver lands.
-#[cfg(test)]
```
paleolimbot, line 86: 🥳

rust/sedona-raster-zarr/src/source_uri.rs
```diff
@@ -40,12 +40,8 @@
 
 use arrow_schema::ArrowError;
 
-/// Parts of a chunk-anchor URI.
-///
-/// `#[cfg(test)]`: no production consumer yet. The async byte
-/// resolver (separate follow-up) will parse `outdb_uri` values back
-/// into this struct.
-#[cfg(test)]
```
paleolimbot, line 54: 🥳

rust/sedona-raster-zarr/src/raster_loader.rs
```diff
@@ -0,0 +1,282 @@
...
+use zarrs_filesystem::FilesystemStore;
+
+use crate::dtype::zarr_to_band_data_type;
+use crate::loader::retrieve_chunk_bytes;
+use crate::source_uri::{group_uri_to_filesystem_path, parse_chunk_anchor};
+
+/// Format key the loader registers under. Keep in sync with
+/// `outdb_format` values emitted by the Zarr reader's band builder
+/// (see `crate::loader`).
+pub const ZARR_FORMAT: &str = "zarr";
```
paleolimbot, line 53: I see...you sync the loader format with the format emitted by the metadata reader. Maybe worth keeping this more related to the file format than the specific loader...some metadata loaders like STAC are maybe not related to the byte loaders (like gdal).

rust/sedona-raster-gdal/src/outdb_loader.rs
```diff
@@ -0,0 +1,548 @@
...
+        if cancel.load(Ordering::Acquire) {
+            return Err(cancelled_err(y_start, height));
+        }
+        let chunk_h = (height - y_start).min(block_h);
+        let byte_off = y_start.saturating_mul(row_bytes);
+        let byte_end = byte_off.saturating_add(chunk_h.saturating_mul(row_bytes));
+        // Sanity: should always hold given the caller's pre-allocated
+        // output slice; defensive in case of arithmetic surprises.
+        if byte_end > output.len() {
+            return Err(sedona_common::sedona_internal_datafusion_err!(
```
paleolimbot:
```suggestion
            return sedona_internal_err!(
```

rust/sedona-raster-functions/src/rs_ensure_loaded.rs
```diff
@@ -0,0 +1,753 @@
...
+    // which plugins are loaded.
+    let registered: Vec<String> = guard.formats().map(String::from).collect();
+    let registered_msg = if registered.is_empty() {
+        "no OutDb loaders are registered in this session".to_string()
+    } else {
+        format!("registered formats: {}", registered.join(", "))
+    };
+    Err(DataFusionError::Plan(format!(
+        "no OutDb loader registered for format '{format}' — {registered_msg}"
+    )))
```
paleolimbot:
```suggestion
    plan_err!(
        "no OutDb loader registered for format '{format}' — {registered_msg}"
    )
```

rust/sedona-raster-functions/src/rs_ensure_loaded.rs
```diff
@@ -0,0 +1,753 @@
...
+    }
+
+    fn return_type(&self, _arg_types: &[DataType]) -> Result<DataType> {
+        // Never called in practice — `return_field_from_args` below is the
+        // authoritative output-type source and carries the raster
+        // extension metadata that a bare `DataType` would drop. Provided
+        // only to satisfy the trait.
+        Err(sedona_internal_datafusion_err!(
+            "RS_EnsureLoaded::return_type should not be called; return_field_from_args is authoritative"
+        ))
```
paleolimbot:
```suggestion
        sedona_internal_err!("use return_field_from_args()")
```

rust/sedona-raster-functions/src/rs_ensure_loaded.rs
```diff
@@ -0,0 +1,753 @@
...
+        Ok(Arc::clone(field))
+    }
+
+    fn invoke_with_args(&self, _args: ScalarFunctionArgs) -> Result<ColumnarValue> {
+        // DataFusion routes async UDFs through `invoke_async_with_args`
+        // on the AsyncFuncExec node; this sync entry should never be
+        // called for an `AsyncScalarUDF`-wrapped impl.
+        Err(sedona_internal_datafusion_err!(
+            "RS_EnsureLoaded is async; AsyncFuncExec should have dispatched to invoke_async_with_args"
+        ))
```
paleolimbot:
```suggestion
       sedona_internal_err!("use invoke_async_with_args()")
```

paleolimbot: Also a few more `Err(sedona_internal_datafusion_err!(...))`s that can be simplified in this file

rust/sedona-raster-functions/src/rs_ensure_loaded.rs
```diff
@@ -0,0 +1,753 @@
...
+                })?;
+            let got = resolved.len();
+            if got as u64 != expected_bytes {
+                return Err(sedona_internal_datafusion_err!(
+                    "RS_EnsureLoaded: band ({raster_idx},{band_idx}) expected {expected_bytes} \
+                     bytes but loader returned {got}"
+                ));
+            }
+
+            builder.band_data_writer().append_value(resolved.as_slice());
```
paleolimbot, line 454:
```suggestion
            // Follow up: ensure loaded bytes are not copied
            // https://github.com/apache/sedona-db/issues/894.
            builder.band_data_writer().append_value(resolved.as_slice());
```

rust/sedona-raster-functions/src/rs_ensure_loaded.rs
```diff
@@ -0,0 +1,753 @@
...
+        // only to satisfy the trait.
+        Err(sedona_internal_datafusion_err!(
+            "RS_EnsureLoaded::return_type should not be called; return_field_from_args is authoritative"
+        ))
+    }
+
+    fn return_field_from_args(&self, args: ReturnFieldArgs) -> Result<FieldRef> {
+        // Identity on schema: the output raster has the same fields as the
+        // input — only the `data` column's bytes change. Return the input
+        // field verbatim so its `"sedona.raster"` extension metadata
```
paleolimbot, line 190: In https://github.com/apache/sedona-db/pull/831 Kristin found that Async UDFs drop metadata (at the physical expression). The optimizer rule that emplaces this may have to add it back via a dummy udf (or, simpler, we don't strictly need `SedonaType::Raster` to emit an extension name...a normal struct is fine).

rust/sedona-raster-functions/src/rs_ensure_loaded.rs
```diff
@@ -0,0 +1,753 @@
...
+        // on the AsyncFuncExec node; this sync entry should never be
+        // called for an `AsyncScalarUDF`-wrapped impl.
+        Err(sedona_internal_datafusion_err!(
+            "RS_EnsureLoaded is async; AsyncFuncExec should have dispatched to invoke_async_with_args"
+        ))
+    }
+}
+
+#[async_trait]
+impl AsyncScalarUDFImpl for RsEnsureLoaded {
```
paleolimbot, line 222: Should we also suggest a batch size here? (I think that's something you can implement on the async udf)

rust/sedona-raster-functions/src/rs_ensure_loaded.rs
```diff
@@ -0,0 +1,753 @@
...
+use sedona_raster::outdb_loader::{
+    AsyncByteLoader, OutDbLoadRequest, OutDbLoaderConfig, OutDbLoaderRegistry,
+};
+use sedona_raster::traits::RasterRef;
+
+/// Re-exported from `sedona-expr` (its canonical home — see
+/// [`sedona_expr::scalar_udf::RS_ENSURE_LOADED_NAME`]) so existing
+/// `crate::rs_ensure_loaded::RS_ENSURE_LOADED_NAME` references keep
+/// resolving.
+pub use sedona_expr::scalar_udf::RS_ENSURE_LOADED_NAME;
```
paleolimbot: I think you can just use `"rs_ensureloaded"` instead of a named constant

rust/sedona-query-planner/src/ensure_loaded.rs
```diff
@@ -0,0 +1,427 @@
...
+            // qualifiers in a self-join). Skip this node rather than
+            // failing the query — a missed wrap surfaces later as a clear
+            // "raster bytes not loaded" error, not a wrong result.
+            return Ok(Transformed::no(plan));
+        };
+        drop(inputs);
+
+        plan.map_expressions(|e| {
+            e.transform_up(|expr| rewrite_expr_node(expr, &schema, &ensure_loaded_udf))
+        })
```
paleolimbot: Does this work?

```suggestion
        plan.map_expressions(|e| {
            e.transform_up(|expr| rewrite_expr_node(expr, plan.schema(), &ensure_loaded_udf))
        })
```

rust/sedona-query-planner/src/ensure_loaded.rs
```diff
@@ -0,0 +1,427 @@
...
+    if !needs_bytes {
+        return Ok(Transformed::no(expr));
+    }
+
+    // Structurally impossible: we matched `expr` as `Expr::ScalarFunction`
+    // a few lines up. Surface it as an internal error rather than a panic
+    // so a future refactor that breaks the invariant fails the query
+    // cleanly instead of crashing a worker.
+    let Expr::ScalarFunction(ScalarFunction { func, args }) = expr else {
+        return Err(sedona_internal_datafusion_err!(
```
paleolimbot:
```suggestion
        return sedona_internal_err!(
```

paleolimbot (commented):
All just details from me!

The one hiccup is that the AsyncUDFs might be dropping metadata (see comment), in which case you'll have to make a dummy scalar function `struct InsertMetadataUdf { field: FieldRef, inner: ScalarUDF }` that wraps RS_EnsureLoaded with an overridden return_field. 😮‍💨

rust/sedona/src/context.rs
james-willis: I don't have a strong opinion here. part of the reason I used "gdal" is because I don't really know what formats:

1. What formats GDAL supports
1. What formats GDAL will support
2. what formats we [will] support with GDAL

calling it GDAL leaves it such that users can just try to load random things with GDAL and it might even work.

paleolimbot: There's a list of drivers, but I do think you want to separate the production of raster metadata from the resolving of metadata bytes (stac is the best reason for this but there could be other metadata catalogs that make sense). You could use a `Vec<AsyncByteLoader>` instead of hash map and call `byte_loader.supports(format)`, where the gdal loader could return `true` based on some runtime call.

rust/sedona-raster-gdal/src/outdb_loader.rs
james-willis: done

rust/sedona-raster-functions/src/rs_ensure_loaded.rs
james-willis: done

rust/sedona-raster-functions/src/rs_ensure_loaded.rs
james-willis: done all of these.

rust/sedona-query-planner/src/ensure_loaded.rs
james-willis: done

rust/sedona-raster-functions/src/rs_ensure_loaded.rs
james-willis: done

rust/sedona-raster/src/raster_loader.rs
paleolimbot: Definitely OK to not support this yet, but I would like the view to be passed to the loader because we have to define an FFI for this and changing the FFI is a pain. When you have the view helpers the loader will be trivial (`views.resolve(all_of_the_bytes)`), or you can push that on to the caller by letting a byte loader signal that it doesn't support views.

rust/sedona-raster-functions/src/rs_ensure_loaded.rs
james-willis: done

rust/sedona-raster-functions/src/rs_ensure_loaded.rs
james-willis: done

rust/sedona-raster-functions/src/rs_ensure_loaded.rs
james-willis: done

rust/sedona-expr/src/scalar_udf.rs
james-willis: done

rust/sedona-raster-zarr/src/raster_loader.rs
james-willis: I am worried that metadata reader and loader are linked. Imagine two zarr loaders.

zarrFormat1 needs uris to contain the inner chunk identifier because of how the underlying lib works.

zarrFormat2 needs uris to contain the explicit element offsets in each dim because the underlying lib is "chunk transparent". 

In this world loading the raster with zarrFormat2 with a metadata created from zarrFormat1 wont work.

What do you think?

rust/sedona/src/context.rs
james-willis: done

rust/sedona/src/context.rs
james-willis: done

rust/sedona-query-planner/src/ensure_loaded.rs
james-willis: This code is needed to support raster that appear in the filter but not the project of a join.

I added a test to capture this. Tested with your proposed change here and the test failed.

rust/sedona-raster-functions/src/rs_ensure_loaded.rs
james-willis: wrapped in a dummy sync UDF that "rerasters" the output.

rust/sedona-raster/src/raster_loader.rs
james-willis: ok I see. ill add this view support

james-willis: done

rust/sedona-expr/src/scalar_udf.rs
```diff
@@ -71,6 +71,15 @@ pub struct SedonaScalarUDF {
...
     kernels: Vec<ScalarKernelRef>,
     aliases: Vec<String>,
+    /// Class-level, string-keyed metadata describing this UDF to the
+    /// planner. Flags are set via [`SedonaScalarUDF::with_metadata`] and
+    /// read back via [`SedonaScalarUDF::metadata`]; the planner keys off
+    /// well-known entries (e.g. the raster `"needs_pixels"` flag, whose
+    /// key is owned by `sedona-raster-functions`) without this crate
+    /// knowing their meaning. The map shape leaves room for further
+    /// planner-visible flags — and a future cross-cdylib FFI carrying
+    /// them — without a new field per flag.
```
paleolimbot: If there's a way to make LLMs stop referring to future PRs that would be great (Jia's PRs keep doing this too)

```suggestion
    /// knowing their meaning. The map shape leaves room for further
    /// planner-visible flags.
```

rust/sedona-query-planner/src/ensure_loaded_reraster.rs
```diff
@@ -0,0 +1,157 @@
...
+use datafusion_common::Result;
+use datafusion_expr::{
+    ColumnarValue, ReturnFieldArgs, ScalarFunctionArgs, ScalarUDF, ScalarUDFImpl, Signature,
+    Volatility,
+};
+use sedona_common::sedona_internal_err;
+use sedona_schema::datatypes::SedonaType;
+
+/// Registry/UDF name of the re-stamp helper.
+pub const RERASTER_NAME: &str = "__rs_reraster";
```
paleolimbot:
```suggestion
/// Registry/UDF name of the re-stamp helper.
pub const RERASTER_NAME: &str = "sd_reraster";
```

(...we have a few other internal functions that start with `sd_`)

rust/sedona-raster/src/raster_loader.rs
```diff
@@ -0,0 +1,459 @@
...
+/// without an extra copy through a `BinaryViewBuilder` block buffer.
+#[async_trait::async_trait]
+pub trait AsyncByteLoader: Send + Sync + std::fmt::Debug {
+    /// Fetch the band's bytes, returning a [`RasterLoadResult`] that
+    /// describes the layout produced. A loader that doesn't prune by the
+    /// view returns [`RasterLoadResult::unresolved`] (full source, length
+    /// `Π source_shape × data_type.byte_size()` in C-order over
+    /// `dim_names`). Errors propagate to the caller of `RS_EnsureLoaded`.
+    async fn load(&self, req: &RasterLoadRequest<'_>) -> Result<RasterLoadResult, ArrowError>;
+}
```
paleolimbot, line 145: I would add this here, so that the definition of the loader and which formats it supports are defined in the same place.

```suggestion
#[async_trait::async_trait]
pub trait AsyncByteLoader: Send + Sync + std::fmt::Debug {
    /// The name of the loader for identifying it in diagnostics/errors (e.g. "gdal")
    fn name(&self) -> Vec<String>;
    
    /// A list of formats supported by this loader. Formats are usually named by their
    /// extension (e.g., tif, zarr).
    fn supported_formats(&self) -> &[String];
    
    /// Fetch the band's bytes, returning a [`RasterLoadResult`] that
    /// describes the layout produced. A loader that doesn't prune by the
    /// view returns [`RasterLoadResult::unresolved`] (full source, length
    /// `Π source_shape × data_type.byte_size()` in C-order over
    /// `dim_names`). Errors propagate to the caller of `RS_EnsureLoaded`.
    async fn load(&self, req: &RasterLoadRequest<'_>) -> Result<RasterLoadResult, ArrowError>;
}
```

rust/sedona-raster/src/raster_loader.rs
```diff
@@ -0,0 +1,459 @@
...
+/// Process-side registry mapping `outdb_format` keys to loader instances.
+///
+/// One registry instance per `SedonaContext`. The owning context wraps it
+/// in `Arc<RwLock<…>>` so extension crates (`sedona-raster-zarr`, future COG /
+/// Icechunk / …) can register their loaders post-construction via a
+/// public `SedonaContext::register_raster_loader` API.
+#[derive(Debug, Default)]
+pub struct RasterLoaderRegistry {
+    loaders: HashMap<String, Arc<dyn AsyncByteLoader>>,
+}
```
paleolimbot, line 162: This can then be a `Vec<dyn AsyncByteLoader>`. You can iterate in reverse to find the most recently registed loader that supports a given format.

rust/sedona-raster/src/raster_loader.rs
```diff
@@ -0,0 +1,459 @@
...
+    /// registration (a process with runtime-registered backends may
+    /// legitimately swap implementations during setup).
+    pub fn register(&mut self, format: impl Into<String>, loader: Arc<dyn AsyncByteLoader>) {
+        self.loaders.insert(format.into(), loader);
+    }
+
+    /// Look up a loader by format key. Returns `None` for keys with no
+    /// registered backend. `RS_EnsureLoaded` surfaces the `None` case as a
+    /// query-time error that names the missing format and points users at
+    /// the install/register step.
```
paleolimbot: It would probably make sense to have a `fn get_or_error(&self)` that populates that here (gives this object more freedom to change how it stores formats/resolves loaders without changing other stuff)

paleolimbot (commented):
This looks good...very close!

I do think the definition of the "format" needs to be the same as the file ("tif"), not the same as the library ("gdal"). This does mean that future loaders that overlap need to use a compatible URI but I think this is fine. If there's ever a metadata reader that would prefer not to play nicely with other loaders, it can return a more specific format string (e.g., "gdal_tif").

rust/sedona-expr/src/scalar_udf.rs
james-willis: made this change. 

if we want to have a better chance of LLM changes taking certain shapes, probably we can add some stuff to the Agents.md file

rust/sedona-query-planner/src/ensure_loaded_reraster.rs
james-willis: done

rust/sedona-raster/src/raster_loader.rs
james-willis: I went with `fn supports_format(&self, format: Option<&str>) -> bool`. This gives more flexibility to handle cases like when format is `None` and avoids constructing a string array unnecessarily.

rust/sedona-raster/src/raster_loader.rs
james-willis: done

rust/sedona-raster/src/raster_loader.rs
james-willis: done

rust/sedona-raster-zarr/src/raster_loader.rs
james-willis: Implemented Dewey's revisions to the loader interface that were described elsewhere.

allows the user to register loaders in certain orders to get precedence that they want.

rust/sedona/src/context.rs
james-willis: did ~this.

discussed elsewhere.

paleolimbot (approved):
Thank you!

The only two design things I see is (1) the AsyncRasterLoader should probably accept a `Vec<Request>` (which amortizes the cost of the implementation spawning a thread or task over some number of elements) and (2) the load request should (probably) provide an `Option<Arc<dyn ObjectStore>>`. I'll make those changes at the FFI layer in a follow up so they can be backported into the implementation if needed.

PR #885, opened 2026-05-27
feat(python/sedonadb): Expose scalar and aggregate udfs from context registry

paleolimbot:
This PR adds accessors to `sd.funcs` that allow extracting any scalar or aggregate function and call it to generate an expression, which had been missing from the expression API.

```python
import sedona.db

sd = sedona.db.connect()
f = sd.funcs

t = sd.funcs.table.sd_random_geometry(num_rows=100)
t.filter(
    f.st_dwithin(t.geometry, f.st_point(50, 50), 10)
).show()
# ┌───────┬────────────────────┬──────────────────────────────────────────────┐
# │   id  ┆        dist        ┆                   geometry                   │
# │ int32 ┆       float64      ┆                   geometry                   │
# ╞═══════╪════════════════════╪══════════════════════════════════════════════╡
# │    54 ┆  59.62435027266706 ┆ POINT(44.832630149284846 42.9320865879236)   │
# ├╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌┤
# │    62 ┆ 43.885794593815966 ┆ POINT(54.858587734052435 51.438644994339946) │
# ├╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌┤
# │    79 ┆  95.62798350158674 ┆ POINT(54.57939637408915 46.18785999007198)   │
# ├╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌┤
# │    89 ┆  26.04589019859811 ┆ POINT(46.642374405687505 43.235146303817395) │
# └───────┴────────────────────┴──────────────────────────────────────────────┘

[...]

PR #887, opened 2026-05-29
feat(python/sedonadb): add DataFrame.agg for global aggregation

jiayuasu:
First DataFrame consumer of the function-registry dispatch landed in #885. Adds global (ungrouped) aggregation; grouped aggregation (`DataFrame.group_by(*keys).agg(*aggs)`) is the next small PR, sharing the same Rust binding.

## API

```python
sd = sedonadb.connect()
df = sd.create_data_frame(pd.DataFrame({"x": [1, 2, 3, 4]}))

df.agg(sd.funcs.sum(col("x")).alias("total"))
df.agg(
    sd.funcs.sum(col("x")).alias("sum_x"),
    sd.funcs.count(col("y")).alias("n"),
    sd.funcs.min(col("x")).alias("lo"),
    sd.funcs.max(col("x")).alias("hi"),
)
```

- Varargs of aggregate `Expr` values, built via `sd.funcs.<name>(args)` from #885.
- Strings are not auto-promoted — a bare column isn't an aggregate.
- Empty `df.agg()` → `ValueError`; non-Expr arg → `TypeError`.
- Returns a one-row DataFrame.

## Why this is so small

The function-registry dispatch in #885 means `sd.funcs.sum`, `sd.funcs.count`, `sd.funcs.min`, `sd.funcs.max`, `sd.funcs.avg` — and every other built-in / plugin / Python-registered aggregate — are already callable. This PR doesn't need any per-aggregate plumbing on either the Rust or Python side. One Rust binding, one Python method, a test file.

## Implementation

| File | Change |
|---|---|
| `python/sedonadb/src/dataframe.rs` | New `InternalDataFrame::aggregate(group_exprs, agg_exprs)`. Generic wrapper over DataFusion's `DataFrame::aggregate`. Shared with the upcoming `group_by` PR — that path passes a populated `group_exprs`. |
| `python/sedonadb/python/sedonadb/dataframe.py` | `DataFrame.agg(*exprs)`. Calls the Rust binding with an empty `group_exprs`. |

## Test plan

9 tests in `tests/expr/test_dataframe_agg.py`:

- **Positive**: single `sum`; single `count`; paired `min`/`max`; `avg` over a compound expression `col("x") + col("y")`; four aggregates yielding a one-row four-column result.
- **Errors**: empty `agg()` → `ValueError`; non-Expr arg → `TypeError`.
- **Plan composition**: chained `filter().agg()` produces the right result.

All assertions use `pd.testing.assert_frame_equal` for outputs.

Local: 9 unit + 22 doctests + `ruff format` + `ruff check` all clean.

### DIFF
--- a/python/sedonadb/python/sedonadb/dataframe.py
+++ b/python/sedonadb/python/sedonadb/dataframe.py
@@ -539,6 +539,63 @@ class DataFrame:
 
         return DataFrame(self._ctx, self._impl.drop_columns(list(cols)), self._options)
 
+    def agg(self, *exprs: Expr, **named_exprs: Expr) -> "DataFrame":
+        """Aggregate the entire DataFrame to a single row.
+
+        Aggregate expressions can be passed positionally or as keyword
+        arguments. With keyword arguments the keyword becomes the
+        output column name — `df.agg(total=sd.funcs.sum(sd.col("x")))`
+        is shorthand for
+        `df.agg(sd.funcs.sum(sd.col("x")).alias("total"))`. The two
+        forms can be mixed in a single call.
+
+        Args:
+            *exprs: Positional aggregate expressions.
+            **named_exprs: Keyword aggregate expressions; each keyword
+                is applied as the output alias of the corresponding
+                expression.
+
+        Examples:
+
+            >>> sd = sedona.db.connect()
+            >>> df = sd.sql("SELECT * FROM (VALUES (1), (2), (3), (4)) AS t(x)")
+            >>> df.agg(sd.funcs.sum(sd.col("x")).alias("total")).show()
+            ┌───────┐
+            │ total │
+            │ int64 │
+            ╞═══════╡
+            │    10 │
+            └───────┘
+            >>> df.agg(total=sd.funcs.sum(sd.col("x"))).show()
+            ┌───────┐
+            │ total │
+            │ int64 │
+            ╞═══════╡
+            │    10 │
+            └───────┘
+        """
+        if not exprs and not named_exprs:
+            raise ValueError("agg() requires at least one aggregate expression")
+
+        for e in exprs:
+            if not isinstance(e, Expr):
+                raise TypeError(f"agg() expects Expr arguments, got {type(e).__name__}")
+
+        all_exprs: List[Expr] = list(exprs)
+        for name, e in named_exprs.items():
+            if not isinstance(e, Expr):
+                raise TypeError(
+                    f"agg() expects Expr keyword values, got {type(e).__name__} "
+                    f"for keyword {name!r}"
+                )
+            all_exprs.append(e.alias(name))
+
+        return DataFrame(
+            self._ctx,
+            self._impl.aggregate([], [e._impl for e in all_exprs]),
+            self._options,
+        )
+
     def limit(self, n: Optional[int], /, *, offset: int = 0) -> "DataFrame":
         """Limit result to n rows starting at offset
 
--- a/python/sedonadb/src/dataframe.rs
+++ b/python/sedonadb/src/dataframe.rs
@@ -235,6 +235,28 @@ impl InternalDataFrame {
         Ok(InternalDataFrame::new(inner, self.runtime.clone()))
     }
 
+    /// Aggregate the rows of the DataFrame, optionally partitioned by
+    /// `group_exprs`. Both inputs are `Vec<PyExpr>` so the same Rust
+    /// method serves global aggregation (`group_exprs` empty, called
+    /// from `DataFrame.agg`) and grouped aggregation.
+    ///
+    /// The Python side guarantees `agg_exprs` is non-empty and that
+    /// every entry is an `Expr` (vs. a string or other type). It does
+    /// not verify that each entry is an aggregate-shaped expression —
+    /// e.g. `col("x")` would pass the Python `isinstance` check but is
+    /// not a valid aggregate. DataFusion's plan-build catches that case
+    /// with a clear error, so we don't reimplement the check here.
+    fn aggregate(
+        &self,
+        group_exprs: Vec<PyExpr>,
+        agg_exprs: Vec<PyExpr>,
+    ) -> Result<InternalDataFrame, PySedonaError> {
+        let group_exprs: Vec<Expr> = group_exprs.into_iter().map(|e| e.inner).collect();
+        let agg_exprs: Vec<Expr> = agg_exprs.into_iter().map(|e| e.inner).collect();
+        let inner = self.inner.clone().aggregate(group_exprs, agg_exprs)?;
+        Ok(InternalDataFrame::new(inner, self.runtime.clone()))
+    }
+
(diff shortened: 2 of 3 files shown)


python/sedonadb/tests/expr/test_dataframe_agg.py
```diff
@@ -0,0 +1,96 @@
...
+        df.agg()
+
+
+def test_agg_non_expr_arg_raises(con):
+    df = con.create_data_frame(pd.DataFrame({"x": [1, 2, 3]}))
+    with pytest.raises(TypeError, match="agg\\(\\) expects Expr arguments"):
+        df.agg("x")
+
+
+def test_agg_chains_with_select(con):
```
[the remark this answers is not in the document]
jiayuasu: Renamed to `test_agg_chains_with_filter` in 6fdc21c7d.

python/sedonadb/src/dataframe.rs
```diff
@@ -235,6 +235,28 @@ impl InternalDataFrame {
...
+    /// `group_exprs`. Both inputs are `Vec<PyExpr>` so the same Rust
+    /// method serves global aggregation (`group_exprs` empty, called
+    /// from `DataFrame.agg`) and grouped aggregation (called from
+    /// `DataFrame.group_by(...).agg(...)` once that lands).
+    ///
+    /// The Python side guarantees `agg_exprs` is non-empty. Argument
+    /// shape validation (every entry being an aggregate-shaped `Expr`)
+    /// happens Python-side. DataFusion's plan-build raises a clear
+    /// error if a non-aggregate Expr is passed in `agg_exprs`, so we
+    /// don't try to enforce that here.
```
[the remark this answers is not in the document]
jiayuasu: Rewrote the comment in 6fdc21c7d to drop the contradictory "Python-side validation" sentence — the Python wrapper only checks `isinstance(e, Expr)`, not aggregate-shapedness, and DataFusion's plan-build catches the rest.

python/sedonadb/python/sedonadb/dataframe.py
```diff
@@ -539,6 +539,47 @@ def drop(self, *cols: str) -> "DataFrame":
...
+        """Aggregate the entire DataFrame to a single row.
+
+        Each argument is an aggregate `Expr` — usually built via the
+        function registry, e.g. `sd.funcs.sum(col("x"))` or
+        `sd.funcs.count(col("y"))`. Aliases set the output column name
+        (`sd.funcs.sum(col("x")).alias("total")`).
+
+        For grouped aggregation use `DataFrame.group_by(...).agg(...)`
+        (lands in a follow-up PR).
+
```
paleolimbot:
```suggestion
```

python/sedonadb/python/sedonadb/dataframe.py
```diff
@@ -539,6 +539,47 @@ def drop(self, *cols: str) -> "DataFrame":
...
+        Args:
+            *exprs: One or more aggregate expressions. At least one is
+                required.
+
+        Examples:
+
+            >>> from sedonadb.expr import col
+            >>> sd = sedona.db.connect()
+            >>> df = sd.sql("SELECT * FROM (VALUES (1), (2), (3), (4)) AS t(x)")
+            >>> df.agg(sd.funcs.sum(col("x")).alias("total")).show()
```
paleolimbot:
```suggestion
            >>> sd = sedona.db.connect()
            >>> df = sd.sql("SELECT * FROM (VALUES (1), (2), (3), (4)) AS t(x)")
            >>> df.agg(sd.funcs.sum(sd.col("x")).alias("total")).show()
```

python/sedonadb/python/sedonadb/dataframe.py
```diff
@@ -539,6 +539,47 @@ def drop(self, *cols: str) -> "DataFrame":
 
         return DataFrame(self._ctx, self._impl.drop_columns(list(cols)), self._options)
 
+    def agg(self, *exprs: Expr) -> "DataFrame":
+        """Aggregate the entire DataFrame to a single row.
```
paleolimbot: - Can we call this `aggregate()`? (Ibis, DuckDB)
- Can we expose `**kwargs` like is done in `select()`? `df.aggregate(x_sum=df.x.sum())` is much more compact than `df.aggregate(df.x.sum().alias("x_sum"))` and is allowed by Ibis.

python/sedonadb/src/dataframe.rs
```diff
@@ -235,6 +235,29 @@ impl InternalDataFrame {
         Ok(InternalDataFrame::new(inner, self.runtime.clone()))
     }
 
+    /// Aggregate the rows of the DataFrame, optionally partitioned by
+    /// `group_exprs`. Both inputs are `Vec<PyExpr>` so the same Rust
+    /// method serves global aggregation (`group_exprs` empty, called
+    /// from `DataFrame.agg`) and grouped aggregation (called from
+    /// `DataFrame.group_by(...).agg(...)` once that lands).
```
paleolimbot:
```suggestion
    /// from `DataFrame.agg`) and grouped aggregation.
```

paleolimbot (commented):
Exciting...thank you!

Mostly nits...I'm hoping we can rename to `aggregate` (which is what DuckDB and Ibis call this).

python/sedonadb/python/sedonadb/dataframe.py
jiayuasu: PySpark, Pandas and Polars all use `agg`. I'd like to keep it that way.

jiayuasu: kwargs added in 2196d03d2 — `df.agg(total=sd.funcs.sum(sd.col("x")))` now desugars to `…sum(sd.col("x")).alias("total")`, and positional + named can mix. Three new tests cover the kwarg path, mixed positional/kwarg, and the non-Expr kwarg value rejection.

python/sedonadb/python/sedonadb/dataframe.py
jiayuasu: Dropped in 2196d03d2 — no more forward-reference to grouped agg in the docstring.

python/sedonadb/python/sedonadb/dataframe.py
jiayuasu: Switched to `sd.col("x")` in 2196d03d2 — the `from sedonadb.expr import col` line is gone from the doctest.

python/sedonadb/src/dataframe.rs
jiayuasu: Simplified in 2196d03d2.

paleolimbot (approved):
> PySpark, Pandas and Polars all use agg. I'd like to keep it that way

At this point our API has little in common with PySpark, Pandas, and Polars but we can always alias it later if LLMs or their humans get confused. Either works for me 🙂 

(Later we can add `sedona.db.pyspark.connect()` if that's a compatibility layer that is important)

PR #888, opened 2026-05-29
feat(rust/sedona-raster-zarr): cloud storage backends (s3, gcs, azure, http) via object_store

james-willis:
## Summary
Adds cloud storage backends to `sedona-raster-zarr`. `ZarrChunkReader` now accepts `file://`, bare paths, `s3://`, `gs://` / `gcs://`, `az://` / `abfs://` / `abfss://`, and `http://` / `https://`.

```python
import sedonadb
import sedonadb_zarr

con = sedonadb.connect()
con.read_format(
    sedonadb_zarr.ZarrFormatSpec(),
    "s3://its-live-data/datacubes/v2/N40W120/ITS_LIVE_vel_EPSG32610_G0120_X250000_Y5450000.zarr",
).show()
```
## All Async
This package has been migrated to use all async for record loading and byte retrieval. This allows a single approach (cloud was always going to be async, but file system was originally sync) without shimming sync fs code into the async RS_EnsureLoaded udf and async cloud access code into the sync record loading machinery.

## Version pinning

`zarrs_object_store` is pinned to `0.5` in the workspace. `0.6` (the latest) depends on `object_store 0.13`, which is semver-incompatible with the `object_store 0.12.x` the rest of the workspace uses through DataFusion 52.

## Merge coordination (with #886)

#886 adds the OutDb byte-loading infrastructure (`AsyncByteLoader`, `OutDbLoadRequest`, the Zarr `ZarrLoader`). Whichever of #886/#888 lands second must reconcile:

- [ ] If #886 has merged: align the `sedona-raster-zarr` `ZarrLoader` onto this PR's all-async storage path (it lands sync in #886).
- [ ] If #886 has merged: wire the session `Arc<dyn ObjectStore>` carried on `OutDbLoadRequest` into the Zarr loader's reads, using `open_storage_from_uri(uri, store)`.

### DIFF
--- a/python/sedonadb-zarr/src/lib.rs
+++ b/python/sedonadb-zarr/src/lib.rs
@@ -104,15 +104,18 @@ impl PyZarrChunkReader {
 /// Build an `Arc<dyn ObjectStore>` for `uri` using env-var credential
 /// discovery, per scheme. This is the temporary bridge that lets the
 /// Python FFI work without going through `con.read_format`; it
-/// reaches into `AWS_*` / `GOOGLE_*` / `AZURE_*` environment
-/// variables via the `object_store` per-backend `Builder::from_env`
-/// helpers.
+/// reaches into `AWS_*` environment variables via the `object_store`
+/// per-backend `Builder::from_env` helpers.
+///
+/// Only the schemes exercised by the crate's tests are wired up (`s3`,
+/// `http`/`https`); `gcs`/`azure` are intentionally omitted until there
+/// is coverage for them.
 ///
 /// **Slated for removal.** When the host wheel surfaces
 /// `args.src.store` to `open_reader` via the `FFI_ObjectStore`
 /// capsule machinery (in progress — see
 /// <https://github.com/apache/sedona-db/pull/890>), this function (and
-/// the `object_store` `aws`/`gcp`/`azure`/`http` features that back it)
+/// the `object_store` `aws`/`http` features that back it)
 /// gets deleted — the store will arrive credentialed from the host's
 /// `ObjectStoreRegistry` and `PyZarrChunkReader::new` will extract it
 /// from the capsule instead. For `file://` and bare paths the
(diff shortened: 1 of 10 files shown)


python/sedonadb-zarr/src/lib.rs
```diff
@@ -39,7 +42,24 @@ impl PyZarrChunkReader {
...
-        let reader = ZarrChunkReader::try_new(uri, arrays.as_deref(), batch_size)
+        let store =
+            default_object_store_for_uri(uri).map_err(|e| PyValueError::new_err(e.to_string()))?;
+        let storage =
+            open_storage_from_uri(uri, store).map_err(|e| PyValueError::new_err(e.to_string()))?;
+        // The crate's async-native loader exposes `try_new` as an
+        // `async fn`; the Python FFI is a sync constructor by
+        // contract, so we bridge with an ad-hoc current-thread tokio
+        // runtime here. `next()` on the returned reader is pure CPU
+        // and stays synchronous.
```
paleolimbot: This is probably fine, although you could also make a shared (global/static) runtime for your whole Python package if this becomes an issue.

rust/sedona-raster-zarr/tests/cloud.rs
```diff
@@ -0,0 +1,127 @@
...
+use std::sync::Arc;
+
+use object_store::aws::AmazonS3Builder;
+use object_store::http::HttpBuilder;
+use object_store::ObjectStore;
+use sedona_raster::array::RasterStructArray;
+use sedona_raster_zarr::{open_storage_from_uri, ZarrChunkReader};
+
+/// ITS_LIVE bucket layout: `s3://<BUCKET>/<KEY>` ↔
+/// `https://<BUCKET>.s3.us-west-2.amazonaws.com/<KEY>`.
```
paleolimbot: Can you add a URL for a site describing what this is / the data?

Or if you are only going to test one object store, maybe just use a filesystem store?

rust/sedona-raster-zarr/Cargo.toml
```diff
@@ -33,17 +33,26 @@ result_large_err = "allow"
 [dependencies]
 arrow-array = { workspace = true }
 arrow-schema = { workspace = true }
+async-trait = { workspace = true }
+datafusion = { workspace = true, default_features = false }
```
paleolimbot: If you can avoid this one by using a subcrate, the compile will probably be faster / wheel will probably be smaller

python/sedonadb-zarr/src/lib.rs
```diff
@@ -70,6 +90,88 @@ impl PyZarrChunkReader {
...
+    // file:// and bare paths use a LocalFileSystem rooted at `/`;
+    // open_storage_from_uri prefixes it at the group's path. This
+    // matches the store the host's ObjectStoreRegistry yields for
+    // file://, so the loader treats local and cloud uniformly.
+    if uri.starts_with("file://") || !uri.contains("://") {
+        return Ok(Arc::new(object_store::local::LocalFileSystem::new()));
+    }
+    let url = Url::parse(uri)
+        .map_err(|e| PyValueError::new_err(format!("group URI {uri:?} is not a valid URL: {e}")))?;
+    match url.scheme().to_ascii_lowercase().as_str() {
```
paleolimbot, line 134: There are some tests in the sedona crate (I think) that smoke test these openers that you can steal. Alternatively, include fewer options since we'll remove some (not sure HTTP applies, unless there's some directory listing stuff I don't know about, and gcs/azure could probably be dropped for now).

python/sedonadb-zarr/src/lib.rs
james-willis: done

python/sedonadb-zarr/src/lib.rs
james-willis: This is the temporary code until we pipe the store through. I'm not sure how much testing we want to add just for that

rust/sedona-raster-zarr/Cargo.toml
james-willis: deleted - wasn't needed at all

rust/sedona-raster-zarr/tests/cloud.rs
james-willis: done

python/sedonadb-zarr/src/lib.rs
paleolimbot: I think you can just remove the untested code here (filesystem and S3 should suffice to actually use this until we pipe through the store)

james-willis: done

paleolimbot (approved):
Thank you!

PR #890, opened 2026-05-30
feat(c/sedona-extension): Add FFI for ObjectStore

paleolimbot:
Major work in progress. This is the most complex possible case (handles full async, non-blocking, no shortcuts). We could make this a lot shorter by making some sacrifices with respect to blocking.


james-willis:
This seems like something that would be valuable to other datafusion users. Could this eventually find a home in datafusion-ffi?

PR #902, opened 2026-06-02
feat(sedonadb/python): Add informative error for len() of a DataFrame

### LINKED ISSUE #553: Implement `__len__` on SedonaDB DataFrame for Pythonic `len(df)` support
It would be great if SedonaDB DataFrames supported Python’s built-in `len()` by implementing `__len__`.

In Python data tooling, it’s common (and expected) that `len(df)` returns the number of rows in a dataframe-like object. For example, DuckDB relations/dataframes support this kind of “Pythonic” ergonomics.

Today, calling `len(df)` on a SedonaDB DataFrame raises a `TypeError` because the DataFrame type doesn’t implement `__len__`.

paleolimbot:
Adds an implementation of __len__ that errors and informs the user of how to get a row count.

```python
import sedona.db

sd = sedona.db.connect()
len(sd.sql("SELECT 1 as one"))
#> ValueError: Can't compute len() of a lazy SedonaDB DataFrame. Use .count() to execute and resolve the number of rows.
```

Closes #553.

### DIFF
--- a/python/sedonadb/python/sedonadb/dataframe.py
+++ b/python/sedonadb/python/sedonadb/dataframe.py
@@ -1216,6 +1216,12 @@ class DataFrame:
         else:
             return super().__repr__()
 
+    def __len__(self):
+        raise ValueError(
+            "Can't compute len() of a lazy SedonaDB DataFrame. "
+            "Use .count() to execute and resolve the number of rows."
+        )
+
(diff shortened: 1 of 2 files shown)


prantogg (approved):
Went through the previous discussions. Agree on all points. Thanks!!

PR #906, opened 2026-06-02
feat(rust/sedona-geoparquet,rust/sedona-datasource)!: Add support for partition columns and discovery

### LINKED ISSUE #765: rust/sedona-geoparquet: Read of hive style partitions are not picked up by default
I believe that our default `read_parquet()` does not automatically add hive-style partitions by default (I'll try to come up with a reproducer soon...this was reported in Discord).

I think this is now supported in DataFusion proper but it's possible that at the time we ported over the listing table/format wrappers it hadn't shown up yet.

paleolimbot:
This PR adds auto-discovery to hive style partitions (can be turned of globally or per read).

This is breaking change because the schema that is read pointing SedonaDB at a directory of partitioned Parquet files is now different. The old behaviour can be resurrected with `SET datafusion.execution.listing_table_factory_infer_partitions = false`.

```python
import sedona.db

sd = sedona.db.connect()

t = sd.funcs.table.sd_random_geometry()
t.select(
    grp=sd.funcs.floor(t.id / 100), geom=sd.funcs.st_setsrid(t.geometry, 3857)
).to_parquet("out_dir", partition_by="grp")

sd.read_parquet("out_dir").show(5)
# ┌──────────────────────────────────────────────┬──────┐
# │                     geom                     ┆  grp │
# │                   geometry                   ┆ utf8 │
# ╞══════════════════════════════════════════════╪══════╡
# │ POINT(59.63310565314948 22.488904786369556)  ┆ 9    │
# ├╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌┤
# │ POINT(83.12076970430842 21.796712564259458)  ┆ 9    │
# ├╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌┤
# │ POINT(30.891948399833048 18.511790453362643) ┆ 9    │
# ├╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌┤
# │ POINT(87.62900829596556 90.36699740467581)   ┆ 9    │
# ├╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌┤
# │ POINT(89.95104383376538 57.37266977967324)   ┆ 9    │
# └──────────────────────────────────────────────┴──────┘
```

Before this PR, the `grp` col wasn't shown (and there was no way to reconstruct it).

I made this the default, because it's been the default in DataFusion since 50.0.0 (about a year).

Closes #765.

### DIFF
--- a/python/sedonadb/tests/io/test_pyogrio.py
+++ b/python/sedonadb/tests/io/test_pyogrio.py
@@ -91,8 +91,7 @@ def test_read_ogr_multi_file(con):
             con.read_parquet(parquet_path).to_pandas().to_file(fgb_path)
 
         # Reading a directory while specifying the extension should work
-        # Disable partitioning since this test focuses on multi-file reading
-        con.read_pyogrio(f"{td}", extension="fgb", partitioning=[]).to_view(
+        con.read_pyogrio(f"{td}", extension="fgb").to_view(
             "gdf_from_dir", overwrite=True
         )
         geopandas.testing.assert_geodataframe_equal(
@@ -101,9 +100,7 @@ def test_read_ogr_multi_file(con):
         )
 
         # Reading using a glob without specifying the extension should work
-        con.read_pyogrio(f"{td}/**/*.fgb", partitioning=[]).to_view(
-            "gdf_from_glob", overwrite=True
-        )
+        con.read_pyogrio(f"{td}/**/*.fgb").to_view("gdf_from_glob", overwrite=True)
         geopandas.testing.assert_geodataframe_equal(
             con.sql("SELECT * FROM gdf_from_glob ORDER BY idx").to_pandas(),
             gdf.filter(["idx", "wkb_geometry"]),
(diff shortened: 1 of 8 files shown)


python/sedonadb/tests/io/test_pyogrio.py
```diff
@@ -91,19 +91,24 @@ def test_read_ogr_multi_file(con):
...
 
         # Reading a directory while specifying the extension should work
+        # Partition columns are auto-discovered from the directory structure
         con.read_pyogrio(f"{td}", extension="fgb").to_view(
             "gdf_from_dir", overwrite=True
         )
         geopandas.testing.assert_geodataframe_equal(
-            con.sql("SELECT * FROM gdf_from_dir ORDER BY idx").to_pandas(),
-            gdf.filter(["idx", "wkb_geometry"]),
+            con.sql(
```
zhangfengcdt, line 99: Looks like due to the default behavior changes and now it needs to explicitly add "partition", the public constructor would be impacted. Maybe we should highlight the changes in the PR description or and change log if we don't want this change to be opt-in?

PR #901, opened 2026-06-02
feat(python/sedonadb): Make expressions context-aware for piped functions

paleolimbot:
In https://github.com/apache/sedona-db/pull/885 we expose scalar and aggregate functions from `sd.funcs` but this still didn't allow for the piped syntax that pandas and polars allow like `sd.col("foofy").sum()` because there was no way for `Expr` to look up the `sum` function from the context's registry.

This PR adds a `_ctx` member on the expression and adds `funcs` to get *any* function from the context (piping `self` in as the first argument).

This was extracted from https://github.com/apache/sedona-db/pull/862 which is a bit more ambitious but not strictly needed (generating source code for inline documentation).

```python
import sedona.db

sd = sedona.db.connect()
t = sd.funcs.table.sd_random_geometry()
t.select(geom=t.geometry.funcs.st_buffer(10)).show(2)
# ┌──────────────────────────────────────────────────────────────────────────────────────────────────┐
# │                                               geom                                               │
# │                                             geometry                                             │
# ╞══════════════════════════════════════════════════════════════════════════════════════════════════╡
# │ MULTIPOLYGON(((77.01165169396903 4.571593672098363,77.20379888993672 2.620690451937078,77.77285… │
# ├╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌┤
# │ MULTIPOLYGON(((2.2288873806768237 68.82846911756472,2.4210345766445194 66.87756589740344,2.9900… │
# └──────────────────────────────────────────────────────────────────────────────────────────────────┘
```

PR #908, opened 2026-06-03
feat(python/sedonadb): add DataFrame.join

jiayuasu:
First of the join sub-PRs from #791. `cross_join` is a separate small follow-up.

## API

```python
df.join(other, on="k")                              # equi-join on common name
df.join(other, on=["k1", "k2"], how="left")         # multi-key
df.join(other, on="k", how="outer")
df.join(other, on=left.k == right.kr)               # arbitrary predicate
df.join(other, on=[left.a == right.a, left.b > right.b])
```

- `on` accepts a column name (`str`), a list of column names, a boolean `Expr`, or a list of boolean `Expr`s combined with logical AND.
- `how`: `inner` (default), `left`, `right`, `outer`, `left_semi`, `left_anti`, `right_semi`, `right_anti`. PySpark aliases also accepted: `full` → outer, `semi` → left_semi, `anti` → left_anti.

## Output shape

- **Column names (`on="k"` / `on=["k1", "k2"]`)**: result has a single copy of each join key — matches pandas / Polars / PySpark, not DataFusion's keep-both-copies default.
- **Predicate `Expr` (`on=left.k == right.k`)**: both sides' columns are kept verbatim — same as PySpark's predicate join. Caller controls disambiguation via `df.alias(...)` when the same column name appears on both sides.

## Worth flagging — the auto-dedup machinery (column-name path)

DataFusion's DataFrame join has two behaviors that diverge from user expectations:

1. **Rejects two unaliased inputs that share column names** with "duplicate qualified field name". Both default to the `?table?` qualifier; the merged schema has unresolvable collisions before the join even runs.
2. **Keeps both copies of the join key** in the result when inputs are aliased. The SQL `USING` parser does the dedup at parse time; the DataFrame API doesn't.

For the column-name path, the Python wrapper:

1. Aliases both sides internally with sentinel qualifiers (`_sd_join_left_` / `_sd_join_right_`).
2. Synthesizes `aliased_left[k] == aliased_right[k]` predicates and calls `join_on`.
3. Projects with fully qualified column refs to dedupe the key columns and strip the sentinel qualifiers from the output.

The unified join-key column:
- **inner / left**: sourced from the left side (always populated).
- **right**: sourced from the right side (left may be NULL for unmatched rows).
- **outer**: `COALESCE(left.k, right.k)` — picks the populated side for rows unmatched on either input.
- **left_semi / left_anti / right_semi / right_anti**: only one side's columns appear; projection is straightforward.

For the predicate path, none of this dedup machinery runs — the Exprs flow straight through to `DataFrame::join_on`.

## Implementation

| File | Change |
|---|---|
| `python/sedonadb/src/dataframe.rs` | New `InternalDataFrame::join_on(right, predicates, how)`. Thin wrapper over DataFusion's `DataFrame::join_on`. Maps the `how` string to `JoinType`; takes a `Vec<PyExpr>` of predicates. |
| `python/sedonadb/python/sedonadb/dataframe.py` | New `DataFrame.join(other, on, how)`. Routes the column-name path through the alias-and-project pipeline; passes Expr predicates straight through. |

## Test plan

20 tests in `tests/expr/test_dataframe_join.py`:

- **Positive (column-name path)**: single-key inner / left / right / outer; multi-key inner; left_semi; left_anti; right_semi; right_anti.
- **PySpark `how` aliases**: `semi` / `anti` / `full`.
- **Positive (predicate path)**: single Expr equi-join; list of Exprs combined with AND; non-equi predicate (`left.x < right.y`) as a spatial-join shape analogue.
- **Errors**: non-DataFrame `other`; bad `on` type; empty `on` list; mixed str / non-str (or str / Expr) in `on` list; invalid `how`; unknown column (caught pre-flight as `KeyError`).

All output assertions use exact `pd.testing.assert_frame_equal` after sorting.

Local: 20 unit + 24 doctests + 187 expr-dir tests + `ruff format` + `ruff check` + `cargo fmt --check` all clean.

## Known limitations (for follow-up)

- **Non-key column-name collisions** in the column-name path are not auto-suffixed (`_x` / `_y` like pandas). The duplicate names propagate and become ambiguous to reference. Deferred to a later PR per the design discussion.
- **Self-join** requires the user to alias one side explicitly. The internal sentinel aliasing handles the unaliased common case but doesn't disambiguate a self-join.

### DIFF
--- a/python/sedonadb/python/sedonadb/dataframe.py
+++ b/python/sedonadb/python/sedonadb/dataframe.py
@@ -634,6 +634,168 @@ class DataFrame:
 
         return GroupedDataFrame(self, coerced)
 
+    def join(
+        self,
+        other: "DataFrame",
+        on: Union[str, List[str], Expr, List[Expr]],
+        how: Literal[
+            "inner",
+            "left",
+            "right",
+            "outer",
+            "full",
+            "left_semi",
+            "semi",
+            "left_anti",
+            "anti",
+            "right_semi",
+            "right_anti",
+        ] = "inner",
+    ) -> "DataFrame":
+        """Join two DataFrames.
+
+        `on` accepts either common column names or arbitrary boolean
+        predicates:
+
+        - **Column names** (`str` or `list[str]`): the named column(s)
+          must exist on both sides. Result has a single copy of each
+          join key — matching pandas / Polars / PySpark output shape.
+        - **Predicate expressions** (`Expr` or `list[Expr]`): each Expr
+          is a boolean predicate combining columns from both sides
+          (e.g. `left.k == right.k`, or `f.st_intersects(left.g, right.g)`).
+          Result keeps both sides' columns verbatim — disambiguate
+          with `df.alias(...)` on either side.
+
+        Args:
+            other: The right-hand DataFrame to join against.
+            on: Join key(s). A column name (`str`), a list of column
+                names, a single boolean `Expr`, or a list of boolean
+                `Expr`s combined with logical AND.
+            how: Join type. Canonical: `"inner"` (default), `"left"`,
+                `"right"`, `"outer"`, `"left_semi"`, `"left_anti"`,
+                `"right_semi"`, `"right_anti"`. PySpark aliases also
+                accepted: `"full"` (= outer), `"semi"` (= left_semi),
+                `"anti"` (= left_anti).
+
+        Examples:
+
+            >>> sd = sedona.db.connect()
+            >>> left = sd.sql(
+            ...     "SELECT * FROM (VALUES (1, 'a'), (2, 'b')) AS t(k, v)"
+            ... )
+            >>> right = sd.sql(
+            ...     "SELECT * FROM (VALUES (1, 'x'), (2, 'y'), (3, 'z')) AS t(k, w)"
+            ... )
+            >>> left.join(right, on="k").sort("k").show()
+            ┌───────┬──────┬──────┐
+            │   k   ┆   v  ┆   w  │
+            │ int64 ┆ utf8 ┆ utf8 │
+            ╞═══════╪══════╪══════╡
+            │     1 ┆ a    ┆ x    │
+            ├╌╌╌╌╌╌╌┼╌╌╌╌╌╌┼╌╌╌╌╌╌┤
+            │     2 ┆ b    ┆ y    │
+            └───────┴──────┴──────┘
+        """
+        if not isinstance(other, DataFrame):
+            raise TypeError(
+                f"join() expects a DataFrame as the first argument, "
+                f"got {type(other).__name__}"
+            )
+
+        HOW_ALIASES = {"semi": "left_semi", "anti": "left_anti", "full": "outer"}
+        valid_canonical = {
+            "inner",
+            "left",
+            "right",
+            "outer",
+            "left_semi",
+            "left_anti",
+            "right_semi",
+            "right_anti",
+        }
+        canonical_how = HOW_ALIASES.get(how, how)
+        if canonical_how not in valid_canonical:
+            accepted = sorted(valid_canonical | set(HOW_ALIASES))
+            raise ValueError(f"join() `how` must be one of {accepted}, got {how!r}")
+
+        if isinstance(on, (str, Expr)):
+            on_list = [on]
+        elif isinstance(on, list):
+            on_list = on
+        else:
+            raise TypeError(
+                f"join() `on` expects str, Expr, or a list of either, "
+                f"got {type(on).__name__}"
+            )
+
+        if not on_list:
+            raise ValueError("join() requires at least one element in `on`")
+
+        is_str_keys = all(isinstance(x, str) for x in on_list)
+        is_expr_keys = all(isinstance(x, Expr) for x in on_list)
+        if not (is_str_keys or is_expr_keys):
+            raise TypeError(
+                "join() `on` list must contain only str or only Expr, not a mix"
+            )
+
+        if is_expr_keys:
+            joined_impl = self._impl.join_on(
+                other._impl, [p._impl for p in on_list], canonical_how
+            )
+            return DataFrame(self._ctx, joined_impl)
+
+        # String-keys path: alias both sides, synthesize equi-join
+        # predicates from the qualified columns, then project to dedupe
+        # the join keys so the output shape matches pandas / PySpark
+        # rather than DataFusion's keep-both-copies default.
+        LEFT_ALIAS = "_sd_join_left_"
+        RIGHT_ALIAS = "_sd_join_right_"
+        left_cols = self._impl.columns()
+        right_cols = other._impl.columns()
+
+        missing_left = [k for k in on_list if k not in left_cols]
+        missing_right = [k for k in on_list if k not in right_cols]
+        if missing_left or missing_right:
+            raise KeyError(
+                f"Join keys missing — left: {missing_left}, "
+                f"right: {missing_right}. "
+                f"Left columns: {left_cols}; right columns: {right_cols}"
+            )
+
+        left_aliased = self.alias(LEFT_ALIAS)
+        right_aliased = other.alias(RIGHT_ALIAS)
+
+        predicates = [(left_aliased[k] == right_aliased[k])._impl for k in on_list]
+        joined_impl = left_aliased._impl.join_on(
+            right_aliased._impl, predicates, canonical_how
+        )
+
+        if canonical_how in ("left_semi", "left_anti"):
+            projection = [left_aliased[c]._impl for c in left_cols]
+        elif canonical_how in ("right_semi", "right_anti"):
+            projection = [right_aliased[c]._impl for c in right_cols]
+        else:
+            # For the unified key column, pick the side that is always
+            # populated: right join takes from the right, outer COALESCEs
+            # so unmatched-on-either-side rows still carry a key value.
+            key_set = set(on_list)
+            projection = []
+            for c in left_cols:
+                if c in key_set and canonical_how == "right":
+                    projection.append(right_aliased[c]._impl)
+                elif c in key_set and canonical_how == "outer":
+                    coalesced = self._ctx.funcs.coalesce(
+                        left_aliased[c], right_aliased[c]
+                    ).alias(c)
+                    projection.append(coalesced._impl)
+                else:
+                    projection.append(left_aliased[c]._impl)
+            for c in right_cols:
+                if c not in key_set:
+                    projection.append(right_aliased[c]._impl)
+
+        return DataFrame(self._ctx, joined_impl.select(projection))
+
     def limit(self, n: Optional[int], /, *, offset: int = 0) -> "DataFrame":
         """Limit result to n rows starting at offset
 
--- a/python/sedonadb/src/dataframe.rs
+++ b/python/sedonadb/src/dataframe.rs
@@ -28,7 +28,7 @@ use datafusion::logical_expr::SortExpr;
 use datafusion::prelude::{DataFrame, SessionContext};
 use datafusion_common::{Column, DataFusionError, ParamValues};
 use datafusion_execution::TaskContextProvider;
-use datafusion_expr::{ExplainFormat, ExplainOption, Expr};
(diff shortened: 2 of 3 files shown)


python/sedonadb/src/dataframe.rs
```diff
@@ -26,7 +26,7 @@ use datafusion::catalog::MemTable;
 use datafusion::config::ConfigField;
 use datafusion::logical_expr::SortExpr;
 use datafusion::prelude::{DataFrame, SessionContext};
-use datafusion_common::{Column, DataFusionError, ParamValues};
+use datafusion_common::{Column, DataFusionError, JoinType, ParamValues};
 use datafusion_execution::TaskContextProvider;
 use datafusion_expr::{ExplainFormat, ExplainOption, Expr};
```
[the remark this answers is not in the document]
jiayuasu: Moved in ef61ea6e5 — `JoinType` is now imported from `datafusion_expr` to match the dominant convention in the workspace (8 sites already use that path).

python/sedonadb/python/sedonadb/dataframe.py
```diff
@@ -640,6 +640,139 @@ def group_by(self, *keys: Union[str, Expr]) -> "GroupedDataFrame":
...
+            projection = []
+            for c in left_cols:
+                if c in key_set and how == "right":
+                    projection.append(_col(c, RIGHT_ALIAS)._impl)
+                elif c in key_set and how == "outer":
+                    # Outer join: rows unmatched on either side have NULL on
+                    # that side's key; COALESCE picks the populated one.
+                    from sedonadb.expr.expression import ScalarUdf
+
+                    coalesce_udf = ScalarUdf(self._ctx.scalar_udf("coalesce"))
```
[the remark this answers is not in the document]
jiayuasu: `self._ctx` in this `DataFrame` class is actually the internal context handle (`_lib.InternalContext`), not the user-facing `SedonaContext` Python class — and `InternalContext` exposes `scalar_udf(name)` (added in #885). The outer-join test `test_join_outer` covers this code path and passes. Added a clarifying comment at the call site in ef61ea6e5 since the naming is genuinely confusing here.

python/sedonadb/python/sedonadb/dataframe.py
```diff
@@ -640,6 +640,139 @@ def group_by(self, *keys: Union[str, Expr]) -> "GroupedDataFrame":
...
+        right_cols = other._impl.columns()
+        left_aliased = self.alias(LEFT_ALIAS)
+        right_aliased = other.alias(RIGHT_ALIAS)
+
+        joined_impl = left_aliased._impl.join(right_aliased._impl, keys, how)
+
+        if how in ("left_semi", "left_anti"):
+            projection = [_col(c, LEFT_ALIAS)._impl for c in left_cols]
+        elif how in ("right_semi", "right_anti"):
+            projection = [_col(c, RIGHT_ALIAS)._impl for c in right_cols]
```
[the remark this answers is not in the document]
jiayuasu: Added `test_join_right_semi` and `test_join_right_anti` in ef61ea6e5, parallel to the existing left-semi/anti cases.

python/sedonadb/python/sedonadb/dataframe.py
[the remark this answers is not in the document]
paleolimbot: For what it's worth I updated this in https://github.com/apache/sedona-db/pull/901 (it was confusing there, too)

python/sedonadb/python/sedonadb/dataframe.py
```diff
@@ -640,6 +640,144 @@ def group_by(self, *keys: Union[str, Expr]) -> "GroupedDataFrame":
 
         return GroupedDataFrame(self, coerced)
 
+    def join(
+        self,
+        other: "DataFrame",
+        on: Union[str, List[str]],
+        how: str = "inner",
```
paleolimbot: Can you add a typing hint here (`Literal["inner", "left", "right", etc]`)? This gives a dropdown when typing this in a notebook.

python/sedonadb/python/sedonadb/dataframe.py
```diff
@@ -640,6 +640,144 @@ def group_by(self, *keys: Union[str, Expr]) -> "GroupedDataFrame":
...
+            # For the unified join-key column, pick the side that's always
+            # populated for that join type. Right joins take the right's
+            # key (left may be NULL for unmatched rows); outer joins COALESCE
+            # the two so unmatched rows on either side still get a key
+            # value.
+            key_set = set(keys)
+            projection = []
+            for c in left_cols:
+                if c in key_set and how == "right":
+                    projection.append(_col(c, RIGHT_ALIAS)._impl)
```
paleolimbot: I added a getter for getting the qualified column...you can just do `right_aliased[c]` (below can be simplified too).

python/sedonadb/python/sedonadb/dataframe.py
```diff
@@ -640,6 +640,144 @@ def group_by(self, *keys: Union[str, Expr]) -> "GroupedDataFrame":
...
+            └───────┴──────┴──────┘
+        """
+        if not isinstance(other, DataFrame):
+            raise TypeError(
+                f"join() expects a DataFrame as the first argument, "
+                f"got {type(other).__name__}"
+            )
+
+        if isinstance(on, str):
+            keys: List[str] = [on]
```
paleolimbot: LLMs love to add these type hints but I would love to avoid them (like LLM-generated comments, they can very easily become type hints that are lying)

python/sedonadb/src/dataframe.rs
```diff
@@ -257,6 +257,49 @@ impl InternalDataFrame {
...
+    /// 2. Borrow the column-name vectors as `&[&str]` since DataFusion's
+    ///    `DataFrame::join` takes that shape.
+    /// 3. Pass `filter=None`; an additional residual predicate is the
+    ///    next sub-PR (expression-based join).
+    fn join(
+        &self,
+        right: &InternalDataFrame,
+        join_cols: Vec<String>,
+        how: &str,
+    ) -> Result<InternalDataFrame, PySedonaError> {
```
paleolimbot: It is probably worth only exposing the expression endpoint here to reduce the binding surface (you can easily generate the expression with `[left_aliased[c] == right_aliased[c] for c in keys]`).

Can you avoid any references to future PRs? These PRs stand nicely on their own.

python/sedonadb/tests/expr/test_dataframe_join.py
```diff
@@ -0,0 +1,165 @@
...
+
+
+def test_join_outer(con):
+    left = con.create_data_frame(pd.DataFrame({"k": [1, 2], "v": ["a", "b"]}))
+    right = con.create_data_frame(pd.DataFrame({"k": [2, 3], "w": ["y", "z"]}))
+    # `outer` returns all rows from both sides; the merged key column has
+    # the union of values.
+    out = left.join(right, on="k", how="outer").sort("k").to_pandas()
+    assert sorted(out["k"].tolist()) == [1, 2, 3]
+    assert len(out) == 3
```
paleolimbot: Is there a reason this isn't testing the actual dataframe output?

python/sedonadb/python/sedonadb/dataframe.py
```diff
@@ -640,6 +640,144 @@ def group_by(self, *keys: Union[str, Expr]) -> "GroupedDataFrame":
...
+        valid_how = {
+            "inner",
+            "left",
+            "right",
+            "outer",
+            "left_semi",
+            "left_anti",
+            "right_semi",
+            "right_anti",
+        }
```
paleolimbot: I don't mind whether we do this or not, but pyspark seems to allow "semi" and "anti" as aliases for "left_semi" and "left_anti", and "full" as an alias for "outer".

paleolimbot (commented):
Thank you!

A few things worth probably worth tackling now. The expr variant should be straightforward/pyspark compatible out of the box:

```python
t = sd.read_parquet(...)
t2 = sd.read_parquet(...)
t.join(t2, on=t.x == t2.x).select(t.a, t.b, t.x, t2.c)
```

python/sedonadb/python/sedonadb/dataframe.py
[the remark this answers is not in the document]
jiayuasu: Thanks — `self._ctx.funcs.coalesce(...)` now in 608cc69, replacing the `InternalContext.scalar_udf` dance. Much cleaner.

python/sedonadb/python/sedonadb/dataframe.py
jiayuasu: Added in 608cc69 — `Literal["inner", "left", "right", "outer", "full", "left_semi", "semi", "left_anti", "anti", "right_semi", "right_anti"]` so the dropdown surfaces both canonical names and the PySpark aliases.

python/sedonadb/python/sedonadb/dataframe.py
jiayuasu: Switched to `left_aliased[c]` / `right_aliased[c]` everywhere in the alias-and-project pipeline. The synthesized equi-join predicates are now `(left_aliased[k] == right_aliased[k])._impl` too — much nicer to read.

python/sedonadb/python/sedonadb/dataframe.py
jiayuasu: Agreed. No internal type annotations on local vars in the new `join()` body. The public signature keeps its hints (`on: Union[str, List[str], Expr, List[Expr]]`, `how: Literal[...]`) since those drive the IDE dropdown — happy to drop those too if you'd rather.

python/sedonadb/src/dataframe.rs
jiayuasu: Done in 608cc69. The Rust side now exposes only `InternalDataFrame::join_on(right, predicates: Vec<PyExpr>, how)` — a thin wrapper over DataFusion's `DataFrame::join_on`. The Python wrapper synthesizes `[left_aliased[k] == right_aliased[k] for k in keys]` for the column-name path and passes Expr predicates through verbatim. Forward-PR references stripped from both the commit message and the code comments.

python/sedonadb/tests/expr/test_dataframe_join.py
jiayuasu: Fixed — `test_join_outer` now asserts the full output via `pd.testing.assert_frame_equal`, including the COALESCE'd key column and the NULL-bearing value columns on both sides. While here I also added tests for the Expr predicate path (single, list, non-equi) and the PySpark `how` aliases.

python/sedonadb/python/sedonadb/dataframe.py
jiayuasu: Added — `full`, `semi`, `anti` route to `outer`, `left_semi`, `left_anti` respectively. Mapping is a small dict at the top of `join()`; `test_join_how_pyspark_aliases` covers all three.

python/sedonadb/python/sedonadb/dataframe.py
```diff
@@ -634,6 +634,158 @@ def group_by(self, *keys: Union[str, Expr]) -> "GroupedDataFrame":
...
+        # String-keys path: alias both sides, synthesize equi-join
+        # predicates from the qualified columns, then project to dedupe
+        # the join keys so the output shape matches pandas / PySpark
+        # rather than DataFusion's keep-both-copies default.
+        LEFT_ALIAS = "_sd_join_left_"
+        RIGHT_ALIAS = "_sd_join_right_"
+        left_cols = self._impl.columns()
+        right_cols = other._impl.columns()
+        left_aliased = self.alias(LEFT_ALIAS)
+        right_aliased = other.alias(RIGHT_ALIAS)
```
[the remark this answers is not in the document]
jiayuasu: Added in 3e00905. Pre-flight check builds `missing_left` / `missing_right` from `on_list` and raises a single `KeyError` naming both sides explicitly, e.g.:

> Join keys missing — left: [], right: ['k']. Left columns: ['k', 'v']; right columns: ['j', 'w']

Covered by `test_join_key_missing_on_one_side_errors` and an updated `test_join_unknown_column_errors`.

paleolimbot (approved):
Thank you! I'm excited to give this a spin!

PR #909, opened 2026-06-03
chore(docker): support multi-arch builds for the GPU image and add a publish script

jiayuasu:
## What

- Make the GPU Dockerfile (`docker/sedonadb-gpu.dockerfile`) build on both `linux/amd64` and `linux/arm64`.
- Add a publishing script (`docker/build.sh`), modeled on Apache Sedona's `docker/build.sh`.

## Why

The Dockerfile hardcoded the amd64-only vcpkg triplet (`installed/x64-linux/lib/pkgconfig`). On arm64, vcpkg installs GEOS/abseil under `arm64-linux`, so the `s2geography` build could not find pkg-config and the build failed. The triplet is now resolved from the host architecture at build time, so the image compiles cleanly on both architectures.

`docker/build.sh` mirrors the Apache Sedona convention:
- `local` — single-arch build loaded into the local Docker daemon (`--load`).
- `release` — multi-arch (`linux/amd64,linux/arm64`) build pushed straight to the registry, defaulting to the `apache/sedona:sedonadb-latest` tag.

## Verification

A multi-arch build was produced and pushed for both `linux/amd64` and `linux/arm64`. The image is now available on Docker Hub:

https://hub.docker.com/r/apache/sedona/tags

Tag: `apache/sedona:sedonadb-latest`

Run it with:
```bash
docker run -it --rm --gpus all -p 8888:8888 apache/sedona:sedonadb-latest
```


jiayuasu:
@zhangfengcdt @pwrliang please give the docker image a try

pwrliang:
> @zhangfengcdt @pwrliang please give the docker image a try

I have tried the image on a `g5.2xlarge` instance, and it works. I can reproduce the [example](https://github.com/apache/sedona-db/blob/main/docs/gpu-acceleration.ipynb). For the arm-based image, I have tried [`g5g.2xlarge`](https://aws.amazon.com/ec2/instance-types/g5g/), and it failed:

```
SedonaError: Execution error: Failed to build spatial refiner on GPU FinishBuild("after determining tmp storage requirements for inclusive_scan: cudaErrorNoKernelImageForDevice: no kernel image is available for execution on the device")
```

The image (built for Compute Capability 8.6) fails on g5g.2xlarge instances because they use the older NVIDIA T4 GPU (7.5), and CUDA JIT is only forward-compatible. To let users run ARM+GPU on AWS out of the box, we can compile the image with CMAKE_CUDA_ARCHITECTURES=75. This enables JIT for newer GPUs, but causes a ~170-second delay on the first run. Alternatively, we can compile for multiple architectures (e.g., 7.5, 8.6, and 8.9) to avoid the startup delay entirely, at the cost of a slightly larger image size.

jiayuasu:
Thanks for testing on both instance types, @pwrliang!

I've rebuilt and republished `apache/sedona:sedonadb-latest` as a fat binary covering compute capabilities 7.5, 8.6, and 8.9 (`CMAKE_CUDA_ARCHITECTURES=75;86;89`), so the T4 on `g5g.2xlarge` should now have a native kernel image and run without the `cudaErrorNoKernelImageForDevice` error — and without the JIT startup delay. This also covers Turing (T4), Ampere (A10G), and Ada Lovelace (L4) out of the box.

Could you re-pull and give the arm64 image another try on `g5g.2xlarge` when you get a chance?

New manifest digest: `sha256:ffcaf6827c563a61eef0becb76d8a768d754d9421cb8b0015a58f47336ed10c3`

pwrliang:
> Thanks for testing on both instance types, @pwrliang!
> 
> I've rebuilt and republished `apache/sedona:sedonadb-latest` as a fat binary covering compute capabilities 7.5, 8.6, and 8.9 (`CMAKE_CUDA_ARCHITECTURES=75;86;89`), so the T4 on `g5g.2xlarge` should now have a native kernel image and run without the `cudaErrorNoKernelImageForDevice` error — and without the JIT startup delay. This also covers Turing (T4), Ampere (A10G), and Ada Lovelace (L4) out of the box.
> 
> Could you re-pull and give the arm64 image another try on `g5g.2xlarge` when you get a chance?
> 
> New manifest digest: `sha256:ffcaf6827c563a61eef0becb76d8a768d754d9421cb8b0015a58f47336ed10c3`

Thanks for the fix. Now, the image works on g5g.2xlarge!

PR #915, opened 2026-06-06
refactor(raster): drop data()/contiguous_data() from Raster interface

james-willis:
After some discussions with @paleolimbot in some recent raster PRs, we reached the conclusion that we don't want the raster type to do transparent allocations. 

Given this I'm simplifying the raster byte access patterns with this PR to be just the `nd_buffer` method, and eliminating the `data` and `contiguous_data` methods. Instead we added two methods to the NDBuffer trait:

* `is_contiguous`: whether or not the data exported byte the buffer is all contiguous in memory
* `as_contiguous`: returns that buffer as a `Result<&'a [u8], ArrowError>`, throwing an error if the data is not contiguous.

This avoids any transparent allocations. We also expose a `try_contiguous_data` method 

## Drive-by: split `is_2d` into `is_spatial_2d` + a contiguity check

While moving the GDAL boundary off `data()`, I noticed `BandRef::is_2d()` was conflating two unrelated things: whether a band is a 2-D spatial raster (dim count + recognized `y/x`, `lat/lon`, … names) *and* whether its view is the identity (a byte-layout question). The name only advertised the first.

So I split it:

* `is_spatial_2d()` — a pure shape/naming predicate (dims only).
* contiguity is now its own question, answered by `NdBuffer::is_contiguous()` / `try_contiguous_data()`.

The GDAL gate now uses `is_spatial_2d()` for dimensionality and lets `try_contiguous_data()` enforce the byte layout where the bytes are actually read (it errors on a strided view). This also relaxes the old strict-identity requirement to plain contiguity — `is_contiguous()` is offset-agnostic, so a contiguous outer-axis slice is acceptable rather than rejected. Behavior-preserving today, since the read boundary still rejects non-identity views.

### DIFF
--- a/rust/sedona-raster-functions/src/rs_ensure_loaded.rs
+++ b/rust/sedona-raster-functions/src/rs_ensure_loaded.rs
@@ -359,13 +359,13 @@ where
                     // boundary rejects non-identity views — but guard it loudly
                     // so that when views land this surfaces as the internal gap
                     // it is, not a confusing downstream error. Fixed alongside
-                    // the zero-copy view-preserving passthrough
-                    // (apache/sedona-db#894 + #897).
+                    // https://github.com/apache/sedona-db/pull/894
+                    // https://github.com/apache/sedona-db/pull/897
                     if !view_is_identity(&view_owned, &source_shape) {
                         return sedona_internal_err!(
                             "RS_EnsureLoaded: InDb band ({raster_idx},{band_idx}) has a \
                              non-identity view; view-preserving passthrough is not \
-                             implemented yet (apache/sedona-db#897)"
+                             implemented yet"
                         );
                     }
                     let bytes = band
--- a/rust/sedona-raster-functions/src/rs_ensure_loaded.rs
+++ b/rust/sedona-raster-functions/src/rs_ensure_loaded.rs
@@ -341,8 +341,8 @@ where
                 let outdb_format: Option<String> = band.outdb_format().map(|s| s.to_string());
                 // For InDb bands, copy bytes into an owned Buffer.
                 // `Buffer::from_vec` is zero-copy ownership transfer of
-                // the Vec; the per-row clone of `band.data()` itself is
-                // the one InDb copy we accept for sequential simplicity.
+                // the Vec; the per-row clone of the band's contiguous bytes
+                // is the one InDb copy we accept for sequential simplicity.
                 //
                 // This passthrough copy (and the equivalent re-copy of
                 // freshly-loaded OutDb bytes through the builder) goes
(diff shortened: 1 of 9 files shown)


rust/sedona-raster-functions/src/rs_ensure_loaded.rs
```diff
@@ -351,7 +351,34 @@ where
...
+                    // A non-identity view can't be passed through yet: the
+                    // rebuild via `start_band_nd` below emits an identity band,
+                    // so the view would be silently dropped (and a contiguous
+                    // outer-slice would trip the source-byte-count check with a
+                    // misleading message). Unreachable today — the read
+                    // boundary rejects non-identity views — but guard it loudly
+                    // so that when views land this surfaces as the internal gap
+                    // it is, not a confusing downstream error. Fixed alongside
+                    // the zero-copy view-preserving passthrough
+                    // (apache/sedona-db#894 + #897).
```
paleolimbot:
```suggestion
                    // it is, not a confusing downstream error. Fixed alongside
                    // https://github.com/apache/sedona-db/pull/894
                    // https://github.com/apache/sedona-db/pull/897
```

rust/sedona-raster-functions/src/rs_ensure_loaded.rs
```diff
@@ -351,7 +351,34 @@ where
...
+                    if !view_is_identity(&view_owned, &source_shape) {
+                        return sedona_internal_err!(
+                            "RS_EnsureLoaded: InDb band ({raster_idx},{band_idx}) has a \
+                             non-identity view; view-preserving passthrough is not \
+                             implemented yet (apache/sedona-db#897)"
```
paleolimbot:
```suggestion
                             implemented yet"
```

rust/sedona-raster-gdal/src/gdal_common.rs
```diff
@@ -239,8 +239,13 @@ pub unsafe fn raster_ref_to_gdal_mem<R: RasterRef + ?Sized>(
...
         let band_type = band_metadata.data_type()?;
         let gdal_type = band_data_type_to_gdal(&band_type);
-        let band_data = band.data();
-        let data_ptr = band_data.as_ptr();
+        // `as_contiguous()` borrows the bytes zero-copy (erroring on a strided
+        // view); GDAL holds the pointer, so `raster` must outlive the dataset.
+        let band_bytes = band
+            .nd_buffer()
+            .and_then(|ndb| ndb.as_contiguous())
+            .map_err(|e| arrow_datafusion_err!(e))?;
```
paleolimbot:
```suggestion
            .and_then(|ndb| ndb.as_contiguous())?;
```

rust/sedona-raster/src/traits.rs
```diff
@@ -947,90 +955,99 @@ mod tests {
         }
     }
 
+    // `is_spatial_2d` is a shape/naming predicate — it inspects only
+    // `dim_names`, never the view. View/byte-layout concerns are covered by
+    // the `NdBuffer::is_contiguous` tests below.
+
```
```suggestion
```

paleolimbot (approved):
Only nits here. Thanks!

PR #916, opened 2026-06-06
feat(python/sedonadb): Enable zarr read via sedonadb-zarr

paleolimbot:
This PR allows actually loading pixels from backends that live in another Python package (notably: sedonadb_zarr). We had discussed in previous PRs implementing an FFI for this, but for now this PR just implements a Python interface on both sides and handles passing bytes via Python buffers.

There is also a tweak to the async load request to allow multiple requests to be issued at once, which works better for backends that have to launch a task (and is more realistic...usually the requests come in batches). For reasonable performance, RS_EnsureLoaded would need to be updated to take advantage of issuing more than one request at a time. That is not implemented here.

Example with zarrs:

```python
import matplotlib.pyplot as plt
import numpy as np

import sedona.db
import sedonadb_zarr

sd = sedona.db.connect()
sd.options.interactive = True
sd.register(sedonadb_zarr.ZarrExtension())

url = "https://atlantis-vis-o.s3-ext.jc.rl.ac.uk/hurricanes/era5/florence"
t = sd.read(url, format="zarr", options={"arrays": ["velocity"]})

tab = t.limit(1).select(
    raster=t.raster.funcs.rs_ensureloaded(),
    env=t.raster.funcs.rs_envelope().geo.set_srid(4326)
).to_arrow_table()
```

And we can look at it!

```python
r = tab["raster"][0].as_py()
data = r.bands[0].to_numpy()
env = tab["env"][0].to_shapely()

# Squeeze extra dimensions to get 2D array
data_2d = np.squeeze(data)

fig, ax = plt.subplots(figsize=(10, 8))
im = ax.imshow(data_2d, extent=[env.bounds[0], env.bounds[2], env.bounds[1], env.bounds[3]], origin='lower')
ax.set_xlabel("Longitude")
ax.set_ylabel("Latitude")
ax.set_title("Velocity Raster")
plt.colorbar(im, ax=ax, label="Velocity")
plt.show()
````

output

### CODE REVIEW

File: python/sedonadb/src/raster_loader.rs [hunk 1]
```diff
@@ -0,0 +1,472 @@
...
+            ArrowError::InvalidArgumentError(format!(
+                "Failed to get buffer from Python object: {e}"
+            ))
+        })?;
+
+        if !buffer.is_c_contiguous() {
+            return Err(ArrowError::InvalidArgumentError(
+                "Buffer must be C-contiguous".to_string(),
+            ));
+        }
```
james-willis, line 267: should we also check `buffer.readonly()`? doc string says we dont allow writable buffers

python/sedonadb-zarr/tests/test_zarr.py
```diff
@@ -194,7 +194,88 @@ def test_format_spec_class_invariants():
...
+        "int32",
+        "uint32",
+        "int64",
+        "uint64",
+        "float32",
+        "float64",
+    ],
+)
+def test_rs_ensure_loaded_with_zarr(tmp_path, numpy_dtype):
+    zarr = pytest.importorskip("zarr", minversion="3.0")
```
james-willis: This should fail if it cant import

python/sedonadb/src/raster_loader.rs
```diff
@@ -0,0 +1,481 @@
...
+                view: req.view.to_vec(),
+                data_type: req.data_type,
+            })
+            .collect();
+
+        let py_load = Python::with_gil(|py| self.py_load.clone_ref(py));
+
+        // Spawn a blocking task to hold the GIL and call Python
+        let results = tokio::task::spawn_blocking(move || {
+            Python::with_gil(|py| -> Result<Vec<PyRasterLoadResultData>, ArrowError> {
```
james-willis, line 104: will gil contention impede perf here?

james-willis (approved):
Couple small things inline! Some questions below to help my understanding

2 things im not totally following:

1. why do we need so many py wrappers? If we went the ABI route would it reduce the amount of pythoning that goes on here? But this one lets users implement extensions in python? Is there perf overhead to this approach vs the FFI/ABI approach?

2. What is the advantage of grouping requests?

python/sedonadb-zarr/tests/test_zarr.py
paleolimbot: I think wheel tests for python <3.11 will fail if this isn't 'skip'

python/sedonadb/src/raster_loader.rs
paleolimbot: It will, which is why we can/will move this to FFI for the next release. As long as RS_EnsureLoaded is grouping these requests properly it will be fine in most cases. The implementation on the other side calls allow_threads, which allows other Python code to run while it's waiting for IO.

python/sedonadb/src/raster_loader.rs
paleolimbot: I removed the comment. It doesn't matter if the buffer happens to be writable (Python does bookkeeping to ensure buffers are not mutated while references are held)

python/sedonadb-zarr/tests/test_zarr.py
james-willis: can we condition the import on pythong version then? I am afraid of silent failure

paleolimbot: I updated this, hopefully better now!

james-willis (approved):

Issue #897, opened 2026-06-01
RS_EnsureLoaded / AsyncByteLoader: round-trip non-identity band views (view in -> realized view out)


james-willis:
## Background

`RS_EnsureLoaded` materialises OutDb band bytes and passes InDb bands through. Today it cannot handle a band with a **non-identity view** (any slice / crop / broadcast / permutation, where the visible `shape()` differs from the backing `raw_source_shape()`):

- `ensure_loaded` rebuilds each band field-by-field via `start_band_nd` and never copies `band.view()`, so the view is silently dropped.
- `OutDbLoadRequest` carries only `source_shape` (the full backing extent); a loader has no way to learn that a sub-window was wanted.
- `AsyncByteLoader::load(req) -> Buffer` returns bare bytes; the caller can't tell what region/layout they represent.

This is **latent, not live**: `RasterRefImpl::band()` currently rejects non-identity views (`"view composition is not yet implemented"`), so such bands are unconstructible. The guardrail must come off as part of view-composition work (#813) — and when it does, `RS_EnsureLoaded` will silently corrupt viewed rasters unless this is addressed. **This issue is a correctness co-requisite of view composition landing.**

## The contract change: view in → realized view out

If the request carries a desired view (so a cloud loader *can* range-read just the sub-window instead of fetching a whole chunk), the loader has a legitimate choice, and the caller can't infer which it made from a bare buffer:

| Loader behavior | buffer | realized source_shape | realized view |
|---|---|---|---|
| Ignores crop (atomic chunk/tile — today's backends) | full backing bytes | full backing shape | input's view (unchanged) |
| Honors crop, returns contiguous cropped bytes | smaller | cropped shape | identity over cropped shape |
| Honors crop, returns full layout + selecting view | full | full | crop view |

`source_shape` and `view` move independently across these rows and neither is derivable from the buffer (its length only constrains `Π source_shape × byte_size`; a view entry doesn't reveal its source axis's true length). So the loader must report **all three** — the response is the data-bearing portion of a band:

```rust
struct LoadedBand {
    buffer: Buffer,
    source_shape: Vec<u64>,   // natural C-order extent of `buffer`
    view: Vec<ViewEntry>,     // how visible axes map onto source_shape
}

async fn load(&self, req: &OutDbLoadRequest<'_>) -> Result<LoadedBand, ArrowError>;
```

[...]

PR #917, opened 2026-06-06
feat(rust/sedona-raster,rust/sedona-raster-functions): zero-copy band-data in RS_EnsureLoaded

### LINKED ISSUE #894: RasterBuilder: zero-copy band-data transfer via BinaryView buffer sharing
## Problem

`RasterBuilder` has no zero-copy path for band data. The single data sink is a `BinaryViewBuilder` (`rust/sedona-raster/src/builder.rs`), written via `band_data_writer().append_value(&[u8])`. `append_value` always copies the slice into the builder's internal block buffer — even when the bytes already live in an Arrow `Buffer` that the output array could reference directly.

Arrow's `BinaryViewArray` is *designed* for sharing: multiple rows can carry `(offset, length, buffer_idx)` views into the same backing `Buffer`. We never exploit this, so every call site that hands the builder band bytes pays a full copy + allocation.

## Affected call sites

1. **`RasterBuilder::with_view` (InDb path).** A view into an existing band copies the source bytes into the output. A 100 MB band sliced 10× in a chain produces an output `StructArray` carrying ~1 GB of duplicated source bytes — the "lazy slice" framing is undermined the moment the source is materialised.
2. **`RS_EnsureLoaded` InDb passthrough.** The UDF rebuilds the whole output batch row-by-row; every unchanged InDb band is copied through via `Buffer::from_vec(band.data().to_vec())` + `append_value`. Untouched bands should be borrowed zero-copy (or skipped entirely).
3. **`RS_EnsureLoaded` freshly-loaded OutDb bytes.** The loader returns an owned `arrow_buffer::Buffer`, which `append_value` then copies *again* into the builder's block — a second copy on the hot byte path the loader worked to keep to one.

## Fix

Give `RasterBuilder` a band-data append that references an existing `arrow_buffer::Buffer` as a `BinaryViewArray` row without copying:

- Track a shared buffer table on the builder alongside the `BinaryViewBuilder`.
- Add an API (e.g. `append_band_data_buffer(buffer, offset, len)`) that records a view row against a shared buffer (dedup by `Buffer::ptr_eq`).
- At `finish()`, construct the `BinaryViewArray` via the low-level constructor, merging the `BinaryViewBuilder`'s own blocks with the shared buffer table.
- Migrate the three call sites above; `start_band`/`start_band_nd` keep the copying `append_value` for genuinely-owned-elsewhere bytes.

## Acceptance

- A chained `with_view` output array's total buffer memory equals the **union** of referenced source buffers, not the per-row sum.
- `RS_EnsureLoaded` over a batch of InDb bands performs zero band-byte copies for passed-through bands.
- A loaded OutDb band's bytes appear in the output backed by the loader's returned `Buffer` (one copy total, storage → loader `Buffer`).

james-willis:
Today in main, we do a new allocation when EnsureLoaded encounters an indb raster. This is slow.

This PR makes those cases zero copy by passing the pointer from the Binary View to the new Raster.

<details>

Today in `main`, `RS_EnsureLoaded` allocates a fresh copy of every band's pixel bytes — even when those bytes already live in an Arrow `Buffer`. For InDb bands it `to_vec()`s the source; for freshly-loaded OutDb bands it re-copies the loader's buffer into the output column. That copy is pure overhead.

This PR makes both paths zero-copy by sharing the existing `Buffer` into the output `BinaryViewArray` as a refcounted block instead of copying the bytes.

### What changed

- **`RasterBuilder` zero-copy API** — `append_band_data_buffer(buffer, offset, len)` attaches an existing `Buffer` as a shared view block (refcount bump, deduped by data pointer); `append_band_data_from(src, row)` passes a row of an existing `BinaryViewArray` through, sharing its backing block. Slices ≤ 12 bytes are stored inline (a block-referencing view that small is non-canonical and fails array validation on roundtrip).
- **`RS_EnsureLoaded` wiring** — InDb bands share the input row's backing buffer; OutDb bands share the loader's returned `Buffer` after the byte-count check.
- **`RasterStructArray` accessors** — `band_data_array()` / `band_data_row(raster, band)` to address a band's bytes for passthrough without re-navigating the struct.

### Notes

- Zero-copy shares *bytes*, not *views*: the rebuilt band carries an identity view, so passthrough is correct only for identity-view bands. A non-identity view is rejected loudly pending view-preserving passthrough (#897); all bands are identity today.
- Tests assert pointer equality between the output band's bytes and the source/loader buffer (proving sharing), plus dedup and inline-small-slice behavior.

Resolves #894.

</details>

### DIFF
--- a/rust/sedona-raster/src/builder.rs
+++ b/rust/sedona-raster/src/builder.rs
@@ -448,6 +448,12 @@ impl RasterBuilder {
     /// pointer, so handing the same backing buffer for many bands attaches it
     /// once and points every view at it.
     ///
+    /// Bytes shorter than the 12-byte inline threshold are stored inline (a
+    /// tiny copy) instead, since the `BinaryViewArray` layout requires `len <=
+    /// 12` views to be inline and a block-referencing view of that size is
+    /// non-canonical (it fails array validation on roundtrip). Band data is
+    /// realistically always larger, so this is the degenerate path.
+    ///
     /// Counts as the one data value for the current band (see [`finish_band`]).
     ///
     /// [`finish_band`]: Self::finish_band
@@ -457,6 +463,11 @@ impl RasterBuilder {
         offset: u32,
         len: u32,
     ) -> Result<(), ArrowError> {
+        if len <= 12 {
+            self.band_data
+                .append_value(&buffer.as_slice()[offset as usize..(offset + len) as usize]);
+            return Ok(());
+        }
         let block = match self.band_data_blocks.get(&(buffer.as_ptr() as usize)) {
             Some(&idx) => idx,
             None => {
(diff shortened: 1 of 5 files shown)


rust/sedona-raster/src/builder.rs
```diff
@@ -431,6 +439,75 @@ impl RasterBuilder {
...
+        let view = src.views()[row];
+        let len = view as u32;
+        if len <= 12 {
+            self.band_data.append_value(src.value(row));
+            Ok(())
+        } else {
+            let buffer_index = (view >> 64) as u32;
+            let offset = (view >> 96) as u32;
+            self.append_band_data_buffer(&src.data_buffers()[buffer_index as usize], offset, len)
+        }
```
paleolimbot: I believe `try_append_view()` will do this for you without hardcoding the struct layout?

paleolimbot (approved):
Thank you!

rust/sedona-raster/src/builder.rs
james-willis: No, it doesnt handle the inline case, attaching the block from the source, or getting the block/offset/len

rust/sedona-raster/src/builder.rs
```diff
@@ -431,6 +444,81 @@ impl RasterBuilder {
...
+        let view = src.views()[row];
+        let len = view as u32;
+        if len <= MAX_INLINE_VIEW_LEN {
+            self.band_data.append_value(src.value(row));
+            Ok(())
+        } else {
+            let buffer_index = (view >> 64) as u32;
+            let offset = (view >> 96) as u32;
+            self.append_band_data_buffer(&src.data_buffers()[buffer_index as usize], offset, len)
+        }
```
paleolimbot, line 519: Might be nice if this also used the ByteView but this works too so let's go with it 🙂

paleolimbot (approved):
Thank you!

PR #919, opened 2026-06-07
Track active spillers in SedonaFairSpillPool

### LINKED ISSUE #918: SedonaFairSpillPool can reject sort memory even when spill memory is available
We're seeing external sort failures where Sedona reports spillable memory is still available, but the allocation is rejected anyway.

The root cause appears to be that `SedonaFairSpillPool` splits the spill budget across all registered spill-capable consumers, including consumers that are idle and have reserved 0 bytes.

Minimal test case:

```rust
#[test]
fn test_idle_spillers_do_not_split_budget() {
    let pool: Arc<dyn MemoryPool> = Arc::new(SedonaFairSpillPool::new(100, 0.0));

    let active_consumer = MemoryConsumer::new(active).with_can_spill(true);
    let mut active = active_consumer.register(&pool);

    let _idle_reservations: Vec<_> = (0..128)
        .map(|i| {
            MemoryConsumer::new(format!(idle-{i}))
                .with_can_spill(true)
                .register(&pool)
        })
        .collect();

    active.try_grow(100).unwrap();
}
```

On current `main`, this fails with:

```text
Failed to allocate additional 100 bytes for active with 0 bytes already allocated - maximum available is 0 bytes.
Current unspillable memory usage: 0 bytes, spillable memory available: 100 bytes
```

Expected behavior: idle spill-capable consumers should not reduce the memory available to the active consumer.

Observed behavior: the active consumer effectively gets a 0-byte limit because the 100-byte pool is divided across 129 registered consumers, even though only one consumer is using memory.

This seems related to sort-heavy queries with many `ExternalSorter` consumers, where some sorters are idle but still reduce the budget for active sorters.

camden-lowrance:
Closes #918.

This changes `SedonaFairSpillPool` fairness from registered spill-capable consumers to active spillers.

Right now, a spill-capable consumer counts against the fair-share calculation as soon as it is registered, even if it has reserved 0 bytes. In sort-heavy plans, that can make an active sorter fail allocation while the pool still has spillable memory available.

After this change:
- idle spill-capable consumers do not reduce the active spill budget
- consumers are counted once they reserve spillable memory
- consumers stop being counted when their spillable reservation returns to zero
- fairness still applies among active spillers
- unspillable reserve behavior is unchanged

Verification:
- `cargo test -p sedona memory_pool --lib`

Draft for discussion: this changes fairness from registered spill-capable consumers to active spillers.

### DIFF
--- a/rust/sedona/src/memory_pool.rs
+++ b/rust/sedona/src/memory_pool.rs
@@ -54,14 +54,14 @@ pub struct SedonaFairSpillPool {
 
 #[derive(Debug)]
 struct FairSpillPoolState {
-    /// The number of consumers that can spill
-    num_spill: usize,
-
     /// The total amount of memory reserved that can be spilled
     spillable: usize,
 
     /// The total amount of memory reserved by consumers that cannot spill
     unspillable: usize,
+
+    /// Spillable memory reserved by each active spillable consumer
+    spillable_consumers: Vec<(usize, usize)>,
 }
 
 impl SedonaFairSpillPool {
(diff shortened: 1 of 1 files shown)


2010YOUY01:
> Right now, a spill-capable consumer counts against the fair-share calculation as soon as it is registered, even if it has reserved 0 bytes. In sort-heavy plans, that can make an active sorter fail allocation while the pool still has spillable memory available.

I think this is an intentional design decision, the assumption is idle consumers will be active shortly after.

> * idle spill-capable consumers do not reduce the active spill budget

This is an interesting idea, but I don't fully understand it now.

I suggest to demonstrate some end-to-end queries, that used to fail, but this new policy make them work.

camden-lowrance:
Thanks I appreciate the discussion, looking through our error traces this has only occured two times on separate datasets.  They process successfully on retry, so it seems to be a transient error.  I will try to pin down this with a reproducible example.

For some context, our production env is memory constrained, only 4gb of memory per kubernetes pod, but we routinely process data in the 10s of millions of rows of geoparquet.

PR #925, opened 2026-06-08
feat(python/sedonadb): add DataFrame.cross_join

jiayuasu:
Small follow-up to #908. Completes the join surface from #791.

## API

```python
df.cross_join(other)
```

- Returns the Cartesian product of `self` and `other`.
- Result row count is `len(self) * len(other)`.
- Both sides' columns are kept verbatim — matches PySpark's `crossJoin` shape. Disambiguate via `df.alias(...)` if column names collide.
- No `on=` / `how=` — those would be `join(...)`.

## Implementation

DataFusion's `DataFrame` doesn't expose a direct `cross_join` method (52.5); the builder lives on `LogicalPlanBuilder`. The Rust binding uses `DataFrame::into_parts` to split off the `SessionState`, runs `LogicalPlanBuilder::cross_join`, and reassembles with `DataFrame::new`.

| File | Change |
|---|---|
| `python/sedonadb/src/dataframe.rs` | New `InternalDataFrame::cross_join(right)`. Plan-builder path; no per-row work. |
| `python/sedonadb/python/sedonadb/dataframe.py` | New `DataFrame.cross_join(other)`. Validates `other` is a DataFrame; otherwise raises `TypeError`. |

## Test plan

7 tests in `tests/expr/test_dataframe_cross_join.py`:

- **Positive**: 2×2; multi-column on both sides; 1×N; empty×N (zero rows); aliased sides sharing a key name (row-count check, since `to_pandas` collapses qualifiers on duplicate names).
- **Error**: non-DataFrame `other` → `TypeError`.

Output assertions use exact `pd.testing.assert_frame_equal` after sorting where row order isn't unique.

Local: 7 unit + 25 doctests + `ruff check` + `cargo fmt --check` all clean. Re-ran the 21 join tests from #908 — no regressions.

### DIFF
--- a/python/sedonadb/python/sedonadb/dataframe.py
+++ b/python/sedonadb/python/sedonadb/dataframe.py
@@ -796,6 +796,44 @@ class DataFrame:
 
         return DataFrame(self._ctx, joined_impl.select(projection))
 
+    def cross_join(self, other: "DataFrame") -> "DataFrame":
+        """Cartesian product of two DataFrames.
+
+        Returns a DataFrame containing every pair of rows from `self`
+        and `other`; the row count is the product of the two input
+        row counts. Both sides' columns are kept verbatim —
+        disambiguate with `df.alias(...)` on either side if column
+        names collide.
+
+        Args:
+            other: The right-hand DataFrame.
+
+        Examples:
+
+            >>> sd = sedona.db.connect()
+            >>> left = sd.sql("SELECT * FROM (VALUES (1), (2)) AS t(x)")
+            >>> right = sd.sql("SELECT * FROM (VALUES ('a'), ('b')) AS t(y)")
+            >>> left.cross_join(right).sort("x", "y").show()
+            ┌───────┬──────┐
+            │   x   ┆   y  │
+            │ int64 ┆ utf8 │
+            ╞═══════╪══════╡
+            │     1 ┆ a    │
+            ├╌╌╌╌╌╌╌┼╌╌╌╌╌╌┤
+            │     1 ┆ b    │
+            ├╌╌╌╌╌╌╌┼╌╌╌╌╌╌┤
+            │     2 ┆ a    │
+            ├╌╌╌╌╌╌╌┼╌╌╌╌╌╌┤
+            │     2 ┆ b    │
+            └───────┴──────┘
+        """
+        if not isinstance(other, DataFrame):
+            raise TypeError(
+                f"cross_join() expects a DataFrame as the first argument, "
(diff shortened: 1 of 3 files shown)


python/sedonadb/python/sedonadb/dataframe.py
```diff
@@ -796,6 +796,43 @@ def join(
 
         return DataFrame(self._ctx, joined_impl.select(projection))
 
+    def cross_join(self, other: "DataFrame") -> "DataFrame":
+        """Cartesian product of two DataFrames.
+
+        Returns a DataFrame containing every pair of rows from `self`
+        and `other`. The result row count is `len(self) * len(other)`.
+        Both sides' columns are kept verbatim — disambiguate with
+        `df.alias(...)` on either side if column names collide.
```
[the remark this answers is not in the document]
jiayuasu: Good catch — fixed in d748394. The docstring now reads:

> Returns a DataFrame containing every pair of rows from `self` and `other`; the row count is the product of the two input row counts.

No `len(...)` reference, so no misleading suggestion of an API that would actually raise on a lazy frame.

python/sedonadb/tests/expr/test_dataframe_cross_join.py
```diff
@@ -0,0 +1,89 @@
...
+    left = con.create_data_frame(pd.DataFrame({"k": [1, 2]})).alias("l")
+    right = con.create_data_frame(pd.DataFrame({"k": [10, 20, 30]})).alias("r")
+    out = left.cross_join(right)
+    assert out.count() == 6
+
+
+def test_cross_join_returns_lazy_dataframe(con):
+    left = con.create_data_frame(pd.DataFrame({"x": [1]}))
+    right = con.create_data_frame(pd.DataFrame({"y": [1]}))
+    assert isinstance(left.cross_join(right), DataFrame)
```
paleolimbot: Unless there's bindings surface you're testing with these, probably just the two tests will do. They can live in the other join file I think since there's only two.

```suggestion
```

paleolimbot (approved):
Thank you!

python/sedonadb/tests/expr/test_dataframe_cross_join.py
jiayuasu: Done in 0e0e5e9 — trimmed to the two tests (`test_cross_join_basic`, `test_cross_join_non_dataframe_other_raises`) and folded them into `test_dataframe_join.py`. Deleted the dedicated `test_dataframe_cross_join.py` file.

PR #927, opened 2026-06-08
refactor(rust/sedona-raster,rust/sedona-schema): source_shape i64 end-to-end (in-memory + Arrow column)

james-willis:
Converts the raster band `source_shape` / visible-shape surface from `u64` to `i64` — both the in-memory representation and the Arrow `source_shape` column (`List<UInt64>` → `List<Int64>`).

### Why

The view/stride machinery (`ViewEntry`, byte strides) is already `i64` (signed, for negative-step / reverse views). Keeping `source_shape` as `u64` forces mixed signed/unsigned arithmetic at every shape↔stride boundary; making the shape `i64` end-to-end removes those casts and lets view composition use uniform signed math.

This is an **Arrow schema change** to the raster type (`source_shape` column element type `UInt64` → `Int64`), landed before the format stabilizes so we don't ship `u64` and break it later.

### DIFF
--- a/rust/sedona-raster-functions/src/rs_ensure_loaded.rs
+++ b/rust/sedona-raster-functions/src/rs_ensure_loaded.rs
@@ -133,14 +133,14 @@ fn lookup_loader(
 /// axis maps to the same source axis, full extent, unit step). An empty view
 /// is treated as identity. Used to detect whether a loader returned an
 /// already-resolved (identity) layout we can build directly.
-fn view_is_identity(view: &[ViewEntry], source_shape: &[u64]) -> bool {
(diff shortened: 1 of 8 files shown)


james-willis:
@paleolimbot What do you think of changing `RasterRef::width()/height() -> Result<u64>` to return signed ?

paleolimbot (approved):
There's also an issue sending u64 Arrow arrays to the JVM but I forget the details. Thank you!

james-willis:
> > What do you think of changing RasterRef::width()/height() -> Result to return signed ?
> 
> I think that is a good idea. There may be some UDFs that return u64s that should probably return i64s as well.

https://github.com/apache/sedona-db/pull/932

PR #928, opened 2026-06-09
feat(rust/sedona-raster-gdal): pass N-D rasters through the GDAL bridge via plane stacking

james-willis:
This allows us to use band stacking to perform warp/resample/reproject operations in GDAL on ND Arrays. 

I don't think this would work well for operations that change the dimensionality of non-xy dims but I'm also thinking we will do those natively rather than rely on GDAL to give us better control over memory allocation.

<details>

GDAL is 2-D-planar — a dataset is a flat stack of 2-D bands — so an N-D raster band (e.g. `[time, y, x]`) can't be handed to GDAL directly. This makes the GDAL bridge N-D-capable by **band stacking**: flatten each band's non-spatial axes into consecutive GDAL bands on the way in, and regroup them back into N-D bands on the way out.

### What changed

- **`raster_ref_to_gdal_mem`** — replaced the hard 2-D reject with "the trailing two dims must be a spatial `(y, x)` pair", and emits **one GDAL band per non-spatial plane** (band-major, then plane-major / C-order). Because the spatial axes are innermost, every plane is a contiguous sub-range of the band buffer, so the existing **zero-copy DATAPOINTER** still holds per plane (no copy). A plain 2-D band is just the single-plane case. The per-band nodata is expanded across that band's planes.
- **`GdalBandLayout` / `GdalBandPlan` + `GdalBandLayout::from_raster`** — the out-of-band record of the N-D structure GDAL is oblivious to (`dim_names`, `nonspatial_shape`, `plane_count`, type, nodata). It's derived from the *input* raster and is invariant under spatial-only ops (warp/reproject/resample), which preserve band count/order and change only the x/y extent.
- **`gdal_dataset_to_nd_raster` / `append_nd_from_dataset`** — the inverse: regroup a GDAL dataset's flat band list back into N-D bands, consuming `plane_count` consecutive bands per source band. The **spatial extent comes from the output dataset** (so warp/resample can change y/x); the **non-spatial structure comes from the layout**.

### Notes

- **Storage:** InDb bands only (OutDb still rejected). The bridge is the input/output primitive; no consuming op (warp/etc.) is wired up here — that's a follow-up. The round-trip (`raster → stacked MEM dataset → regrouped raster`) is the testable unit.
- A defensive guard rejects a band whose spatial extent doesn't match the raster grid, so the public `unsafe` bridge can't silently mis-stack a directly-constructed `RasterRef`.
- **Out of scope:** the mixed InDb/OutDb VRT path (`gdal_dataset_provider`, currently dead code) keeps its 2-D gate; and ops that change non-spatial dimensionality (per the description above) are intended to be native, not GDAL.

### Tests

Round-trip byte/shape/name identity for: a single N-D band; a mixed 2-D + N-D raster with heterogeneous plane counts and per-plane nodata; multiple non-spatial dims (C-order flatten); plus rejection when the spatial pair isn't the trailing two dims.

</details>

### DIFF
--- a/rust/sedona-raster-gdal/src/gdal_common.rs
+++ b/rust/sedona-raster-gdal/src/gdal_common.rs
@@ -235,9 +235,24 @@ pub unsafe fn raster_ref_to_gdal_mem<R: RasterRef + ?Sized>(
             );
         }
 
+        // The plane's 2-D extent must equal the MEM dataset's (and the raster's
+        // spatial grid); otherwise the per-plane byte slicing below would
+        // disagree with the GDAL band size and silently mis-stack the planes.
+        // `finish_raster` already enforces this for builder-made rasters, but
+        // re-check so the public bridge stays sound for any `RasterRef`.
+        let shape = band.shape();
+        if shape[ndim - 2] as usize != height || shape[ndim - 1] as usize != width {
+            return exec_err!(
+                "band spatial extent {}x{} does not match the raster grid \
+                 {width}x{height} (dim_names={dims:?})",
+                shape[ndim - 1],
+                shape[ndim - 2]
+            );
+        }
+
         if band.metadata().storage_type()? != StorageType::InDb {
             return Err(DataFusionError::NotImplemented(
-                "OutDb bands are not supported in raster_to_mem_dataset".to_string(),
+                "OutDb bands are not supported by raster_ref_to_gdal_mem".to_string(),
             ));
         }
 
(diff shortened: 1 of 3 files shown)


paleolimbot (approved):
Thank you! This seems reasonable and a useful way to squeeze our ND idioms into GDAL's 2D idioms.

At some point we should invest in some testing infrastructure (Python and Rust) to make the tests a bit more obvious. I think these tests look good but I also am not sure I'd be able to tell if they weren't.

PR #929, opened 2026-06-09
fix(rust/sedona-raster): classify 0-element bands as InDb; make is_indb required

james-willis:
This was an outstanding issue where a band of 0 bytes could never return true from is_indb. Probably not a common case but I wanted to make sure it works in the cases where it arises instead of failing the whole pipeline for one row.

<details>

The InDb/OutDb discriminator was `!data.is_empty()`. That conflated two different 0-byte cases:
- an **InDb** band with an empty visible region (e.g. a `time = 0` dim) — genuinely 0 bytes, but in-database, and
- an **OutDb** band — 0 bytes because the pixels live elsewhere.

So a legitimately-empty InDb band was misclassified as OutDb (and `nd_buffer()`/`as_contiguous()` would error on it), failing a whole pipeline over one such row.

### Fix

A `data` column is empty for an InDb band **only** when the band has zero visible pixels, and `Π(shape) == 0` detects exactly that — so the discriminator becomes total without any schema change:

```rust
fn is_indb(&self) -> bool {
    self.shape().iter().product::<u64>() == 0       // 0-element band → nothing to load → InDb
        || !self.data_array.value(self.band_row).is_empty()
}
```

- `Π(shape) == 0` → InDb (a 0-element band holds 0 bytes legitimately; there's nothing to fetch).
- `Π(shape) > 0` + empty data → OutDb; + non-empty data → InDb.

Uses the **visible** `shape()` rather than the raw source shape on purpose: `is_indb` answers "can reads be satisfied from RAM without a loader?", and an empty visible region needs no load. (Moot today — identity views mean visible == source — but it's the cleaner semantics and avoids a pointless fetch for a future sliced-to-empty OutDb band.)

### Also

Made `is_indb` a **required** trait method, dropping the old `nd_buffer`-based default. That default was a recursion footgun (`nd_buffer()` itself depends on `is_indb()`) that every real impl had to override anyway; a band's storage location is a primitive each impl answers directly.

### Not a schema change

This resolves the ambiguity locally — no `data`-column nullability flip and no on-disk format change (which the original issue had proposed). Test: a 0-element InDb band now classifies as InDb and its `storage_type` follows.

</details>

### DIFF
--- a/rust/sedona-raster/src/array.rs
+++ b/rust/sedona-raster/src/array.rs
@@ -108,7 +108,11 @@ impl<'a> BandRef for BandRefImpl<'a> {
     }
 
     fn is_indb(&self) -> bool {
-        !self.data_array.value(self.band_row).is_empty()
+        // A 0-element visible region (any visible dim is 0) holds no readable
+        // bytes — trivially fully in-RAM — so it's InDb, not the OutDb
+        // empty-`data` sentinel. Otherwise the discriminator is buffer presence.
+        self.shape().iter().product::<u64>() == 0
+            || !self.data_array.value(self.band_row).is_empty()
     }
 
     fn nd_buffer(&self) -> Result<NdBuffer<'_>, ArrowError> {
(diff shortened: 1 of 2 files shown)


paleolimbot (approved):
Thank you!

PR #932, opened 2026-06-09
refactor(rust/sedona-raster-functions): width/height and RS_Width/RS_Height return i64

james-willis:
Follow up to https://github.com/apache/sedona-db/pull/927

trying to eliminate a bunch of the casting between u64 and i64. This is something of a breaking change as it changes the return type of these RS functions

### Drive By
`main` was broken by a semantic conflict between #927 and #928. This PR fixes that issue with its second commit

<details>

`RS_Width` / `RS_Height` and the underlying `width()` / `height()` accessors returned `u64`. This switches them to `i64`, matching the signed `spatial_shape` (and band `shape`) from the source_shape work and the `Int64` Arrow type used for dimension sizes elsewhere — so values stop being round-tripped through `u64` at the boundary.

### What changed

- `RasterRef::width()` / `height()` now return `Result<i64, ArrowError>`. Since `spatial_shape()` is already `i64`, the value is returned directly; the old non-negative guard and `as u64` cast are gone. The only remaining error is an empty / too-short `spatial_shape`, which is an invariant violation rather than a "no value" state.
- `RasterMetadata`'s `width` / `height` fields, the `MetadataRef` trait, and the inherent accessors are all `i64`.
- `RS_Width` / `RS_Height` now produce an `Int64` column (`Int64Builder`).
- The construction surface follows so no casts are reintroduced: `RasterBuilder::start_raster_2d` parameters and the builder's `current_width` / `current_height`, plus the `VrtKey` GDAL cache key.
- Call sites updated to match: the `Display` impl, the envelope / convex-hull / spatial-predicate corner math, and the test helpers.

### Also repairs `main`

#927 (source_shape → `i64`) and #928 (GDAL plane stacking) each merged green against a `main` that lacked the other, so their combination left `sedona-raster-gdal` not compiling: the plane-stacking path still did `Vec<u64> = band.shape()…` and `product::<u64>()` over a now-`i64` `shape()`. This flips `GdalBandPlan::nonspatial_shape` to `Vec<i64>` and the regroup path to push `i64` extents, restoring the build.

</details>

### DIFF
--- a/rust/sedona-raster-functions/src/executor.rs
+++ b/rust/sedona-raster-functions/src/executor.rs
@@ -697,8 +697,8 @@ impl<'a, 'b> RasterExecutor<'a, 'b> {
 #[cfg(test)]
 mod tests {
     use super::*;
-    use arrow_array::builder::UInt64Builder;
-    use arrow_array::UInt64Array;
+    use arrow_array::builder::Int64Builder;
+    use arrow_array::Int64Array;
     use arrow_schema::Field;
     use sedona_raster::traits::RasterRef;
     use sedona_schema::crs::{deserialize_crs, lnglat};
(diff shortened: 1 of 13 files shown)


paleolimbot (approved):
Thanks! On green CI this looks good to me

PR #934, opened 2026-06-09
feat(rust/sedona-raster): add ViewEntries view-spec module

james-willis:
This PR isolates just the machinery for the lazy view support. A subsequent PR will contain the integration(s) into the raster type.

This is spun out from #813

<details>

Adds `view_entries.rs`, the per-axis view-spec layer for N-D raster bands. No consumers are wired up yet — the reader/builder integration is the follow-up — so there is no behavior change.

### `ViewEntry`

The per-axis spec `(source_axis, start, step, steps)`, moved here from `traits.rs` and re-exported as `crate::traits::ViewEntry` so existing import paths are unchanged. `step == 0` is broadcast, negative `step` is reverse iteration, and `steps == 0` is an empty axis.

### `ViewEntries`

A newtype over `Vec<ViewEntry>` that owns the view logic:

- **`validate(source_shape)`** — the load-bearing check: view length matches `source_shape`, each `source_axis` is in range and the axes form a permutation of `0..ndim`, `steps >= 0`, and the addressed extent `start + (steps - 1) * step` stays within `[0, source_shape[axis])`. Uses `checked_*` arithmetic so a pathological spec errors rather than wrapping.
- **`visible_shape()`** — the visible extent (`steps` per axis); this is what a band reports as its `shape()`.
- **`is_identity(source_shape)`** — whether the view is the canonical no-op, so it can be stored as a NULL view row.
- **`compose(next)`** — stacks a new view over an already-viewed band, translating `next`'s axes (which index the *visible* axes) back onto the source axes. This is what lets views nest without materializing.
- **`identity_for_shape`**, plus `len` / `is_empty` / `as_slice` / `iter` accessors.

### `view_entries!` macro

A terse constructor taking `start:stop` half-open ranges per axis: `view_entries![0:4, 1:5]` yields `source_axis = position`, `step = 1`, `steps = stop - start` for each axis. Convenience for call sites and tests.

### Scope

Pure spec logic — no I/O, no Arrow columns; the only allocation is the small `Vec<ViewEntry>` (ndim is a handful). Fully unit-tested in-module (validation rejections, composition, identity, visible-shape). It's the substrate the integration PR builds on.

</details>

### DIFF
--- a/rust/sedona-raster/src/traits.rs
+++ b/rust/sedona-raster/src/traits.rs
@@ -18,6 +18,8 @@
 use arrow_schema::ArrowError;
 use sedona_schema::raster::BandDataType;
 
+use crate::view_entries::ViewEntry;
+
 /// Recognized spatial dimension-name pairs, in band C-order: the slower-
 /// varying Y-like (row) axis first, the faster-varying X-like (column) axis
 /// second. A band whose trailing two `dim_names` match one of these pairs is
@@ -117,12 +119,6 @@ impl<'a> NdBuffer<'a> {
         }
     }
 }
-
-// `ViewEntry` (and the `ViewEntries` newtype with its validation /
-// composition / visible-shape machinery) lives in `view_entries.rs`.
-// Re-exported here so the `crate::traits::ViewEntry` import path keeps
-// working.
-pub use crate::view_entries::ViewEntry;
 
 /// Concrete raster metadata returned by `RasterRef::metadata()`.
 ///
--- a/rust/sedona-raster/src/traits.rs
+++ b/rust/sedona-raster/src/traits.rs
@@ -118,26 +118,11 @@ impl<'a> NdBuffer<'a> {
     }
 }
 
-/// One per-dimension entry of a band's logical view. Describes how a
-/// visible axis maps onto an axis of the underlying source buffer.
-///
-/// - `source_axis`: index into the band's `source_shape` that this visible
-///   axis reads from. Across a band's full view, `source_axis` values must
-///   form a permutation of `0..ndim` — axis-dropping and axis-introducing
-///   views are not supported today.
-/// - `start`: starting index along the source axis (in elements, not bytes).
-/// - `step`: stride between consecutive visible elements along the source
-///   axis. `step == 0` means broadcast (the same source element is
-///   exposed `steps` times); negative `step` means reverse iteration.
-/// - `steps`: number of visible elements along this axis. `steps == 0` is
-///   allowed (empty axis).
-#[derive(Debug, Clone, Copy, PartialEq, Eq)]
-pub struct ViewEntry {
-    pub source_axis: i64,
-    pub start: i64,
-    pub step: i64,
-    pub steps: i64,
-}
+// `ViewEntry` (and the `ViewEntries` newtype with its validation /
+// composition / visible-shape machinery) lives in `view_entries.rs`.
+// Re-exported here so the `crate::traits::ViewEntry` import path keeps
+// working.
+pub use crate::view_entries::ViewEntry;
 
 /// Concrete raster metadata returned by `RasterRef::metadata()`.
 ///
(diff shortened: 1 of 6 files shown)


rust/sedona-raster/src/traits.rs
```diff
@@ -118,26 +118,11 @@ impl<'a> NdBuffer<'a> {
...
-    pub source_axis: i64,
-    pub start: i64,
-    pub step: i64,
-    pub steps: i64,
-}
+// `ViewEntry` (and the `ViewEntries` newtype with its validation /
+// composition / visible-shape machinery) lives in `view_entries.rs`.
+// Re-exported here so the `crate::traits::ViewEntry` import path keeps
+// working.
+pub use crate::view_entries::ViewEntry;
```
paleolimbot: llms love to leave these pub uses around but I'd rather remove them (these are all internal crates)

rust/sedona-raster/src/view_entries.rs
```diff
@@ -0,0 +1,594 @@
...
+/// and a single place to attach helpers and tests.
+#[derive(Debug, Clone, PartialEq, Eq, Default)]
+pub struct ViewEntries(Vec<ViewEntry>);
+
+impl ViewEntries {
+    /// Wrap a pre-built vector of entries. The view is not validated
+    /// here — call [`Self::validate`] before relying on it.
+    pub fn new(inner: Vec<ViewEntry>) -> Self {
+        Self(inner)
+    }
```
paleolimbot, line 63: Should there be a `try_new()` that does both?

rust/sedona-raster/src/view_entries.rs
```diff
@@ -0,0 +1,594 @@
...
+    /// - When `steps > 0`: `start ∈ [0, source_shape[source_axis])`,
+    ///   and when `step != 0` the last addressed element
+    ///   `start + (steps - 1) * step` is also in that range.
+    pub fn validate(&self, source_shape: &[i64]) -> Result<(), ArrowError> {
+        let ndim = source_shape.len();
+        if self.0.len() != ndim {
+            return Err(ArrowError::InvalidArgumentError(format!(
+                "view length ({}) must equal source_shape length ({ndim})",
+                self.0.len()
+            )));
```
paleolimbot, line 150: Not for this PR, but we have probably reached the limit on how far we have pushed the ArrowError in this crate. We can either have a RasterError struct like we do for seonda-geometry, or we can have a sedona-error crate with a `datafusion` and/or `arrow` feature that handles the impl From to reduce the number of `map_err`s needed (they collapse to `?` when From is implemented).

paleolimbot (approved):
Thank you!

I am still a tiny bit confused as to why `source_axis` is part of the view entry, since it seems there's a bit of work spent validating it only appears once. Can it be dropped and the index of the ViewEntry in the Vec<ViewEntry> be used instead? (Or if not, can a comment get added explaining why that can't happen?)

james-willis:
> I am still a tiny bit confused as to why source_axis is part of the view entry

One functionality of this views feature is to allow bands to be permuted, e.g. `[x, y, z]` -> `[z, y, x]`. The indices of the entries in ViewEntries defines the target band ordering. the values of `source_axis` defines the source band location. 

**The index cannot be used to define the source axis because it is used to define the target axis.**

One example: GDAL-backed functions with band stacking require y then x are the last two dimensions. If the source raster does not have this behavior a permutation is neccessary.

rust/sedona-raster/src/view_entries.rs
james-willis: done

rust/sedona-raster/src/traits.rs
james-willis: removed the reexport all together. not sure if that is what you had in mind.

rust/sedona-raster/src/view_entries.rs
james-willis: added a task to my personal item tracker for this

PR #936, opened 2026-06-10
chore(ci): build, test, and publish sedonadb-zarr wheels

james-willis:
Build the wheels and run tests for sedona-zarr. This parallels the build for sedona db but is simpler because theres no native code dependencies here. 

I also refactored out a common action that is shared between sedona db and the zarr extension.


paleolimbot (approved):
Thank you for this!

PR #940, opened 2026-06-11
refactor(python/sedonadb)!: Refactor registration of extension components

paleolimbot:
Before we only had a single component that could be registered (UDFs), but now we have UDAFs, extension formats, and external Python packages that may have custom registration requirements and/or version constraints on SedonaDB.

This PR implements `sd.register()`, renamed from `sd.register_udf()`, and implements a generic extension in the form of any object implementing `__sedonadb_extension__(<the sd object>)`. The motivation is to make the `sedonadb_zarr` registration a little more ergonomic.

I also added the ability to register `ExternalFormatSpec` with a context, which enables them to be used in SQL (or generally from a DataFusion FileFormatFactory, which may help with a generic "read" function).

It is a breaking change because `register_udf()` was removed. It has only existed for one version for a somewhat obscure feature and we're early in 0.xes, so I think this is OK (but could be added back if disruptive).

The motivating example is:

```python
sd = sedona.db.connect()
sd.register(sedonadb_zarr)
```

...to reduce the number of register_x methods we need on the context.

### DIFF
--- a/python/sedonadb-zarr/python/sedonadb_zarr/__init__.py
+++ b/python/sedonadb-zarr/python/sedonadb_zarr/__init__.py
@@ -53,7 +53,7 @@ class ZarrExtension:
         if kwargs:
             raise ValueError("Registration options not supported for ZarrExtension")
 
-        # Regiser the Zarr() format as a FileFormatFactory for SQL support
+        # Register the Zarr() format as a FileFormatFactory for SQL support
         ctx.register(Zarr())
 
 
@@ -100,9 +100,7 @@ class Zarr(ExternalFormatSpec):
     def open_reader(self, args: Any) -> PyZarrChunkReader:
         uri = args.src.to_url()
         if uri is None:
-            raise ValueError(
-                "Zarr: could not resolve a URL from the source object"
-            )
+            raise ValueError("Zarr: could not resolve a URL from the source object")
         arrays = self._options.get("arrays")
         batch_size = args.batch_size if args.batch_size is not None else 8192
         return PyZarrChunkReader(uri, arrays, batch_size)
--- a/python/sedonadb-zarr/python/sedonadb_zarr/__init__.py
+++ b/python/sedonadb-zarr/python/sedonadb_zarr/__init__.py
@@ -18,11 +18,11 @@
 """Zarr support for SedonaDB.
 
 ```python
-import sedonadb
 import sedonadb_zarr
 
-con = sedonadb.connect()
-con.read_format(sedonadb_zarr.ZarrFormatSpec(), "file:///path/to/foo.zarr").show()
+sd = sedona.db.connect()
+sd.register(sedonadb_zarr)
+sd.read_format(sedonadb_zarr.Zarr(), "file:///path/to/foo.zarr").show()
 ```
 
 Importing `sedonadb_zarr` is opt-in — applications that don't import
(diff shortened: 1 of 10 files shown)


python/sedonadb-zarr/python/sedonadb_zarr/__init__.py
```diff
@@ -18,11 +18,11 @@
...
 
 ```python
-import sedonadb
 import sedonadb_zarr
 
-con = sedonadb.connect()
-con.read_format(sedonadb_zarr.ZarrFormatSpec(), "file:///path/to/foo.zarr").show()
+sd = sedona.db.connect()
+sd.register(sedonadb_zarr)
+sd.read_format(sedonadb_zarr.Zarr(), "file:///path/to/foo.zarr").show()
```
paleolimbot: @james-willis What do you think about this change?

- ZarrFormatSpec -> Zarr to make it look nicer
- `sd.register(sedonadb_zarr)` instead of `sedonadb_zarr.register(sd)`

python/sedonadb-zarr/python/sedonadb_zarr/__init__.py
```diff
@@ -31,23 +31,43 @@
...
+
+    Args:
+        ctx: A SedonaDB context (e.g., `sedona.db.connect()`)
+
+    Examples:
+        >>> import sedonadb_zarr
+        >>> sd = sedona.db.connect()
+        >>> sd.register(sedonadb_zarr)
+    """
+    ctx.register(Zarr())
```
paleolimbot: Registering zarr like this might enable `SELECT * FROM 'temp.zarr'` in SQL, but is mostly ceremony (zarr can trivially be a DataFusion `FileFormatFactory` so we let it be one to get the registration party started).

python/sedonadb-zarr/python/sedonadb_zarr/__init__.py
```diff
@@ -57,23 +77,23 @@ def __init__(self, options: Optional[Mapping[str, Any]] = None):
 
     @property
     def extension(self) -> str:
-        return ".zarr"
+        return "zarr"
```
paleolimbot, line 82: This is what in theory allows `SELECT * FROM 'stuff.zarr'` (without the dot)

python/sedonadb/python/sedonadb/context.py
```diff
@@ -498,13 +498,22 @@ def sql(
...
+        """Register an extension component
+
+        The following types of components are currently supported:
+
+        - Python UDFs annotated with arrow_aggregate_udf or arrow_udf
+        - An ExternalFormatSpec implementing a custom datasource type
+        - An object implementing __sedonadb_extension__(ctx, **kwargs), which
+          is called with this context and any keyword arguments passed.
+
+        The extension interface is experimental and may change.
```
paleolimbot, line 511: This is the main change

python/sedonadb/src/context.rs
```diff
@@ -257,53 +258,62 @@ impl InternalContext {
...
+            let spec = component
+                .call_method0("__sedonadb_external_format__")?
+                .extract::<PyExternalFormat>()?;
+            let state_ref = self.inner.ctx.state_ref();
+            let Some(mut writable) = state_ref.try_write() else {
+                return Err(PySedonaError::SedonaPython(
+                    "Can't get writable session state".to_string(),
+                ));
+            };
+
```
paleolimbot: This is an addition: Python `ExternalFormatSpec`s are trivially a DataFusion `FileFormatFactory`, so we can register them. This can in theory improve the consistency of SQL vs Python but is mostly to consolidate the things we can register here.

python/sedonadb/src/import_from.rs
```diff
@@ -69,13 +69,6 @@ pub fn import_ffi_table_provider(
     Ok(provider)
 }
 
-pub fn import_ffi_scalar_udf(obj: &Bound<PyAny>) -> Result<ScalarUDF, PySedonaError> {
-    let capsule = obj.getattr("__datafusion_scalar_udf__")?.call0()?;
-    let udf_ptr = check_pycapsule(&capsule, "datafusion_scalar_udf")? as *mut FFI_ScalarUDF;
-    let udf_impl: Arc<dyn ScalarUDFImpl> = unsafe { udf_ptr.as_ref().unwrap().into() };
-    Ok(ScalarUDF::new_from_shared_impl(udf_impl))
-}
-
```
paleolimbot, line 78: I dropped support for `__datafusion_scalar_udf__` here because for it to work DataFusion major versions have to align and there is no great way to enforce that. The failure mode is a crash and so I am not sure I would like to support this until we can improve that failure mode.

python/sedonadb-zarr/python/sedonadb_zarr/__init__.py
james-willis: I think its probably better. no strong opinion.

The only thing about the old approach that I dont think the new approach allows is the old registration allowed arbitrary code execution which was IMO a good feature to maintain

python/sedonadb-zarr/python/sedonadb_zarr/__init__.py
james-willis: This is good. I want this to work eventually

python/sedonadb-zarr/python/sedonadb_zarr/__init__.py
james-willis: I guess to me it is a little weird to pass the module instead of the class

james-willis (approved):

python/sedonadb-zarr/python/sedonadb_zarr/__init__.py
paleolimbot: I tried and unfortunately DataFusion attempts to directory list. Battle for another day 🙂

python/sedonadb-zarr/python/sedonadb_zarr/__init__.py
paleolimbot: I made the "Extension" (== the format plus the future loader, which will also need to be registered) a class. I think we still get arbitrary code execution (you can do whatever you wan with the context or otherwise in `__sedonadb_extension__(ctx)`.

james-willis (approved):

PR #943, opened 2026-06-11
fix(rust/sedona-raster-zarr): discover child arrays via consolidated metadata

### LINKED ISSUE #941: Basic Zarr read example errors
I think this is an issue with .show() (i.e., sd_format()), but in any case a raw read example seems not to work (could be my error too!)

```python
import sedona.db
import sedonadb_zarr
import zarr
import numpy as np

root = zarr.open_group("temp.zarr", mode="w")
arr = root.create_array(
    "temperature",
    shape=(2, 2),
    chunks=(1, 2),
    dtype="uint8",
    dimension_names=["y", "x"],
)
arr[:] = np.array([[10, 11], [20, 21]], dtype=np.uint8)


sd = sedona.db.connect()
sd.options.interactive = True

sd.read_format(sedonadb_zarr.ZarrFormatSpec(), "temp.zarr")
#> SedonaError: SedonaDB internal error: Extension type sedona.raster has unexpected storage type: Struct("crs": Utf8View, "transform": non-null List(non-null Float64), "spatial_dims": non-null List(non-null Utf8View), "spatial_shape": non-null List(non-null Int64), "bands": List(non-null Struct("name": Utf8, "dim_names": non-null List(non-null Utf8), "source_shape": non-null List(non-null UInt64), "data_type": non-null UInt32, "nodata": Binary, "view": List(non-null Struct("source_axis": non-null Int64, "start": non-null Int64, "step": non-null Int64, "steps": non-null Int64)), "outdb_uri": Utf8, "outdb_format": Utf8View, "data": non-null BinaryView), field: 'band')).
```

Another example we'd discussed offline I can't get to work

```python
import sedona.db
import sedonadb_zarr

sd = sedona.db.connect()

url = "https://atlantis-vis-o.s3-ext.jc.rl.ac.uk/hurricanes/era5/florence"
cube = sd.read_format(sedonadb_zarr.ZarrFormatSpec(), url)
#> SedonaError: ValueError: External error: External error: SedonaDB internal error: failed to list child nodes of Zarr group at https://atlantis-vis-o.s3-ext.jc.rl.ac.uk/hurricanes/era5/florence: Generic HTTP error: Error performing PROPFIND https://atlantis-vis-o.s3-ext.jc.rl.ac.uk/hurricanes/era5/florence in 159.455584ms - Server returned non-2xx status code: 405 Method Not Allowed: .
```

james-willis:
## Summary

Zarr group child-array discovery hand-rolled a `storage.list_dir()` call, which bypassed zarrs's built-in consolidated-metadata support entirely. On backends that don't support directory listing — plain HTTPS / S3 behind `HttpStore`, which answer `PROPFIND` with `405 Method Not Allowed` — discovery always failed, even when the group's `zarr.json` carried an inline `consolidated_metadata` block naming every child array (the common case for cloud-hosted v3 groups). This is (kind of) #941.

## Changes

Discovery now routes through zarrs's `Group::async_children`, which:

1. **Uses consolidated metadata when present** — child nodes are populated from the inline `consolidated_metadata` block with no per-node storage reads and no listing. It only falls back to listing the store when there is no consolidated metadata (inevitable) and no array option. 
3. **Errors clearly when neither works** — a store with no consolidated metadata that also can't list now yields a message pointing at the `arrays` option, instead of a raw store error.

Each array is built from its already-parsed metadata via `Array::new_with_metadata`, so a single malformed sibling array (e.g. an xarray-style coord variable with a dtype zarrs can't parse) is still skipped rather than poisoning the whole group — preserving the prior permissive behavior.

Fixes #941

### DIFF
--- a/rust/sedona-raster-zarr/tests/cloud.rs
+++ b/rust/sedona-raster-zarr/tests/cloud.rs
@@ -129,3 +129,30 @@ async fn https_zarr_smoke() {
     let rows = count_rows(reader);
     assert!(rows > 0, "expected at least one chunk row from {uri}");
 }
+
+/// Consolidated-metadata discovery over plain HTTPS, with NO `arrays`
+/// filter — the exact shape of issue #941. The ERA5 Hurricane Florence
+/// group (public, anonymous, UK JASMIN/CEDA) is a Zarr v3 group whose
+/// `zarr.json` carries an inline `consolidated_metadata` block, so child
+/// arrays are discovered from that block without ever listing the store.
+/// The host answers directory listing (`PROPFIND`) with 405, so this read
+/// would fail if discovery fell back to listing.
+#[tokio::test]
+#[ignore]
+async fn https_consolidated_metadata_discovery_no_filter() {
+    let uri = "https://atlantis-vis-o.s3-ext.jc.rl.ac.uk/hurricanes/era5/florence";
+    let authority = "https://atlantis-vis-o.s3-ext.jc.rl.ac.uk";
+    let store: Arc<dyn ObjectStore> = Arc::new(
+        HttpBuilder::new()
+            .with_url(authority)
+            .build()
+            .expect("build HttpStore"),
+    );
+    let storage = open_storage_from_uri(uri, store).expect("open_storage_from_uri");
+    // No arrays filter: discovery must come from consolidated metadata.
+    let reader = ZarrChunkReader::try_new(storage, uri, None, 1024)
+        .await
+        .expect("ZarrChunkReader::try_new against Florence over https:// (consolidated metadata)");
+    let rows = count_rows(reader);
+    assert!(rows > 0, "expected at least one chunk row from {uri}");
+}
--- a/rust/sedona-raster-zarr/src/loader.rs
+++ b/rust/sedona-raster-zarr/src/loader.rs
@@ -37,9 +37,8 @@ use zarrs::array::Array;
 #[cfg(test)]
 use zarrs::array::ArrayBytes;
 use zarrs::group::Group;
-use zarrs::storage::{
-    AsyncReadableListableStorage, AsyncReadableListableStorageTraits, StorePrefix,
-};
+use zarrs::node::NodeMetadata;
+use zarrs::storage::{AsyncReadableListableStorage, AsyncReadableListableStorageTraits};
 
 use crate::dtype::zarr_to_band_data_type;
 use crate::geozarr::GroupGeoMetadata;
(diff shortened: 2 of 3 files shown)


rust/sedona-raster-zarr/tests/cloud.rs
```diff
@@ -129,3 +129,30 @@ async fn https_zarr_smoke() {
...
+/// Consolidated-metadata discovery over plain HTTPS, with NO `arrays`
+/// filter — the exact shape of issue #941. The ERA5 Hurricane Florence
+/// group (public, anonymous, UK JASMIN/CEDA) is a Zarr v3 group whose
+/// `zarr.json` carries an inline `consolidated_metadata` block, so child
+/// arrays are discovered from that block without ever listing the store.
+/// The host answers directory listing (`PROPFIND`) with 405, so this read
+/// would fail if discovery fell back to listing.
+#[tokio::test]
+#[ignore]
+async fn https_consolidated_metadata_discovery_no_filter() {
```
paleolimbot, line 142: I would rather remove the test than `#[ignore]` it (since you also test this below). We could add a test zarr to sedona-testing that does this and use its github http url if it's an important case to cover.

paleolimbot (approved):
Thank you!

rust/sedona-raster-zarr/tests/cloud.rs
james-willis: This whole file is ignored tests because I dont like unit tests that rely on the external network; theyre flakey.

but its nice for developers to have these when they need a regression test. 

The tests below are local file tests so they dont catch the https behavior. A good https test would have caught this earlier.

paleolimbot: I didn't catch those...we have plenty of tests that rely on an external network. We can add to sedona-testing to make sure the fixtures are mostly available, or we could add "slow tests" and a CI job that runs them less frequently. Tests that don't run can live in GitHub (you can link to the issue in a comment).

james-willis: ok maybe just remove the ignore for now in that case. probably don't need to move them right now if we dont mind external network requesting integ tests. LMK your preference

paleolimbot: Let's remove the ignores for now and reevaluate if they become flaky. Adding some zarr test data to sedona-testing would be beneficial anyway as a route to fix that (you can test via the submodule or test via the http urls). A quick way to reduce the number or requests would be to set some environment variable like SEDONADB_RUN_SLOW_TESTS only on `main` (i.e., don't run them on PRs).

paleolimbot: I added a follow up so I can use this to test the actual byte loading 🙂

PR #944, opened 2026-06-12
feat(rust/sedona-raster-functions): add RS_IsEmpty

james-willis:
## Summary

Adds `RS_IsEmpty(raster) -> boolean`, true when a raster has **zero visible volume** — some dimension has length 0, so it holds no cells.

## Example Use case

Filtering the zero-length slabs that an absolute-position slice emits for non-matching chunks, e.g.:

```sql
SELECT plane FROM (SELECT RS_Slice_Abs_Pos(raster, 'time', 3) AS plane FROM cube)
WHERE NOT RS_IsEmpty(plane)
```

In this case `RS_Slice_Abs_Pos` is an imagined function which slices by the index in the absolute grid, not just the individual chunk.

It's also generally useful for discarding any empty/zero-volume raster.

### DIFF
--- a/rust/sedona-raster-functions/src/rs_isempty.rs
+++ b/rust/sedona-raster-functions/src/rs_isempty.rs
@@ -1,0 +1,149 @@
+// Licensed to the Apache Software Foundation (ASF) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership.  The ASF licenses this file
+// to you under the Apache License, Version 2.0 (the
+// "License"); you may not use this file except in compliance
+// with the License.  You may obtain a copy of the License at
+//
+//   http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+use std::{sync::Arc, vec};
+
+use crate::executor::RasterExecutor;
+use arrow_array::builder::BooleanBuilder;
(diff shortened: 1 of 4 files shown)


rust/sedona-raster-functions/src/rs_isempty.rs
```diff
@@ -0,0 +1,149 @@
...
+fn raster_is_empty(raster: &impl RasterRef) -> Result<bool> {
+    let bands = raster.bands();
+    if bands.is_empty() {
+        return Ok(true);
+    }
+    // Band numbers are 1-based.
+    for i in 1..=bands.len() {
+        let band = bands
+            .band(i)
+            .map_err(|e| datafusion_common::DataFusionError::External(Box::new(e)))?;
```
paleolimbot: If `band()` returns an ArrowError this probably works:

```suggestion
        let band = bands.band(i)?;
```

rust/sedona-raster-functions/src/rs_isempty.rs
```diff
@@ -0,0 +1,149 @@
...
+        if band.shape().iter().product::<i64>() == 0 {
+            return Ok(true);
+        }
```
paleolimbot, line 92: Just double checking that shape here refers to the viewed shape (not the source shape)

paleolimbot (approved):
Thank you!

rust/sedona-raster-functions/src/rs_isempty.rs
james-willis: that was the intent (I made a similar change yesterday about this). I will double check though!

james-willis: yes confirmed

rust/sedona-raster-functions/src/rs_isempty.rs
james-willis: done

PR #945, opened 2026-06-12
feat(rust/sedona-testing): add ergonomic raster function test harness

james-willis:


rust/sedona-testing/src/raster_spec.rs
```diff
@@ -0,0 +1,688 @@
...
+        self.crs = crs.map(|c| c.to_string());
+        self
+    }
+
+    /// Override the GDAL geotransform
+    /// `[origin_x, scale_x, skew_x, origin_y, skew_y, scale_y]`.
+    pub fn transform(mut self, transform: [f64; 6]) -> Self {
+        self.transform = transform;
+        self
+    }
```
paleolimbot, line 226: No need to do this here, but Wwhen we get to some more spatial interactions, setting this from a BoundingBox might also be nice

python/sedonadb/tests/functions/conftest.py
```diff
@@ -0,0 +1,57 @@
...
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+import pytest
+import sedonadb
+
+
+@pytest.fixture
+def zarr_group(tmp_path):
```
paleolimbot: It would be great to find a way to have an ND fixture without the zarr dependency (for the sedonadb package...for sedonadb-zarr this will be great)

paleolimbot (commented):
This is great, thanks for working on it! Feel free to ignore any of this if you're still working through changes

python/sedonadb/tests/functions/conftest.py
james-willis: yeah this is just Fable's first go, I havent even look at this code yet. probably will just use numpy for creating rasters

rust/sedona-testing/src/raster_spec.rs
james-willis: I'm adding some logics this week to build transforms from bboxes to support `spatial:bbox` attributes as a way to georeference datasets.

paleolimbot (approved):
Thank you!

PR #946, opened 2026-06-12
feat(python/sedonadb): add pandas-style Series (df["x"] -> Series)

jiayuasu:
First step of the pandas `Series` surface (the Series/Scalar phase of #791). `df["x"]` and `df.x` now return a `Series` instead of a bare `Expr`.

## What changes

`Series` **subclasses `Expr`**, so it is accepted unchanged everywhere an `Expr` already is — `select`, `filter`, `sort`, `agg`, `group_by`, `join` all keep working with no signature changes (they gate on `isinstance(.., Expr)` and unwrap `._impl`, both still true). The difference is that operators and transforms are overridden to return `Series` rather than `Expr`, so element-wise work chains naturally:

```python
df["x"].fillna(0).clip(upper=5).astype(pa.int64())
df.filter(df["x"].between(2, 4))
df.select(z=(df["x"] + df["y"]).alias("z"))
```

## API in this PR

pandas-flavored element-wise methods (all return `Series`):

- `astype(target)` — pandas spelling of `cast`
- `between(low, high)` — inclusive
- `fillna(value)` — via `coalesce`
- `clip(lower=None, upper=None)` — via `greatest` / `least`

Inherited `Expr` transforms (`alias`, `cast`, `is_null`, `is_not_null`, `isin`, `negate`) and all operators are re-wrapped to return `Series`. `asc`/`desc` still return `SortExpr`. `__bool__`/`__len__` raise with `Series`-specific guidance; `Series` stays unhashable (inherited `__hash__ = None`). Repr is `Series(<expr>)`.

`Series` lives at `sedonadb.series.Series` — consistent with `Expr` at `sedonadb.expr` (neither is a top-level export yet).

## Design note

The alternative was composition (`Series` *wraps* an `Expr`). Subclassing was chosen to avoid touching the 8 integration points and to reuse the operator machinery; the tradeoff is that `Series` inherits a few non-pandas `Expr` methods (`asc`/`desc`/`cast`/`is_null`). Happy to revisit if you'd prefer composition.

## Deferred to follow-up sub-PRs

- Parent-`DataFrame` link + `Scalar` + lazy reductions (`mean`/`sum`/`min`/`max`/`std`/`count`/`nunique`, auto-materializing on Python coercion).
- `isna`/`notna` with NaN-aware semantics (per the #791 decision that `isna` matches NULL *and* NaN — wants deliberate handling, `isnan` is available).
- `unique` / `value_counts`.
- `.str.*` / `.dt.*` accessors.

## Tests

New `tests/expr/test_series.py`: operators/transforms return `Series`; `astype`/`between`/`fillna`/`clip`; chaining; `select`/`filter`/`agg` accept `Series`; `bool`/`len`/`hash` guards; repr. `tests/expr/test_dataframe_getitem.py` updated to assert the `Series` contract.

Local: 207 expr tests + 63 doctests + `ruff` all clean. No regressions (the broader suite's PostGIS/DuckDB failures are pre-existing/environmental, identical on main).

### DIFF
--- a/python/sedonadb/python/sedonadb/dataframe.py
+++ b/python/sedonadb/python/sedonadb/dataframe.py
@@ -23,6 +23,7 @@ from sedonadb.expr import Expr, SortExpr
 from sedonadb.expr import Literal as _SedonaLit
 from sedonadb.expr import col as _col
 from sedonadb.expr.expression import _to_expr
+from sedonadb.series import Series
 from sedonadb.utility import sedona  # noqa: F401
 
 if TYPE_CHECKING:
@@ -200,12 +201,13 @@ class DataFrame:
         """
         return DataFrame(self._ctx, self._impl.alias(name))
 
-    def __getitem__(self, key: Union[str, int]) -> Expr:
+    def __getitem__(self, key: Union[str, int]) -> "Series":
         """Reference a single column by name or position.
 
-        Returns an `Expr` referencing the requested column. The
-        return-type is always `Expr` (no `DataFrame ⏐ Expr` union) so that
-        IDEs and type-aware tools can resolve `df["x"].<method>` cleanly.
+        Returns a `Series` (a pandas-flavored column) referencing the
+        requested column. The return-type is always `Series` (no
+        `DataFrame ⏐ Series` union) so that IDEs and type-aware tools can
+        resolve `df["x"].<method>` cleanly.
 
         For row filtering use `df.filter(predicate)`. For multi-column
         projection use `df.select(*cols)`.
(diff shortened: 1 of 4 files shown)


paleolimbot (commented):
This is a separate interface that should go in a separate package (e.g., `sedonadb-pandas`). Our `DataFrame` interface is a tool that can be used to expose compatibility layers for a few engines if we want, not the pandas compatibility layer itself.

```python
import sedona.db.geopandas as gpd

df = gpd.read_file(...)
```

My personal vote would be to put this in a separate repository or to put it in apache/sedona (with the other GeoPandas code).

jiayuasu:
Closing for now — the pandas `Series` surface needs a decision on whether it lives on the core `DataFrame` or in a separate package (`sedonadb-pandas` / apache/sedona), per @paleolimbot's review. Parking until that's settled rather than building more of P3 on the contested `df["x"] -> Series` core change. The code stays on the `feature/series-core` branch and this can be reopened once the direction is decided.

PR #948, opened 2026-06-12
chore: add raster label to GitHub repository topics

james-willis:
I think this'll be helpful as we build out our raster backend


paleolimbot (approved):
Thanks!

PR #950, opened 2026-06-12
fix(rust/sedona-raster-zarr): parse spatial:transform as affine order, add proj:code and spatial:dimensions

james-willis:
2 changes here:

1. fix a bug with how we interpret the affine transformation. geozarr does not use GDAL norm, so we need to reorder.
2. geozarr has changes their field names a bit so we now support the proposal + the updated field names


james-willis:
CI failure is unrelated, raised #951 to fix.

paleolimbot (approved):
Thank you!

PR #951, opened 2026-06-12
chore(examples): pin transitive deps via committed Cargo.lock

james-willis:
## Problem

The example crates under `examples/sedonadb-rust` and `examples/sedonadb-rust-pointcloud` ship a `Cargo.toml` but no committed `Cargo.lock` (it was gitignored). On every CI run the `rust (sedonadb-rust)` and `rust (sedonadb-rust-pointcloud)` jobs re-resolve all transitive dependencies from scratch, so the examples drift away from the versions the main workspace lock pins and validates.

That fresh resolution recently selected `time 0.3.48` alongside `aws-smithy-types 1.5.0`, which fails to compile:

```
error[E0119]: conflicting implementations of trait `From<...HourBase>` ...
  = note: conflicting implementation in crate `time`
```

This is the known `time`-crate format-description macro conflict: `time 0.3.47` builds cleanly, while `0.3.48` reintroduces the clash with `aws-smithy-types`' generic `From` impl.

## Fix

- Commit a `Cargo.lock` for each example so CI builds against pinned, known-good versions instead of re-resolving from scratch.
- Pin `time` to `0.3.47` in both lockfiles. Each now resolves `time 0.3.47` / `aws-smithy-types 1.5.0` and passes `cargo check` locally.
- Update `examples/.gitignore` to stop ignoring these two specific lockfiles so they can be tracked (the blanket `Cargo.lock` ignore is what allowed the drift). Without this the lockfiles cannot be committed or honored by CI.


examples/.gitignore
```diff
@@ -16,3 +16,8 @@
 # under the License.
 
 Cargo.lock
+
+# Example projects commit their lockfile so CI builds against pinned,
+# known-good transitive deps instead of re-resolving from scratch.
+!sedonadb-rust/Cargo.lock
+!sedonadb-rust-pointcloud/Cargo.lock
```
paleolimbot, line 23: Would removing Cargo.lock work too?

paleolimbot (approved):
Thank you!

examples/.gitignore
james-willis: commiting lock files everywhere would prevent issues like this because CI on the PR would tend to catch such issues.

But more PRs would merge conflict with each other. 

I dont know enough about rust to know what the common practice with lock files is these days.

PR #953, opened 2026-06-12
feat(rust/sedona-schema): support WKT1/WKT2 CRS strings in deserialize_crs

james-willis:
This PR aims to support WKT1/2 through all uses of the crs.rs functionality. This includes ST_Transform, ST_CRS, and ST_SetCRS. It rebases on the merged CRS cleanup (#962) and builds on that PR's "preserve the definition" behavior. A couple things worth pointing out that are somewhat opinionated or potentially controversial:

- **WKT CRS Equality:** We define equality here first through authority comparison if available. But if not available on both sides, we fall back to strict string equality so differences in representation (json vs wkt, wkt1 vs wkt2) and differences in the absolute string (whitespace, field ordering) will indicate inequality, arguably erroneously.
- **WKT detection vs. opaque fallback:** WKT is currently gated behind a keyword check (`looks_like_wkt`), keeping the else branch as PROJJSON so `deserialize_crs` still errors on unrecognized input. The alternative — drop the keyword list and treat any non-PROJJSON string as an opaque, PROJ-consumable definition (also accepting proj4 strings and any WKT2 keyword the list misses) — is simpler but makes `deserialize_crs` lenient (defers validation to PROJ, PostGIS-style) and would rename `WktCrs`. I lean toward the opaque fallback, but want a deliberate decision since `deserialize_crs` is core with many callers.

**I am happy to revise these behaviors based on what the group thinks!** The equality one specifically I'm not happy with, but I didn't want to bring proj into sedona-schema.


rust/sedona-functions/src/st_setsrid.rs
```diff
@@ -952,6 +952,52 @@ mod test {
...
+        .unwrap();
+        let SedonaType::Wkb(_, Some(crs)) = &return_type else {
+            panic!("expected a Wkb type with a CRS, got {return_type:?}");
+        };
+        assert_eq!(
+            crs.to_authority_code().unwrap().as_deref(),
+            Some("EPSG:3857")
+        );
+        assert_eq!(crs.srid().unwrap(), Some(3857));
+        assert_eq!(crs.to_crs_string(), "EPSG:3857");
```
paleolimbot: I don't think this should be true...`to_crs_string()` should give the original string back verbatim.

rust/sedona-functions/src/st_srid.rs
```diff
@@ -177,7 +177,14 @@ impl SedonaScalarKernel for StCrs {
...
             SedonaType::Wkb(_, Some(crs)) | SedonaType::WkbView(_, Some(crs)) => {
-                Some(crs.to_authority_code()?.unwrap_or_else(|| crs.to_json()))
+                // Prefer the authority code; otherwise the unescaped CRS string
+                // (`to_crs_string`, not `to_json`) so the result round-trips
+                // back through ST_SetCRS / deserialize_crs — e.g. a raw WKT
+                // rather than a JSON-quoted one.
+                Some(
+                    crs.to_authority_code()?
+                        .unwrap_or_else(|| crs.to_crs_string()),
+                )
```
paleolimbot: I don't think we should be canonicalizing on the way in...this looses information.

rust/sedona-functions/src/st_srid.rs
```diff
@@ -421,6 +472,35 @@ mod test {
...
+        const WKT_3857: &str = r#"PROJCS["WGS 84 / Pseudo-Mercator",GEOGCS["WGS 84",DATUM["WGS_1984",SPHEROID["WGS 84",6378137,298.257223563,AUTHORITY["EPSG","7030"]],AUTHORITY["EPSG","6326"]],AUTHORITY["EPSG","4326"]],PROJECTION["Mercator_1SP"],AUTHORITY["EPSG","3857"]]"#;
+        const WKT_LCC_NO_AUTHORITY: &str = r#"PROJCS["Custom LCC",GEOGCS["WGS 84",DATUM["WGS_1984",SPHEROID["WGS 84",6378137,298.257223563]]],PROJECTION["Lambert_Conformal_Conic_2SP"],PARAMETER["standard_parallel_1",33],PARAMETER["standard_parallel_2",45],PARAMETER["latitude_of_origin",39],PARAMETER["central_meridian",-96],UNIT["metre",1]]"#;
+
+        let udf: ScalarUDF = st_crs_udf().into();
+
+        // WKT with an EPSG authority -> ST_CRS returns the authority code.
+        let crs = deserialize_crs(WKT_3857).unwrap();
+        let tester = ScalarUdfTester::new(udf.clone(), vec![SedonaType::Wkb(Edges::Planar, crs)]);
+        let result = tester.invoke_scalar("POINT (0 1)").unwrap();
+        tester.assert_scalar_result_equals(result, "EPSG:3857");
```
paleolimbot: This is definitely consistent with previous behaviour, although I don't actually like the previous behaviour (I think it's confusing that the CRS you'd get by inspecting the type is different than the CRS you'd get from calling `ST_Crs()`.

rust/sedona-raster-functions/src/rs_setsrid.rs
```diff
@@ -505,6 +505,33 @@ mod tests {
...
+
+        // A WKT carrying an EPSG authority canonicalizes to that code via normalize().
+        let rasters = generate_test_rasters(1, None).unwrap();
+        let result = tester
+            .invoke_array_scalar(Arc::new(rasters), WKT_3857)
+            .unwrap();
+        let result_struct = result.as_any().downcast_ref::<StructArray>().unwrap();
+        let raster = RasterStructArray::new(result_struct).get(0).unwrap();
+        assert_eq!(raster.crs(), Some("EPSG:3857"));
+
```
paleolimbot: For rasters I'm not sure...we always store rasters as an item-level crs where the compactness of authority:code matters (although with the string view representation and the optimization that was just added to this file, there shouldn't be a memory or performance issue). On the other hand, collapsing to authority:code on the way in looses information. Probably authority:code is fine until somebody notices.

rust/sedona-schema/src/crs.rs
```diff
@@ -682,6 +698,125 @@ impl CoordinateReferenceSystem for ProjJSON {
...
+
+/// A CRS given as a WKT1/WKT2 string.
+///
+/// The WKT is preserved verbatim (`to_crs_string`) so it can be handed to PROJ
+/// for transforms, while an `authority:code` is extracted once at construction
+/// for equality, SRID, and geographic-parameter lookups.
+#[derive(Debug)]
+struct Wkt {
+    wkt: String,
+    authority: Option<String>,
```
paleolimbot: There are a few places here that just say "authority" which should probably be named "authority_code" to minimize confusion (unless you are only caching the authority)

rust/sedona-schema/src/crs.rs
```diff
@@ -682,6 +698,125 @@ impl CoordinateReferenceSystem for ProjJSON {
...
+    })
+}
+
+/// The CRS-level `AUTHORITY["auth","code"]` (WKT1) or `ID["auth",code]` (WKT2)
+/// tag. The `\b` keeps `ID` from matching inside `ELLIPSOID[` / `GEOID[`; the
+/// code is quoted in WKT1 and bare in WKT2, so the second group accepts either.
+static WKT_AUTHORITY_RE: LazyLock<Regex> = LazyLock::new(|| {
+    Regex::new(r#"(?i)\b(?:AUTHORITY|ID)\s*[\[(]\s*"([^"]+)"\s*,\s*"?([^"\],)]+)"?"#)
+        .expect("valid WKT authority regex")
+});
```
paleolimbot: This is a bit of a hack and I don't think it's too much extra work to do a bit better. WKT is very JSONish

```rust
enum WktNode<'a> {
   Unquoted(&'a str), // usually a number but you don't have to parse it; starts with `[0-9.-]`
   Quoted(&'a str), // a string, can be serialized with a JSON parser; starts with `"`
   Tagged(&'a str, Vec<Box<WktNode<'a>>>) // TAG[<nodes, comma separated>]; starts with anything else but must contain a `[]`.
}
```

Feel free to open a follow up ticket for that.

rust/sedona-schema/src/crs.rs
```diff
@@ -682,6 +698,125 @@ impl CoordinateReferenceSystem for ProjJSON {
...
+    Some(format!("{}:{}", authority.to_uppercase(), code))
+}
+
+/// A CRS given as a WKT1/WKT2 string.
+///
+/// The WKT is preserved verbatim (`to_crs_string`) so it can be handed to PROJ
+/// for transforms, while an `authority:code` is extracted once at construction
+/// for equality, SRID, and geographic-parameter lookups.
+#[derive(Debug)]
+struct Wkt {
```
paleolimbot: Can we call this `WktCrs`? (`Wkt` is already a thing we use...the concrete geo-traits class that WKT parses into)

rust/sedona-schema/src/crs.rs
```diff
@@ -682,6 +698,125 @@ impl CoordinateReferenceSystem for ProjJSON {
...
+}
+
+impl CoordinateReferenceSystem for Wkt {
+    /// Emit the extracted `authority:code` when available (so round-tripped
+    /// metadata is compact and recognized), otherwise the WKT itself. Both are
+    /// re-parseable by [`deserialize_crs`].
+    fn to_json(&self) -> String {
+        let value = self.authority.as_deref().unwrap_or(&self.wkt);
+        Value::String(value.to_string()).to_string()
+    }
```
paleolimbot: I think you actually want `Value::String(self.wkt).to_string()` here. The `Crs` doesn't take sides about representations...if the consumer wants authority:code they can ask for it.

rust/sedona-schema/src/crs.rs
```diff
@@ -682,6 +698,125 @@ impl CoordinateReferenceSystem for ProjJSON {
...
+    fn to_crs_string(&self) -> String {
+        self.wkt.clone()
+    }
+
+    fn geographic_params(&self) -> Result<Option<GeographicCrsParams>> {
+        Ok(self
+            .authority
+            .as_deref()
+            .and_then(geographic_params_for_authority))
+    }
```
paleolimbot, line 961: If we had some basic parsing we could extract these rather than restrict to the hard-coded authority/code combos

rust/sedona-schema/src/crs.rs
```diff
@@ -965,3 +1100,95 @@ mod test {
...
+    const WKT_NO_AUTHORITY: &str = r#"PROJCS["Custom",GEOGCS["Custom datum",DATUM["D",SPHEROID["S",6378137,298.257223563]]],PROJECTION["Mercator_1SP"]]"#;
+
+    #[test]
+    fn detects_wkt_strings() {
+        assert!(looks_like_wkt(WKT1_PSEUDO_MERCATOR));
+        assert!(looks_like_wkt(WKT2_WGS84));
+        assert!(looks_like_wkt("  geogcs[\"x\"]")); // leading ws + lowercase
+        assert!(!looks_like_wkt("EPSG:4326"));
+        assert!(!looks_like_wkt(r#"{"type":"GeographicCRS"}"#));
+        assert!(!looks_like_wkt(r#"PROJECTILE["not a crs"]"#));
```
paleolimbot, line 1353: I would like this to be a CRS even if it isn't one 🙂

rust/sedona-schema/src/crs.rs
```diff
@@ -682,6 +685,118 @@ impl CoordinateReferenceSystem for ProjJSON {
...
+static WKT_AUTHORITY_RE: LazyLock<Regex> = LazyLock::new(|| {
+    Regex::new(r#"(?i)\b(?:AUTHORITY|ID)\s*[\[(]\s*"([^"]+)"\s*,\s*"?([^"\],)]+)"?"#)
+        .expect("valid WKT authority regex")
+});
+
+/// Best-effort `authority:code` from a WKT CRS string. Returns the *last* tag —
+/// the outermost (CRS-level) authority, since inner tags belong to the datum,
+/// ellipsoid, etc. `None` when the WKT carries no authority tag.
+fn extract_authority_from_wkt(wkt: &str) -> Option<String> {
+    let caps = WKT_AUTHORITY_RE.captures_iter(wkt).last()?;
```
jiayuasu: [P1] This extraction can return the wrong CRS authority. It assumes the last AUTHORITY/ID tag is the CRS-level identifier, but WKT can contain nested authorities for the datum, spheroid, or inner GEOGCS. A custom projected CRS with no top-level authority but an inner GEOGCS authority such as EPSG:4326 would be classified as EPSG:4326 here, which can give the wrong SRID, equality result, and geography validation outcome. This is more serious than the PROJJSON id question because it is a parsing correctness bug; we should either parse nesting and only use an outer CRS-level authority, ask PROJ to identify the CRS, or avoid extracting WKT authority ids.

rust/sedona-schema/src/crs.rs
```diff
@@ -582,29 +600,14 @@ impl CoordinateReferenceSystem for ProjJSON {
         if let (Ok(Some(auth_code)), Ok(Some(other_auth_code))) =
             (self.to_authority_code(), other.to_authority_code())
         {
-            auth_code == other_auth_code
+            authority_codes_equal(&auth_code, &other_auth_code)
```
jiayuasu, line 566: [P2] This is less severe than the WKT extraction issue because a top-level PROJJSON id is structurally unambiguous, but treating a matching id as sufficient CRS equality is still a policy choice. PROJJSON can carry definition details that differ from the local authority database or from another document with the same id. If the id is meant to be authoritative, that should be explicit; otherwise definition-bearing PROJJSON should compare by full parsed value or a PROJ-backed equivalence check.

rust/sedona-schema/src/crs.rs
```diff
@@ -682,6 +685,118 @@ impl CoordinateReferenceSystem for ProjJSON {
...
+        }
+    }
+}
+
+impl CoordinateReferenceSystem for WktCrs {
+    /// Emit the extracted `authority:code` when available (so round-tripped
+    /// metadata is compact and recognized), otherwise the WKT itself. Both are
+    /// re-parseable by [`deserialize_crs`].
+    fn to_json(&self) -> String {
+        let value = self.authority_code.as_deref().unwrap_or(&self.wkt);
```
jiayuasu: [P2] This still collapses a WKT CRS with an extracted authority to the authority code when serializing metadata, which loses the original WKT definition. That overlaps with the concern about not canonicalizing definition-bearing CRS values on the way in: the CRS object should probably preserve its representation here, and callers that want the compact code can ask for to_authority_code() explicitly.

rust/sedona-functions/src/st_setsrid.rs
james-willis: dealt with this in #962

rust/sedona-functions/src/st_srid.rs
james-willis: fixed this, mostly by the changes from #962

rust/sedona-raster-functions/src/rs_setsrid.rs
james-willis: we decided in #962 to leave CRSes as given. One thing we did in that PR to save memory usage is use StringViews with deduplicated buffers to reduce the memory impact of this. NDV of CRS strings is hopefully pretty low in most cases.

rust/sedona-schema/src/crs.rs
james-willis: done

rust/sedona-schema/src/crs.rs
james-willis: done

rust/sedona-schema/src/crs.rs
james-willis: fixed in #962, just hadn't finished the rebased on this PR

rust/sedona-schema/src/crs.rs
james-willis: I think it is the right thing to do to treat the authority as authoritative about the identity of the CRS. 

Do you disagree?

rust/sedona-schema/src/crs.rs
james-willis: good idea. will do

rust/sedona-schema/src/crs.rs
james-willis: done here

rust/sedona-schema/src/crs.rs
james-willis: done in prior pr

rust/sedona-functions/src/st_srid.rs
james-willis: dne in earlier pr

rust/sedona-schema/src/crs.rs
james-willis: fixed

rust/sedona-schema/src/crs.rs
jiayuasu: agree. IIRC this is a decision Dewey made a while ago so just throw this out here. And I am fine with that. BTW, should we document this somewhere?

paleolimbot: > agree. IIRC this is a decision Dewey made a while ago so just throw this out here.

Mostly it's just annoying when `crsa != crsb` so it's worth being as lenient as reasonable here. FWIW GeoPandas uses a proj comparison with an 80% threshold (you have to pick between 0 and 100 🤷 ).

> BTW, should we document this somewhere

It would be great to document our CRS model!

rust/sedona-schema/src/crs.rs
```diff
@@ -634,6 +648,302 @@ impl CoordinateReferenceSystem for ProjJSON {
...
+}
+
+/// Best-effort `authority:code` from a WKT CRS string, read from the **CRS-level**
+/// `AUTHORITY["auth","code"]` (WKT1) or `ID["auth",code]` (WKT2) tag — i.e. a
+/// direct child of the root node. Authorities on nested datums, ellipsoids,
+/// projection methods, parameters, or units are deliberately ignored: a custom
+/// CRS with no top-level authority must not be misidentified by an inner code
+/// (e.g. a Lambert Conformal whose only `ID`s are the `EPSG:9001` metre unit and
+/// projection operation codes). `None` when the root carries no authority tag.
+fn extract_authority_from_wkt(wkt: &str) -> Option<String> {
```
paleolimbot, line 821: Thanks! I think this is better than a regex, and is certainly a lot cooler (for me, as a CRS nerd).

rust/sedona-schema/src/crs.rs
```diff
@@ -634,6 +648,302 @@ impl CoordinateReferenceSystem for ProjJSON {
...
+    }
+
+    fn parse_node(&mut self, depth: usize) -> Option<WktNode<'a>> {
+        if depth > WKT_MAX_DEPTH {
+            log::warn!(
+                "WKT CRS string nests beyond {WKT_MAX_DEPTH} levels; \
+                 giving up on authority extraction"
+            );
+            return None;
+        }
```
paleolimbot: Probably better to have this return a `Result` with an error (one level up in the authority extraction you can decide if you want to error or pretend there's no authority).

rust/sedona-schema/Cargo.toml
```diff
@@ -34,6 +34,7 @@ result_large_err = "allow"
 arrow-schema = { workspace = true }
 arrow-array = { workspace = true }
 datafusion-common = { workspace = true }
+log = { workspace = true }
```
paleolimbot: It would be great if we could avoid this because it's a low-level crate (but also OK if it's essential)

paleolimbot (approved):
It has always bothered me that we don't have WKT CRS support and I'm glad this adds it in a way that feels like proper support. Thank you!

rust/sedona-schema/src/crs.rs
james-willis: regex is bad for this! i managed to reproduce the exact same bug here (now fixed) that I fixed in some zarr code in another project last week.

rust/sedona-schema/Cargo.toml
james-willis: maybe we let the max depth WKT err propagate up then. honestly don't think anyone will ever hit that...

jiayuasu (approved):

rust/sedona-schema/src/crs.rs
james-willis: pinged you offline dewey about where to put docs. tracking it privately

PR #955, opened 2026-06-14
feat(rust/sedona-raster-gdal): add RS_Polygonize

Kontinuation:
## Summary
- add the GDAL-backed `RS_Polygonize` raster function
- add SQL reference docs and a dedicated Criterion benchmark
- keep the change limited to the new function and its incremental registration

## Dependency

## Testing
- `cargo bench -p sedona-raster-gdal --bench rs_polygonize --no-run`
  - blocked in this environment by existing missing raster fixture files in other tests
- `PATH="/home/kontinuation/workspace/github/apache/sedona-db/.venv/bin:$PATH" quarto render docs/reference/sql/rs_polygonize.qmd`

### DIFF
--- a/rust/sedona-raster-gdal/src/rs_polygonize.rs
+++ b/rust/sedona-raster-gdal/src/rs_polygonize.rs
@@ -278,7 +278,7 @@ mod tests {
     fn test_polygonize_raster() {
         let result = with_gdal(|gdal| {
             let raster_array = test_raster_spec().build();
-            let raster_struct = RasterStructArray::new(&raster_array);
+            let raster_struct = RasterStructArray::try_new(&raster_array).unwrap();
             let raster = raster_struct.get(0).unwrap();
             polygonize_raster(gdal, &raster, 1)
         })
--- a/rust/sedona-raster-gdal/src/rs_polygonize.rs
+++ b/rust/sedona-raster-gdal/src/rs_polygonize.rs
@@ -304,7 +304,9 @@ mod tests {
     use super::*;
 
     use arrow_array::Array;
-    use datafusion_common::cast::{as_list_array, as_string_view_array, as_struct_array};
+    use datafusion_common::cast::{
+        as_float64_array, as_list_array, as_string_view_array, as_struct_array,
+    };
     use datafusion_common::ScalarValue;
     use datafusion_expr::{ScalarUDF, ScalarUDFImpl};
     use sedona_gdal::raster::types::Buffer;
@@ -319,6 +321,13 @@ mod tests {
             .transform([0.0, 1.0, 0.0, 3.0, 0.0, -1.0])
             .band_values(&[1u8, 1, 0, 1, 2, 2, 0, 2, 2])
             .nodata(255u8)
+    }
+
+    fn test_float_raster_spec() -> RasterSpec {
+        RasterSpec::d2(3, 3)
+            .transform([0.0, 1.0, 0.0, 3.0, 0.0, -1.0])
+            .band_values(&[1.5f64, 1.5, 0.0, 1.5, 2.7, 2.7, 0.0, 2.7, 2.7])
+            .nodata(255.0f64)
     }
 
     fn build_polygonize_test_raster() -> arrow_array::StructArray {
@@ -413,7 +422,11 @@ mod tests {
         let udf: ScalarUDF = rs_polygonize_udf().into();
         let tester = ScalarUdfTester::new(udf, vec![RASTER, SedonaType::Arrow(DataType::Int32)]);
 
-        let rasters = vec![Some(test_raster_spec()), None, Some(test_raster_spec())];
+        let rasters = vec![
+            Some(test_raster_spec()),
+            None,
+            Some(test_float_raster_spec()),
+        ];
         let raster_array = Arc::new(raster_array(rasters));
         let band_array = Arc::new(Int32Array::from(vec![Some(1), Some(1), Some(1)]));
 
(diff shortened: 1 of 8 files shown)


rust/sedona-raster-gdal/src/rs_polygonize.rs
```diff
@@ -0,0 +1,375 @@
...
+        ));
+
+        with_gdal(|gdal| {
+            configure_thread_local_options(gdal, config_options)?;
+            let provider = thread_local_provider(gdal)
+                .map_err(|e| exec_datafusion_err!("Failed to init GDAL provider: {e}"))?;
+            // Note: We deliberately use "Memory" instead of "MEM" to be compatible with older GDAL versions.
+            let mem_driver = gdal
+                .get_driver_by_name("Memory")
+                .map_err(|e| exec_datafusion_err!("Failed to get Memory driver: {e}"))?;
```
paleolimbot, line 102: In earlier work we tried to avoid getting a driver by name in a loop to avoid contention on the global lock guarding the driver registry. Is it possible to avoid that lock here using the lower level mem dataset constructor?

(edit: I see this is for vector data. We're still doing this once per batch...is there anything we can do to avoid resolving the driver that frequently?)

rust/sedona-raster-gdal/src/rs_polygonize.rs
```diff
@@ -0,0 +1,375 @@
...
+    DataType::List(Arc::new(Field::new(
+        "item",
+        DataType::Struct(polygon_value_struct_fields()),
+        true,
+    )))
+}
+
+fn polygon_value_struct_fields() -> Fields {
+    Fields::from(vec![
+        Field::new("geom", DataType::Binary, false),
```
paleolimbot: I know that it is annoying, but this should be an `item_crs` and propagate the string Crs verbatim from the raster crs field. You could propagate this array after iterating if you'd like because it should be an identical arc reference to the raster field.

rust/sedona-raster-gdal/src/rs_polygonize.rs
```diff
@@ -0,0 +1,375 @@
...
+            None => {
+                let idx = feature
+                    .field_index("value")
+                    .map_err(|e| exec_datafusion_err!("Missing 'value' field: {e}"))?;
+                value_field_idx = Some(idx);
+                idx
+            }
+        };
+
+        polygon_values.push((wkb, feature.field_as_double(idx)));
```
paleolimbot: Can we use a `BinaryBuilder` and `Float64Builder` and write these incrementally to amortize the allocations?

paleolimbot (commented):
Thank you! A few questions but this looks great.

rust/sedona-raster-gdal/src/rs_polygonize.rs
```diff
@@ -0,0 +1,375 @@
...
+
+    let field_defn = gdal
+        .create_field_defn("value", OGRFieldType::OFTReal)
+        .map_err(|e| exec_datafusion_err!("Failed to create field definition: {e}"))?;
+    layer
+        .create_field(&field_defn)
+        .map_err(|e| exec_datafusion_err!("Failed to add field to layer: {e}"))?;
+
+    // Polygonize the raster band into the vector layer, using the "value" field to store pixel values.
+    gdal.polygonize(&raster_band, None, &layer, 0, &PolygonizeOptions::default())
```
james-willis: I believe our polygonize implementation is backed by GDALPolygonize.

I think we need to make sure to use GDALFPolygonize for cases where the raster is a float type and GDALPolygonize when it is an int. Otherwise we round floats to the nearest int and polygonize incorrectly

rust/sedona-raster-gdal/src/rs_polygonize.rs
```diff
@@ -0,0 +1,375 @@
...
+
+                let raster = match (raster_opt, band_opt) {
+                    (Some(raster), Some(_)) => raster,
+                    _ => {
+                        list_builder.append_null();
+                        return Ok(());
+                    }
+                };
+
+                let band_num: usize = band_opt.unwrap().max(1).try_into().unwrap_or(1);
```
james-willis: wont this clamp all values `[1, inf)`? so the validation below will never trigger?

rust/sedona-raster-gdal/src/rs_polygonize.rs
james-willis: I think also our behavior here is divergent from GDAL with regard to nodata values. We will polygonize them by default, but generally the GDAL will ignore them

james-willis:
This one is my fault but we need to consider how GDAL-backed RS function should work with the native outdb loader functionality I built to integrate zarr support. 

The simple solution is to annotate the function with `needs_pixels` metadata, but that will eliminate the VRT path for these functions: all rasters will wind up indb when they reach the functions.

Without a benchmark it is not clear to me whether the VRT outdb path is preferable to the needs_pixels/RS_EnsureLoaded solution. probably we need more data to make some determination on solution here.

cc @paleolimbot @zhangfengcdt

james-willis:
> I think we can merge this without a benchmark

We still need to decide what we want to do for non-GDAL loaders before merging

rust/sedona-raster-gdal/src/rs_polygonize.rs
Kontinuation: No. The dataset created by `MEMDatasetCreate` does not work well with vector layers in old GDAL versions such as 3.8 for some reason. We've hit a CI error because of this and switched to using the "Memory" driver. We use "Memory" instead of "MEM" here for the same reason.

Relevant GitHub workflow job: https://github.com/apache/sedona-db/actions/runs/27903364821/job/82567585551

```
failures:

---- rs_polygonize::tests::test_polygonize_invoke_array_and_nulls stdout ----

thread 'rs_polygonize::tests::test_polygonize_invoke_array_and_nulls' (46126) panicked at rust/sedona-raster-gdal/src/rs_polygonize.rs:348:14:
called `Result::unwrap()` on an `Err` value: Execution("Failed to create layer: GDAL method 'GDALDatasetCreateLayer' returned a NULL pointer. Error msg: 'CreateLayer() not supported by this dataset.'")
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

---- rs_polygonize::tests::test_polygonize_raster stdout ----

thread 'rs_polygonize::tests::test_polygonize_raster' (46129) panicked at rust/sedona-raster-gdal/src/rs_polygonize.rs:297:10:
called `Result::unwrap()` on an `Err` value: Execution("Failed to create layer: GDAL method 'GDALDatasetCreateLayer' returned a NULL pointer. Error msg: 'CreateLayer() not supported by this dataset.'")

---- rs_polygonize::tests::test_polygonize_invoke_scalar stdout ----

thread 'rs_polygonize::tests::test_polygonize_invoke_scalar' (46127) panicked at rust/sedona-raster-gdal/src/rs_polygonize.rs:327:14:
called `Result::unwrap()` on an `Err` value: Execution("Failed to create layer: GDAL method 'GDALDatasetCreateLayer' returned a NULL pointer. Error msg: 'CreateLayer() not supported by this dataset.'")
```

Kontinuation (commented):
We have addressed the suggestions regarding CRS propagation, builder usage, array extraction and interleaving testing in a follow-up commit.

rust/sedona-raster-gdal/src/rs_polygonize.rs
Kontinuation: Switched the geom field to be item_crs. The field name is still "geom".

rust/sedona-raster-gdal/src/rs_polygonize.rs
Kontinuation: The result array construction code has been refactored. The polygons were directly appended to the output arrow arrays in a closure passed into `polygonize_raster`.

rust/sedona-raster-gdal/src/rs_polygonize.rs
Kontinuation: I reviewed how we normalize and validate `band_num` in other RS functions and found that we should only do `.unwrap().unwrap_or(1)`. No `max(1)` should be performed. I'll remove the `.max(1)` here.

rust/sedona-raster-gdal/src/rs_polygonize.rs
Kontinuation: I've added fpolygonize (backed by GDALFPolygonize) to sedona-gdal and dynamically dispatch to it when the raster is of a floating-point type (Float32/Float64), addressing this in a follow-up commit.

rust/sedona-raster-gdal/src/rs_polygonize.rs
Kontinuation: I've removed the .max(1) call completely and rewritten the band validation logic. It now checks for <= 0 and out-of-range band values properly without silently wrapping/clamping negative values during conversion, addressing this in a follow-up commit.

paleolimbot (approved):
Thank you!

This looks great to me but would benefit from 👀 from @james-willis before merging.

james-willis (approved):
looks great! thanks

PR #956, opened 2026-06-14
feat(rust/sedona-raster-gdal): add RS_AsRaster for rasterizing geometries

Kontinuation:
## Summary
- add the GDAL-backed `rs_asraster` raster function
- normalize the SQL name to `rs_asraster`
- add SQL reference docs and a dedicated Criterion benchmark

## Dependency

## Testing
- `cargo bench -p sedona-raster-gdal --bench rs_as_raster --no-run`
  - blocked in this environment by existing missing raster fixture files in other tests
- `PATH="/home/kontinuation/workspace/github/apache/sedona-db/.venv/bin:$PATH" quarto render docs/reference/sql/rs_asraster.qmd`

### DIFF
--- a/docs/reference/sql/rs_asraster.qmd
+++ b/docs/reference/sql/rs_asraster.qmd
@@ -1,0 +1,125 @@
+---
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+title: RS_AsRaster
+description: Rasterizes a geometry using the grid of a reference raster.
+kernels:
+  - returns: raster
+    args:
+    - name: geometry
+      type: geometry
+    - name: reference_raster
+      type: raster
+    - name: pixel_type
+      type: string
+  - returns: raster
+    args:
+    - name: geometry
+      type: geometry
+    - name: reference_raster
+      type: raster
+    - name: pixel_type
+      type: string
+    - name: all_touched
+      type: boolean
+  - returns: raster
+    args:
+    - name: geometry
+      type: geometry
+    - name: reference_raster
+      type: raster
+    - name: pixel_type
+      type: string
+    - name: all_touched
+      type: boolean
+    - name: burn_value
+      type: double
+  - returns: raster
+    args:
+    - name: geometry
+      type: geometry
+    - name: reference_raster
+      type: raster
+    - name: pixel_type
+      type: string
+    - name: all_touched
+      type: boolean
+    - name: burn_value
+      type: double
+    - name: nodata_value
+      type: double
+  - returns: raster
+    args:
+    - name: geometry
+      type: geometry
+    - name: reference_raster
+      type: raster
+    - name: pixel_type
+      type: string
+    - name: all_touched
+      type: boolean
+    - name: burn_value
+      type: double
+    - name: nodata_value
+      type: double
+    - name: use_geometry_extent
+      type: boolean
+  - returns: raster
+    args:
+    - name: geometry
+      type: geometry
+    - name: reference_raster
+      type: raster
+    - name: pixel_type
+      type: string
+    - name: all_touched
+      type: boolean
+    - name: burn_value
+      type: double
+    - name: nodata_value
+      type: double
+    - name: use_geometry_extent
+      type: boolean
+---
+
+## Description
+
+Creates a single-band in-db raster by rasterizing `geometry` onto the grid
+defined by `reference_raster`.
+
+The output pixel type is controlled by `pixel_type`. Optional arguments control
+whether pixels touched by the geometry are burned, the value burned into covered
+pixels, the nodata fill value, and whether the output extent is cropped to the
+geometry extent or kept equal to the full reference raster extent.
+
+## Examples
+
+```sql
+SELECT
+  RS_Width(
+    RS_AsRaster(
+      ST_GeomFromWKT('POLYGON((1 1, 1 2, 2 2, 2 1, 1 1))'),
+      RS_FromPath('../../../submodules/sedona-testing/data/raster/test4.tiff'),
+      'D',
+      false,
+      1.0,
+      0.0,
+      true
+    )
+  );
+```
--- a/docs/reference/sql/rs_asraster.qmd
+++ b/docs/reference/sql/rs_asraster.qmd
@@ -110,16 +110,19 @@ geometry extent or kept equal to the full reference raster extent.
 ## Examples
 
 ```sql
-SELECT
-  RS_Width(
-    RS_AsRaster(
-      ST_GeomFromWKT('POLYGON((1 1, 1 2, 2 2, 2 1, 1 1))'),
-      RS_FromPath('../../../submodules/sedona-testing/data/raster/test4.tiff'),
-      'D',
-      false,
-      1.0,
-      0.0,
-      true
-    )
-  );
+WITH r AS (
+  SELECT RS_FromPath('https://download.osgeo.org/geotiff/samples/gdal_eg/cea.tif') AS raster
+)
+SELECT RS_Width(
+  RS_AsRaster(
+    RS_ConvexHull(raster),
+    raster,
+    'float64',
+    false,
+    1.0,
+    0.0,
+    false
+  )
+)
+FROM r;
 ```
--- a/docs/reference/sql/rs_asraster.qmd
+++ b/docs/reference/sql/rs_asraster.qmd
@@ -97,6 +97,10 @@ kernels:
       type: boolean
 ---
 
+::: callout-warning
+**Experimental.** This function is experimental; its behavior may change without notice.
+:::
+
 ## Description
 
 Creates a single-band in-db raster by rasterizing `geometry` onto the grid
--- a/docs/reference/sql/rs_asraster.qmd
+++ b/docs/reference/sql/rs_asraster.qmd
@@ -111,6 +111,10 @@ whether pixels touched by the geometry are burned, the value burned into covered
 pixels, the nodata fill value, and whether the output extent is cropped to the
 geometry extent or kept equal to the full reference raster extent.
 
+When both `geometry` and `reference_raster` have a CRS and they differ,
+`geometry` is reprojected into the reference raster's CRS before rasterization.
+Geometries without a CRS are rasterized using their coordinates as given.
+
 ## Examples
 
 ```sql
--- a/docs/reference/sql/rs_asraster.qmd
+++ b/docs/reference/sql/rs_asraster.qmd
@@ -113,7 +113,7 @@ geometry extent or kept equal to the full reference raster extent.
 
 When both `geometry` and `reference_raster` have a CRS and they differ,
 `geometry` is reprojected into the reference raster's CRS before rasterization.
-Geometries without a CRS are rasterized using their coordinates as given.
+An error is returned if exactly one of them has a CRS.
 
 ## Examples
 
--- a/python/sedonadb/tests/functions/test_raster_functions.py
+++ b/python/sedonadb/tests/functions/test_raster_functions.py
@@ -16,8 +16,10 @@
 # under the License.
 
 import pytest
+import numpy as np
 
 from sedonadb.testing import SedonaDB
+from sedonadb.raster import Raster
 
 
 @pytest.mark.parametrize(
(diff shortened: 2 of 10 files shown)


docs/reference/sql/rs_asraster.qmd
```diff
@@ -0,0 +1,125 @@
...
+geometry extent or kept equal to the full reference raster extent.
+
+## Examples
+
+```sql
+SELECT
+  RS_Width(
+    RS_AsRaster(
+      ST_GeomFromWKT('POLYGON((1 1, 1 2, 2 2, 2 1, 1 1))'),
+      RS_FromPath('../../../submodules/sedona-testing/data/raster/test4.tiff'),
```
paleolimbot: Can this example be user-runnable using a URL (can be URL to the sedona-testing repo tiff)

docs/reference/sql/rs_asraster.qmd
```diff
@@ -0,0 +1,125 @@
...
+description: Rasterizes a geometry using the grid of a reference raster.
+kernels:
+  - returns: raster
+    args:
+    - name: geometry
+      type: geometry
+    - name: reference_raster
+      type: raster
+    - name: pixel_type
+      type: string
```
paleolimbot, line 29: I don't think we should solve this now but these signatures result in pretty much unreadable SQL (we can probably improve this with some unified "options" handling in the future using struct arrays or JSON).

rust/sedona-raster-gdal/benches/rs_as_raster.rs
```diff
@@ -0,0 +1,169 @@
...
+            .unwrap();
+        dataset.set_projection("EPSG:4326").unwrap();
+        let band = dataset.rasterband(1).unwrap();
+        band.set_no_data_value(Some(0.0)).unwrap();
+        let mut buffer = Buffer::new((4, 3), vec![0u8; 12]);
+        band.write((0, 0), (4, 3), &mut buffer).unwrap();
+        sedona_raster_gdal::dataset_to_indb_raster(&dataset).unwrap()
+    })
+    .unwrap()
+}
```
paleolimbot: Can we use the `RasterSpec` to generate this a little more readably?

rust/sedona-raster-gdal/src/rs_as_raster.rs
```diff
@@ -0,0 +1,702 @@
...
+use sedona_raster::builder::RasterBuilder;
+use sedona_raster::traits::{BandMetadata, RasterMetadata, RasterRef};
+use sedona_schema::datatypes::{SedonaType, RASTER};
+use sedona_schema::matchers::ArgMatcher;
+use sedona_schema::raster::{BandDataType, StorageType};
+
+use crate::gdal_common::{band_data_type_to_gdal, nodata_f64_to_bytes, with_gdal};
+use crate::gdal_dataset_provider::{configure_thread_local_options, thread_local_provider};
+
+pub fn rs_as_raster_udf() -> SedonaScalarUDF {
```
paleolimbot, line 52: Not for now, but this seems like it would be most effective as an aggregate function (ST_Collect + AsRaster is a workaround but probably much more intensive)

rust/sedona-raster-gdal/src/rs_as_raster.rs
```diff
@@ -0,0 +1,702 @@
...
+        "B" => Ok(BandDataType::UInt8),
+        "I8" | "INT8" => Ok(BandDataType::Int8),
+        "U64" | "UINT64" => Ok(BandDataType::UInt64),
+        "I64" | "INT64" => Ok(BandDataType::Int64),
+        other => exec_err!(
+            "Unsupported pixelType: {} (expected one of D, F, I, S, US, B, I8, U64, I64)",
+            other
+        ),
+    }
+}
```
paleolimbot, line 286: These are pretty cryptic...can we also accept the strings we use elsewhere (`int8`, `uint8`, etc.)?

rust/sedona-raster-gdal/src/rs_as_raster.rs
```diff
@@ -0,0 +1,702 @@
...
+            return exec_err!("Reference raster has zero scale");
+        }
+
+        let start_col = ((env.MinX - ulx) / scale_x).floor() as isize;
+        let end_col_excl = ((env.MaxX - ulx) / scale_x).ceil() as isize;
+        let start_row = ((env.MaxY - uly) / scale_y).floor() as isize;
+        let end_row_excl = ((env.MinY - uly) / scale_y).ceil() as isize;
+
+        let width = (end_col_excl - start_col).max(0) as usize;
+        let height = (end_row_excl - start_row).max(0) as usize;
```
paleolimbot: Should these use a checked cast? I may be reading this incorrectly but it is probably possible to accidentally pass a huge geometry and a very small reference raster and overflow an `isize`.

rust/sedona-raster-gdal/src/rs_as_raster.rs
```diff
@@ -0,0 +1,702 @@
...
+) -> Result<()> {
+    match band_type {
+        BandDataType::UInt8 => initialize_band_t::<u8>(dataset, width, height, init_value as u8),
+        BandDataType::Int8 => initialize_band_t::<i8>(dataset, width, height, init_value as i8),
+        BandDataType::UInt16 => initialize_band_t::<u16>(dataset, width, height, init_value as u16),
+        BandDataType::Int16 => initialize_band_t::<i16>(dataset, width, height, init_value as i16),
+        BandDataType::UInt32 => initialize_band_t::<u32>(dataset, width, height, init_value as u32),
+        BandDataType::Int32 => initialize_band_t::<i32>(dataset, width, height, init_value as i32),
+        BandDataType::UInt64 => initialize_band_t::<u64>(dataset, width, height, init_value as u64),
+        BandDataType::Int64 => initialize_band_t::<i64>(dataset, width, height, init_value as i64),
```
paleolimbot: These two could generate potentially lossy values. It may be worth casting the nodata/init value based on the requested data type (or erroring for now until that can be supported).

rust/sedona-raster-gdal/src/rs_as_raster.rs
```diff
@@ -0,0 +1,702 @@
...
+    let band = dataset
+        .rasterband(1)
+        .map_err(|e| exec_datafusion_err!("Failed to get output band: {}", e))?;
+    let mut buffer = Buffer::new((width, height), vec![init_value; width * height]);
+    band.write((0, 0), (width, height), &mut buffer)
+        .map_err(|e| exec_datafusion_err!("Failed to initialize band: {}", e))?;
+    Ok(())
+}
+
+fn calc_num_iterations(args: &[ColumnarValue]) -> usize {
```
paleolimbot: This is fine for now but we should improve our pattern for iterating over pairs of raster(s)/geometry(ies) without expanding scalars.

rust/sedona-raster-gdal/src/rs_as_raster.rs
```diff
@@ -0,0 +1,702 @@
...
+        let geom = gdal.geometry_from_wkt(wkt).unwrap();
+        geom.wkb().map_err(|e| exec_datafusion_err!("{e}"))
+    }
+
+    fn bytes_to_f64_vec(bytes: &[u8]) -> Vec<f64> {
+        bytes
+            .chunks_exact(8)
+            .map(|chunk| f64::from_le_bytes(chunk.try_into().unwrap()))
+            .collect()
+    }
```
paleolimbot: I have seen both of these functions before. Can we use the previous versions of them?

paleolimbot (commented):
Thank you!

The tests here are pretty light given the complexity of the functionality here. We now have some Python utilities for input and output of rasters that can make a wider variety of tests more readable and/or compare them to something like rasterio to ensure correctness.

Like the other raster functions, I'm a little concerned about doing something expensive/potentially non-cancellable in a synchronous UDF; however, we need some working + correct functions before we can evaluate alternatives 🙂

docs/reference/sql/rs_asraster.qmd
paleolimbot: What we settled on in other PRs is to mark the raster SQL signatures as experimental before we release them to give us the freedom to update the signatures and break compatibility. It might be that with the generated R and Python APIs, users can pass these as keyword args (maybe we can make a list of them and force that behaviour in a future change).

python/sedonadb/tests/functions/test_raster_functions.py
```diff
@@ -268,3 +268,191 @@ def test_rs_setbandnodatavalue_two_arg_requires_single_band():
...
+    path = sedona_testing / "data/raster/test4.tiff"
+    transform = (0.0, 1.0, 0.0, 10.0, 0.0, -1.0)
+
+    result = (
+        con.sql(
+            _rs_as_raster_sql("float64", all_touched, 7.0, 0.0, use_geometry_extent),
+            params=(str(path),),
+        )
+        .to_arrow_table()["raster"][0]
+        .as_py()
```
paleolimbot: It doesn't matter for the test, but because of some pyarrow behaviour the only way to get a zero copy numpy array out is `.to_pandas().iloc[...]` or `Raster(arrow_table["raster"], 0)`. It might be good to update the syntax so that future contributors pick up a better pattern.

python/sedonadb/tests/functions/test_raster_functions.py
```diff
@@ -268,3 +268,191 @@ def test_rs_setbandnodatavalue_two_arg_requires_single_band():
...
+
+
+@pytest.mark.parametrize(
+    ("all_touched", "use_geometry_extent"),
+    [(False, False), (False, True), (True, False)],
+)
+def test_rs_as_raster_matches_rasterio(
+    con, sedona_testing, all_touched, use_geometry_extent
+):
+    import numpy as np
```
paleolimbot: I think we can move this to the top level (it's in test dependencies and should always be around)

paleolimbot (approved):
Thank you!

rust/sedona-raster-gdal/src/rs_as_raster.rs
Kontinuation: Why do you prefer defining RS_AsRaster as an aggregate function? Is it because it is usually used for dealing with mosaics where we want to stitch polygons extracted on multiple tiles together?

docs/reference/sql/rs_asraster.qmd
Kontinuation: I have marked this RS_AsRaster function as experimental as well.

rust/sedona-raster-gdal/src/rs_as_raster.rs
Kontinuation: The code added for casting in rs_as_raster.rs is massive. The newly added type `TypedBandValue` and the related functions seems to be useful for implementing other RS features. We can create one follow-up PR to extract these functions to a common module such as gdal_common.rs

Kontinuation:
Updated RS_AsRaster to handle ItemCrs geometries and follow the CRS auto-reprojection rule mentioned in https://github.com/apache/sedona-db/pull/974#discussion_r3455680920.

python/sedonadb/tests/functions/test_raster_functions.py
Kontinuation: Switched to `Raster(table["raster"], 0)`

python/sedonadb/tests/functions/test_raster_functions.py
Kontinuation: Hoisted the local imports generated by LLM

rust/sedona-raster-gdal/src/rs_as_raster.rs
paleolimbot: I'm not sure I prefer it, but it seems like it might be a useful addition on the day that we can allocate the rasterizer and burn in values lazily as they come.

rust/sedona-raster-gdal/Cargo.toml
```diff
@@ -45,6 +45,7 @@ sedona-gdal = { workspace = true }
 sedona-raster = { workspace = true }
 sedona-raster-functions = { workspace = true }
 sedona-schema = { workspace = true }
+sedona-proj = { workspace = true, features = ["proj-sys"] }
```
paleolimbot: I think this is causing the R build failures...we can't use proj-sys the feature (we need lazily defined proj). The CrsEngine is now available in the SedonaOptions, so hopefully that removes the need for this!

rust/sedona-raster-gdal/src/rs_as_raster.rs
```diff
@@ -0,0 +1,1014 @@
...
+
+        let mut builder = RasterBuilder::new(num_iterations);
+
+        let mut pixel_type_iter = pixel_type_array.iter();
+        let mut all_touched_iter = all_touched_array.iter();
+        let mut burn_value_iter = burn_value_array.iter();
+        let mut nodata_value_iter = nodata_value_array.iter();
+        let mut use_geom_extent_iter = use_geom_extent_array.iter();
+
+        with_global_proj_engine(|engine| {
```
paleolimbot: No need to do it here, but the CrsEngine is now available from the options. I can also take a pass through all the raster UDFs sometime before the next release.

rust/sedona-raster-gdal/src/rs_as_raster.rs
```diff
@@ -0,0 +1,1014 @@
...
+    Int8(i8),
+    UInt16(u16),
+    Int16(i16),
+    UInt32(u32),
+    Int32(i32),
+    UInt64(u64),
+    Int64(i64),
+    Float32(f32),
+    Float64(f64),
+}
```
paleolimbot, line 300: After this merges perhaps @james-willis can move these to a reusable location (this concept also shows up in one of the functions I just reviewed and this does a better job handling i64 and u64s)

paleolimbot (approved):
Thank you!

I think the proj-sys feature needs to be removed or moved to the dev-dependencies to fix the R build, but other than that this look good!

paleolimbot (approved):

PR #957, opened 2026-06-14
feat(rust/sedona-raster-gdal,c/sedona-gdal): add rs_geotiff_tiles

Kontinuation:
## Summary
- Added the `RS_GeoTiff_Tiles` table function. This function efficiently reads GeoTIFF files (or directories containing them) and loads the rasters as distinct tiles by leveraging the native internal tiling scheme of the source files.
  **Usage Example:**
  ```sql
  SELECT path, x, y, RS_MetaData(rast)
  FROM rs_geotiff_tiles('path/to/raster.tiff');
  ```
- Added the minimal VSI directory traversal support needed by the UDTF.
- Added SQL reference docs and a dedicated Criterion benchmark.

## Dependency

## Testing
- `cargo test -p sedona-raster-gdal rs_geotiff_tiles`
- `cargo bench -p sedona-raster-gdal --bench rs_geotiff_tiles --no-run`
- `PATH="/home/kontinuation/workspace/github/apache/sedona-db/.venv/bin:$PATH" quarto render docs/reference/sql/rs_geotiff_tiles.qmd`

## Future Work
- **Expanded Format Support:** Although the function is named `RS_GeoTiff_Tiles`, it might actually be capable of handling a wider range of GDAL-supported raster formats. We conservatively restricted the name to GeoTIFF since our testing has exclusively focused on this format.
- **Parallel Loading:** Currently, the function produces a single-partition relation, which does not distribute the data load. Enhancing this to allow for multi-partition loading is a logical next step to improve scalability.


docs/reference/sql/rs_geotiff_tiles.qmd
```diff
@@ -0,0 +1,51 @@
...
+
+```sql
+SELECT path, x, y
+FROM rs_geotiff_tiles('../../../submodules/sedona-testing/data/raster/test4.tiff');
+```
+
+```sql
+SELECT RS_MetaData(rast)
+FROM rs_geotiff_tiles('../../../submodules/sedona-testing/data/raster/test4.tiff');
+```
```
paleolimbot, line 53: Can you ensure these are user-runnable examples using a URL? If we end up getting transient failures we could do something fancy with the doc renderer that transforms sedona-testing urls into submodule paths.

rust/sedona-raster-gdal/src/rs_geotiff_tiles.rs
```diff
@@ -0,0 +1,625 @@
...
+use sedona_schema::raster::StorageType;
+
+use crate::gdal_common::{
+    convert_gdal_err, gdal_to_band_data_type, nodata_f64_to_bytes, normalize_outdb_source_path,
+    open_gdal_dataset, with_gdal,
+};
+
+pub fn rs_geotiff_tiles_udtf() -> Arc<dyn TableFunctionImpl> {
+    Arc::new(RsGeoTiffTilesFunction {})
+}
```
paleolimbot: This is probably better as a FileFormat than a table function...DataFusion can already take care of the listing and can kick off multi-file reads on separate partitions without us asking (plus has a built-in mechanism to handle key/value options). The `ExernalFormatSpec` should be able to abstract away the hard parts for you.

In Python this means you can get `sd.read(".../*.tif")` and `sd.read("some_dir", format="tiff")` for free.

rust/sedona-raster-gdal/src/rs_geotiff_tiles.rs
```diff
@@ -0,0 +1,625 @@
...
+    schema: SchemaRef,
+    properties: PlanProperties,
+}
+
+impl GeoTiffTilesExec {
+    fn new(dir: String, recursive: bool, schema: SchemaRef) -> Self {
+        let properties = PlanProperties::new(
+            EquivalenceProperties::new(schema.clone()),
+            // TODO: allow split paths to load into multiple partitions to enable parallelism.
+            Partitioning::UnknownPartitioning(1),
```
paleolimbot: This comes for free with ExternalFormatSpec I believe (but worth checking)

rust/sedona-raster-gdal/Cargo.toml
```diff
@@ -35,8 +35,11 @@ arrow-array = { workspace = true }
 arrow-buffer = { workspace = true }
 arrow-schema = { workspace = true }
 async-trait = { workspace = true }
+datafusion = { workspace = true, default_features = false }
```
paleolimbot: If we can avoid the full `datafusion` crate as a dependency that would be helpful for compilation order

paleolimbot (commented):
The tiling implementation looks good here, although I think we should implement this as a FileFormat + ListingTable rather than a table function. We haven't done it yet, but we can write a generic table function that wraps a FileFormat + ListingTable (e.g., `sd_read_format()`) if we want a table function in the future.

Among other things, this will get you directory listing and partitioning (I think) for free.

PR #962, opened 2026-06-15
fix(rust): clean up CRS string handling (preserve definitions, consolidate equality/SRID)

james-willis:
Cleans up how CRS strings flow through `ST_Crs`/`RS_Crs` and `ST_SetCRS`/`RS_SetCRS` so a CRS is preserved as the user gave it rather than rewritten, and consolidates the CRS equality/SRID logic. Groundwork before adding WKT CRS support.

Note: I didnt change the numeric SRID path (ST_SRID/RS_SRID, EWKT SRID=4326) still maps 4326→OGC:CRS84, so `RS_SetSRID(r, 4326)` stores OGC:CRS84 while `RS_SetCRS(r, 'EPSG:4326')` stores EPSG:4326 — equal CRSes, different strings.

<details>
<summary>Implementation notes (AI-generated)</summary>

- `ST_Crs`/`RS_Crs` and `ST_SetCRS`/`RS_SetCRS` emit `to_crs_string()`: an `authority:code` stays compact, but a PROJJSON/WKT definition with an embedded `id` is kept in full instead of being collapsed to that code. Same fix when widening a fixed-CRS geometry to an item-level CRS.
- The `EPSG:4326`/`OGC:CRS84` lon/lat aliases are kept verbatim in the round-trippable string (`deserialize_crs` no longer folds `4326`→`CRS84`), but are canonicalized to `OGC:CRS84` in the `to_json` metadata path so GeoParquet/GeoArrow stay axis-order-explicit. The two still compare equal.
- CRS equality is consolidated into `crs_equals` (the lon/lat alias rule lives in one `authority_codes_equal` helper); `PartialEq` delegates to it. `CoordinateReferenceSystem::srid()` is now a trait default derived from `to_authority_code()`.
- `RS_SetCRS` normalization handles the scalar case (one definition fanned out via `try_append_value_n`) and the array case (deduplicating builder) distinctly, so a high-row-count / few-distinct-CRS batch stays compact.
- `CachedCrsNormalization` is replaced by a one-line `normalize_crs()` free function; parsing is already memoized by `deserialize_crs`'s thread-local cache.
- "Kept in full" is semantic, not byte-for-byte: a PROJJSON re-serializes from its parsed tree (keys may reorder, whitespace normalizes); it round-trips and compares equal.

</details>

### DIFF
--- a/rust/sedona-functions/src/st_setsrid.rs
+++ b/rust/sedona-functions/src/st_setsrid.rs
@@ -658,12 +658,12 @@ mod test {
         assert!(out.value(0).contains("GeographicCRS"));
         assert_ne!(out.value(0), "EPSG:4269");
         assert_eq!(out.value(0), out.value(2));
-        // EPSG:4326 folds to OGC:CRS84 (<=12 bytes, inlined into the view).
-        assert_eq!(out.value(3), "OGC:CRS84");
+        // EPSG:4326 is kept verbatim (<=12 bytes, inlined into the view).
+        assert_eq!(out.value(3), "EPSG:4326");
 
         // Deduplication: the repeated PROJJSON lives in the shared data buffer
         // exactly once, so total buffer bytes equal a single copy (the inlined
-        // OGC:CRS84 contributes nothing to the buffer).
+        // EPSG:4326 contributes nothing to the buffer).
         let buffer_bytes: usize = out.data_buffers().iter().map(|b| b.len()).sum();
         assert_eq!(buffer_bytes, out.value(0).len());
     }
@@ -906,7 +906,7 @@ mod test {
                     None
                 ],
                 [
-                    Some("OGC:CRS84"),
+                    Some("EPSG:4326"),
                     Some("EPSG:3857"),
                     Some("EPSG:3857"),
                     None,
@@ -959,7 +959,7 @@ mod test {
                     Some("POINT (4 5)"),
                     None
                 ],
-                [Some("OGC:CRS84"), Some("EPSG:4269"), None, None],
+                [Some("EPSG:4326"), Some("EPSG:4269"), None, None],
                 &WKB_GEOGRAPHY
             )
         );
--- a/rust/sedona-functions/src/st_setsrid.rs
+++ b/rust/sedona-functions/src/st_setsrid.rs
@@ -39,7 +39,7 @@ use sedona_expr::{
 };
 use sedona_geometry::transform::CrsEngine;
 use sedona_schema::{
-    crs::{deserialize_crs, CachedCrsNormalization, CachedSRIDToCrs, Crs},
+    crs::{deserialize_crs, normalize_crs, CachedSRIDToCrs, Crs},
     datatypes::{Edges, SedonaType},
     matchers::ArgMatcher,
 };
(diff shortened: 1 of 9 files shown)


rust/sedona-schema/src/crs.rs
```diff
@@ -226,8 +229,11 @@ impl CachedCrsNormalization {
...
     ///
-    /// Returns the abbreviated `authority:code` form if available, otherwise the
-    /// original string. Returns `None` for `"0"`, `""`, or CRS strings that
+    /// Returns the round-trippable CRS definition (`to_crs_string`) — the
+    /// authority code for an `authority:code` CRS, or the full PROJJSON/WKT for
+    /// a definition-based one. The full definition is preserved rather than
+    /// collapsed to its embedded authority code, so no information is lost on
+    /// the way in. Returns `None` for `"0"`, `""`, or CRS strings that
     /// deserialize to `None`.
     pub fn normalize(&mut self, crs_str: &str) -> Result<Option<String>> {
```
paleolimbot: Do we still need this function if it is just unserializing and then reserializing?

paleolimbot (approved):
Thank you! I think this makes the most sense for `ST_Crs()` (which we invented)...we could add `ST_AuthorityCode()` or something else to specifically canonicalize item-level CRSes if `ST_SRID` isn't sufficient.

I have one question, and if there's an opportunity to integrate the `StringViewBuilder::try_append_n()` optimization ( https://github.com/apache/sedona-db/pull/903 ) this might be a good time to do it.

rust/sedona-functions/src/st_setsrid.rs
```diff
@@ -507,18 +507,19 @@ fn normalize_crs_array(
...
-                })
-                .collect::<Result<StringViewArray>>()?;
+            // Deduplicate: when many rows carry the same (possibly large)
+            // PROJJSON/WKT definition, the bytes are stored once in the view
+            // buffer and shared across rows.
+            let mut builder = StringViewBuilder::with_capacity(string_view_array.len())
+                .with_deduplicate_strings();
+            for maybe_crs in string_view_array.iter() {
+                match maybe_crs {
+                    Some(crs_str) => builder.append_option(crs_norm.normalize(crs_str)?),
```
james-willis: @paleolimbot does this not accomplish the same thing as `StringViewBuilder::try_append_n()`?

paleolimbot: I believe the code you've written will hash the 4KB WKT/PROJJSON string `n` times, whereas try_append_n will hash it once and append the same view `n` times.

james-willis: Now I avoid the hash in the Scalar case but do the hash to capture repeated CRSes in the Array case.

rust/sedona-schema/src/crs.rs
james-willis: I've revised the method down to just a single line. I think thats better than inlining that LOC into the call sites

rust/sedona-functions/src/st_setsrid.rs
james-willis: I think I need to explain how this works more and why its different than the RS function. 

RS function does not have type level CRS, so the scalar case hits the function implementation. We have branching logic that does the hash + dedupe for the Array case there so we don't duplicate the string n times to save RAM. Zarrs often encode the whole CRS WKT so this is important! 

In the ST function, the Scalar case is handled at the type level so it doesn't hit this implementation. The code you see hereis only run with row level CRS (ie non-Scalar case). My assumption is that is it worth it to add this dedupe logic here for the same reason as the RS function: avoid allocating for long CRS WKTs on every row. I believe in this case the hash is worth it!

paleolimbot (approved):
Thank you!

I *think* preserving the source CRS more faithfully is a good idea in these places...we don't have a ton of usage or benchmarks to compare to see to what extent this matters in real usage (item crs was primarily added for alignment with raster, which is always item-crs, so the use cases are only starting to be realized).

PR #963, opened 2026-06-15
fix(c/sedona-gdal): resolve MEMDataset::Create on GDAL 3.13 + use MEMCreate C-API when available

james-willis:
I'm not really sure what this does but it makes it so that GDAL 3.13 doesnt break the build/runtime anymore.

I integrated the fix Kristin recommended below as well

<details>
<summary>Summary (AI-generated)</summary>

The GDAL MEM-dataset bridge resolves the MEM `Create` entry point dynamically in `c/sedona-gdal/src/dyn_load.rs`. It originally relied solely on the C++ `MEMDataset::Create` symbol, looked up by its mangled name. That mangling is fragile — it depends on the platform *and* on the parameters' constness — so GDAL 3.13 changed the final parameter from `char**` to `char const* const*` (`…GDALDataTypePPc` → `…GDALDataTypePKS1_`), the candidate list only had the ≤3.12 name, and the bridge panicked at load time ("Failed to resolve … under any known mangled name"), breaking every GDAL raster op. Each future re-mangling would break it again.

GDAL has since added `MEMCreate` as a stable, unmangled public C API. This change resolves `MEMCreate` first and falls back to the mangled `MEMDataset::Create` only when it is absent (older GDAL), erroring only if neither resolves. Because `MEMCreate` drops the `pszFilename` parameter, it gets its own fn-pointer type and call path: `create_mem_dataset` calls `MEMCreate` (no filename) when present, otherwise the mangled C++ method with a leading empty filename.

Verified against a local GDAL 3.13.1 install: `cargo test -p sedona-gdal --features gdal-sys/bindgen` → 102 passed.

</details>

### DIFF
--- a/c/sedona-gdal/src/dyn_load.rs
+++ b/c/sedona-gdal/src/dyn_load.rs
@@ -52,10 +52,13 @@ macro_rules! load_fn {
     };
 }
 
-/// Try to load a symbol under one of several names (e.g. for C++ mangled symbols).
-/// Writes the first successful match into `$api.$field`. Returns an error only if
-/// *none* of the names resolve.
-macro_rules! load_fn_any {
+/// Try to load a symbol under one of several names (e.g. for C++ mangled
+/// symbols) and return whether a match was found, writing the first successful
+/// match into `$api.$field`. Unlike [`load_fn!`], it does not error when none
+/// resolve — used when a symbol has acceptable alternatives whose presence
+/// varies by GDAL version (e.g. the public `MEMCreate` C API vs. the mangled
+/// `MEMDataset::Create`), so the caller decides what counts as a fatal miss.
+macro_rules! try_load_fn_any {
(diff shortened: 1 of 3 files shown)


paleolimbot (approved):
The `const` changes the mangling of the C++ name; however, Kristin added this to the C API which I believe was also released in 3.13 ( https://github.com/OSGeo/gdal/pull/14108 ).

This may need a similar change for MSVC mangling, but we can/should probably go straight to the C symbol if it exists (now that there is one).

Kontinuation:
Each time the constness of pointer parameter changes, the ABI breaks. Fortunatelly the upstream accepted our request of adding `MEMCreate` as its public C API, so we can first try resolving to `MEMCreate`. If it fails, we fallback to the ugly C++ mangled name lookup.

paleolimbot (approved):
I checked with local GDAL 3.13 and specifically checked that the new C symbol was getting used and this works great! Thank you!

PR #964, opened 2026-06-16
feat(rust/sedona-raster): BandRef::copy_into derive-with-overrides (zero-copy)

james-willis:


zhangfengcdt:
@james-willis Is this ready to review?

james-willis:
> @james-willis Is this ready to review?


No, its not prioritized for 0.4.0 so I haven't finished it. 

Its part of the lazy slicing feature which will land later.

rust/sedona-raster/src/traits.rs
```diff
@@ -667,6 +685,59 @@ pub trait BandRef {
...
+    /// The default copies the visible source bytes via `append_value`. Arrow-
+    /// backed implementations override this to share the source row's backing
+    /// `Buffer` zero-copy (a refcount bump via
+    /// [`RasterBuilder::append_band_data_from`]), keeping the buffer plumbing
+    /// encapsulated rather than exposing a raw `Buffer` accessor. Call after the
+    /// band's schema has been written (e.g. by [`Self::copy_into`]).
+    fn append_data_into(&self, builder: &mut RasterBuilder) -> Result<(), ArrowError> {
+        if self.is_indb() {
+            let ndb = self.nd_buffer()?;
+            builder.band_data_writer().append_value(ndb.buffer);
```
zhangfengcdt, line 761: Would this be `buffer[offset..offset+len]` since it is the visible portion of the buffer?

james-willis: I tweaked this a little but that was not a bug.

The purpose of the copy is not to in any way modify the data buffer. Its to pass it as is.

The view will be passed separately which tells the new copy how to step through the nd_buffer.

rust/sedona-raster/src/builder.rs
```diff
@@ -387,6 +388,53 @@ impl RasterBuilder {
...
+        &mut self,
+        name: Option<&str>,
+        dim_names: &[&str],
+        shape: &[i64],
+        data_type: BandDataType,
+        nodata: Option<&[u8]>,
+        outdb_uri: Option<&str>,
+        outdb_format: Option<&str>,
+        view: &[ViewEntry],
+    ) -> Result<(), ArrowError> {
```
paleolimbot: Clippy is right about this one, and it's a public API and probably worth dealing with before it is used. A common way to handle this is to have a `struct NewBandArgs<'a> { ... }`, which forces code to name arguments.

paleolimbot (approved):
Thank you!

Possibly worth dealing with the clippy allow before you go further (there are a lot of follow-ups already)

paleolimbot (approved):
Thank you!

PR #965, opened 2026-06-16
feat(python/sedonadb): add DataFrame.union and union_distinct

jiayuasu:
Adds two relational set-union methods to the Python `DataFrame`, continuing the Ibis/DuckDB-style relational surface (after `distinct`/`distinct_on` in #961).

## API

```python
a.union(b)            # vertical concat, keeps duplicates   (SQL UNION ALL)
a.union_distinct(b)   # vertical concat, drops duplicates   (SQL UNION)
```

- `union` keeps duplicate rows (matches **SQL `UNION ALL`** and **PySpark's `union`**).
- `union_distinct` removes duplicate rows (**SQL `UNION`**).
- Columns are matched **by position**; the inputs must share a schema. A mismatch surfaces as a `SedonaError` at call time (DataFusion builds the plan eagerly), e.g. *"UNION queries have different number of columns"*.

Naming note: the engine-neutral convention (SQL ALL-vs-not, Spark) is that bare `union` keeps duplicates and the `_distinct` variant dedupes — documented prominently in the docstrings since a SQL user might expect bare `UNION` (dedup) semantics from `union`. A by-name variant (DataFusion's `union_by_name`) can follow later if wanted.

## Implementation

| File | Change |
|---|---|
| `python/sedonadb/src/dataframe.rs` | `InternalDataFrame::union` / `union_distinct` — thin wrappers over DataFusion's `DataFrame::union` / `union_distinct`. |
| `python/sedonadb/python/sedonadb/dataframe.py` | `DataFrame.union` / `union_distinct` with `isinstance` validation. |

## Test plan

9 tests in `tests/expr/test_dataframe_union.py`:
- `union` keeps duplicates; multi-column; position-based column matching (result takes the left's names).
- `union_distinct` dedupes across inputs and within each input.
- Lazy return for both.
- Errors: non-DataFrame argument (both methods); column-count mismatch.

Local: 9 unit + 29 doctests + full expr suite + `ruff` + `cargo fmt --check` + `clippy -Dwarnings` all clean.

### DIFF
--- a/python/sedonadb/tests/expr/test_dataframe_union.py
+++ b/python/sedonadb/tests/expr/test_dataframe_union.py
@@ -1,0 +1,99 @@
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+import pandas as pd
+import pandas.testing as pdt
+import pytest
+
+from sedonadb.dataframe import DataFrame
+from sedonadb.expr import col
+
+
+def test_union_keeps_duplicates(con):
+    a = con.create_data_frame(pd.DataFrame({"x": [1, 2]}))
+    b = con.create_data_frame(pd.DataFrame({"x": [2, 3]}))
+    out = a.union(b).sort("x").to_pandas()
+    # UNION ALL: the shared value 2 appears twice.
+    pdt.assert_frame_equal(out, pd.DataFrame({"x": [1, 2, 2, 3]}))
+
+
+def test_union_multi_column(con):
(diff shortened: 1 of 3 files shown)


python/sedonadb/tests/expr/test_dataframe_union.py
```diff
@@ -0,0 +1,88 @@
...
+        a.union_distinct({"x": 1})
+
+
+def test_union_column_count_mismatch_raises(con):
+    from sedonadb._lib import SedonaError
+
+    a = con.create_data_frame(pd.DataFrame({"x": [1]}))
+    b = con.create_data_frame(pd.DataFrame({"x": [1], "y": [2]}))
+    with pytest.raises(SedonaError, match="different number of columns"):
+        a.union(b)
```
paleolimbot: Can you also check that unioning `pd.DataFrame({"x": [1]})` with `pd.DataFrame({"y": [1]})` (i.e., same number of columns but different names) is an error? (If it is not an error, this should be opt-in because it's a footgun otherwise).

Can you ensure that this test and the one I suggested also applies to `union_distinct()`?

paleolimbot (approved):
Thank you!

One suggestion for a test to verify what I think is the case but other than that this is good to go!

python/sedonadb/tests/expr/test_dataframe_union.py
jiayuasu: Good call — it was *not* an error before (DataFusion's `union` aligns by position and silently takes the left's names), so I made matching column names the requirement rather than leaving the footgun in. Both `union` and `union_distinct` now raise a clear `ValueError` when the two inputs don't have the same column names in the same order; a positional union of differently-named columns is opt-in (align names first, e.g. `a.union(b.select(col('y').alias('x')))`).

Tests added/updated in d599900, parametrized over both methods: different-names raises, different-count raises, plus a test for the opt-in alignment path. Also rebased on latest main.

paleolimbot (approved):
Thank you!

PR #967, opened 2026-06-16
chore(rust/sedona-raster): Ensure creating a RasterStructArray does not panic

paleolimbot:
Extracted from #916. Some code paths create a RasterStructArray from  data that hasn't yet been validated. Because the `new()` function paniced for unexpected child array types, this meant that malformed data could crash the process.

This PR updates the `new()` function to be `try_new()` and ensures it errors for unexpected input.

### DIFF
--- a/rust/sedona-raster/src/array.rs
+++ b/rust/sedona-raster/src/array.rs
@@ -20,6 +20,11 @@ use arrow_array::{
     StringViewArray, StructArray, UInt32Array,
 };
 use arrow_schema::ArrowError;
+use datafusion_common::cast::{
+    as_binary_array, as_binary_view_array, as_float64_array, as_int64_array, as_list_array,
+    as_string_array, as_string_view_array, as_struct_array, as_uint32_array,
+};
+use datafusion_common::DataFusionError;
 
 use crate::traits::{BandRef, Bands, NdBuffer, RasterRef};
 use crate::view_entries::ViewEntry;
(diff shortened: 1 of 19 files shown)


rust/sedona-raster/src/array.rs
```diff
@@ -402,115 +407,64 @@ pub struct RasterStructArray<'a> {
 
 impl<'a> RasterStructArray<'a> {
     /// Create a new RasterStructArray from an existing StructArray.
+    ///
+    /// Returns an error if the array doesn't have the expected raster schema.
     #[inline]
-    pub fn new(raster_array: &'a StructArray) -> Self {
+    pub fn try_new(raster_array: &'a StructArray) -> Result<Self, ArrowError> {
```
james-willis, line 412: One comment from claude:

> try_new still panics on a struct with the wrong number of columns. The cast helpers guard against wrong types, but raster_array.column(raster_indices::CRS) and bands_struct.column(band_indices::...) are called before the cast — and StructArray::column(i) panics on out-of-bounds index. So malformed input that has too few top-level or band fields (a plausible shape for the "data that hasn't been validated" threat model from #916) will still panic, defeating the PR's stated goal. Consider using column_by_name or guarding num_columns() / columns().get(i) first, and at minimum verifying the band sub-struct has the expected field count. Worth confirming whether upstream type-checking already guarantees arity for every try_new caller — if it does, this is moot and a comment saying so would help; if not, it's a residual panic vector.

james-willis:
LGTM other than the one comment

PR #968, opened 2026-06-16
refactor(python/sedonadb): Extract common read functionality into read accessor

paleolimbot:
This PR extracts common functionality between `read_format()`, `read_parquet()`, and `read_pyogrio()` into a `read` accessor that is easier to use:

```python
import sedona.db
import sedonadb_zarr

sd = sedona.db.connect()
sd.register(sedonadb_zarr.ZarrExtension())

t = sd.read(
    "https://atlantis-vis-o.s3-ext.jc.rl.ac.uk/hurricanes/era5/florence",
    options={"arrays": ["velocity"]},
    format="zarr",
)

t.select(t.raster.funcs.rs_envelope().geo.set_srid(0)).to_pandas().plot(
    facecolor="none", edgecolor="black"
)
```

output

I'd like to push some of this into Rust (using the session's FileFormatFactory) but that change is too disruptive for the pressing need, which is to solidify the read command for an externally registered format.

### DIFF
--- a/python/sedonadb/python/sedonadb/read.py
+++ b/python/sedonadb/python/sedonadb/read.py
@@ -1,0 +1,306 @@
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+import json
+from typing import Any, Dict, Iterable, Optional, Union
+
+from pathlib import Path
+
+from sedonadb.dataframe import DataFrame
+from sedonadb.utility import sedona  # noqa: F401
+from sedonadb.datasource import ExternalFormatSpec
+
+
+class Read:
+    """Read external data into SedonaDB
+
+    This class is the entrypoint to reading any path or URL as
+    a SedonaDB DataFrame.
+    """
+
+    def __init__(self, ctx):
+        self._ctx = ctx
+
+    def __call__(
+        self,
+        table_paths: Union[str, Path, Iterable[str]],
+        *,
+        options: Optional[Dict[str, Any]] = None,
+        partitioning: Union[str, Iterable[str], None] = None,
+        format: Union[str, "ExternalFormatSpec", None] = None,
+        check_extension: bool = False,
+    ) -> DataFrame:
+        """Read one or more paths
+
+        This is the generic read path, which guesses the appropriate format.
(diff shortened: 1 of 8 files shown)


python/sedonadb/python/sedonadb/read.py
```diff
@@ -0,0 +1,354 @@
...
+        if options is None:
+            options = {}
+
+        # If the format is a Python object, call with_options() on the Python objects
+        # to eliminate the serialization that would otherwise happen
+        if isinstance(format, ExternalFormatSpec):
+            return DataFrame(
+                self._ctx,
+                self._ctx._impl.read_external_format(
+                    format.with_options(options),
```
james-willis: ExternalFormatSpec documents that "implementations are only required to implement open_reader()", and the base with_options raises NotImplementedError for any call.

So we should either implement a default with_options OR not call with_options unless options is provided as a non-None arg.

python/sedonadb/python/sedonadb/read.py
```diff
@@ -0,0 +1,354 @@
...
+            <sedonadb.dataframe.DataFrame object at ...>
+        """
+        if options is None:
+            options = {}
+
+        if geometry_columns is not None and not isinstance(geometry_columns, str):
+            geometry_columns = json.dumps(geometry_columns)
+
+        options["geometry_columns"] = geometry_columns
+        options["validate"] = validate
```
james-willis, line 253: copy options first to avoid mutating the caller's dict.

(im only like 75% sure on this one, python pass by reference semantics confuse me)

james-willis: One from claude:

> geometry_columns not JSON-serialized on the generic path

>  .parquet() does json.dumps(geometry_columns) for non-str values, but the generic __call__ parquet branch pops geometry_columns straight out of options and forwards it to _impl.read_parquet un-serialized. So sd.read("x.parquet",  options={"geometry_columns": {...}}) passes a dict where sd.read.parquet(geometry_columns={...}) passes a JSON string. Depending on what the Rust binding expects, the former likely errors or misbehaves. Worth either normalizing in one place or documenting  that geometry_columns via raw options must be a JSON string.

python/sedonadb/python/sedonadb/read.py
```diff
@@ -0,0 +1,354 @@
...
+    def _guess_format(self, table_paths):
+        """A heuristic to guess a format when not provided. This doesn't handle
+        URL suffixes like query strings or fragments (these types of inputs should
+        use an explicit format)."""
+        if not table_paths:
+            raise ValueError("Can't guess table paths from empty path list")
+
+        formats = set()
+        for path in table_paths:
+            try:
```
james-willis: does anything in this block throw a Value Error?

james-willis (changes requested):
I think we need to fix at least the with_options thing. maybe test the geoparquet thing

python/sedonadb/python/sedonadb/read.py
paleolimbot: Good point...I changed this to avoid the call if there aren't options

python/sedonadb/python/sedonadb/read.py
paleolimbot: These are both good suggestions...I fixed both and added a test for the second one

python/sedonadb/python/sedonadb/read.py
paleolimbot: Apparently `Path` can throw if you have invalid unicode in your string (which probably deserves to surface to the user asap)

PR #969, opened 2026-06-17
fix(rust/sedona-query-planner): Ensure user provided RS_EnsureLoaded call preserves metadata

paleolimbot:
We can't show off the fact that we can load pixels if we can't call it!

In https://github.com/apache/sedona-db/pull/886 we added an optimizer rule that wraps pixel-requiring functions in RS_EnsureLoaded AND an extra function that works around async UDFs dropping metadata at the physical plan level. This doesn't handle normal calls to RS_EnsureLoaded, which are useful for testing the loaders or generally moving back into Python land.

This PR separates separates out the rule that wraps async udfs in the metadata-fixing function and tests it separately.

```python
import sedona.db

sd = sedona.db.connect()
sd.options.interactive = True

path = "submodules/sedona-testing/data/raster/sentinel2.tif"
t = sd.sql("SELECT RS_FromPath($1) AS raster", params=(path, ))
t
# ┌───────────────────────────────────────────────────────────────┐
# │                             raster                            │
# │                             raster                            │
# ╞═══════════════════════════════════════════════════════════════╡
# │ [512x512/1] @ [330720 4138560 335840 4143680] / {...} <outdb> │
# └───────────────────────────────────────────────────────────────┘

t.select(t.raster.funcs.rs_ensureloaded())
# ┌───────────────────────────────────────────────────────┐
# │                rs_ensureloaded(raster)                │
# │                         raster                        │
# ╞═══════════════════════════════════════════════════════╡
# │ [512x512/1] @ [330720 4138560 335840 4143680] / {...} │
# └───────────────────────────────────────────────────────┘
```


rust/sedona-query-planner/src/wrap_async_udf.rs
```diff
@@ -0,0 +1,321 @@
...
+use datafusion_expr::{Expr, LogicalPlan};
+use datafusion_optimizer::{ApplyOrder, OptimizerConfig, OptimizerRule};
+
+use crate::restore_metadata::{restore_metadata_udf, RESTORE_METADATA_NAME};
+
+/// Logical optimizer rule that wraps async scalar UDF calls with
+/// `sd_restore_metadata` to preserve field metadata stripped at the
+/// physical layer (DF-22662).
+#[derive(Default, Debug)]
+pub struct WrapAsyncUdfRule;
```
paleolimbot, line 44: This is the real addition...we wrap AsyncUdfs with a UDF that applies extra metadata to the output (based on `expr.to_field()`, which is correct for an async udf at the logical level.

rust/sedona-query-planner/src/restore_metadata.rs
```diff
@@ -0,0 +1,192 @@
...
+/// Registry/UDF name of the metadata restoration helper.
+pub const RESTORE_METADATA_NAME: &str = "sd_restore_metadata";
+
+/// Build the sync metadata-restore UDF with the given metadata map. The rule
+/// constructs one per rewrite, embedding the metadata that should be stamped
+/// onto the output field. `Stable` volatility so CSE can deduplicate identical
+/// `sd_restore_metadata(...)` subtrees.
+pub fn restore_metadata_udf(metadata: HashMap<String, String>) -> Arc<ScalarUDF> {
+    Arc::new(ScalarUDF::new_from_impl(RestoreMetadata::new(metadata)))
+}
```
paleolimbot, line 56: This was extracted out of the previous "reraster" (since it could in theory do things that aren't raster)

python/sedonadb/tests/functions/test_raster_functions.py
```diff
@@ -15,43 +15,44 @@
...
+    assert r.height == 512
+    assert r.width == 512
+
+    assert len(r.bands) == 1
+    b = r.bands[0]
+    assert b.shape == (512, 512)
+    arr = b.to_numpy()
+    assert arr.shape == (512, 512)
+    assert arr.dtype == "uint16"
+    assert arr[0, 0] == 2324
```
paleolimbot, line 58: It works!

python/sedonadb/python/sedonadb/raster.py
```diff
@@ -232,7 +232,8 @@ def source_data(self) -> memoryview:
     @property
     def data(self) -> memoryview:
         """The band data as a typed, shaped memoryview."""
-        if self.outdb_uri is not None:
+        source_data = self.source_data
+        if self.outdb_uri is not None and len(source_data) == 0:
```
james-willis, line 264: you should also make sure that this is not a 0 byte raster. if hte product of the shape is 0 then it is always indb and [] is the byte array

rust/sedona-query-planner/src/wrap_async_udf.rs
james-willis, line 44: can we add an idempotency test for this rule? we want to make sure we dont rewrap

rust/sedona-query-planner/src/wrap_async_udf.rs
```diff
@@ -0,0 +1,321 @@
...
+        .is_some();
+
+    if !is_async {
+        return Ok(Transformed::no(expr));
+    }
+
+    let Ok((_, return_field)) = expr.to_field(schema) else {
+        // Can't determine return field; skip wrapping.
+        return Ok(Transformed::no(expr));
+    };
```
james-willis: should we throw an internal error in this case? I dont think it should happen but if it does, its probably an internal sedona db bug

james-willis: actually maybe this will tend to be caused by the user, like an ambiguous column name in the join result

james-willis (approved):

PR #974, opened 2026-06-17
feat(rust/sedona-raster-functions): add RS_Value for spatial 2d rasters

james-willis:


rust/sedona-raster-functions/src/rs_value.rs
```diff
@@ -0,0 +1,583 @@
...
+                )?))
+            }
+        }
+        (None, None) => Ok(None),
+        (Some(_), None) => {
+            exec_err!("RS_Value: point has a CRS but the raster does not")
+        }
+        (None, Some(_)) => {
+            exec_err!("RS_Value: raster has a CRS but the point does not")
+        }
```
james-willis, line 364: is erroring the right behavior or should we assume that they match?

docs/reference/sql/rs_value.qmd
```diff
@@ -0,0 +1,85 @@
...
+      description: Band index (1-based). Defaults to 1 if not specified.
+  - returns: float64
+    args:
+    - raster
+    - name: colX
+      type: int
+      description: Column index (1-based).
+    - name: rowY
+      type: int
+      description: Row index (1-based).
```
paleolimbot: Indexing is typically rows, cols (e.g., numpy, R) instead of cols, rows.

docs/reference/sql/rs_value.qmd
```diff
@@ -0,0 +1,85 @@
...
+```sql
+SELECT RS_Value(RS_Example(), 2, 1);
+```
+
+```sql
+SELECT RS_Value(RS_Example(), 2, 1, 2);
+```
+
+```sql
+SELECT RS_Value(RS_Example(), ST_SetCRS(ST_Point(74.58, 110.57), 'OGC:CRS84'));
```
paleolimbot:
```suggestion
SELECT RS_Value(RS_Example(), ST_Point(74.58, 110.57, 'OGC:CRS84');
```

rust/sedona-raster-functions/src/rs_value.rs
```diff
@@ -0,0 +1,583 @@
...
+// specific language governing permissions and limitations
+// under the License.
+
+//! `RS_Value` — sample a raster's pixel value at a point or grid cell.
+//!
+//! ```text
+//! RS_Value(raster, point)              -> Double  -- band defaults to 1
+//! RS_Value(raster, point, band)        -> Double
+//! RS_Value(raster, colX, rowY)         -> Double  -- 1-based grid coords, band 1
+//! RS_Value(raster, colX, rowY, band)   -> Double
```
paleolimbot: These two should really be a separate function (indexing order is unintuitive and conflicts with PostGIS ST_Value)

rust/sedona-raster-functions/src/rs_value.rs
```diff
@@ -0,0 +1,583 @@
...
+                )?
+                .clone(),
+            )
+        } else {
+            None
+        };
+        let mut band_iter = band_array.as_ref().map(|a| a.iter());
+
+        // Reprojecting the point into the raster CRS needs a PROJ engine.
+        with_global_proj_engine(|engine| {
```
paleolimbot, line 119: This can stay but I'll open a PR so that we don't have to do this...we shouldn't have to need a PROJ dependency for this to work (e.g., in the likely event all CRSes are equal, no need for PROJ).

rust/sedona-raster-functions/src/rs_value.rs
```diff
@@ -0,0 +1,583 @@
...
+                let wkb = reprojected.as_deref().unwrap_or(point_wkb);
+
+                let (x, y) = read_point_xy(wkb)?;
+                // Floor (not truncate toward zero) so a point just outside the
+                // top/left edge maps to a negative index and is rejected as out
+                // of bounds, rather than truncating to 0 and sampling an edge pixel.
+                let (raster_x, raster_y) = AffineMatrix::from_metadata(&raster.metadata())
+                    .inv_transform(x, y)
+                    .map_err(|e| exec_datafusion_err!("RS_Value: {e}"))?;
+                let (col, row) = (raster_x.floor() as i64, raster_y.floor() as i64);
```
paleolimbot: Can this math overflow an i64 and panic?

rust/sedona-raster-functions/src/rs_value.rs
```diff
@@ -0,0 +1,583 @@
...
+        let num_iterations = executor.num_iterations();
+        let mut builder = Float64Builder::with_capacity(num_iterations);
+
+        let col_array = as_int32_array(
+            &args[1]
+                .clone()
+                .cast_to(&DataType::Int32, None)?
+                .into_array(num_iterations)?,
+        )?
+        .clone();
```
paleolimbot: I trust the linter here but I am surprised you need to clone this (it's vaguely more expensive to clone an Int32Array than an ArrayRef).

rust/sedona-raster-functions/src/rs_value.rs
paleolimbot: It is probably the right behaviour for now. If this becomes annoying we can consider options (e.g., an "elided" CRS like `wk::wk_crs_inherit()` in R).

rust/sedona-raster-functions/src/rs_value.rs
```diff
@@ -0,0 +1,583 @@
...
+        (None, Some(_)) => {
+            exec_err!("RS_Value: raster has a CRS but the point does not")
+        }
+    }
+}
+
+/// Read the (x, y) coordinates of a WKB Point geometry.
+fn read_point_xy(wkb: &[u8]) -> Result<(f64, f64)> {
+    let geom = read_wkb(wkb).map_err(|e| DataFusionError::External(Box::new(e)))?;
+    match geom.as_type() {
```
paleolimbot: You can use `WkbHeader` for this, which is possibly faster

rust/sedona-raster-functions/src/rs_value.rs
```diff
@@ -0,0 +1,583 @@
...
+        .map_err(|_| exec_datafusion_err!("RS_Value: pixel byte offset overflow"))?;
+    let bytes = buffer.buffer.get(start..end).ok_or_else(|| {
+        exec_datafusion_err!("RS_Value: pixel is out of the band's buffer bounds")
+    })?;
+
+    // Decode the pixel to f64. The lossless converter errors (rather than
+    // silently rounding) on Int64/UInt64 values beyond f64's exact-integer
+    // range (2^53) — RS_Value returns a Double, so such a pixel can't be
+    // represented faithfully; failing loudly is preferred over a wrong value.
+    let value = nodata_bytes_to_f64_lossless(bytes, &buffer.data_type)
```
paleolimbot, line 518: 👍  (we'll have to figure out what to do about i64/u64 outputs at some point)

paleolimbot (commented):
Thank you! I took a first pass...I am not sure how far you want to go in optimizing this now, but there are a few things I think you'll want to consider even before we take a general optimization pass through all the functions.

In most of our functions we assume that whatever operation we're doing is sufficiently expensive that the per-iteration overhead of things like null checks and input validation are not measurable. Extracting a value out of a buffer is a nanosecond-scale operation depending mostly on whether it's cached or not, so per-item null checks (and especially the branching introduced by them) in this case are something that a production-grade raster operator worth its salt needs to avoid.

We can (probably) improve this by composing operations such that the actual sample loop has no branches.

- Validate raster band selection and build a `Vec<(&[u8], AffineMatrix)>`. In the possibly common case of a scalar raster, this check only has to happen once (instead of once per point).
- For the geometry case, validate the point geometry and extract the x, y as floats. You can use `WkbHeader` for this.
- Resolve whether there needs to be a CRS transform (I think we have a "are all these CRSes equal" function, which could be better optimized to check if a vector of CRSes has repeated values). You may want to punt on the CRSes not equal front for now.
- Build a "selection vector" to figure out which values of the geometry/row+col items you need to consider. You can start by intersecting the NullBuffers, and then you also need to exclude empty points. A selection vector is often `Vec<usize>` (DuckDB uses this trick to unify dictionary, run-end-encoding, and execution of scalar functions after a filter).
- Resolve the logical row/col in a branchless loop (separate cases for scalar vs. array raster)
- When you support views, make the logical indices physical ones
- Extract from the buffer in a branchless loop.

A good way to know you're on the right track is to benchmark `sd.sql("SELECT RS_Value($1, x, y)"`, params=(Raster.from_numpy(arr),)` against `arr.take()`, particularly after #999.

(Definitely prototype that suggestion before taking on the complexity!)

rust/sedona-raster-functions/src/rs_value.rs
paleolimbot: Also, `NaN` values from a POINT EMPTY might cause this to panic (worth testing if it is not tested already)

rust/sedona-raster-functions/src/rs_value.rs
james-willis: Ill defer these then.

rust/sedona-raster-functions/src/rs_value.rs
james-willis: done

docs/reference/sql/rs_value.qmd
james-willis: punted the col/row version of the function for now

docs/reference/sql/rs_value.qmd
james-willis: done

rust/sedona-raster-functions/src/rs_value.rs
james-willis: POINT EMPTY should now just return null

james-willis: I think overflow is very unlikely but I added a checked_mul operator here

rust/sedona-raster-functions/src/rs_value.rs
james-willis: refactored a bit to avoid the allocation

james-willis:
My inclination is to punt on the optimization work in order to add functionality. Also will eventually add a benchmark for these functions

docs/reference/sql/rs_value.qmd
james-willis: But the signature here is based on trying to match postgis as much as possible:

> Here are PostGIS's ST_Value (raster) signatures, straight from the current docs:
>  * ST_Value(raster rast, geometry pt, boolean exclude_nodata_value=true)
>  * ST_Value(raster rast, integer band, geometry pt, boolean exclude_nodata_value=true, text
>           resample='nearest')
>  * ST_Value(raster rast, integer x, integer y, boolean exclude_nodata_value=true)
>  * ST_Value(raster rast, integer band, integer x, integer y, boolean exclude_nodata_value=true)

james-willis: also sedona spark: https://sedona.apache.org/latest/api/sql/Raster-Operators/RS_Value/

paleolimbot: Sure, but nobody asked me to review those 🙂 . Compared to GDAL and xarray the PostGIS and Sedona Spark APIs are not particularly popular...part of porting that to SedonaDB is to make it fast, but if anybody is ever going to use this we're also going to have to make it less awkward.

rust/sedona-raster-functions/src/rs_value.rs
james-willis: been playing with the profiler. WkbHeader is slower.

```
  ┌─────────────────────────────────────────────────────────┬──────────┬─────────────────┐
  │                         option                          │ ns/point │   vs current    │
  ├─────────────────────────────────────────────────────────┼──────────┼─────────────────┤
  │ (a) WkbHeader::try_new + first_xy — current             │ 19.1     │ —               │
  ├─────────────────────────────────────────────────────────┼──────────┼─────────────────┤
  │ (b) wkb read_wkb + coord — what I replaced this session │ 7.5      │ −11.6 ns (2.5×) │
  ├─────────────────────────────────────────────────────────┼──────────┼─────────────────┤
  │ (c) hand-rolled minimal                                 │ 0.9      │ −18.2 ns (21×)  │
  └─────────────────────────────────────────────────────────┴──────────┴─────────────────┘
```

going to consider if I want to revert orproductionize the hand rolled point parser

james-willis: I went with c to build the point-optimized parser. I thought the perf gains were worth it: 50% e2e on the Scalar Raster case:

```
  ┌──────────────────────────────┬────────────────────────┐
  │             case             │         change         │
  ├──────────────────────────────┼────────────────────────┤
  │ ScalarArray (primary target) │ −56% (≈10.6 → 4.66 ms) │
  ├──────────────────────────────┼────────────────────────┤
  │ ArrayArray                   │ −11.7%                 │
  ├──────────────────────────────┼────────────────────────┤
  │ ArrayArrayScalar             │ −12.2%                 │
  └──────────────────────────────┴────────────────────────┘
```

We can reuse this implementation on other functions that only accept points

james-willis:
I spent a few hours doing benchmarking/profiling. I found that branch elimination did basically nothing. The big gains were in the Scalar raster case hoist, type level crs hoist, and the use of a point-specific wkb parser.

<details>
<summary>Detailed write-up (generated by Claude, the AI pair on this session)</summary>

### What worked

**CRS-decision hoist** — per point we were calling `crs_equals`, which allocates a `String` via `to_authority_code` (~15 ns/point, uniform across CRS types: OGC:CRS84, EPSG:4326, EPSG:3857 all ~14.7–14.9 ns). Resolving the reproject decision once per batch when the point CRS is column-level eliminates it.

**Band/affine/buffer hoist** — the scalar fast path resolves band buffer, nodata, affine transform, and raster CRS once for the whole batch instead of per point.

**Point-specific WKB reader** — the parse was the dominant remaining cost. A three-way comparison settled it:

| parse approach | ns/point |
| --- | --- |
| `WkbHeader::try_new` + `first_xy` (general) | 19.1 |
| `wkb` crate `read_wkb` + coord | 7.5 |
| hand-rolled point-only `read_point_xy` | 0.9 |

`read_point_xy` reads the coordinate from a fixed offset because a Point WKB has a trivial fixed layout.

### Benchmark result (criterion)

| case | change |
| --- | --- |
| ScalarArray (one raster, many points) | **−56%** |
| ArrayArray (per-row raster) | −11.7% |
| ArrayArrayScalar | −12.2% |

### What did NOT help

**Branch elimination** — profiling showed the branches guard operations that are already tiny: `inv_transform` ~1.5–2.0 ns/point, `read_pixel` ~3.2–3.4 ns/point. No headroom around 2–3 ns ops when the path is dominated by a ~19 ns parse.

### Takeaway

The two levers that mattered were both about doing work once instead of per-point (CRS decision, band/affine resolution) and not doing general-purpose work for a specific case (point-only parse). The micro-loop mechanics were never the bottleneck. Follow-up [DB-97](https://linear.app/wherobots/issue/DB-97/migrate-point-coord-consumers-to-wkb-headerread-point-xy) tracks spreading `read_point_xy` to other Point-only consumers.

</details>

rust/sedona-geometry/src/wkb_header.rs
```diff
@@ -1059,4 +1130,102 @@ mod tests {
...
+            0x00, 0x00, 0x00, 0x01, // type code 1 (Point)
+            0x3F, 0xF0, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, // x = 1.0
+            0x40, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, // y = 2.0
+        ];
+        assert_eq!(read_point_xy(&wkb).unwrap(), Some((1.0, 2.0)));
+    }
+
+    #[test]
+    fn read_point_xy_ewkb_with_srid() {
+        use sedona_testing::fixtures::*;
```
paleolimbot: This could probably be moved up to `mod test` since it shows up a few times

python/sedonadb/tests/functions/test_raster_functions.py
```diff
@@ -98,3 +98,36 @@ def test_rs_ensureloaded(con, sedona_testing):
...
     assert arr[0, 0] == 2324
+
+
+# Point sampling. RS_Example fills band `b` with the constant value `b`, except
+# the top-left pixel which is set to the nodata value (127). (74.58, 110.57) is
+# the centroid of pixel (10, 10) (0-based) in the raster's OGC:CRS84 space; the
+# point and raster share a CRS so no reprojection happens. A point far outside
+# the footprint yields NULL. (The `needs_pixels` -> RS_EnsureLoaded planner path
+# is covered against a real OutDb raster by `test_rs_ensureloaded`.)
+@pytest.mark.parametrize(
```
paleolimbot, line 113: One other test that would be valuable would be filling a numpy array with random values and sampling (1) every corner, center, and edge point (which is just the same grid shifted by 0.5 four ways) and (2) random points, maybe via sd_random_geometry. Then check both against rasterio for identical results. This could also use a fixture (I think we already have one for the sedona_testing directory).

paleolimbot (approved):
Thank you for the optimization goose chase here...50% better is great and this is also useful for non-raster PR reviews. WKB parsing and CRS equality are likely places we can improve performance elsewhere, too.

docs/reference/sql/rs_value.qmd
zhangfengcdt: Would it be better to keep st functions in sync between sedona and sedona_db? I understand the concern and bad decisions in the past on the orders of params, but if we'd make them different in signatures, users would need to change their SQL to adapt in between if they ever want to run on both engines.

rust/sedona-geometry/src/wkb_header.rs
james-willis: done

python/sedonadb/tests/functions/test_raster_functions.py
james-willis: done

paleolimbot (approved):
Thank you!

If the R failure persists I'll fix in a separate PR.

PR #975, opened 2026-06-17
feat(rust/sedona-raster-zarr): derive Zarr transform from spatial:bbox

james-willis:


rust/sedona-raster-zarr/src/geozarr.rs
```diff
@@ -122,16 +137,139 @@ fn parse_transform(
...
+///
+/// `bbox` is `[xmin, ymin, xmax, ymax]`. `registration` (default `"pixel"`) sets
+/// how the bbox relates to the grid, per the GeoZarr spatial convention
+/// (<https://github.com/zarr-conventions/spatial>): `"pixel"` means the bbox is
+/// the grid's outer edge, spanning all `N` cells (`scale = extent / N`) with the
+/// top-left corner on the bbox edge `(xmin, ymax)`; `"node"` means the bbox
+/// endpoints are the *centers* of the border cells, so `N` centers span `N - 1`
+/// intervals (`scale = extent / (N - 1)`) and the corner sits half a cell outside
+/// the bbox. `scale_y` is negative (rows increase downward).
+pub(crate) fn derive_transform_from_bbox(
```
paleolimbot: While you're here, can you move this function to sedona-raster (e.g., `AffineMatrix::from_bbox_and_spatial_shape()` or something). It would be especially nice to support this in `Raster.from_numpy()` (probably easier to do it in this PR but OK if you want it to be a follow-up). A bonus of doing that would be that you can do a sanity check against a test fixture (ask rasterio for the bounds of something and check its affine matrix).

rust/sedona-raster-zarr/src/loader.rs
paleolimbot, line 281: Not essential, but this free function is now a few hundred lines long and would benefit from being split (`try_from_bbox()`, `try_from_transform()`)

paleolimbot (approved):
A few things maybe worth taking care of here but in general looks good...thank you!

I am not sure if we're keeping a list of real-life public zarrs that we can support, but it might be a good idea to at least keep them in the doc header for open_and_validate if we don't have an analogue in sedona-testing.

james-willis:
> it might be a good idea to at least keep them in the doc header for open_and_validate if we don't have an analogue in sedona-testing.

They're like Pokemon, im going to catch them all 😄

rust/sedona-raster-zarr/src/loader.rs
james-willis: done

rust/sedona-raster-zarr/src/geozarr.rs
james-willis: I moved the function but did not do the from_numpy integration. tracked as a followup

paleolimbot (approved):
Thank you!

PR #979, opened 2026-06-18
feat(rust/sedona-query-planner): skip RS_EnsureLoaded on args returning loaded bytes

james-willis:
I noticed in #750 that rounds trips (ie `RS_DimToBand(RS_BandToDim(..),...)`) were failing because of this datafusion bug with nested Asyc UDFs: https://github.com/apache/datafusion/issues/20031

This mitigates this. Theres no reason to have multiple EnsureLoaded calls today so this is sufficient


paleolimbot (approved):
Thank you!

If the async UDFs/optimizer rule approach becomes too much we can reevaluate approach. It is very possible that what we really want is a dedicated `ExcecutionPlan` that collects its input and schedules byte loads + operations completely independently of scalar function execution.

PR #985, opened 2026-06-19
feat(rust/sedona-raster-zarr): resolve CRS from CF/rioxarray conventions

james-willis:
Reads CRS declared the CF / rioxarray way (`crs_wkt` / `spatial_ref`) in addition to GeoZarr `proj:*` — both at the group level and from a separate scalar grid-mapping variable (linked via `grid_mapping` or `coordinates`). Verified end-to-end against public CarbonPlan Zarr cubes.


paleolimbot (approved):
Thank you!

PR #989, opened 2026-06-20
feat(rust/sedona-functions): add ST_NumPoints alias for ST_NPoints

ajaypadwal73:
## Summary

- Adds `ST_NumPoints` as an alias for `ST_NPoints` in `st_points.rs`
- Adds a test verifying the alias is registered correctly

Closes part of #200.

## Changes

- `rust/sedona-functions/src/st_points.rs`: added `.with_aliases(vec!["st_numpoints".to_string()])` to `st_npoints_udf()` and a corresponding alias test


docs/reference/sql/st_npoints.qmd
```diff
@@ -29,6 +29,8 @@ kernels:
 
 Returns the total number of coordinate points in the geometry, counting all vertices across all components.
 
+This function also has the alias `ST_NumPoints`.
```
paleolimbot:
```suggestion
Aliases: `ST_NumPoints`.
```

Just following the existing pattern we have for that:

https://github.com/apache/sedona-db/blob/ce74a892ea7d8030cc48803c7e8ffa106bfb816f/docs/reference/sql/st_geomfromwkt.qmd#L38

paleolimbot (commented):
Thank you!

We actually have an implementation of `ST_NumPoints` in c/sedona-geos that we should remove if we go ahead with this. I believe there are some tests in python/sedonadb/tests/functions that assert the behaviour of ST_NumPoints and STNPoints against PostGIS, whose documentation says they are aliases but we should make sure (or deliberately treat them as aliases anyway).

I think the alias behaviour is the one we want (BigQuery and possibly others treat them as exact aliases too).

docs/reference/sql/st_npoints.qmd
ajaypadwal73: Done.

ajaypadwal73:
@paleolimbot Thanks for the context! Agreed - treating them as exact aliases (consistent with BigQuery) makes more sense than strict PostGIS LineString-only behavior. The old GEOS st_numpoints implementation has been removed and the Python tests updated to reflect the alias semantics.

python/sedonadb/tests/functions/test_functions.py
```diff
@@ -3772,23 +3772,23 @@ def test_st_numinteriorrings_basic(eng, geom, expected):
     ("geom", "expected"),
     [
         (None, None),
-        ("POINT EMPTY", None),
+        ("POINT EMPTY", 0),
         ("LINESTRING EMPTY", 0),
-        ("POLYGON EMPTY", None),
-        ("MULTIPOINT EMPTY", None),
-        ("MULTILINESTRING EMPTY", None),
-        ("MULTIPOLYGON EMPTY", None),
```
paleolimbot: Sorry to bikeshed here, but can we just remove this test and add a second line to the ST_NPoints test? That will ensure that the behaviour stays in sync and any added parameters always apply to both.

paleolimbot (approved):
Thank you! The PostGIS docs indicate that these functions are supposed to be aliases but they don't seem to be (as you just found out). I think an alias is the right move.

One note on the test but other that this is good to go!

python/sedonadb/tests/functions/test_functions.py
ajaypadwal73: @paleolimbot Done - standalone test_st_numpoints removed and ST_NumPoints assertion added directly into test_st_points alongside ST_NPoints, sharing the same parametrized cases.

paleolimbot (approved):

PR #990, opened 2026-06-20
feat: implement ST_BuildArea, ST_DelaunayTriangles, ST_ExteriorRing, ST_PointOnSurface, ST_NumInteriorRing alias

ajaypadwal73:
## Summary

Implements missing GEOS-backed functions from #224:

- `ST_BuildArea` — builds a polygon from linework using `build_area()`
- `ST_DelaunayTriangles` — Delaunay triangulation with optional tolerance using `delaunay_triangulation()`
- `ST_ExteriorRing` — returns exterior ring of a polygon using `get_exterior_ring()`
- `ST_PointOnSurface` — returns a point guaranteed to lie on/inside the geometry using `point_on_surface()`
- `ST_NumInteriorRing` — alias for `ST_NumInteriorRings` (both spellings registered)

Each function follows the existing pattern: WKB in → GEOS → WKB out via `GeosExecutor` + `write_geos_geometry`.

## Changes

- `c/sedona-geos/src/lib.rs` — module declarations (alphabetical order)
- `c/sedona-geos/src/register.rs` — kernel registrations (alphabetical order)
- `c/sedona-geos/src/st_buildarea.rs` — new
- `c/sedona-geos/src/st_delaunaytriangles.rs` — new
- `c/sedona-geos/src/st_exteriorring.rs` — new
- `c/sedona-geos/src/st_pointonsurface.rs` — new

Closes part of #224.


paleolimbot (commented):
Thank you!

I'll take a proper look on Monday, but I think the only missing component is Python integration tests that check this behaviour against PostGIS. Those are in python/sedonadb/tests/functions.

ajaypadwal73:
Hi @paleolimbot, added Python integration tests in python/sedonadb/tests/functions/test_functions.py covering ST_BuildArea, ST_DelaunayTriangles, ST_ExteriorRing, ST_NumInteriorRing, and ST_PointOnSurface - each parametrized to run against both SedonaDB and PostGIS.

CI caught a few mismatches in the initial expected values (vertex ordering, ST_PointOnSurface on a polygon-with-hole, and ST_DelaunayTriangles flag type); those are fixed in the latest commit. Ready for your review.

docs/reference/sql/st_buildarea.qmd
```diff
@@ -0,0 +1,42 @@
...
+    args: [geometry]
+  - returns: geography
+    args: [geography]
+---
+
+## Description
+
+Creates an areal geometry from linework. The input may be any collection of lines, polygons,
+or points. Lines are used to build the polygon boundary, and any unclosed lines are automatically
+closed. Returns NULL for non-polygon input that produces no area.
```
paleolimbot: Can you rewrite this in your own words? (Parts of this are a verbatim copy of the PostGIS documentation)

docs/reference/sql/st_exteriorring.qmd
```diff
@@ -0,0 +1,41 @@
...
+
+## Examples
+
+```sql
+SELECT ST_ExteriorRing(ST_GeomFromWKT('POLYGON ((0 0, 4 0, 4 4, 0 4, 0 0), (1 1, 1 3, 3 3, 3 1, 1 1))'));
+```
+
+```sql
+SELECT ST_ExteriorRing(ST_GeogFromWKT('POLYGON ((0 0, 4 0, 4 4, 0 4, 0 0))'));
+```
```
paleolimbot, line 41: Can you align these examples exactly so that the difference (geom vs geog) is more obvious?

python/sedonadb/tests/functions/test_functions.py
```diff
@@ -750,6 +750,23 @@ def test_st_buffer_style_parameters(
...
+@pytest.mark.parametrize(
+    ("geom", "expected"),
+    [
+        (None, None),
+        ("LINESTRING (0 0, 1 0, 1 1, 0 0)", "POLYGON ((0 0, 1 1, 1 0, 0 0))"),
+        (
+            "MULTILINESTRING ((0 0, 1 0, 1 1, 0 0), (2 2, 3 2, 3 3, 2 2))",
+            "MULTIPOLYGON (((1 1, 1 0, 0 0, 1 1)), ((3 3, 3 2, 2 2, 3 3)))",
+        ),
+    ],
```
paleolimbot, line 771: Can you add cases here for LINESTRING EMPTY and MULTILINESTRING EMPTY?

Can you a separate test to confirm what happens when a POINT or POLYGON arrives in this function? (I am guessing an error. You may have to do something like

```python
with pytest.raises(..., match="(sedona_error)|(postgis_error)"):`
```

...since the message might be slightly different.

We may have to decide how we handle non linestring EMPTY here (some PostGIS functions will error for an EMPTY of the wrong geometry type...in general in SedonaDB we handle them all the same, so an input of POINT EMPTY would give the same output as a LINESTRING EMPTY here).

python/sedonadb/tests/functions/test_functions.py
```diff
@@ -1421,6 +1438,28 @@ def test_st_dump(eng):
...
 
+@pytest.mark.parametrize("eng", [SedonaDB, PostGIS])
+@pytest.mark.parametrize(
+    ("geom", "expected"),
+    [
+        (None, None),
+        (
+            "POLYGON ((0 0, 1 0, 0.5 1, 0 0))",
+            "GEOMETRYCOLLECTION (POLYGON ((0.5 1, 0 0, 1 0, 0.5 1)))",
+        ),
```
paleolimbot, line 1496: Can you add cases for EMPTY input and LINESTRING input here?

Can you add a separate test with similar cases for the tolerance parameter?

python/sedonadb/tests/functions/test_functions.py
```diff
@@ -1451,6 +1490,29 @@ def test_st_envelope(eng, geom, expected):
...
+@pytest.mark.parametrize("eng", [SedonaDB, PostGIS])
+@pytest.mark.parametrize(
+    ("geom", "expected"),
+    [
+        (None, None),
+        (
+            "POLYGON ((0 0, 1 0, 1 1, 0 1, 0 0))",
+            "LINESTRING (0 0, 1 0, 1 1, 0 1, 0 0)",
+        ),
+        (
```
paleolimbot, line 1602: Can you add cases for EMPTY input?

python/sedonadb/tests/functions/test_functions.py
```diff
@@ -2659,6 +2721,26 @@ def test_st_pointn(eng, geometry, n, expected):
...
+@pytest.mark.parametrize("eng", [SedonaDB, PostGIS])
+@pytest.mark.parametrize(
+    ("geom", "expected"),
+    [
+        (None, None),
+        ("POLYGON ((0 0, 4 0, 4 4, 0 4, 0 0))", "POINT (2 2)"),
+        (
+            "POLYGON ((0 0, 4 0, 4 4, 0 4, 0 0), (1 1, 1 2, 2 2, 2 1, 1 1))",
+            "POINT (2 3)",
+        ),
```
paleolimbot, line 2836: Can you add cases for EMPTY input?

Can you cases for non-empty MULTIPOINT, MULTILINESTRING, and GEOMETRYCOLLECTION input?

python/sedonadb/tests/functions/test_functions.py
```diff
@@ -3767,6 +3849,28 @@ def test_st_numinteriorrings_basic(eng, geom, expected):
...
+        ("POINT (0 0)", None),
+        ("LINESTRING (0 0, 1 1)", None),
+    ],
+)
+def test_st_numinteriorring(eng, geom, expected):
+    eng = eng.create_or_skip()
+    eng.assert_query_result(
+        f"SELECT ST_NumInteriorRing({geom_or_null(geom)})",
+        expected,
+    )
```
paleolimbot: Because this is a true alias, we should just add another assert_query_result to the existing test for this function (ensures exact behaviour for the exact same cases)

c/sedona-geos/src/st_delaunaytriangles.rs
```diff
@@ -0,0 +1,152 @@
...
+
+impl SedonaScalarKernel for STDelaunayTriangles {
+    fn return_type(&self, args: &[SedonaType]) -> Result<Option<SedonaType>> {
+        let arg_matchers = if self.tolerance {
+            vec![ArgMatcher::is_geometry(), ArgMatcher::is_numeric()]
+        } else {
+            vec![ArgMatcher::is_geometry()]
+        };
+        let matcher = ArgMatcher::new(arg_matchers, WKB_GEOMETRY);
+        matcher.match_args(args)
```
paleolimbot: Can you implement this using two separate `SedonaScalarKernel` implementations rather than crunching them both into the same one? You can split out the logic into a common `invoke_scalar()`.

While you're here, is it easy to add support for the "flags" parameter? We can't support the 2 (TIN) option but I believe GEOS has the MULTILINESTRING output option.

paleolimbot (commented):
Thank you! A few notes but in general this looks great and I'm excited to have these.

Can you add an `Aliases: ` note for ST_NumInteriorRing documentation?

c/sedona-geos/src/st_exteriorring.rs
```diff
@@ -0,0 +1,153 @@
...
+        sedona_type: SedonaType,
+    ) {
+        let udf = SedonaScalarUDF::from_impl("st_exteriorring", st_exterior_ring_impl());
+        let tester = ScalarUdfTester::new(udf.into(), vec![sedona_type.clone()]);
+        tester.assert_return_type(sedona_type);
+
+        let result = tester
+            .invoke_scalar("POLYGON ((0 0, 4 0, 4 4, 0 4, 0 0))")
+            .unwrap();
+        tester.assert_scalar_result_equals(result, "LINEARRING (0 0, 4 0, 4 4, 0 4, 0 0)");
```
paleolimbot: This probably works if the wkt crate accepts LINEARRING, but LINESTRING is probably better:

```suggestion
        tester.assert_scalar_result_equals(result, "LINESTRING (0 0, 4 0, 4 4, 0 4, 0 0)");
```

c/sedona-geos/src/st_delaunaytriangles.rs
```diff
@@ -0,0 +1,234 @@
...
+#[derive(Debug)]
+struct STDelaunayTrianglesWithFlags;
+
+impl SedonaScalarKernel for STDelaunayTrianglesWithFlags {
+    fn return_type(&self, args: &[SedonaType]) -> Result<Option<SedonaType>> {
+        ArgMatcher::new(
+            vec![
+                ArgMatcher::is_geometry(),
+                ArgMatcher::is_numeric(),
+                ArgMatcher::is_boolean(),
```
paleolimbot: This works because 0 or 1 are the two values we accept, but probably this is better as `is_integer`.

c/sedona-geos/src/st_delaunaytriangles.rs
```diff
@@ -0,0 +1,234 @@
...
+    use sedona_schema::datatypes::{WKB_GEOMETRY, WKB_GEOMETRY_ITEM_CRS};
+    use sedona_testing::testers::ScalarUdfTester;
+
+    use super::*;
+
+    #[rstest]
+    fn udf_no_tolerance(#[values(WKB_GEOMETRY)] sedona_type: SedonaType) {
+        let udf = SedonaScalarUDF::from_impl("st_delaunaytriangles", st_delaunay_triangles_impl());
+        let tester = ScalarUdfTester::new(udf.into(), vec![sedona_type.clone()]);
+
```
paleolimbot, line 225: Now that you have the flags version, you should check to make sure it works (because the Rust tests' scope is to cover the branches of the Rust implementation)

c/sedona-geos/src/st_exteriorring.rs
```diff
@@ -0,0 +1,153 @@
...
+        })?;
+
+        executor.finish(Arc::new(builder.finish()))
+    }
+}
+
+fn invoke_scalar(geom: &Geometry) -> Result<Option<Vec<u8>>> {
+    let geom_type = geom
+        .geometry_type()
+        .map_err(|e| DataFusionError::Execution(format!("Failed to get geometry type: {e}")))?;
```
paleolimbot: Our other GEOS functions have an `invoke_scalar()` that accept a `mut impl Write` so that the GEOS output can be written to the output builder in place (rather than allocating a wkb byte buffer, then copying it into a new wkb byte buffer). I believe all the other functions that return a geometry do this but we may have missed some.

c/sedona-geos/src/st_pointonsurface.rs
```diff
@@ -0,0 +1,137 @@
...
+
+/// ST_PointOnSurface() implementation using the geos crate
+pub fn st_point_on_surface_impl() -> Vec<ScalarKernelRef> {
+    ItemCrsKernel::wrap_impl(vec![
+        Arc::new(STPointOnSurface {
+            matcher: ArgMatcher::new(vec![ArgMatcher::is_geometry()], WKB_GEOMETRY),
+        }),
+        Arc::new(STPointOnSurface {
+            matcher: ArgMatcher::new(vec![ArgMatcher::is_geography()], WKB_GEOGRAPHY),
+        }),
```
paleolimbot:
```suggestion
```

This one shouldn't apply to geography, since the algorithm is very much a planar one and may not be correct. We also implement this for geography already.

docs/reference/sql/st_delaunaytriangles.qmd
```diff
@@ -0,0 +1,46 @@
...
+description: Returns a geometry collection of Delaunay triangles covering the vertices of the input geometry.
+kernels:
+  - returns: geometry
+    args: [geometry]
+  - returns: geometry
+    args:
+      - geometry
+      - name: tolerance
+        type: float64
+        description: Snap-rounding tolerance for vertices. Use 0.0 for exact computation.
```
paleolimbot, line 29: Can you add the flags kernel here too now that it's implemented?

docs/reference/sql/st_delaunaytriangles.qmd
```diff
@@ -0,0 +1,46 @@
...
+---
+
+## Description
+
+Computes a Delaunay triangulation of the vertices of the input geometry. The result is a
+`GEOMETRYCOLLECTION` of `POLYGON` triangles. An optional tolerance value can be specified to
+snap nearby vertices together before triangulation.
```
paleolimbot:
```suggestion
Computes a Delaunay triangulation of the vertices of the input geometry. The result is a
`GEOMETRYCOLLECTION` of `POLYGON` triangles unless the `flags` parameter is set
to request `MULTILINESTRING` output. An optional tolerance value can be specified to
snap nearby vertices together before triangulation.
```

docs/reference/sql/st_delaunaytriangles.qmd
```diff
@@ -0,0 +1,46 @@
...
+
+## Examples
+
+```sql
+SELECT ST_DelaunayTriangles(ST_GeomFromWKT('MULTIPOINT ((0 0), (1 0), (0 1))'));
+```
+
+```sql
+SELECT ST_DelaunayTriangles(ST_GeomFromWKT('MULTIPOINT ((0 0), (1 0), (0 1))'), 0.0);
+```
```
paleolimbot, line 56: A flags example would be good, too

docs/reference/sql/st_exteriorring.qmd
```diff
@@ -0,0 +1,41 @@
...
+kernels:
+  - returns: geometry
+    args: [geometry]
+  - returns: geography
+    args: [geography]
+---
+
+## Description
+
+Returns the exterior ring (outer boundary) of a polygon as a `LINEARRING`. Returns NULL for
```
paleolimbot:
```suggestion
Returns the exterior ring (outer boundary) of a polygon as a `LINESTRING`. Returns NULL for
```

docs/reference/sql/st_exteriorring.qmd
```diff
@@ -0,0 +1,41 @@
...
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+title: ST_ExteriorRing
+description: Returns the exterior ring of a polygon as a linear ring geometry.
```
paleolimbot:
```suggestion
description: Returns the exterior ring of a polygon as a linestring geometry.
```

python/sedonadb/tests/functions/test_functions.py
```diff
@@ -750,6 +750,51 @@ def test_st_buffer_style_parameters(
...
+def test_st_buildarea_empty_linework(eng, geom, sedona_expected, postgis_expected):
+    is_postgis = eng is PostGIS
+    eng = eng.create_or_skip()
+    expected = postgis_expected if is_postgis else sedona_expected
+    eng.assert_query_result(f"SELECT ST_BuildArea({geom_or_null(geom)})", expected)
+
+
+@pytest.mark.parametrize("eng", [SedonaDB, PostGIS])
+def test_st_buildarea_non_linework(eng):
+    """POINT input forms no closed ring: PostGIS returns NULL, SedonaDB returns GEOMETRYCOLLECTION EMPTY."""
```
paleolimbot: We should catch this case and return NULL as well

paleolimbot (commented):
Thank you! A few comments but this is looking good!

ajaypadwal73:
@paleolimbot Thanks - addressed the latest round:

**ST_DelaunayTriangles**
- Switched the flags matcher from `is_boolean` to `is_integer`.
- Added a Rust test covering the flags (MULTILINESTRING output) branch.
- Docs: updated the description to mention the `flags` parameter and added a flags example.

**ST_ExteriorRing**
- Reworked `invoke_scalar` to write GEOS output directly into the builder via `mut impl Write`, avoiding the intermediate WKB buffer copy.
- Fixed the test assertion and docs (description + metadata) to use LINESTRING instead of LINEARRING.

**ST_PointOnSurface**
- Dropped the geography matcher — it's a planar algorithm and geography is already implemented separately.

**ST_BuildArea**
- Now catches the non-linework (POINT) case and returns NULL, matching PostGIS.

Let me know if anything else needs tweaking.

python/sedonadb/tests/functions/test_functions.py
```diff
@@ -750,6 +750,59 @@ def test_st_buffer_style_parameters(
...
+@pytest.mark.parametrize(
+    ("geom", "sedona_expected", "postgis_expected"),
+    [
+        # Both engines return an empty geometry for empty linework, not NULL.
+        # SedonaDB returns GEOMETRYCOLLECTION EMPTY; PostGIS returns POLYGON EMPTY.
+        ("LINESTRING EMPTY", "GEOMETRYCOLLECTION EMPTY", "POLYGON EMPTY"),
+        ("MULTILINESTRING EMPTY", "GEOMETRYCOLLECTION EMPTY", "POLYGON EMPTY"),
+    ],
+)
+def test_st_buildarea_empty_linework(eng, geom, sedona_expected, postgis_expected):
```
paleolimbot: Can we fix this so that we return a POLYGON EMPTY? This makes sense to me and should be possible.

python/sedonadb/tests/functions/test_functions.py
```diff
@@ -750,6 +750,59 @@ def test_st_buffer_style_parameters(
...
+    eng = eng.create_or_skip()
+    expected = postgis_expected if is_postgis else sedona_expected
+    eng.assert_query_result(f"SELECT ST_BuildArea({geom_or_null(geom)})", expected)
+
+
+@pytest.mark.parametrize("eng", [SedonaDB, PostGIS])
+@pytest.mark.parametrize(
+    ("geom", "sedona_expected", "postgis_expected"),
+    [
+        ("POINT (0 0)", None, None),
```
paleolimbot: Can we align our behaviour here? I think this is just checking geometry types.

python/sedonadb/tests/functions/test_functions.py
```diff
@@ -1421,6 +1474,88 @@ def test_st_dump(eng):
...
+def test_st_delaunaytriangles_tolerance(eng, geom, tolerance, expected):
+    eng = eng.create_or_skip()
+    if tolerance is None:
+        eng.assert_query_result(
+            f"SELECT ST_DelaunayTriangles({geom_or_null(geom)}, NULL)", expected
+        )
+    else:
+        eng.assert_query_result(
+            f"SELECT ST_DelaunayTriangles({geom_or_null(geom)}, {tolerance})", expected
+        )
```
paleolimbot: You may have to import this from sedonadb.testing but we have a helper for this one:

```suggestion
    eng.assert_query_result(
        f"SELECT ST_DelaunayTriangles({geom_or_null(geom)}, {val_or_null(tolerance)})", expected
    )
```

c/sedona-geos/src/st_pointonsurface.rs
```diff
@@ -0,0 +1,127 @@
...
+        })?;
+
+        executor.finish(Arc::new(builder.finish()))
+    }
+}
+
+fn invoke_scalar(geom: &Geometry, writer: &mut impl std::io::Write) -> Result<()> {
+    let result = geom
+        .point_on_surface()
+        .map_err(|e| DataFusionError::Execution(format!("ST_PointOnSurface failed: {e}")))?;
```
paleolimbot, line 82:
```suggestion
        .map_err(|e| exec_datafusion_err!("ST_PointOnSurface failed: {e}"))?;
```

c/sedona-geos/src/st_delaunaytriangles.rs
```diff
@@ -0,0 +1,288 @@
...
+fn invoke_scalar(
+    geom: &Geometry,
+    tolerance: f64,
+    only_edges: bool,
+    writer: &mut impl std::io::Write,
+) -> Result<()> {
+    let result = geom
+        .delaunay_triangulation(tolerance, only_edges)
+        .map_err(|e| DataFusionError::Execution(format!("ST_DelaunayTriangles failed: {e}")))?;
```
paleolimbot:
```suggestion
        .map_err(|e| exec_datafusion_err!("ST_DelaunayTriangles failed: {e}"))?;
```

paleolimbot (commented):
A few optional nits and there are a few place I think that it's easy enough to align our behaviour with PostGIS. Thank you!

c/sedona-geos/src/st_buildarea.rs
```diff
@@ -0,0 +1,156 @@
...
+
+/// ST_BuildArea() implementation using the geos crate
+pub fn st_build_area_impl() -> Vec<ScalarKernelRef> {
+    ItemCrsKernel::wrap_impl(vec![
+        Arc::new(STBuildArea {
+            matcher: ArgMatcher::new(vec![ArgMatcher::is_geometry()], WKB_GEOMETRY),
+        }),
+        Arc::new(STBuildArea {
+            matcher: ArgMatcher::new(vec![ArgMatcher::is_geography()], WKB_GEOGRAPHY),
+        }),
```
paleolimbot:
```suggestion
```

Sorry for missing this, but we can't support Geography with this one because the definition of "enclosing" isn't the same on the sphere.

c/sedona-geos/src/st_buildarea.rs
```diff
@@ -0,0 +1,156 @@
...
+    match geom_type {
+        GeometryTypes::LineString
+        | GeometryTypes::MultiLineString
+        | GeometryTypes::GeometryCollection => {}
+        _ => return Ok(false),
+    }
+
+    let result = geom
+        .build_area()
+        .map_err(|e| DataFusionError::Execution(format!("ST_BuildArea failed: {e}")))?;
```
paleolimbot:
```suggestion
        .map_err(|e| exec_datafusion_err!("ST_BuildArea failed: {e}"))?;
```

c/sedona-geos/src/st_delaunaytriangles.rs
paleolimbot:
```suggestion
        .map_err(|e| exec_datafusion_err!("ST_DelaunayTriangles failed: {e}"))?;
```

c/sedona-geos/src/st_exteriorring.rs
```diff
@@ -0,0 +1,159 @@
...
+        })?;
+
+        executor.finish(Arc::new(builder.finish()))
+    }
+}
+
+fn invoke_scalar(geom: &Geometry, writer: &mut impl std::io::Write) -> Result<bool> {
+    let geom_type = geom
+        .geometry_type()
+        .map_err(|e| DataFusionError::Execution(format!("Failed to get geometry type: {e}")))?;
```
paleolimbot:
```suggestion
        .map_err(|e| exec_datafusion_err!("Failed to get geometry type: {e}"))?;
```

c/sedona-geos/src/st_exteriorring.rs
```diff
@@ -0,0 +1,159 @@
...
+
+    if geom_type != GeometryTypes::Polygon {
+        return Ok(false);
+    }
+
+    let ring = geom
+        .get_exterior_ring()
+        .map_err(|e| DataFusionError::Execution(format!("ST_ExteriorRing failed: {e}")))?;
```
paleolimbot:
```suggestion
        .map_err(|e| exec_datafusion_err!("ST_ExteriorRing failed: {e}"))?;
```

c/sedona-geos/src/st_exteriorring.rs
```diff
@@ -0,0 +1,159 @@
...
+    let line =
+        Geometry::create_line_string(ring.get_coord_seq().map_err(|e| {
+            DataFusionError::Execution(format!("Failed to get ring coordinates: {e}"))
```
paleolimbot:
```suggestion
            exec_datafusion_err!("Failed to get ring coordinates: {e}")
```

c/sedona-geos/src/st_exteriorring.rs
```diff
@@ -0,0 +1,159 @@
...
+        })?)
+        .map_err(|e| {
+            DataFusionError::Execution(format!("Failed to create exterior linestring: {e}"))
```
paleolimbot:
```suggestion
            exec_datafusion_err!("Failed to create exterior linestring: {e}")
```

docs/reference/sql/st_buildarea.qmd
```diff
@@ -0,0 +1,41 @@
...
+# specific language governing permissions and limitations
+# under the License.
+
+title: ST_BuildArea
+description: Returns a geometry that encloses the area formed by the given linework.
+kernels:
+  - returns: geometry
+    args: [geometry]
+  - returns: geography
+    args: [geography]
```
paleolimbot:
```suggestion
```

paleolimbot (commented):
Thanks for fixing the R error!

A few outstanding comments on the PostGIS alignment and a few things I missed in a previous review (sorry!)

ajaypadwal73:
@paleolimbot could you please take a look at the latest changes? I’ve addressed your comments. Let me know if you have any remaining concerns.

python/sedonadb/tests/functions/test_functions.py
paleolimbot: Is there any objection to using the `val_or_null()` helper here?

python/sedonadb/tests/functions/test_functions.py
paleolimbot: Bumping this comment

python/sedonadb/tests/functions/test_functions.py
paleolimbot: Bumping this comment

paleolimbot (commented):
Thank you!

I think a few comments got lost in the mix. In particular, while we're here I'd like to align the behaviour in the two places I bumped: for empty input, we return POLYGON empty; for something that is already a POLYGON/MULTIPOLYGON, we return the input geometry. This behaviour seems intentional and well thought out, so I think we should replicate it.

ajaypadwal73:
@paleolimbot can you please take a stab at this?

paleolimbot (approved):
Thank you!

PR #991, opened 2026-06-20
chore: update Cargo.lock file 

blackrez:
I was trying to build sedondb-js and on linux it fails because aws-lc-rs used an old version in Cargo.lock. 

The bug seems to be fixed after 1.15.3.


blackrez:
It seems it breaks sedona-adbc. Will try to fix it when the CI of js binding will work.

blackrez:
> Thanks for opening!
> 
> We just updated a number of dependencies if you'd like to rebase and try again. If it doesn't work, I am guessing that the update to DataFusion 53 (sometime next week) will let us update the aws versions.

Let me retry the CI. I will check if the release of 0.4.0 solve the build issue of sedonadb-js.

PR #993, opened 2026-06-21
feat(rust/sedona-functions): implement ST_XxxFromText typed geometry constructors (issue #205)

### LINKED ISSUE #205: rust/sedona-functions: Implement `ST_xxxFromText()` functions
There are a number of variants of ST_GeomFromText that are identical except they error if a specific data type was not passed. These functions are:

- ST_GeomCollFromText
- ST_MPointFromText
- ST_LineFromText
- ST_LineStringFromText
- ST_MLineFromText
- ST_MPointFromText
- ST_MPolyFromText
- ST_PointFromText
- ST_PolygonFromText

I believe these could all be implemented by adding an (optional) "expected geometry type" here:

https://github.com/apache/sedona-db/blob/6d0b218844fe50d6c4d2b3a2d15ee124baecf110/rust/sedona-functions/src/st_geomfromwkt.rs#L127-L133

...and add an `expected_geometry_type: Option<GeometryTypeId>` here:

https://github.com/apache/sedona-db/blob/6d0b218844fe50d6c4d2b3a2d15ee124baecf110/rust/sedona-functions/src/st_geomfromwkt.rs#L87-L90

Then these functions can be constructed and declared here with their documentation:

https://github.com/apache/sedona-db/blob/6d0b218844fe50d6c4d2b3a2d15ee124baecf110/rust/sedona-functions/src/st_geomfromwkt.rs#L37-L41

We should ensure that this doesn't slow down the general case, although I don't think that in the context of WKT parsing this will be measurable.

I don't think we need Geography constructors like this, because nobody else provides them either (these functions would be for OGC/PostGIS compatibility).

ajaypadwal73:
## Summary

Closes #205.

Implements the typed WKT constructor functions listed in the issue:

- `ST_GeomCollFromText` — errors if input is not a GeometryCollection
- `ST_LineFromText` — errors if input is not a LineString (alias: `ST_LineStringFromText`)
- `ST_MLineFromText` — errors if input is not a MultiLineString
- `ST_MPointFromText` — errors if input is not a MultiPoint
- `ST_MPolyFromText` — errors if input is not a MultiPolygon
- `ST_PointFromText` — errors if input is not a Point
- `ST_PolygonFromText` — errors if input is not a Polygon

## Implementation

Added `ExpectedGeomType` enum and an `expected_geom_type: Option<ExpectedGeomType>` field to the existing `STGeoFromWKT` struct. All parsing, SRID, and CRS propagation logic is shared with `ST_GeomFromWKT` — no duplication.

A `make_typed_geom_udf` helper constructs each function with one line.

## Tests

- `typed_constructors_accept_correct_type` — all 7 functions accept their matching WKT type
- `typed_constructors_reject_wrong_type` — error message includes expected type name
- `typed_constructors_accept_srid` — all 7 functions work with the optional SRID argument

## Docs

Reference `.qmd` added for each function including SRID examples. CI doc check passes (`Missing docs: 0, Missing alias mentions: 0`).


docs/reference/sql/st_linefromtext.qmd
```diff
@@ -0,0 +1,50 @@
...
+      type: string
+    - name: srid
+      type: crs
+---
+
+## Description
+
+Parses a WKT string and returns a geometry. Raises an error if the WKT does not represent a LineString.
+
+This function also has the alias `ST_LineStringFromText`.
```
paleolimbot:
```suggestion
Aliases: `ST_LineStringFromText`.
```

docs/reference/sql/st_polygonfromtext.qmd
```diff
@@ -0,0 +1,48 @@
...
+
+Parses a WKT string and returns a geometry. Raises an error if the WKT does not represent a Polygon.
+
+## Examples
+
+```sql
+SELECT ST_PolygonFromText('POLYGON ((0 0, 4 0, 4 4, 0 4, 0 0))');
+```
+
+With an SRID:
```
paleolimbot:
```suggestion
With an SRID or CRS:
```

(For the others as well)

rust/sedona-functions/src/st_geomfromwkt.rs
```diff
@@ -38,13 +38,26 @@ use wkt::Wkt;
 use crate::executor::WkbExecutor;
 use crate::st_setsrid::SRIDifiedKernel;
 
+/// Geometry type constraint used by ST_XxxFromText functions.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+enum ExpectedGeomType {
```
paleolimbot: Can you use `sedona_geometry::...::GeometryTypeId` for this to avoid the duplicated enum?

rust/sedona-functions/src/st_geomfromwkt.rs
```diff
@@ -468,5 +573,89 @@ mod tests {
...
         let udf: ScalarUDF = st_geogfromwkt_udf().into();
         assert!(udf.aliases().contains(&"st_geogfromtext".to_string()));
+
+        let udf: ScalarUDF = st_linefromtext_udf().into();
+        assert!(udf.aliases().contains(&"st_linestringfromtext".to_string()));
+    }
+
+    #[test]
+    fn typed_constructors_accept_correct_type() {
+        let cases: &[(&str, SedonaScalarUDF)] = &[
```
paleolimbot: Can you use `rstest` parameterized tests for consistency here?

paleolimbot (commented):
Thank you!

Can you add Python tests to confirm parity with PostGIS here?

Can you also confirm the behaviour for EMPTY types? (i.e., is `LINESTRING EMTPY` rejected by ST_GeomCollFromText()`)?

ajaypadwal73:
Hi @paleolimbot , I've addressed all the feedback:

ExpectedGeomType → GeometryTypeId: Replaced the custom enum with sedona_geometry::types::GeometryTypeId and switched error messages to use .geojson_id().
rstest parametrize: All three Rust test functions (accept_correct_type, accept_srid, reject_wrong_type) now use #[rstest] + #[case] covering all 7 constructors.
Python integration tests: Added tests for correct type acceptance, wrong type rejection (including wrong EMPTY types for all 7 constructors - covers your question about LINESTRING EMPTY in ST_GeomCollFromText), SRID, matching EMPTY input, null input, and the ST_LineStringFromText alias - all parametrized against both SedonaDB and PostGIS.
Docs: Aliases: \ST_LineStringFromText`inst_linefromtext.qmd; With an SRID or CRS:` across all 7 reference pages.

paleolimbot (approved):
Thank you!

PR #994, opened 2026-06-21
feat(c/sedona-geos): add geometry support for ST_LineLocatePoint, ST_MaxDistance, ST_ReducePrecision

### LINKED ISSUE #900: Implement functions for geometry that are already supported in geography
A few functions have geography support with no corresponding geometry support:

- ST_PointOnSurface
- ST_ReducePrecision
- ST_MaxDistance
- ST_LineLocatePoint

ajaypadwal73:
## Summary

Closes #900.

These three functions had geography implementations in `c/sedona-s2geography` but no geometry counterpart in `c/sedona-geos`.

- **ST_LineLocatePoint**: uses GEOS `project_normalized()`; returns NULL for non-LineString first argument (consistent with PostGIS behavior)
- **ST_MaxDistance**: recursive vertex extraction across geometry trees, returns max pairwise 2D Cartesian distance
- **ST_ReducePrecision**: uses GEOS `set_precision()` with `Precision::ValidOutput` flag

## Test plan

- [x] Unit tests for all three functions (correct output, NULL input, array batch, item-CRS passthrough for `ST_ReducePrecision`)
- [x] `ST_LineLocatePoint` non-LineString guard tested (returns NULL)
- [x] `cargo clippy -p sedona-geos` — zero warnings
- [x] Doc CI check: `cargo run -p sedona-cli -- list-functions | python3 ci/scripts/check_sql_function_docs.py -` — 0 missing docs


c/sedona-geos/src/st_line_locate_point.rs
```diff
@@ -0,0 +1,147 @@
...
+use geos::{Geom, GeometryTypes};
+use sedona_expr::scalar_udf::{ScalarKernelRef, SedonaScalarKernel};
+use sedona_schema::{datatypes::SedonaType, matchers::ArgMatcher};
+
+use crate::executor::GeosExecutor;
+
+/// ST_LineLocatePoint() implementation using the geos crate
+pub fn st_line_locate_point_impl() -> ScalarKernelRef {
+    Arc::new(STLineLocatePoint {})
+}
```
paleolimbot, line 35: I believe this should also handle item_crs via ItemCrsKernel

c/sedona-geos/src/st_line_locate_point.rs
```diff
@@ -0,0 +1,147 @@
...
+
+        executor.finish(Arc::new(builder.finish()))
+    }
+}
+
+fn invoke_scalar(line: &geos::Geometry, point: &geos::Geometry) -> Result<Option<f64>> {
+    match line
+        .geometry_type()
+        .map_err(|e| DataFusionError::Execution(format!("Failed to get geometry type: {e}")))?
+    {
```
paleolimbot, line 87: This can also check is_empty before we get here to consistently handle empty inputs (regardless of geometry type). This is a slight deviation from PostGIS but is more consistent with our overall approach (empty input is empty input).

c/sedona-geos/src/st_line_locate_point.rs
```diff
@@ -0,0 +1,147 @@
...
+    use sedona_schema::datatypes::{WKB_GEOMETRY, WKB_VIEW_GEOMETRY};
+    use sedona_testing::compare::assert_array_equal;
+    use sedona_testing::create::create_array;
+    use sedona_testing::testers::ScalarUdfTester;
+
+    use super::*;
+
+    #[rstest]
+    fn udf(#[values(WKB_GEOMETRY, WKB_VIEW_GEOMETRY)] sedona_type: SedonaType) {
+        let udf = SedonaScalarUDF::from_impl("st_linelocatepoint", st_line_locate_point_impl());
```
paleolimbot, line 115: Because your error behaviour is implement here and not coming from GEOS, can you also check your error here?

When you handle the empties, that test can go here too.

c/sedona-geos/src/st_max_distance.rs
```diff
@@ -0,0 +1,173 @@
...
+use geos::{GResult, Geom, GeometryTypes};
+use sedona_expr::scalar_udf::{ScalarKernelRef, SedonaScalarKernel};
+use sedona_schema::{datatypes::SedonaType, matchers::ArgMatcher};
+
+use crate::executor::GeosExecutor;
+
+/// ST_MaxDistance() implementation using the geos crate
+pub fn st_max_distance_impl() -> ScalarKernelRef {
+    Arc::new(STMaxDistance {})
+}
```
paleolimbot: Because you are using a brute force approach here, that implementation is not a good fit for GEOS. Instead, sedona-geo-generic-alg is probably a better fit (that is where distance is implemented as well). The difference is that GEOS has quite a bit of per-geometry overhead (full copy!) that we've figured out how to avoid elsewhere.

docs/reference/sql/st_linelocatepoint.qmd
```diff
@@ -19,16 +19,28 @@
...
   - returns: double
     args: [geography, geography]
 ---
 
 ## Description
 
-Returns a double between 0 and 1, representing the location of the closest point on the LineString as a fraction of its total length. The first argument must be a LINESTRING, and the second argument is a POINT geometry or geography. For geography, the fraction is computed based on spherical interpolation between vertices.
+Returns a double between 0 and 1, representing the location of the closest point on the LineString as a fraction of its total 2D length. The first argument must be a LineString.
+
+For geometry, the fraction is computed using 2D Cartesian distance. For geography, the fraction is computed based on spherical interpolation between vertices.
```
paleolimbot: For silly reasons, this will render better in some of the generated forms of this documentation if you collapse it into a single paragraph

```suggestion
Returns a double between 0 and 1, representing the location of the closest point on the LineString as a fraction of its total 2D length. The first argument must be a LineString. For geometry, the fraction is computed using 2D Cartesian distance. For geography, the fraction is computed based on spherical interpolation between vertices.
```

docs/reference/sql/st_linelocatepoint.qmd
```diff
@@ -19,16 +19,28 @@
...
 
 ## Examples
 
+```sql
+SELECT ST_LineLocatePoint(
+    ST_GeomFromText('LINESTRING (0 0, 0 1)'),
+    ST_GeomFromText('POINT (0 0.5)')
+);
+-- returns 0.5
```
paleolimbot: This is rendered automatically by the doc build script

```suggestion
```

docs/reference/sql/st_maxdistance.qmd
```diff
@@ -19,16 +19,27 @@
...
+    args: [geometry, geometry]
   - returns: double
     args: [geography, geography]
 ---
 
 ## Description
 
+For geometry, returns the 2D Cartesian maximum distance between any pair of vertices across the two geometries, in the units of the coordinate reference system.
+
 For geography, returns the maximum distance in meters calculated using a spherical approximation.
```
paleolimbot:
```suggestion
For geometry, returns the 2D Cartesian maximum distance between any pair of vertices across the two geometries, in the units of the coordinate reference system. For geography, returns the maximum distance in meters calculated using a spherical approximation.
```

paleolimbot (commented):
Thank you!

Can you add Python integration tests for these functions that verify PostGIS compatibility. You should be able to draw from the Geography tests for these functions.

ajaypadwal73:
Hi @paleolimbot, I've addressed all the review comments. Here's a summary of what changed:

1. ST_MaxDistance: moved from sedona-geos to sedona-geo, now uses GeoTypesExecutor + geo::CoordsIter to avoid the per-geometry GEOS copy overhead
2. ST_LineLocatePoint: wrapped with ItemCrsKernel::wrap_impl for CRS-tagged geometry support; added is_empty() guard returning NULL for empty inputs; non-LineString first arg now returns NULL
3. Python integration tests: added tests for ST_LineLocatePoint, ST_MaxDistance, and ST_ReducePrecision; PostGIS-incompatible case (non-LineString first arg raises in PostGIS but returns NULL in SedonaDB) isolated in a SedonaDB-only test
4. Docs: collapsed multi-paragraph descriptions in ST_LineLocatePoint and ST_MaxDistance reference pages

Let me know if anything needs to be changed!

ajaypadwal73:
@paleolimbot Is the SedonaDB NULL-on-empty behavior the intended long-term design, or should we align with PostGIS's 0.0 return? The current approach seems safer, but worth confirming.

c/sedona-geos/src/st_line_locate_point.rs
```diff
@@ -0,0 +1,187 @@
...
+
+        executor.finish(Arc::new(builder.finish()))
+    }
+}
+
+fn invoke_scalar(line: &geos::Geometry, point: &geos::Geometry) -> Result<Option<f64>> {
+    // Empty input -> NULL (consistent with SedonaDB's general approach to empties)
+    let line_empty = line
+        .is_empty()
+        .map_err(|e| DataFusionError::Execution(format!("Failed to check empty: {e}")))?;
```
paleolimbot, line 76: I *think* this is the right thing to do but honestly either is great and we can update if we get a request one way or the other.

python/sedonadb/tests/functions/test_functions.py
```diff
@@ -2304,6 +2304,116 @@ def test_st_length(eng, geom, expected):
...
+def test_st_linelocatepoint_empty_inputs(
+    eng, line, point, expected_sedona, expected_postgis
+):
+    is_postgis = eng is PostGIS
+    eng = eng.create_or_skip()
+    expected = expected_postgis if is_postgis else expected_sedona
+    eng.assert_query_result(
+        f"SELECT ST_LineLocatePoint({geom_or_null(line)}, {geom_or_null(point)})",
+        expected,
+    )
```
paleolimbot: I think we can just test our own behaviour if it's different than PostGIS (there are a few places we do this)

```suggestion
@pytest.mark.parametrize("eng", [SedonaDB])
@pytest.mark.parametrize(
    ("line", "point", "expected"),
    [
        ("LINESTRING EMPTY", "POINT (0 0)", None),
        ("LINESTRING (0 0, 1 1)", "POINT EMPTY", None),
    ],
)
def test_st_linelocatepoint_empty_inputs(
    eng, line, point, expected
):
    eng = eng.create_or_skip()
    eng.assert_query_result(
        f"SELECT ST_LineLocatePoint({geom_or_null(line)}, {geom_or_null(point)})",
        expected,
    )
```

python/sedonadb/tests/functions/test_functions.py
```diff
@@ -2304,6 +2304,116 @@ def test_st_length(eng, geom, expected):
...
+
+
+@pytest.mark.parametrize("eng", [SedonaDB])
+def test_st_linelocatepoint_non_linestring_returns_null(eng):
+    # PostGIS raises an error for non-LineString first arg; SedonaDB returns NULL
+    eng = eng.create_or_skip()
+    eng.assert_query_result(
+        "SELECT ST_LineLocatePoint(ST_GeomFromText('POLYGON ((0 0, 1 0, 1 1, 0 0))'), ST_GeomFromText('POINT (0.5 0.5)'))",
+        None,
+    )
```
paleolimbot: I think Geography's implementation of this errors but I'm not sure. We should probably do whatever geography does here...it's also reasonable to error I think for non-empty non-linestrings.

rust/sedona-functions/src/st_max_distance.rs
```diff
@@ -0,0 +1,170 @@
...
+        matcher.match_args(args)
+    }
+
+    fn invoke_batch(
+        &self,
+        arg_types: &[SedonaType],
+        args: &[ColumnarValue],
+    ) -> Result<ColumnarValue> {
+        let executor = GeoTypesExecutor::new(arg_types, args);
+        let mut builder = Float64Builder::with_capacity(executor.num_iterations());
```
paleolimbot, line 65: If I recall correctly I believe this will error for POINT EMPTY. The other kernels we have here special case those, possibly using a WkbExecutor. In either case, we should ensure that POINT EMPTY works here.

python/sedonadb/tests/functions/test_functions.py
```diff
@@ -2304,6 +2304,116 @@ def test_st_length(eng, geom, expected):
...
+
+
+@pytest.mark.parametrize("eng", [SedonaDB, PostGIS])
+@pytest.mark.parametrize(
+    ("geom1", "geom2", "expected"),
+    [
+        (None, None, None),
+        ("POINT (0 0)", "POINT (3 4)", 5.0),
```
paleolimbot, line 2359: Can you add a test for POINT EMPTY here, too?

paleolimbot (commented):
Thank you...this is very close! Just some tiny details 🙂

ajaypadwal73:
@paleolimbot Thanks - addressed the remaining items:

- **ST_LineLocatePoint non-LineString**: aligned with Geography's behavior (errors on a non-empty, non-LineString first arg rather than returning NULL). Updated the test assertion to match.
- **ST_LineLocatePoint empty-input test**: dropped the PostGIS parametrization, now a SedonaDB-only test for the NULL-on-empty behavior, per your suggestion.
- **ST_MaxDistance POINT EMPTY**: special-cased empty geometries so they no longer error in the GeoTypesExecutor path.
- **ST_MaxDistance test**: added a POINT EMPTY case alongside the existing ones.
- Left the NULL-on-empty behavior in ST_LineLocatePoint as-is, since you noted either is fine.

Let me know if anything else needs tweaking.

rust/sedona-geo/src/st_max_distance.rs
```diff
@@ -0,0 +1,239 @@
...
+    fn invoke_batch(
+        &self,
+        arg_types: &[SedonaType],
+        args: &[ColumnarValue],
+    ) -> Result<ColumnarValue> {
+        let executor = WkbExecutor::new(arg_types, args);
+        let mut builder = Float64Builder::with_capacity(executor.num_iterations());
+        executor.execute_wkb_wkb_void(|lhs, rhs| {
+            match (lhs, rhs) {
+                (Some(lhs), Some(rhs)) => match invoke_scalar(lhs, rhs) {
```
paleolimbot: This may not make a difference, but in general we try to make a passing effort to reuse allocations between iterations of a loop over geometries. In this case it would mean declaring the coordinate vectors up front and `.clear()`ing them on each iteration (which keep the allocated capacity but resets the length to zero).

```suggestion
        let mut coords_lhs_scratch = Vec::new();
        let mut coords_rhs_scratch = Vec::new();
        executor.execute_wkb_wkb_void(|lhs, rhs| {
            match (lhs, rhs) {
                (Some(lhs), Some(rhs)) => match invoke_scalar(lhs, rhs, &mut coords_lhs_scratch, &mut coords_rhs_scratch) {
```

rust/sedona-geo/src/st_max_distance.rs
```diff
@@ -0,0 +1,239 @@
...
+            }
+            Ok(())
+        })?;
+        executor.finish(Arc::new(builder.finish()))
+    }
+}
+
+fn invoke_scalar(lhs: &Wkb, rhs: &Wkb) -> Option<f64> {
+    let mut lhs_coords = Vec::new();
+    let mut rhs_coords = Vec::new();
```
paleolimbot:
```suggestion
fn invoke_scalar(lhs: &Wkb, rhs: &Wkb, lhs_coords: &mut Vec<(f64, f64)>, rhs_coords: &mut Vec<(f64, f64)>) -> Option<f64> {
    lhs_coords.clear();
    rhs_coords.clear();
```

rust/sedona-geo/src/st_max_distance.rs
```diff
@@ -0,0 +1,239 @@
...
+};
+use sedona_expr::{
+    item_crs::ItemCrsKernel,
+    scalar_udf::{ScalarKernelRef, SedonaScalarKernel},
+};
+use sedona_functions::executor::WkbExecutor;
+use sedona_schema::{datatypes::SedonaType, matchers::ArgMatcher};
+use wkb::reader::Wkb;
+
+/// ST_MaxDistance() — max pairwise 2D vertex distance using geo-types CoordsIter
```
paleolimbot:
```suggestion
/// ST_MaxDistance() — max pairwise 2D vertex distance using geo-traits iteration
```

rust/sedona-functions/src/st_max_distance.rs
paleolimbot, line 1: This is no longer using the geo crate, so it can live in rust/sedona-functions!

paleolimbot (commented):
Thank you for these...just a few minor comments on the new max distance implementation!

ajaypadwal73:
Hey @paleolimbot, thanks for the review! I’ve addressed all the comments. Could you please take another look when you get a chance?

c/sedona-geos/src/st_line_locate_point.rs
```diff
@@ -0,0 +1,185 @@
...
+        return Ok(None);
+    }
+
+    match line
+        .geometry_type()
+        .map_err(|e| DataFusionError::Execution(format!("Failed to get geometry type: {e}")))?
+    {
+        GeometryTypes::LineString => {
+            let fraction = line.project_normalized(point).map_err(|e| {
+                DataFusionError::Execution(format!("Failed to locate point on line: {e}"))
```
paleolimbot, line 90: In general we try to use the abbreviated forms:

```suggestion
                exec_datafusion_err!("Failed to locate point on line: {e}")
```

paleolimbot (approved):
Thank you!

Let's merge and if we have time later we can clean up the `Err(DataFusionError::Execution(format!`s a another time.

PR #999, opened 2026-06-22
feat(sedonadb/python): Add zero copy raster IO to/from numpy

paleolimbot:
This PR adds zero-coy support to/from numpy. I had thought my initial support was zero copy but there are a few operations that force a binaryview copy in pyarrow that make it tricky to make this work.

```python
import sedona.db
from sedonadb.raster import Raster
import numpy as np

arr = np.arange(100 * 100).reshape((100, 100))
r = Raster.from_numpy(arr) 
assert np.shares_memory(r.bands[0].to_numpy(), arr)

sd = sedona.db.connect()
df = sd.sql("SELECT $1 as raster", params=(r, )).to_pandas()

r_from_df = df.iloc[0, 0]
assert np.shares_memory(r_from_df.bands[0].to_numpy(), arr)
```


python/sedonadb/python/sedonadb/raster.py
```diff
@@ -390,13 +311,17 @@ def data(self) -> memoryview:
...
-        """Convert this band's data to a numpy array."""
+        """Convert this band's data to a numpy array (zero-copy when possible)."""
         import numpy as np
 
         if self.data_size == 0:
             return np.empty(self.shape, dtype=self.data_type)
 
-        return np.array(self.data)
+        # Use frombuffer for zero-copy when data is loaded
+        if self.outdb_uri is not None and len(self.source_data) == 0:
```
james-willis: the visible shape is how we determine if data is loaded or not. `len(self.source_data) == 0` can mean loaded if the product of the elements of the shape vector is 0.

paleolimbot: Ah sorry, you fixed that for me in a previous PR but it got lost in the shuffle. I'll fix.

python/sedonadb/python/sedonadb/raster.py
```diff
@@ -476,3 +456,246 @@ def register_extension_type():
...
+        data = np.ascontiguousarray(data)
+        view = memoryview(data)
+
+    nbytes = view.nbytes
+    if nbytes == 0:
+        return pa.array([b""], type=pa.binary_view())
+
+    # Build BinaryArray zero-copy using from_buffers, then cast to binary_view
+    # BinaryArray buffers: [validity, offsets, data]
+    offsets = pa.py_buffer(np.array([0, nbytes], dtype=np.int32))
```
james-willis, line 574: I think we need to check if nbytes is addressable by the int32 from the BinaryView. 

```python
if nbytes > 2**31 - 1:
    raise ValueError(
        f"Cannot store {nbytes} bytes in a single raster band: "
        "the Arrow BinaryView storage type limits a band to 2 GiB. "
        "Split the raster into multiple bands or tiles."
    )
```

python/sedonadb/python/sedonadb/raster.py
```diff
@@ -476,3 +456,246 @@ def register_extension_type():
...
+        )
+
+    # spatial_dims / spatial_shape reference the trailing (y, x) axes, in x,y order.
+    y_name, x_name = dim_names[-2], dim_names[-1]
+    height, width = shape[-2], shape[-1]
+
+    # Build band struct array with zero-copy data field
+    band_data_array = _wrap_data_zero_copy(data)
+
+    band_struct = pa.StructArray.from_arrays(
```
james-willis, line 635: do we really need to redefine the schema here or can we use RASTER_STORAGE_TYPE?

python/sedonadb/python/sedonadb/raster.py
```diff
@@ -390,13 +311,17 @@ def data(self) -> memoryview:
...
 
         if self.data_size == 0:
             return np.empty(self.shape, dtype=self.data_type)
 
-        return np.array(self.data)
+        # Use frombuffer for zero-copy when data is loaded
+        if self.outdb_uri is not None and len(self.source_data) == 0:
+            raise ValueError("Can't extract buffer from a reference to external data.")
+
+        return np.frombuffer(self.source_data, dtype=self.data_type).reshape(self.shape)
```
james-willis: is this cast duplicate to line 311?

python/sedonadb/python/sedonadb/raster.py
paleolimbot: Unfortunately, we do. This is because we need to use StructArray.from_arrays() specifically to avoid the copy.

james-willis: claude seems to think you can put a dummy data field in to avoid redefining the schema, then update the data field in a zero copy manner. My arrow knowledge is pretty limited here.

<details>
the claude code:

```python
def _replace_child(struct_arr, name, new_child):
    """Return a copy of a StructArray with one named child replaced, deriving
    every field definition (type, order, nullability) from the array's own
    type — nothing about the schema is restated here."""
    t = struct_arr.type
    children = [
        new_child if t.field(i).name == name else struct_arr.field(i)
        for i in range(t.num_fields)
    ]
    fields = [t.field(i) for i in range(t.num_fields)]
    return pa.StructArray.from_arrays(children, fields=fields)


def _with_band_data(raster_array, data_array):
    """Inject `data_array` as the single band's `data` field, zero-copy,
    without touching any other buffer."""
    bands = raster_array.field("bands")
    new_band = _replace_child(bands.values, "data", data_array)
    # pass type= to preserve the list's non-null `item` field
    new_bands = pa.ListArray.from_arrays(bands.offsets, new_band, type=bands.type)
    return _replace_child(raster_array, "bands", new_bands)


def _build_raster(dim_names, shape, data_type_id, *, data, crs=None,
                  nodata=None, transform=None, outdb_uri=None, outdb_format=None):
    if crs is not None:
        crs = gat.type_spec(crs=crs).crs.to_json()
    nodata_bytes = None
    if nodata is not None:
        nodata_bytes = struct.pack("<" + BAND_DATA_TYPE_STRUCT_CHARS[data_type_id], nodata)

    y_name, x_name = dim_names[-2], dim_names[-1]
    height, width = shape[-2], shape[-1]

    # Build the whole structure declaratively with an empty `data` placeholder,
    # so types / field order / nullability all come from the canonical type.
    raster = {
        "crs": crs,
        "transform": list(transform) if transform is not None
                     else [0.0, 1.0, 0.0, 0.0, 0.0, -1.0],
        "spatial_dims": [x_name, y_name],
        "spatial_shape": [width, height],
        "bands": [{
            "name": None,
            "dim_names": list(dim_names),
            "source_shape": list(shape),
            "data_type": data_type_id,
            "nodata": nodata_bytes,
            "view": None,
            "outdb_uri": outdb_uri,
            "outdb_format": outdb_format,
            "data": b"",  # placeholder, replaced zero-copy below
        }],
    }
    raster_array = pa.array([raster], type=RASTER_STORAGE_TYPE)

    # Swap in the pixel data zero-copy (the one field pa.array() would copy).
    return Raster(_with_band_data(raster_array, _wrap_data_zero_copy(data))) 
``` 
</details>

paleolimbot: Thanks for that...you/Claude is/are probably correct here. This errors loudly at the planning stage if incorrect and so I'm inclined to keep it as is because it reads better (but I'm happy to change if you feel strongly).

python/sedonadb/python/sedonadb/raster.py
paleolimbot: Good catch...I updated this to just reuse the casted memoryview from `data`.

python/sedonadb/python/sedonadb/raster.py
paleolimbot: The reuse of `data` should solve this...we do have a test for an empty outdb view.

python/sedonadb/python/sedonadb/raster.py
james-willis: Fine by me. I’m not planning on changing the schema any time soon. 

Maybe a test that’ll fail if they drift is a good idea

paleolimbot: They all fail if this schema doesn't match 🙂  (used to be a crash before we fixed the panics!)

PR #1000, opened 2026-06-23
feat(rust/sedona-raster-gdal): add RS_Clip

james-willis:
Migrates RS_Clip out of the GDAL raster draft (#704) into its own module, with N-D plane-broadcast support, SQL reference docs, and a benchmark.

Also adds `sedonadb.raster_testing` — a cross-engine raster parity harness (`RasterEngine`, implemented for SedonaDB and rasterio) used to validate RS_Clip against rasterio across the option permutations — plus `Band.nodata` and `Raster.to_numpy()` accessors.

## Deviations from Sedona Spark's RS_Clip

Checked against `RasterBandEditors.clip` in apache/sedona master; recorded here so future contributors can tell which differences are intentional.

- **Default nodata when the argument is omitted.** Spark ignores the band's own nodata and fills with `Integer.MIN_VALUE` for every integral band type (silently truncated on narrower bands, e.g. byte) and Java's `Double.MIN_VALUE` for floating bands — the smallest positive subnormal (4.9e-324), not the most-negative double. This PR resolves the fill as: explicit argument → the band's own nodata → the exact minimum of the band's data type, produced in band-native bytes so Int64/UInt64 sentinels are exact. That matches PostGIS's precedence (band nodata first) and avoids the Java `MIN_VALUE` quirks.
- **An explicit nodata is validated, not truncated.** Spark casts the double into the band's integer type silently (`(int) noDataValue` plus the AWT sample write), so e.g. -9999 on a byte band wraps. Here a value that is not exactly representable in the band's data type is an error.
- **Band 0 selects all bands.** Spark clips exactly one band and always emits a single-band raster. Here band 0 clips every band and preserves the multiband layout; a negative band errors.
- **A geometry that selects no pixels yields NULL, or errors under `lenient => false`.** In Spark, `lenient` only guards the raster/geometry intersection check; a geometry that intersects the raster but selects no pixels (a sliver passing between pixel centers with `allTouched => false`) returns an all-nodata raster. Here it yields NULL under the default `lenient => true`, and when strict the error message points at `allTouched => true`.

One behavior confirmed to **match** Spark, noted because rasterio differs: pixels inside the geometry are copied verbatim, including pixels whose value equals the source band's nodata — they are not remapped to the output nodata the way `rasterio.mask.mask` does (Spark's copy loop writes raw `Raster.getPixel` values with no source-nodata check; the Python parity tests pin this deliberately).


james-willis:
@paleolimbot should we hoist the scalar geometry case? Is that a common case we want to optimize?

james-willis:
> can be zero copy using views

I was thinking this but never finished the views feature

rust/sedona-raster-gdal/src/rs_clip.rs
```diff
@@ -0,0 +1,1367 @@
...
+        return None; // no non-zero pixels (shouldn't happen if caller checked)
+    }
+
+    Some(CropWindow {
+        col_off: min_col,
+        row_off: min_row,
+        width: max_col - min_col + 1,
+        height: max_row - min_row + 1,
+    })
+}
```
prantogg: The crop window here is the tight bbox of the selected mask pixels, but the prior art crops to the geometry envelope ∩ raster extent, snapped to the grid — PostGIS `ST_Clip` (which the module doc cites), `gdalwarp -crop_to_cutline`, and Sedona Spark's `RS_Clip` all do the latter, keeping unselected envelope pixels as nodata padding.

prantogg (changes requested):

rust/sedona-raster-gdal/src/rs_clip.rs
prantogg: There's a performance angle too. Tight crop means the output extent isn't known until after the mask is burned, so every clip pays for a full-raster W×H mask (MEM dataset + the full scans in `has_intersection` and here) even when the geometry covers a handful of pixels. 

Envelope crop knows the window up front, which enables three wins: 
- the mask shrinks to the window
- this scan goes away
- and disjoint geometries can be rejected with a cheap envelope check before any GDAL work

Sedona Spark's implementation is structured exactly this way.

docs/reference/sql/rs_clip.qmd
```diff
@@ -0,0 +1,131 @@
...
+    - name: lenient
+      type: boolean
+      description: >
+        If true, a geometry that does not intersect the raster yields NULL;
+        if false, it raises an error. Defaults to true.
+---
+
+## Description
+
+`RS_Clip` clips a raster to a geometry, in the spirit of PostGIS `ST_Clip`.
```
paleolimbot, line 90: Can you mark this as experimental so we can update the slew of positional arguments in a future update?

rust/sedona-raster-gdal/src/rs_clip.rs
```diff
@@ -0,0 +1,1491 @@
...
+        let mut builder = RasterBuilder::new(num_iterations);
+
+        let exec_arg_types = vec![arg_types[0].clone(), arg_types[geom_arg_idx].clone()];
+        let exec_args = vec![args[0].clone(), args[geom_arg_idx].clone()];
+        let executor =
+            RasterExecutor::new_with_num_iterations(&exec_arg_types, &exec_args, num_iterations);
+
+        with_gdal(|gdal| {
+            configure_thread_local_options(gdal, config_options)?;
+            with_global_proj_engine(|engine| {
```
paleolimbot: You can use the engine from the `SedonaOptions` now that it's available!

rust/sedona-raster-gdal/src/rs_clip.rs
```diff
@@ -0,0 +1,1491 @@
...
+
+                Ok(())
+            })
+            })?;
+
+            // Decide array-vs-scalar over *all* args, not just the raster/geom
+            // the executor was given: a per-row band/option column over a scalar
+            // raster+geom must still yield an N-row array. (`executor.finish`
+            // only inspects its two exec args and would collapse to row 0.)
+            let out: ArrayRef = Arc::new(builder.finish()?);
```
paleolimbot, line 333: Should you just put all the args into the exec and make the exec flexible enough to handle the arg order here?

rust/sedona-raster-gdal/src/rs_clip.rs
```diff
@@ -0,0 +1,1491 @@
...
+        .finish_raster()
+        .map_err(|e| exec_datafusion_err!("Failed to finish raster: {}", e))?;
+
+    Ok(())
+}
+
+/// The minimum representable value of a band data type, as `f64` — the default
+/// nodata sentinel when neither an explicit value nor the band's own nodata is
+/// available, so masked-out pixels stay distinguishable from real data.
+fn band_data_type_min(data_type: &BandDataType) -> f64 {
```
paleolimbot: Can these be associated functions on BandaDataType to reduce the chance the will get duplicated?

rust/sedona-raster-gdal/src/rs_clip.rs
```diff
@@ -0,0 +1,1491 @@
...
+    match data_type {
+        BandDataType::UInt8
+        | BandDataType::UInt16
+        | BandDataType::UInt32
+        | BandDataType::UInt64 => 0.0,
+        BandDataType::Int8 => i8::MIN as f64,
+        BandDataType::Int16 => i16::MIN as f64,
+        BandDataType::Int32 => i32::MIN as f64,
+        BandDataType::Int64 => i64::MIN as f64,
```
paleolimbot: I am not sure that int64::MIN is non-lossy when it goes through f64 (which would mean an unrecognized nondata value). You could template this function on output type (or go the GDAL route, where there's often dedicated functions for i64/u64

rust/sedona-raster-gdal/src/rs_clip.rs
```diff
@@ -0,0 +1,1491 @@
...
+        BandDataType::Float32 => f32::MIN as f64,
+        BandDataType::Float64 => f64::MIN,
+    }
+}
+
+/// The maximum representable value of a band data type, as `f64` — paired with
+/// [`band_data_type_min`] to range-check a custom `no_data_value` before it is
+/// narrowed to the band's type (which would otherwise saturate silently).
```
paleolimbot: This should also do a better job with i64/u64 nodata values (or error for these cases).

In general the nodata handling of i64 and u64 has come up a lot and will probably come up in every function that takes a nodata value argument...I am not sure where the utilities should go, but there should be some so that we don't reinvent/make this mistake every time.

The "right" way to handle the nodata value is to cast the nodata_value argument to the band's data type (rather than f64). These functions can use associated types to be generic on the output type.

rust/sedona-raster-gdal/src/rs_clip.rs
```diff
@@ -0,0 +1,1491 @@
...
+}
+
+fn calc_num_iterations(args: &[ColumnarValue]) -> usize {
+    for arg in args {
+        if let ColumnarValue::Array(array) = arg {
+            return array.len();
+        }
+    }
+    1
+}
```
paleolimbot: This is probably something that needs to be fixed with the executor rather than duplicated here

rust/sedona-raster-gdal/src/rs_clip.rs
```diff
@@ -0,0 +1,1491 @@
...
+    use sedona_expr::scalar_udf::SedonaScalarKernel;
+    use sedona_raster::array::RasterStructArray;
+    use sedona_schema::crs::deserialize_crs;
+    use sedona_schema::datatypes::Edges;
+    use sedona_testing::raster_spec::RasterSpec;
+
+    fn wkb_from_wkt(gdal: &sedona_gdal::gdal::Gdal, wkt: &str) -> Result<Vec<u8>> {
+        let geometry = gdal.geometry_from_wkt(wkt).unwrap();
+        geometry.wkb().map_err(|e| exec_datafusion_err!("{e}"))
+    }
```
paleolimbot: We almost certainly have this utility in sedona-testing

paleolimbot (commented):

I think this needs a Python integration test to check that this version of clip (and all its permutations of optional arguments) match what Rasterio would give on real and/or non-trivial input. Or if we are prioritizing PostGIS compatibility, building out the testing framework to be able to run tests against it.

I'm hesitant to go further without solving the nodata value rounding issues, which have also come up in other PRs and apply to all functions.

docs/reference/sql/rs_clip.qmd
james-willis: done

python/sedonadb/tests/functions/raster_parity.py
```diff
@@ -0,0 +1,169 @@
...
+            "crop": pa.array([r["crop"] for r in rows], type=pa.bool_()),
+        }
+    )
+    view = "raster_parity_clip_rows"
+    con.create_data_frame(table).to_view(view, overwrite=True)
+    try:
+        result = con.sql(
+            "SELECT RS_Clip(RS_FromPath(path), band, ST_GeomFromText(wkt),"
+            f"               all_touched, nodata, crop) AS r FROM {view} ORDER BY idx"
+        ).to_arrow_table()["r"]
```
paleolimbot: Does it make more sense to just use the python api instead of an f-string?

```python
t = con.create_data_frame(table)
t.select(
    r=t.path.funcs.rs_from_path().funcs.rs_clip(
        t.band,
        con.funcs.st_geom_from_text(t.wkt),
        t.all_touched,
        t.nodata,
        t.crop
    )
)
```

python/sedonadb/tests/functions/raster_parity.py
```diff
@@ -0,0 +1,169 @@
...
+            all_touched=all_touched,
+            invert=True,
+        )
+        fill = np.asarray(nodata, dtype=data.dtype)
+        out = np.where(inside, data, fill)
+    return out, tuple(transform.to_gdal())
+
+
+def band_nodata(band):
+    """A band's nodata sentinel unpacked in its dtype, or None."""
```
paleolimbot: Should this be a method on `Band`?

python/sedonadb/tests/functions/raster_parity.py
```diff
@@ -0,0 +1,169 @@
...
+        count=bands,
+        dtype=str(data.dtype),
+        transform=Affine.from_gdal(*gdal_transform),
+        nodata=nodata,
+    ) as dst:
+        dst.write(data)
+
+
+def rasterio_clip(path, geom_wkt: str, *, all_touched, nodata, crop):
+    """Reference clip composed from rasterio primitives over a GeoTIFF.
```
paleolimbot: Should this go in `testing.py` (or maybe a new `raster_testing.py`)? Maybe `class RasterEngine` with methods `clip()`, etc that could be implemented for SedonaDB and rasterio?

python/sedonadb/tests/functions/raster_parity.py
```diff
@@ -0,0 +1,169 @@
...
+    """A band's nodata sentinel unpacked in its dtype, or None."""
+    raw = band._py_field("nodata")
+    if raw is None:
+        return None
+    type_id = BAND_DATA_TYPE_IDS[band.data_type]
+    return struct.unpack("<" + BAND_DATA_TYPE_STRUCT_CHARS[type_id], raw)[0]
+
+
+def raster_to_numpy(raster: Raster):
+    """Extract ``(array(bands, h, w), gdal_transform, [nodata per band])``."""
```
paleolimbot: Should this be a method on `Raster`? I suppose the stacking would wipe out zero copy but that could be documented.

python/sedonadb/tests/functions/test_rs_clip.py
```diff
@@ -0,0 +1,331 @@
...
+    "uint16": 60000.0,
+    "int16": -8888.0,
+    "int32": -88888.0,
+    "float32": -8888.5,
+    "float64": -23456.5,
+}
+
+
+def _random_data(dtype, band_nodata=None):
+    """Random (BANDS, HEIGHT, WIDTH) pixels with adversarial plants.
```
paleolimbot: This would also be a good addition to `raster_testing.py`

python/sedonadb/tests/functions/test_rs_clip.py
```diff
@@ -0,0 +1,331 @@
...
+        gdal_transform=GDAL_TRANSFORM,
+        nodata=BAND_NODATA["uint8"],
+    )
+
+    def clip(args):
+        tab = con.sql(
+            f"SELECT RS_Clip(RS_FromPath($1), 1, ST_GeomFromText($2){args}) AS r",
+            params=(str(tiff), GEOM_TRIANGLE),
+        ).to_arrow_table()
+        return raster_to_numpy(tab["r"][0].as_py())
```
paleolimbot: The Python API would make this  f-string/parameterized SQL/inline `def` combination less hacky feeling. If you use the `expr.rs.clip()` accessor (which should have been automatically generated in sedona-expr), you can even use named params.

rust/sedona-raster-gdal/src/rs_clip.rs
james-willis: found it. using it now

rust/sedona-raster-gdal/src/rs_clip.rs
james-willis: done

rust/sedona-raster-gdal/src/rs_clip.rs
james-willis: done

rust/sedona-raster-gdal/src/rs_clip.rs
james-willis: done

rust/sedona-raster-gdal/src/rs_clip.rs
james-willis: avoided casting by just writing as bytes. no more casting in this flow

rust/sedona-raster-gdal/src/rs_clip.rs
james-willis: Won't datafusion cast the input argument for the no data value to a f64 before we ever get to see the original value? so were kind of screwed here in terms of e.g. representing 2^60 faithfully.

Anyway, the shared utilities now live in `sedona_raster::traits` and are used by `rs_set_band_nodata` and `rs_value` as well.

rust/sedona-raster-gdal/src/rs_clip.rs
james-willis: done

python/sedonadb/tests/functions/raster_parity.py
james-willis: done

python/sedonadb/tests/functions/raster_parity.py
james-willis: done

python/sedonadb/tests/functions/raster_parity.py
james-willis: created the raster_testing and RasterEngine components to hold the rasterio and sedonad db comparisons. will extend later this week with a harness for more engines to compare behavior.

python/sedonadb/tests/functions/raster_parity.py
james-willis: done. made the 1-band case zero copy but multiband cases copy like you describe

python/sedonadb/tests/functions/test_rs_clip.py
james-willis: done

python/sedonadb/tests/functions/test_rs_clip.py
james-willis: done

james-willis:
Heads up: I've pushed the review follow-ups (Band.nodata / Raster.to_numpy accessors, the sedonadb.raster_testing engine harness, expr-accessor tests) but GitHub is being laggy syncing the branch into the PR — jw/rs-clip is at 171fd09d4 if the Commits/Files tabs look stale.

james-willis:
merging to continue working on the parity harness, happy to take more feedback into a new PR.

rust/sedona-raster-functions/src/crs_utils.rs
```diff
@@ -17,11 +17,32 @@
...
+use sedona_common::option::SedonaOptions;
 use sedona_geometry::transform::{transform, CrsEngine};
+use sedona_proj::transform::with_global_proj_engine;
 use sedona_schema::crs::{deserialize_crs, CoordinateReferenceSystem, Crs, CrsRef};
 use wkb::reader::read_wkb;
 
+/// Run `f` with the session's CRS engine: the [`SedonaOptions`] runtime engine
+/// when config options are available (the query path), falling back to the
+/// process-global PROJ engine otherwise (e.g. direct `invoke_batch` calls).
+pub fn with_crs_engine<T>(
```
paleolimbot, line 31: I assume the direct calls are coming from the SedonaUDF tester or tests generally, which would use a default config options. We should ensure our testing config options have `SedonaOptions` in them (I think this is easy) and perhaps a `tester.with_crs_engine()` so that UDFs that use that particular aspect can add `sedona-proj` as a `dev-dependencies` and populate it for their tests.

rust/sedona-raster-gdal/src/rs_clip.rs
```diff
@@ -0,0 +1,1435 @@
...
+        // noDataValue at index 4 (when arg_count >= 5)
+        let nodata_array = if self.arg_count >= 5 {
+            args[4]
+                .clone()
+                .cast_to(&arrow_schema::DataType::Float64, None)?
+                .into_array(num_iterations)?
+        } else {
+            ScalarValue::Float64(None).to_array_of_size(num_iterations)?
+        };
+        let nodata_array = as_float64_array(&nodata_array)?.clone();
```
paleolimbot, line 195: We can file a follow-up to fix this cast, which will loose precision for i64 and u64 nodata values and is handled differently in a few PRs. The solution is to resolve the arraydata and add add an accessor for the appropriate slice of the data buffer (which will be lossless little endian bytes regardless of type), and use a NullsBuffer to check for nulls.

rust/sedona-raster-gdal/src/rs_clip.rs
```diff
@@ -0,0 +1,1435 @@
...
+    let mut max_col = f64::NEG_INFINITY;
+    let mut min_row = f64::INFINITY;
+    let mut max_row = f64::NEG_INFINITY;
+    for (x, y) in corners {
+        let (col, row) = inverse.apply(x, y);
+        min_col = min_col.min(col);
+        max_col = max_col.max(col);
+        min_row = min_row.min(row);
+        max_row = max_row.max(row);
+    }
```
paleolimbot, line 659: Are you sure this logic survives a rotated and/or skewed raster? Is there a test that confirms this works?

rust/sedona-raster-gdal/src/rs_clip.rs
```diff
@@ -0,0 +1,1435 @@
...
+    // upper-left by the pixel offset; scale/skew are unchanged.
+    // Layout: [upper_left_x, scale_x, skew_x, upper_left_y, skew_y, scale_y].
+    let src = original_raster.transform();
+    let transform: [f64; 6] = if let Some(cw) = clipped_data.crop_window {
+        let new_ulx = src[0] + cw.col_off as f64 * src[1] + cw.row_off as f64 * src[2];
+        let new_uly = src[3] + cw.row_off as f64 * src[5] + cw.col_off as f64 * src[4];
+        [new_ulx, src[1], src[2], new_uly, src[4], src[5]]
+    } else {
+        [src[0], src[1], src[2], src[3], src[4], src[5]]
+    };
```
paleolimbot, line 786: I'm still not sure how this works with rotated input but if there's a test that confirms it works I trust you

rust/sedona-raster-gdal/src/rs_clip.rs
```diff
@@ -0,0 +1,1435 @@
...
+    }
+}
+
+/// One clipped band: masked/cropped bytes plus the N-D layout needed to rebuild
+/// it. The clip is a 2-D `(y, x)` operation broadcast across every non-spatial
+/// plane, so `dim_names` are unchanged from the source and only the trailing
+/// `(y, x)` extent of `shape` shrinks when cropping.
+struct ClippedBand {
+    /// Masked/cropped bytes, plane-major in the band's dim order.
+    data: Vec<u8>,
```
paleolimbot, line 345: Here, and in a few other places, there's an allocation of a `Vec<u8>` of the size of the input that isn't reused. If possible we should allocate one `Vec<u8>` as scratch space and reuse it for all clip operations in a single batch. I'm not sure if there's an opportunity to save the final output and wrap it as a view that we append to the builder...I didn't see that here but I may have missed it.

paleolimbot (commented):
Thank you! The parity work is important too...feel free to open issues for these or incorporate them in one of your other PRs.

PR #1001, opened 2026-06-24
refactor(rust/sedona-common)!: Consolidate CrsEngine and Bounder into SedonaOptions

paleolimbot:
This PR consolidates two concepts that are needed in multiple internal crates such that they are accessible from the SedonaOptions (i.e., the ConfigOptions). The two concepts are

- 2D bounding (e.g., for geography bounding), which notably had resulted in sedona-geoparquet having an "s2geography" feature, which shouldn't really be needed (and doesn't scale to other edge types if we need them, like Karney, provided by other dependencies, in that case PROJ).
- CRS transformations. These are very finicky and should always be derived from the Context, because we might want to let them be overridden in the future (e.g., faster special-cased transforms for common pipelines) or more specifically customized (e.g., allow PROJ to access its CDN for network derived gridshift files). This resulted in some awkward interactions (raster dependency on PROJ, the sd_order implementation specifically existing in the proj crate.

There are some cleanups we can do after this merges, which I didn't put here to keep the PR smaller:

- RS_Value and the RS predicates can get the CRS engine from the options instead of depending on the sedona-proj crate
- Geography bounds functions can move back to sedona-functions (and be generic on the edge type instead of being duplicated for any crate that supports an edge type).


rust/sedona-functions/src/st_setsrid.rs
```diff
@@ -1041,5 +1040,14 @@ mod test {
         ) -> Result<Rc<dyn CrsTransform>, SedonaGeometryError> {
             Err(SedonaGeometryError::Unknown)
         }
+
+        fn to_projjson(
+            &self,
+            _crs_string: &str,
+        ) -> std::prelude::v1::Result<String, SedonaGeometryError> {
+            Err(SedonaGeometryError::Invalid(
+                "don't even think about it".to_string(),
```
james-willis, line 1046: 😆

rust/sedona-geoparquet/src/statistics_accumulator.rs
```diff
@@ -254,63 +170,11 @@ impl GeoStatsAccumulator for GeographyGeoStatsAccumulator {
 
 #[cfg(test)]
 mod test {
+    #[allow(unused_imports)]
```
james-willis: !!

james-willis (approved):
My review didn't surface anything to change but our friend claude had some ideas. will leave his comments separately

james-willis:
claude's take:

**Behavior / feature-gating**

1. Spherical bounder registration is nested inside #[cfg(feature = "spatial-join")] — rust/sedona/src/context.rs:167. The only with_bounder(Edges::Spherical, …) call sits inside the s2geography block within the spatial-join block. A build with --no-default-features --features s2geography compiles fine but silently loses geography Parquet statistics (falls to VoidGeoStatsAccumulator) and geography row-group pruning — both of which that feature alone enabled pre-PR. Since spatial-join is a default feature the combo is unusual, but the bounder registration has nothing to do with join planning; move it to its own #[cfg(feature = "s2geography")] block.
2. First-session-wins global stats factory — statistics_accumulator.rs:41 + context.rs:188. try_init snapshots the first context's bounder_factory into arrow-rs's process-wide OnceLock; re-inits fail with a deliberately swallowed ParquetError. Meanwhile the read path re-reads live options per plan (format.rs:356). So a second context with a different runtime — or a bounder registered after context creation via the public SedonaRuntime::with_bounder — writes stats with stale bounders while pruning uses fresh ones. Arrow-rs's global API forces some of this, but it deserves a doc comment on try_init stating the snapshot semantics.
3. sedona-geoparquet's published s2geography feature is now a runtime no-op. Downstream crates.io users who enabled it for geography statistics get silently different behavior (they must now register a bounder through SedonaOptions). The Cargo.toml comment covers this, but it's worth a line in release notes.

**Design / API**

4. SedonaRuntime::PartialEq ignores bounder_factory — option.rs:479. with_bounder keeps the same crs_engine Arc, so a runtime before and after registering a completely different bounder compares equal — and this feeds DataFusion's UDF equality via NormalizeForGeoParquet's derived PartialEq/Eq. Benign today (that UDF never touches bounders) but a footgun; compare both fields.
5. SHOW ALL now leaks raw FFI pointer addresses — option.rs:487. visit debug-prints the whole SedonaRuntime; with s2geography registered that includes RectBounder { ptr: 0x… }, making sedona.runtime in information_schema.df_settings non-deterministic per session. Consider a stable, hand-written summary string.
6. with_bounder returns Result<Self> but cannot fail — option.rs:452. The inner WkbBounder2DFactory::with_bounder is infallible; the ? at the call site is noise. Drop the Result (or document what future failure it reserves).
7. Edges::Display is now load-bearing for on-disk metadata, but deserialize_edges still hardcodes "planar"/"spherical". The other => serialization arms mean a hypothetical future variant would be written into GeoArrow/GeoParquet metadata that SedonaDB itself then rejects on read. Worth a comment tying Display to deserialize_edges as a round-trip contract.
8. Error double-wrapping in LazyProjEngine — transform.rs:150-182. Each method maps to DataFusionError::External inside SedonaGeometryError::External, prefixing user-facing CRS errors with a redundant "External error:" layer. Map straight to SedonaGeometryError.
9. Nits: stale comment at context.rs:218 still describes the stats-accumulator init that moved ~35 lines up; the two consecutive if  let Some(LogicalType::Geography{…}) blocks in make_accumulator collapse into one match with None | Some(SPHERICAL); std::prelude::v1::Result auto-import artifact in st_setsrid.rs:1047; the planar-falls-back/spherical-doesn't asymmetry of bounder_for_edge_type is undocumented; one shared WkbGeographyBounder in the options makes the pre-existing unsafe impl Sync for RectBounder load-bearing for the first time (safe as written, since only create_instance() is called on the shared prototype — worth a comment).

Test coverage (the weakest part of the PR)

The deleted test_literal_bounder (both feature configs) covered geography literal bounding end-to-end: antimeridian wraparound, ~100km distance expansion, and NULL scalars. Nothing replaces it:

- No test exercises geography literal bounding through the new WkbBounder2DFactory → SpatialFilterFactory::literal_bounds path; the new test_bounder_factory_basic is planar-only, and with_bounder/create_instance are never exercised by any test
- ST_DWithin distance expansion for geography literals is now untested anywhere
- NULL literal → BoundingBox::empty() (a genuine behavior improvement — the old planar path hit an internal error) is untested
- SedonaGeoStatsAccumulatorFactory::try_init(runtime) dispatch, including the silent Void fallback, has no test
- LazyProjEngine has zero tests; the deleted proj_crs_provider test was the only coverage of the global-engine to_projjson path

A single s2-gated test in sedona-geoparquet or sedona-expr that registers WkbGeographyBounder via the factory and asserts wraparound + expansion bounds (i.e., a port of the deleted test) would restore most of this.

rust/sedona-geoparquet/src/statistics_accumulator.rs
paleolimbot: 😬

rust/sedona-geometry/src/types.rs
```diff
@@ -22,6 +22,36 @@ use serde_with::{DeserializeFromStr, SerializeDisplay};
...
+/// layer we can reuse implementations for structural manipulation more efficiently if
+/// we consider the edge interpolation as a parameter of the physical type. This maps to
+/// the concept of "edges" in GeoArrow and "algorithm" in Parquet and Iceberg (where the
+/// planar case would be resolved to a geometry instead of a geography).
+#[derive(Debug, PartialEq, Clone, Copy)]
+#[non_exhaustive]
+pub enum Edges {
+    Planar,
+    Spherical,
+}
```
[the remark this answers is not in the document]
paleolimbot: Yes, this is intentional. There are more edge cases coming.

PR #1003, opened 2026-06-24
feat(rust/sedona-raster-functions): add RS_SetGeoReference and RS_SetBandNoDataValue

james-willis:
Metadata-only setter companions to the `RS_GeoReference` and `RS_BandNoDataValue` getters.

### DIFF
--- a/docs/reference/sql/rs_setgeoreference.qmd
+++ b/docs/reference/sql/rs_setgeoreference.qmd
@@ -44,8 +44,10 @@ kernels:
     - name: format
       type: string
       description: >
-        Coordinate convention for the upper-left values, either 'GDAL' (default,
-        pixel corner) or 'ESRI' (pixel center).
+        Coordinate convention for the upper-left values: 'GDAL' (default; alias
+        'PIXEL') places them at the pixel corner, 'ESRI' (alias 'NODE') at the
+        pixel center. The center convention ('ESRI'/'NODE') is rejected for
+        skewed (rotated) rasters, where the half-pixel shift would be inexact.
 ---
 
 ## Examples
--- a/docs/reference/sql/rs_setgeoreference.qmd
+++ b/docs/reference/sql/rs_setgeoreference.qmd
@@ -1,0 +1,59 @@
+---
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+title: RS_SetGeoReference
+description: >
+  Sets the affine geotransform of a raster from a world-file georeference string
+  of six whitespace-separated numbers in the order scalex, skewy, skewx, scaley,
+  upperleftx, upperlefty — the same order RS_GeoReference emits. The format is
+  'GDAL' (default) or 'ESRI'; in GDAL the upper-left coordinates refer to the
+  corner of the upper-left pixel, while in ESRI they refer to its center. This is
+  metadata-only: raster cell values are not resampled.
+kernels:
+  - returns: raster
+    args:
+    - raster
+    - name: georeference
+      type: string
+      description: >
+        Six whitespace-separated numbers — scalex, skewy, skewx, scaley,
+        upperleftx, upperlefty.
+  - returns: raster
+    args:
+    - raster
+    - name: georeference
+      type: string
+      description: >
+        Six whitespace-separated numbers — scalex, skewy, skewx, scaley,
+        upperleftx, upperlefty.
+    - name: format
+      type: string
+      description: >
+        Coordinate convention for the upper-left values, either 'GDAL' (default,
+        pixel corner) or 'ESRI' (pixel center).
+---
+
+## Examples
+
+```sql
+SELECT RS_GeoReference(RS_SetGeoReference(RS_Example(), '2 0 0 -3 100 200'));
+```
+
+```sql
+SELECT RS_GeoReference(RS_SetGeoReference(RS_Example(), '2 0 0 -3 101 198.5', 'ESRI'));
+```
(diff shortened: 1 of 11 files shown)


rust/sedona-raster-functions/src/rs_set_band_nodata.rs
```diff
@@ -0,0 +1,301 @@
...
+            None
+        };
+        let band_values = band_array.as_ref().map(|a| a.as_primitive::<Int64Type>());
+
+        let value_arg = if self.with_band { &args[2] } else { &args[1] };
+        let value_array = value_arg
+            .clone()
+            .cast_to(&DataType::Float64, None)?
+            .into_array(n)?;
+        let value_values = value_array.as_primitive::<Float64Type>();
```
paleolimbot, line 120: If you want to handle i64 or u64, you can register an is_numeric kernel first, an is_integer kernel second, and a u64 kernel last (overload resolution goes backwards). You may be able to make this generic on the Arrow array type (e.g., `Float64Array`) but I find sometimes that gets verbose quickly.

Maybe put a note here that these values are checked later and will error for an out-of-bounds i64 or u64 that is casted losslessly to float here.

rust/sedona-raster-functions/src/rs_set_band_nodata.rs
```diff
@@ -0,0 +1,301 @@
...
+
+            // A null raster, band, or value yields a null raster.
+            let Some(raster) = raster_opt else {
+                return null_out(&mut builder);
+            };
+            let band = match band_values {
+                Some(bands) if bands.is_null(i) => return null_out(&mut builder),
+                Some(bands) => bands.value(i),
+                None => 1,
+            };
```
paleolimbot, line 135: Setting the band to 1 by default is great when there is only one band, but possibly a footgun for multiband. Can we error for this case for now and relax the error if we see evidence that this is causing an issue? Or update the behaviour to set all the bands' nodata values?

rust/sedona-raster-functions/src/rs_set_band_nodata.rs
```diff
@@ -0,0 +1,301 @@
...
+            if value_values.is_null(i) {
+                return null_out(&mut builder);
+            }
+            set_band_nodata(&mut builder, raster, band, value_values.value(i))
+        })?;
+
+        executor.finish(Arc::new(
+            builder.finish().map_err(|e| arrow_datafusion_err!(e))?,
```
paleolimbot:
```suggestion
            builder.finish()?,
```

rust/sedona-raster-functions/src/rs_set_band_nodata.rs
```diff
@@ -0,0 +1,301 @@
...
+        return exec_err!(
+            "RS_SetBandNoDataValue: band {band} out of range (raster has {num_bands} band(s))"
+        );
+    }
+
+    // Copy the raster header (transform/dims/crs) verbatim; we rebuild the bands
+    // below so we can override the addressed band's nodata.
+    builder
+        .start_raster_from(raster, RasterOverrides::default())
+        .map_err(|e| arrow_datafusion_err!(e))?;
```
paleolimbot:
```suggestion
    builder
        .start_raster_from(raster, RasterOverrides::default())?;
```

rust/sedona-raster-functions/src/rs_set_band_nodata.rs
```diff
@@ -0,0 +1,301 @@
...
+
+    for band_idx in 0..num_bands {
+        let band_ref = raster
+            .band(band_idx)
+            .map_err(|e| arrow_datafusion_err!(e))?;
+        // Override the nodata only on the addressed band; others inherit.
+        let nodata_bytes = if band_idx + 1 == band as usize {
+            Some(
+                nodata_f64_to_bytes(value, &band_ref.data_type())
+                    .map_err(|e| arrow_datafusion_err!(e))?,
```
paleolimbot:
```suggestion
                nodata_f64_to_bytes(value, &band_ref.data_type())?,
```

paleolimbot: ...and others in this file

docs/reference/sql/rs_setgeoreference.qmd
```diff
@@ -0,0 +1,59 @@
...
+    - name: georeference
+      type: string
+      description: >
+        Six whitespace-separated numbers — scalex, skewy, skewx, scaley,
+        upperleftx, upperlefty.
+    - name: format
+      type: string
+      description: >
+        Coordinate convention for the upper-left values, either 'GDAL' (default,
+        pixel corner) or 'ESRI' (pixel center).
```
paleolimbot: Can we additionally accept the "pixel" and "node" values (or whatever zarr called them).

docs/reference/sql/rs_setgeoreference.qmd
```diff
@@ -0,0 +1,59 @@
...
+        Six whitespace-separated numbers — scalex, skewy, skewx, scaley,
+        upperleftx, upperlefty.
+  - returns: raster
+    args:
+    - raster
+    - name: georeference
+      type: string
+      description: >
+        Six whitespace-separated numbers — scalex, skewy, skewx, scaley,
+        upperleftx, upperlefty.
```
paleolimbot, line 43: Six whitespace separated values is kind of gross, but I get why it happens. A cleaner future solution is an affine matrix type (we could use it for ST_Affine() as well).

rust/sedona-raster-functions/src/rs_georeference.rs
```diff
@@ -46,9 +46,16 @@ pub fn rs_georeference_udf() -> SedonaScalarUDF {
...
 /// Both formats output six lines: scalex, skewy, skewx, scaley, upperleftx, upperlefty.
-/// The difference is how the upper-left coordinate is reported:
+/// The difference is how the upper-left coordinate is reported. Shared with the
+/// [`RS_SetGeoReference`](crate::rs_set_georeference) setter so the two agree on
+/// accepted format strings.
+///
+/// NOTE: the ESRI center shift uses `scale * 0.5` only and ignores skew; it is
+/// exact for north-up (zero-skew) rasters but off by `skew * 0.5` for rotated
+/// ones. The getter and setter share this approximation, so they round-trip;
+/// interop with a true ESRI world file for a skewed raster is not exact.
```
paleolimbot: Should we be tracking this approximation with a GitHub issue or erroring in the case of a skewed raster? (Both would be my preference unless this is common in other software)

rust/sedona-raster-functions/src/rs_set_georeference.rs
```diff
@@ -0,0 +1,311 @@
...
+            None
+        };
+        let format_array = format_array
+            .as_ref()
+            .map(|a| as_string_array(a))
+            .transpose()?;
+
+        // The executor yields `None` for null raster rows, which short-circuit to
+        // a null output before the georef is parsed — so a malformed georef
+        // aligned to an already-null raster doesn't raise a spurious error.
```
paleolimbot: This is probably fine but I'm not sure the error is spurious (probably would be less confusing to error for all invalid georeferences for the unlikely case where one of them happens to line up like this)

paleolimbot (approved):
Thank you! All nits here.

paleolimbot (approved):
Thank you!

PR #1005, opened 2026-06-24
perf(rust/sedona-functions): fast-path ST_X/ST_Y Point coords via read_point_xy

james-willis:
ST_X/ST_Y read a Point's (x, y) from the fixed WKB offset via `read_point_xy` instead of a full parse; non-Point inputs fall back to the general parser. Benchmark `native-st_{x,y}-Array(Point)`: ~1.90 ms → ~0.53 ms (−72%).

### DIFF
--- a/rust/sedona-functions/src/st_xyzm.rs
+++ b/rust/sedona-functions/src/st_xyzm.rs
@@ -16,7 +16,7 @@
 // under the License.
 use std::{sync::Arc, vec};
 
-use crate::executor::WkbExecutor;
+use crate::executor::{WkbBytesExecutor, WkbExecutor};
 use arrow_array::builder::Float64Builder;
 use arrow_schema::DataType;
 use datafusion_common::error::{DataFusionError, Result};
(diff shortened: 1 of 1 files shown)


james-willis:
Scope note: this only fast-paths **ST_X / ST_Y**. ST_Z / ST_M deliberately stay on the general parse.

I did prototype accelerating Z/M with a Point-specific `read_point_ordinate` (reads the point's dimensionality from the WKB type code to locate/rule out the Z/M ordinate). It measured well — `native-st_z/st_m-Array(Point)` ~**−54%** — but folding all four accessors into one closure regressed X/Y by ~**+12%** (the per-row dimensionality branch bloats the X/Y hot path). Keeping X/Y fast requires separate per-call loops plus a new public helper in the shared `sedona-geometry` crate, so I punted Z/M to keep this change focused. Easy to revisit if Z/M perf matters.

paleolimbot (approved):
Thank you!

ST_Distance for point to point might be fast enough to see improvement from this too if you're having fun, and I believe our DWithin is based on this/gets used in the join. I'm not sure if checking that all values are points is fast or not but I think it might be.

james-willis:
@paleolimbot I can benchmark both cases for that and see if it regresses the non-piont case. Also profiling it first might give us a hint if parse is even a substantial piece of the distance calc

PR #1006, opened 2026-06-24
feat(python/sedonadb): add DataFrame.intersect, intersect_distinct, except_

jiayuasu:
Completes the relational set-op family on the Python `DataFrame`, after `distinct`/`distinct_on` (#961) and `union`/`union_distinct` (#965).

## API

```python
a.intersect(b)            # rows in both, multiset      (SQL INTERSECT ALL)
a.intersect_distinct(b)   # distinct rows in both       (SQL INTERSECT)
a.except_(b)              # distinct rows in a not in b (SQL EXCEPT)
```

- `intersect` preserves multiplicity; `intersect_distinct` de-duplicates.
- `except_` uses a trailing underscore to avoid Python's `except` keyword (matching DuckDB's Python relational API).
- All three require the two inputs to have the **same column names in the same order** (shared `_check_set_op_compatible` guard, generalized from the union-only helper introduced in #965), so a positional set op can't silently misalign differently-named columns. A positional combination of differently-named columns is opt-in: align names first with `select`.

## Why no `except_distinct`

DataFusion's `EXCEPT ALL` does **not** preserve multiplicity — `[1,1,1] EXCEPT ALL [1]` returns `[]` (verified, including via raw SQL), the same result as `EXCEPT`. So a separate ALL/distinct pair for `except` would be two public methods that behave identically today. `except_` therefore exposes the distinct semantics only (wired to DataFusion's `except_distinct`), documented as such. An `except_all` can be added if/when the engine gains multiplicity-preserving `EXCEPT ALL`.

(For `intersect`, `INTERSECT ALL` *does* work, so both variants are kept.)

## Implementation

| File | Change |
|---|---|
| `python/sedonadb/src/dataframe.rs` | `InternalDataFrame::intersect` / `intersect_distinct` / `except_` — thin wrappers over DataFusion's `intersect` / `intersect_distinct` / `except_distinct`. |
| `python/sedonadb/python/sedonadb/dataframe.py` | `DataFrame.intersect` / `intersect_distinct` / `except_`; `_check_union_compatible` renamed to `_check_set_op_compatible` and reused by all set ops (union included). |

## Test plan

`tests/expr/test_dataframe_set_ops.py`:
- `intersect` preserves multiplicity; `intersect_distinct` dedupes; multi-column intersect.
- `except_` distinct difference; multi-column except.
- Lazy return for all three.
- Parametrized over the three methods: different-column-names raises; non-DataFrame argument raises.

Local: full set-op + union suite + 31 doctests + expr suite + `ruff` + `cargo fmt --check` + `clippy -Dwarnings` all clean.


python/sedonadb/python/sedonadb/dataframe.py
```diff
@@ -905,10 +905,10 @@ def distinct_on(self, *cols: Union[str, Expr]) -> "DataFrame":
...
         return DataFrame(self._ctx, self._impl.distinct_on(coerced))
 
-    def _check_union_compatible(self, other: "DataFrame", method: str) -> None:
+    def _check_set_op_compatible(self, other: "DataFrame", method: str) -> None:
         # Both inputs must line up by column name (in order) so a positional
-        # union can't silently misalign differently-named columns — the
-        # classic footgun. Callers that want a positional union of
+        # set operation can't silently misalign differently-named columns —
+        # the classic footgun. Callers that want a positional combination of
         # differently-named columns opt in by aliasing/selecting to align
```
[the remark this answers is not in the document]
jiayuasu: Tightened in the latest push — the comment now says "positional set operation" to match the shared helper's name and its union/intersect/except callers.

python/sedonadb/python/sedonadb/dataframe.py
```diff
@@ -989,9 +989,86 @@ def union_distinct(self, other: "DataFrame") -> "DataFrame":
...
+
+        Args:
+            other: The DataFrame to intersect with. Must have the same
+                column names, in the same order, as this DataFrame.
+        """
+        self._check_set_op_compatible(other, "intersect_distinct")
+        return DataFrame(self._ctx, self._impl.intersect_distinct(other._impl))
+
+    def except_(self, other: "DataFrame") -> "DataFrame":
+        """Distinct rows in this DataFrame that are not in `other`.
```
paleolimbot: This should probably be called `except_distinct()`? (and we can't do "except" unless you want to write the requisite `join(how="anti")` to make it happen).

python/sedonadb/src/dataframe.rs
```diff
@@ -342,6 +342,32 @@ impl InternalDataFrame {
...
+        let inner = self.inner.clone().intersect_distinct(other.inner.clone())?;
+        Ok(InternalDataFrame::new(inner, self.runtime.clone()))
+    }
+
+    /// `EXCEPT` (distinct) — the distinct rows in this DataFrame that are
+    /// not in the other. Uses DataFusion's `except_distinct`; multiplicity-
+    /// preserving `EXCEPT ALL` is not currently supported by the engine.
+    /// Named `except_` (trailing underscore) so the exposed Python method
+    /// doesn't clash with the `except` keyword.
+    fn except_(&self, other: &InternalDataFrame) -> Result<InternalDataFrame, PySedonaError> {
```
paleolimbot: This should probably be called `except_distinct()` since that's what it wraps

python/sedonadb/tests/expr/test_dataframe_set_ops.py
paleolimbot, line 1: Can you merge this file with the union file (since that's also a set op?)

paleolimbot (commented):
Thank you!

python/sedonadb/python/sedonadb/dataframe.py
jiayuasu: Agreed — renamed to `except_distinct()` in d47d9a1. It's clearer (matches what it wraps + the `union_distinct`/`intersect_distinct` family) and drops the trailing-underscore keyword hack. No bare `except` for now, with a docstring note that a multiplicity-preserving anti-difference would be a `join(how="anti")`.

python/sedonadb/src/dataframe.rs
jiayuasu: Done — Rust binding renamed `except_distinct` to match.

python/sedonadb/tests/expr/test_dataframe_set_ops.py
jiayuasu: Merged — `test_dataframe_union.py` is gone and all set-op tests (union + intersect + except) now live in `test_dataframe_set_ops.py`. The shared-guard tests (lazy return, different-names, count-mismatch, non-DataFrame) are parametrized over all five methods.

paleolimbot: I don't see any union tests here, did you just delete them by accident?

jiayuasu: Good catch — they were dropped by accident. My consolidation commit didn't stage cleanly (a stray `git add` of the already-removed union file aborted the staging), so the prior revision kept the old set-ops file without the union tests. Fixed in e7ad981: the union tests (union/union_distinct, multi-column, the positional-alignment opt-in) are now in `test_dataframe_set_ops.py` alongside intersect/except, and the shared-guard tests are parametrized over all five methods. Verified the committed file this time.

paleolimbot (approved):
Thank you!

PR #1008, opened 2026-06-25
perf(rust): point-coordinate fast paths for ST_Distance/ST_DWithin/ST_Azimuth

james-willis:
Introduces `PointXYGeometryFactory` (in `sedona-functions`): a `GeometryFactory` whose `Geom` is a `PointOrWkb` — a finite Point decoded straight from its fixed WKB offset (no full parse), or a parsed `Wkb` for anything else (incl. `POINT EMPTY`). Because `GenericExecutor` parses each scalar once, kernels built on it get parse-once-for-scalars for free. Replaces the hand-rolled `PreparedScalar` scalar-prep that was here before.

Adopted in the planar point-coordinate distance functions:
- **ST_Distance** — Point/Point computes Euclidean directly; mixed Point/geometry uses a `geo_types::Point` so the Point is never re-parsed.
- **ST_DWithin** — shares ST_Distance's backend (`distance <= bound`); this also speeds the spatial-join distance predicate transitively.

Results vs `main` (criterion):

| config | ST_Distance | ST_DWithin |
|---|---|---|
| Array/Array Point,Point | −68% | −63% |
| Array/Scalar Point,Point | −67% | — |
| Array Point / Scalar Polygon | −41% | −33% |
| Array Polygon / Scalar Point | −24% | — |
| Array/Array Point,Polygon | −34% | — |

(The mixed Point/geometry cases improve too, since the factory computes point-to-geometry without re-parsing the Point.)

**ST_Azimuth** applies the same "skip the parse for points" idea but lands on the lighter primitive. It is defined only for two non-empty POINTs (non-Point and `POINT EMPTY` are errors, not a fallback), so it is coordinate-extraction-shaped rather than a distance kernel: it reads each operand's `(x, y)` via `read_point_xy` on a `WkbBytesExecutor` — no parse, no enum. Benchmarked head-to-head against the factory, the accessor wins the dominant two-point-column shape:

| ST_Azimuth config | accessor vs `main` | accessor vs factory |
|---|---|---|
| Array/Array Point,Point | −40% | ~8% faster |
| Array/Scalar Point,Point | −14% | ~8% slower |
| Scalar/Array Point,Point | −15% | ~8% slower |

Both variants beat the full-parse baseline everywhere; the accessor is chosen for the ArrayArray win (the largest absolute saving and the canonical origin→destination shape) and to keep the split consistent — accessors read via `read_point_xy`, distance/fallback kernels use the factory.

**ST_X / ST_Y are deliberately left on their existing `read_point_xy` fast path.** Migrating them to the factory *regressed* them ~30% in benchmarks: the `PointOrWkb` enum (sized to its `Wkb` variant) adds per-element overhead that, for the already-optimal unary case, isn't offset by any parse savings. So the factory is the right tool for the binary distance kernels but not for the unary accessors.

Geography distance/dwithin/maxdistance (s2geography) are a separate follow-up.

### DIFF
--- a/rust/sedona-geometry/src/wkb_header.rs
+++ b/rust/sedona-geometry/src/wkb_header.rs
@@ -152,74 +152,114 @@ impl WkbHeader {
     }
 }
 
+/// The coordinate layout of a WKB Point, discovered by reading only its header.
+///
+/// This is the fast-path primitive for sampling/indexing by Point. A Point WKB
+/// has a fixed layout (`byte order`, `type code`, optional SRID, then the
+/// coordinate), so [`try_from_wkb`](Self::try_from_wkb) can cheaply decide
+/// whether a buffer is a Point — reading only the first few header bytes, never
+/// a coordinate — and a later [`read_xy`](Self::read_xy) reads the coordinate
+/// from the offset it already found.
+///
+/// Splitting the header inspection from the coordinate read lets a binary
+/// function (e.g. distance) test *both* operands first and read coordinates only
+/// when both are Points, without re-parsing the header or wasting a coordinate
+/// read on rows that fall back to the general parser.
+#[derive(Debug, Clone, Copy)]
+pub struct WkbPointLayout {
+    little_endian: bool,
+    coord_offset: usize,
+}
+
+impl WkbPointLayout {
+    /// Inspects a WKB header, returning the Point layout or `Ok(None)` if `buf`
+    /// is not a Point. Reads only the byte order and type code (recognizing the
+    /// EWKB SRID prefix) — never a coordinate. Errors only on a missing/invalid
+    /// byte order or a truncated type code. Accepts both ISO (with Z/M dimension
+    /// flags) and EWKB (with Z/M/SRID flags) encodings.
+    #[inline]
+    pub fn try_from_wkb(buf: &[u8]) -> Result<Option<Self>, SedonaGeometryError> {
+        let little_endian = match buf.first() {
+            Some(1) => true,
+            Some(0) => false,
+            _ => {
+                return Err(SedonaGeometryError::Invalid(
+                    "WKB: missing or invalid byte order".to_string(),
+                ))
+            }
+        };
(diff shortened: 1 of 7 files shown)


paleolimbot (commented):
Cool!

I know you're still working here, but we have something called a `GeometryFactory` (parameterized by executor) that might be a good fit here...it handles the "preparation" of scalars to ensure we don't parse WKB more than needed. Maybe a `PointXYGeometryFactory` / `GenericExecutor<PointXYGeometryFactory>` would make this trivial to scale out to other places.

rust/sedona-geometry/src/wkb_header.rs
```diff
@@ -152,74 +152,114 @@ impl WkbHeader {
...
-        let bytes: [u8; 4] = buf
-            .get(offset..offset + 4)
+impl WkbPointLayout {
+    /// Inspects a WKB header, returning the Point layout or `Ok(None)` if `buf`
+    /// is not a Point. Reads only the byte order and type code (recognizing the
+    /// EWKB SRID prefix) — never a coordinate. Errors only on a missing/invalid
+    /// byte order or a truncated type code. Accepts both ISO (with Z/M dimension
+    /// flags) and EWKB (with Z/M/SRID flags) encodings.
+    #[inline]
+    pub fn try_from_wkb(buf: &[u8]) -> Result<Option<Self>, SedonaGeometryError> {
```
paleolimbot, line 181: This is great!

Because it's doing some byte parsing / is indexing into buffers in a way that could panic for an out of bounds read, I think it needs at least one test to cover the branches. Hopefully there are some in this file to draw from (if not, there are some fixtures in sedona-testing).

paleolimbot (commented):
Thank you!

The low level WkbPointLayout would benefit from a quick test covering the type of points it supports (especially errors resulting from incomplete buffers, including EWKB), with apologies if I missed this test.

paleolimbot (approved):
Thank you!

PR #1009, opened 2026-06-26
feat(rust/sedona-raster-zarr): read CF spatial_ref GeoTransform georeferencing

james-willis:
Read CF / rioxarray georeferencing from a Zarr scalar grid-mapping variable (commonly `spatial_ref`): the reader already resolved the CRS from `crs_wkt` there; this also reads its `GeoTransform` and uses it as the group transform.

The CF `GeoTransform` is a space-separated string **already in GDAL order** `[origin_x, scale_x, skew_x, origin_y, skew_y, scale_y]`, so it's used verbatim — no affine reorder (unlike GeoZarr `spatial:transform`). Priority: explicit `spatial:transform` → CF `GeoTransform` → `spatial:bbox` → coordinate arrays → identity.

This closes the last open Zarr georeferencing source (GeoZarr transform, bbox, and CF coordinate arrays already landed). Covered by a synthetic `spatial_ref` round-trip fixture; the real-world cloud dataset is not public so it isn't included as a test.

### DIFF
--- a/rust/sedona-raster-zarr/src/geozarr.rs
+++ b/rust/sedona-raster-zarr/src/geozarr.rs
@@ -147,6 +147,62 @@ pub fn crs_from_cf_attributes(
     Ok(None)
 }
 
+/// Parse a CF `GeoTransform` attribute (as written by GDAL / rioxarray on a
+/// `spatial_ref` / grid-mapping variable) into a GDAL-order transform.
+///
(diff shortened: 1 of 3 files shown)


paleolimbot (approved):
Thank you! I believe I've seen this version in NetCDF files before.

PR #1010, opened 2026-06-27
feat(python/sedonadb): add DataFrame.mutate and rename

jiayuasu:
Adds two relational column operations to the Python `DataFrame`, in the engine-neutral (Spark/Polars) naming, continuing the Ibis/DuckDB-style relational surface.

## API

```python
df.with_column("c", df.a + df.b)     # add a column
df.with_column("b", df.b * 2)        # replace an existing column in place
df.with_column("k", lit(9))          # constant column
df.with_column("a_copy", "a")        # copy a column by name
df.with_column_renamed("b", "c")     # rename one column
```

- `with_column(name, expr)` — add a column, or replace the existing column of that name in place (order preserved). `expr` accepts an `Expr`, a column-name `str` (copied), or a `lit()` literal. For building several columns at once, `select(..., **kwargs)` is usually more convenient.
- `with_column_renamed(old, new)` — rename a single column, preserving order and all other columns.

## Footgun guard

DataFusion's `with_column_renamed` **silently no-ops** when the source column doesn't exist, which hides typos. The Python wrapper validates the source name and raises a `KeyError` listing the available columns — same pattern already used by `drop`.

## Implementation

| File | Change |
|---|---|
| `python/sedonadb/src/dataframe.rs` | `InternalDataFrame::with_column` / `with_column_renamed` — thin wrappers over DataFusion. |
| `python/sedonadb/python/sedonadb/dataframe.py` | `DataFrame.with_column` (Expr/str/Literal coercion + name-type check) and `with_column_renamed` (KeyError guard). |

## Test plan

`tests/expr/test_dataframe_with_column.py`:
- `with_column`: append; in-place replace (order preserved); from a column-name str; from a literal; bad name type; bad value type; lazy return.
- `with_column_renamed`: rename; order preservation; unknown-column `KeyError` (listing available columns); lazy return.

Local: 11 unit + 33 doctests + expr suite + `ruff` + `cargo fmt --check` + `clippy -Dwarnings` all clean.

### DIFF
--- a/python/sedonadb/python/sedonadb/dataframe.py
+++ b/python/sedonadb/python/sedonadb/dataframe.py
@@ -22,7 +22,7 @@ from typing import TYPE_CHECKING, Any, Dict, Iterable, List, Literal, Optional, 
 from sedonadb.expr import Expr, SortExpr
 from sedonadb.expr import Literal as _SedonaLit
 from sedonadb.expr import col as _col
-from sedonadb.expr.expression import _to_expr
+from sedonadb.expr.expression import collect_exprs
 from sedonadb.utility import sedona  # noqa: F401
 
 if TYPE_CHECKING:
@@ -337,38 +337,8 @@ class DataFrame:
         if not exprs and not kwargs:
             raise ValueError("select() requires at least one column or expression")
 
-        coerced = []
-        for e in exprs:
-            if isinstance(e, Expr):
-                coerced.append(e._impl)
-            elif isinstance(e, str):
-                coerced.append(_col(e)._impl)
-            elif isinstance(e, _SedonaLit):
-                # `lit(value)` returns Literal; route it through the same
-                # coercion path operators use so the resulting Expr lands
-                # in the plan with metadata preserved.
-                coerced.append(_to_expr(e)._impl)
-            else:
-                raise TypeError(
-                    f"select() expects str, Expr, or Literal arguments, "
-                    f"got {type(e).__name__}"
-                )
-
-        # Process keyword arguments: alias each value with its key
-        for name, e in kwargs.items():
-            if isinstance(e, Expr):
-                coerced.append(e.alias(name)._impl)
-            elif isinstance(e, str):
-                coerced.append(_col(e).alias(name)._impl)
-            elif isinstance(e, _SedonaLit):
-                coerced.append(_to_expr(e).alias(name)._impl)
-            else:
-                raise TypeError(
-                    f"select() expects str, Expr, or Literal keyword arguments, "
-                    f"got {type(e).__name__} for '{name}'"
-                )
-
-        return DataFrame(self._ctx, self._impl.select(coerced))
+        coerced = collect_exprs("select", exprs, kwargs)
+        return DataFrame(self._ctx, self._impl.select([e._impl for e in coerced]))
 
     def filter(self, *exprs: Expr) -> "DataFrame":
         """Filter rows by one or more boolean expressions.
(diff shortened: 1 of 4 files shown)


python/sedonadb/python/sedonadb/dataframe.py
```diff
@@ -1067,6 +1067,89 @@ def except_distinct(self, other: "DataFrame") -> "DataFrame":
...
+            │ int64 ┆ int64 │
+            ╞═══════╪═══════╡
+            │     1 ┆     2 │
+            └───────┴───────┘
+        """
+        # DataFusion silently no-ops when the column is missing, which hides
+        # typos. Validate Python-side so the caller gets a clear KeyError
+        # listing the available columns (matching `drop`).
+        columns = self._impl.columns()
+        if old_name not in columns:
```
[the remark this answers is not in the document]
jiayuasu: Good catch — added type checks for both `old_name` and `new_name` (raising a clear `TypeError`) before the membership lookup, matching `drop()`'s pattern. Covered by a parametrized test (`None`/`int` for each argument).

python/sedonadb/tests/expr/test_dataframe_with_column.py
```diff
@@ -0,0 +1,106 @@
...
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+import pandas as pd
+import pandas.testing as pdt
+import pytest
+
+from sedonadb.dataframe import DataFrame
+from sedonadb.expr import col, lit
```
paleolimbot: Tests should use `con.col()` and `con.lit()` to set a good example...these versions aren't user-accessible via `sedona.db.xx` and don't have am embedded context (so chaining via `.geo.xxx()` doesn't work). There are some preexisting tests that I missed before this got added but at least for new ones we should use the constructors from the context.

paleolimbot (commented):
Thank you!

These functions are nicely implemented and I have no issues with the code...I do want to check quickly on whether or not we want to expose these or simply the bulk versions (`mutate()` and `rename()`). As the documentation notes, multiple `.with_column()`/`with_column_renamed()` calls are an anti-pattern, but there's not a great alternative (hand-rolled `select()`s with a mix of splatted `.columns`).

Ibis and dplyr solve this with `mutate()`: https://ibis-project.org/reference/expression-tables#ibis.expr.types.relations.Table.mutate and `rename()`: https://ibis-project.org/reference/expression-tables#ibis.expr.types.relations.Table.rename (different syntax than pandas, which is not very flexible). I like the bulk versions better because they're more efficient, format more nicely, and allow bulk mutations/renames via selectors when we implement them, and I'd rather have our Python API have a Pythonesque exactly one obvious way to do something. Do you have thoughts on this?

jiayuasu:
Good points — I'm convinced. The chained-`with_column` anti-pattern (my own docstring flags it), the efficiency win of a single projection, the "exactly one obvious way" principle, and the Ibis/DuckDB framing this API already advertises all line up. Let me pivot this PR to the bulk versions:

- **`mutate(**exprs)`** — keep all existing columns plus the named/replaced ones, mirroring how `select(**kwargs)` already coerces (and accepting positional aliased `Expr`s too). This fills the gap `select` doesn't: "all columns plus these" without re-listing everything.
- **`rename(...)`** — bulk rename.

One thing I want your call on before I implement `rename`: the **direction/signature**, since the popular libraries disagree and I'd rather not bake in a silent-swap footgun:

1. **`rename({"old": "new"})`** — dict, old→new. Matches **pandas and Polars**.
2. **`rename(new_name="old_name")`** — kwargs, new→old. Matches **Ibis and dplyr** (the framing this API cites), though the new name must be a valid identifier and the reversal surprises pandas users.

You linked Ibis and noted pandas rename is "not very flexible," so I'm guessing you lean toward (2) — but which would you like? (Or a dict in the Ibis `{new: old}` direction for arbitrary names?)

Also — good note on `con.col()`/`con.lit()` in the tests; I'll switch the new tests to the context constructors as part of the rework.

python/sedonadb/python/sedonadb/dataframe.py
```diff
@@ -1067,6 +1067,61 @@ def except_distinct(self, other: "DataFrame") -> "DataFrame":
         self._check_set_op_compatible(other, "except_distinct")
         return DataFrame(self._ctx, self._impl.except_distinct(other._impl))
 
+    def mutate(self, **exprs: Union[Expr, str, _SedonaLit]) -> "DataFrame":
```
paleolimbot: Should this accept positional expressions as well like `select()`?

python/sedonadb/python/sedonadb/dataframe.py
```diff
@@ -1067,6 +1067,61 @@ def except_distinct(self, other: "DataFrame") -> "DataFrame":
...
+                mutations[name] = e.alias(name)
+            elif isinstance(e, str):
+                mutations[name] = _col(e).alias(name)
+            elif isinstance(e, _SedonaLit):
+                mutations[name] = _to_expr(e).alias(name)
+            else:
+                raise TypeError(
+                    f"mutate() expects Expr, str, or Literal values, "
+                    f"got {type(e).__name__} for '{name}'"
+                )
```
paleolimbot: Should there be a helper in the expressions file that collects named + positional expressions for `select()` and `mutate()` (which as far as I can tell share the same rules for input)?

python/sedonadb/python/sedonadb/dataframe.py
```diff
@@ -1067,6 +1067,61 @@ def except_distinct(self, other: "DataFrame") -> "DataFrame":
...
+
+        # Build a single projection: existing columns (replaced in place
+        # where a mutation matches the name), then the new columns appended.
+        # One projection node rather than a chain of per-column ones.
+        current = self._impl.columns()
+        projection = [mutations[c] if c in mutations else _col(c) for c in current]
+        projection += [m for name, m in mutations.items() if name not in current]
```
paleolimbot: The hiccup here is that existing columns should be updated in place rather than appended to the end of the projection (like `with_column()`).

It's a little cryptic how this works, but this is how we do it in R, although I forget if we specifically tested this case.

https://github.com/apache/sedona-db/blob/046fdede7e207c3d127f958af7b2612afe51a5e1/r/sdplyr/R/mutate.R#L35-L47

paleolimbot (commented):
> One thing I want your call on before I implement rename

How about we detect the case where `rename()` is called with a dictionary and error with a helpful message? If we get reports of this being annoying we can update the implementation to just go through with the rename. The error message should work for LLMs that are iterating to just update the syntax and move on.

python/sedonadb/python/sedonadb/dataframe.py
jiayuasu: Done — `mutate` now takes positional expressions too (`mutate(expr, y=...)`), with the same input rules as `select`. For a positional expr, its own output name is used to match existing columns for in-place replacement.

python/sedonadb/python/sedonadb/dataframe.py
jiayuasu: Added `collect_exprs()` in `expr/expression.py` — it collects named + positional expressions with one set of coercion rules, and both `select` and `mutate` now use it. (`select`'s inline coercion is gone.)

python/sedonadb/python/sedonadb/dataframe.py
jiayuasu: In-place replacement is handled: the projection walks the existing columns and substitutes a mutation where its output name matches (only genuinely-new columns append). Now that positional exprs are supported, the match uses each expr's output name (new `Expr._output_name()`, via DataFusion's `schema_name`) rather than just the kwarg key. Covered by `test_mutate_replaces_in_place`, `test_mutate_positional_expr`, and `test_mutate_preserves_column_order`.

python/sedonadb/tests/expr/test_dataframe_with_column.py
jiayuasu: Addressed — the with_column test file is gone; the new `test_dataframe_mutate.py` uses `con.col()`/`con.lit()` throughout.

jiayuasu:
Reworked per the discussion — this PR is now `mutate` + `rename` (the single-column `with_column`/`with_column_renamed` and their Rust bindings are gone):

- **`mutate(*exprs, **named)`** — positional + keyword, same input rules as `select` (shared `collect_exprs()` helper in `expr/expression.py`), in-place replacement by output name, one projection.
- **`rename(new_name="old_name")`** — Ibis/dplyr direction. A positionally-passed dict (the pandas/Polars `{old: new}` habit) raises a clear error that prints the corrected keyword call, e.g. `rename({"a": "x"})` → *"...Rewrite {'a': 'x'} as rename(x=\"a\")."* — actionable for hand-editing or an iterating LLM. Unknown old columns raise a `KeyError` listing available columns.

On the rename direction: went with your suggestion — Ibis-style kwargs as the one true form, dict detected-and-redirected rather than silently accepted. Easy to relax to "just do the rename" later if it proves annoying.

Local: mutate/rename tests + select regression (shared-helper refactor) + doctests + ruff + cargo fmt + clippy all clean.

paleolimbot (approved):
Thank you!

PR #1011, opened 2026-06-27
Normalize expected geometry WKT in Python tests

### LINKED ISSUE #815: python/sedonadb: Investigate formatting issues when using WKT as an expected value

ajaypadwal73:
Closes #815.

## Summary

- Normalize expected WKT values for geometry result columns before tuple/list comparisons in the Python test harness.
- Use the same `geoarrow.pyarrow.format_wkt()` path for expected geometry strings that the harness already uses for actual geometry results.
- Keep ordinary string columns as exact string comparisons, so mixed geometry/text result assertions still catch real text differences.
- Leave expected strings unchanged when GeoArrow cannot parse them, preserving existing expectations for known GeoArrow rendering artifacts.

## Why

Some tests failed only because the expected WKT used compact formatting while GeoArrow rendered the actual geometry with spaces after commas. For example, these represent the same geometry but previously compared as different strings:

```text
POLYGON Z ((0 0 5,0 1 5,1 1 5,1 0 5,0 0 5))
POLYGON Z ((0 0 5, 0 1 5, 1 1 5, 1 0 5, 0 0 5))
```

This change makes geometry assertions compare against the harness' normalized WKT representation instead of failing on formatting-only differences.

## Verification

- `PYTHONPYCACHEPREFIX=/private/tmp/sedona-pycache .venv/bin/python -m py_compile python/sedonadb/python/sedonadb/testing.py python/sedonadb/tests/test_testing.py`
- `.venv/bin/python -m pytest python/sedonadb/tests/test_testing.py`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -49,6 +49,9 @@ __pycache__
 
 # .env file for release management
 dev/release/.env
+.mcp.env
 
 /.luarc.json
 .positai
+
+.venv/*
--- a/python/sedonadb/python/sedonadb/testing.py
+++ b/python/sedonadb/python/sedonadb/testing.py
@@ -225,6 +225,31 @@ class DBEngine:
 
         return list(zip(*columns))
 
+    def normalize_expected_tuples(
+        self, result, expected, *, wkt_precision=None, **kwargs
+    ) -> List[Tuple[str]]:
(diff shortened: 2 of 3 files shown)


ajaypadwal73:
@paleolimbot can you please take a look at this?

PR #1012, opened 2026-06-27
chore(python/sedonadb): Upgrade PyO3 and enable free-threaded Python support

### LINKED ISSUE #236: Upgrade pyo3 and enable free-threaded Python support
There's a great guide on how to enable free-threaded support here: https://pyo3.rs/v0.26.0/free-threading.html ...for us that mostly means protecting the SedonaContext and DataFrame with locks and using `Python::attach()` instead of `with_gil()`. After #228 this in theory opens up Python UDFs that are executed in parallel.

ajaypadwal73:
## Summary
- upgrade PyO3 to 0.26 for the Python extension crates
- mark the Python modules as free-threading compatible with gil_used = false
- protect the Python InternalContext state with explicit locking and update callback paths to use attach/detach
- add a threaded context/dataframe regression test

Closes #236

## Testing
- cargo check -p sedonadb
- cargo check -p sedonadb-zarr
- .venv/bin/python -m pytest python/sedonadb/tests/test_context.py::test_shared_context_and_dataframe_from_threads -vv
- .venv/bin/python -m pytest python/sedonadb/tests/test_dataframe.py::test_to_view python/sedonadb/tests/test_dataframe.py::test_count -vv
- .venv/bin/python -m pytest python/sedonadb/tests/test_context.py -vv (10 passed; 2 network-backed tests failed locally while reading GitHub/S3 URLs)

cargo test for sedonadb and sedonadb-zarr reached linking locally but failed because libpython3.9 was not available in this local macOS Python environment.

### DIFF
--- a/python/sedonadb/src/context.rs
+++ b/python/sedonadb/src/context.rs
@@ -174,7 +174,7 @@ impl InternalContext {
         Ok(InternalDataFrame::new(df, self.runtime.clone()))
     }
 
-    pub fn drop_view(&self, _py: Python<'_>, table_ref: &str) -> Result<(), PySedonaError> {
+    pub fn drop_view(&self, table_ref: &str) -> Result<(), PySedonaError> {
         self.inner.ctx.deregister_table(table_ref)?;
         Ok(())
     }
@@ -243,11 +243,11 @@ impl InternalContext {
         }
     }
 
-    pub fn list_scalar_udfs(&self, _py: Python<'_>) -> Result<Vec<String>, PySedonaError> {
+    pub fn list_scalar_udfs(&self) -> Result<Vec<String>, PySedonaError> {
         Ok(self.inner.scalar_udf_names()?)
     }
 
-    pub fn list_aggregate_udfs(&self, _py: Python<'_>) -> Result<Vec<String>, PySedonaError> {
+    pub fn list_aggregate_udfs(&self) -> Result<Vec<String>, PySedonaError> {
         Ok(self
             .inner
             .ctx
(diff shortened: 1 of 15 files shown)


python/sedonadb/src/context.rs
```diff
@@ -36,10 +37,50 @@ use crate::{
 
 #[pyclass]
 pub struct InternalContext {
-    pub inner: SedonaContext,
+    pub inner: Arc<Mutex<SedonaContext>>,
     pub runtime: Arc<Runtime>,
 }
```
paleolimbot, line 40: I think it may be possible to avoid the Mutex here, which is causing some complexity.

The DataFusion SessionContext, on which the SedonaContext is based, is thread safe (I think) because its `state()` already uses a `RwLock`. We may have to put one or more members of the SedonaContext behind a RwLock as well, which would only require operations that update the context to lock it (e.g., registering things).

paleolimbot (commented):
I will take a closer look Monday, but I do think it's worth some effort to avoid `with_context()` (in particular, waiting for a future just to kick off an operation!)

python/sedonadb/src/context.rs
ajaypadwal73: @paleolimbot Thanks! I agreed that the Python-side `Arc<Mutex<SedonaContext>>` was too broad.

I reworked this to keep normal `SedonaContext` / `SessionContext` operations lock-free from the Python wrapper, and moved synchronization to the mutable function registry instead. `SedonaContext.functions` is now a private `RwLock<FunctionSet>` with narrow helper methods for lookup and registration. `InternalContext` owns `SedonaContext` directly again, and `register_component()` uses those helpers.

I also removed the unused Python-token parameters after the refactor.

paleolimbot (approved):
Thank you!

PR #1013, opened 2026-06-28
fix(rust/sedona): accept aws.allow_http / aws.session_token; reject dead AWS options

adonm:
## Problem

`read_parquet()` can't read from HTTP-only S3 endpoints (MinIO, LocalStack, `s3s`). Setting `aws.endpoint` to `http://...` fails:

```
HTTP is not allowed for S3 endpoints. To allow HTTP, set 'aws.allow_http' to true
```

…but `aws.allow_http` is itself rejected before it gets there:

```
Unknown AWS option 'aws.allow_http'. Valid options are: aws.access_key_id, aws.secret_access_key, aws.region, aws.endpoint, aws.skip_signature, aws.nosign, aws.bucket_name, aws.use_ssl, aws.force_path_style
```

## Root cause

`GeoParquetReadOptions::from_table_options()` hard-codes an `aws.*` allowlist that drifted out of sync with `AwsOptions::set` (the actual consumer in `sedona::object_storage`). Two mismatches:

- **Rejected but consumed:** `aws.allow_http`, `aws.session_token` — both applied by `AwsOptions` but blocked by the allowlist.
- **Accepted but never consumed:** `aws.bucket_name`, `aws.use_ssl`, `aws.force_path_style`, `aws.nosign` — pass validation, then hit the `_` arm of `AwsOptions::set` and are silently dropped.

## Fix

Align the allowlist with `AwsOptions::set` exactly:

| option | before | after |
| --- | --- | --- |
| `aws.allow_http` | rejected | **accepted** |
| `aws.session_token` | rejected | **accepted** |
| `aws.bucket_name` | silently ignored | **rejected** |
| `aws.use_ssl` | silently ignored | **rejected** |
| `aws.force_path_style` | silently ignored | **rejected** |
| `aws.nosign` | silently ignored | **rejected** |

The four newly-rejected options never worked, so erroring surfaces the misconfiguration instead of hiding it. Azure already had `azure.allow_http`; this brings AWS to parity.

## Test

Rust unit test pinning the accepted vs. rejected set against `from_table_options`.

### DIFF
--- a/rust/sedona-geoparquet/src/provider.rs
+++ b/rust/sedona-geoparquet/src/provider.rs
@@ -393,8 +393,13 @@ mod test {
         ] {
             let mut opts = HashMap::new();
             opts.insert(rejected.to_string(), "x".to_string());
-            let err =
-                GeoParquetReadOptions::from_table_options(opts).expect_err(rejected);
+            // Avoid `expect_err`/`unwrap_err`: they require `Debug` on the
+            // `Ok` type, but `GeoParquetReadOptions` (via `ParquetReadOptions`)
+            // does not implement `Debug`.
+            let err = match GeoParquetReadOptions::from_table_options(opts) {
+                Ok(_) => panic!("{rejected} should be rejected"),
+                Err(e) => e,
+            };
             assert!(
                 err.contains("Unknown AWS option"),
                 "{rejected} should be rejected, got: {err}"
--- a/rust/sedona-geoparquet/src/provider.rs
+++ b/rust/sedona-geoparquet/src/provider.rs
@@ -125,16 +125,23 @@ impl GeoParquetReadOptions<'_> {
     pub fn from_table_options(options: HashMap<String, String>) -> Result<Self, String> {
         for key in options.keys() {
             if key.starts_with("aws.") {
+                // Keep this list in sync with the options consumed by
+                // `AwsOptions::set` in `sedona::object_storage`. The previous
+                // list accepted several options (`aws.bucket_name`,
+                // `aws.use_ssl`, `aws.force_path_style`, `aws.nosign`) that are
+                // never consumed there -- they passed validation here but were
+                // silently dropped by `to_listing_options`, so a user setting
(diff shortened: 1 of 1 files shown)


paleolimbot (commented):
Thank you!

It looks like there's a minor issue with the test compiling and `cargo fmt` is needed, but I think this is a reasonable change!

adonm:
Thanks will refine and update (was a bit rushed heh just getting patch working for some testing with local s3)

adonm:
Ran the CI matrix locally on `86947d2`, since the workflow runs are currently awaiting maintainer approval (PR from a fork). All previously-failing checks pass.

**Whole repo:**
- `cargo fmt --all -- --check` ✅ — CI `fmt`
- `pre-commit run --all-files` ✅ — CI `pre-commit` (incl. rustfmt, the original failure)

**Affected crate `sedona-geoparquet` (the only file changed):**
- `cargo clippy -p sedona-geoparquet --all-targets --all-features -- -Dwarnings` ✅ — CI `clippy` flags; compiles the `s2geography` feature with no warnings
- `cargo test -p sedona-geoparquet --all-targets` ✅ — **63 passed, 0 failed** (includes the new `aws_option_allowlist_matches_consumer` plus all data-driven tests)

**Two checks couldn't be fully reproduced locally, both due to dependency-version mismatches in crates this PR does not touch:**
- `--all-features` test-binary *link* of the `s2geography` feature: Homebrew's abseil is built as shared libs, which `rust-lld` can't resolve here (CI links static vcpkg libs). clippy compiles this path clean.
- `cargo check --workspace`: `gdal-sys` 0.12.0 ships no pre-built bindings for Homebrew's GDAL 3.13 (CI's ubuntu `libgdal-dev` is older). Affects only `sedona-raster`.

The change is test-only in `GeoParquetReadOptions::from_table_options` and cannot reach the raster or s2 crates, so these two local blockers are environmental, not code.

paleolimbot (approved):
Thank you!

PR #1016, opened 2026-06-29
chore(python/sedonadb): Use existing WKT normalization in ST_Normalize test

### LINKED ISSUE #815: python/sedonadb: Investigate formatting issues when using WKT as an expected value

paleolimbot:
This PR fixes a test for ST_Normalize that used ST_AsText in the query, which in general we don't need to do (unless we're testing ST_AsText!). The Engine testing harness knows about geometry return types and serializes them for use in `assert_query_result()`.

Apologies to @oglego  and @ajaypadwal73...I did not do a very good job writing that ticket and you both solved a slightly different problem because of my vague phrasing!

Closes #815.

### DIFF
--- a/python/sedonadb/tests/functions/test_functions.py
+++ b/python/sedonadb/tests/functions/test_functions.py
@@ -2525,7 +2525,7 @@ def test_st_reduceprecision(eng, geom, grid_size, expected):
     ("geom", "expected"),
     [
         (None, None),
-        ("POINT EMPTY", "POINT EMPTY"),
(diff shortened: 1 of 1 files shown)


oglego:
@paleolimbot no worries at all!  Thanks for fixing this!

james-willis (approved):

ajaypadwal73:
No worries, thanks for clarifying, on to the next one! @paleolimbot

PR #1017, opened 2026-06-29
chore(python/sedonadb,python/sedonadb-zarr): Update pyo3 to 0.29

paleolimbot:
Quick follow-up to #1012 (which took care of the hard part of enabling free threading) and taking care of the harder upgrade...this updates to the latest version fixes compile errors in the process.

Closes #949.


james-willis (approved):
nice work

PR #1019, opened 2026-06-30
build(deps): bump adbc to 0.23 and adapt Statement::execute

james-willis:
`adbc_core` 0.23 changed `Statement::execute` to return `Box<dyn RecordBatchReader + Send + 'static>` instead of `impl RecordBatchReader` (our impl body already produced the boxed reader, so it's a one-line signature change). Bumps the `adbc_core`/`adbc_ffi` floor to `>=0.23.0` to match.

Supersedes #892 (the Dependabot bump), which couldn't build without this code change.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -52,9 +52,9 @@ dependencies = [
 
 [[package]]
 name = "adbc_core"
-version = "0.22.0"
+version = "0.23.0"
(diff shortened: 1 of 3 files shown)


james-willis:
@paleolimbot this should be the last of my dependency clean up pings. PTAL

paleolimbot (approved):
Thank you!

PR #1021, opened 2026-07-02
fix: arrow version alignment

viktornordling:
The bump by dependabot was failing, so here's my attempt to fix it.


viktornordling:
Closing, will wait for DataFusion to bump to version 59 of arrow.

PR #1022, opened 2026-07-03
fix(c/sedona-tg): Fix polygon hole containment predicates

### LINKED ISSUE #709: ST_Covers and ST_Contains output does not match DuckDB or Shapely
Hi,

I was porting some DuckDB and Shapely code to SedonaDB to see if I could get better performance. While it does run faster, I found that the output of `ST_Covers` was not identical to PostGIS, DuckDB and Shapely in some cases.

I narrowed it down to the case when the child geometry is identical to a hole of the parent geometry.

Here's the smallest reproduction I could create:

```
# /// script
# requires-python = ">=3.14"
# dependencies = [
#     "apache-sedona[db]>=1.8.1",
#     "duckdb>=1.5.0",
#     "shapely>=2.1.2",
# ]
# ///

import duckdb
import sedona.db
import shapely

parent_wkt = "POLYGON ((0 0, 3 0, 3 3, 0 3, 0 0), (1 1, 1 2, 2 1, 1 1))"
child_wkt = "POLYGON ((1 1, 1 2, 2 1, 1 1))"

duckdb_connection = duckdb.connect()
duckdb_connection.execute("install spatial")
duckdb_connection.execute("load spatial")

sedona_connection = sedona.db.connect()

parent_geometry = shapely.from_wkt(parent_wkt)
child_geometry = shapely.from_wkt(child_wkt)

for predicate, shapely_value in [
    ("ST_Intersects", shapely.intersects(parent_geometry, child_geometry)),
    ("ST_Covers", shapely.covers(parent_geometry, child_geometry)),
    ("ST_Contains", shapely.contains(parent_geometry, child_geometry)),
]:
    duckdb_value = duckdb_connection.execute(
        f"select {predicate}(ST_GeomFromText('{parent_wkt}'), ST_GeomFromText('{child_wkt}'))"
    ).fetchone()[0]
    sedona_value = sedona_connection.sql(
        f"select {predicate}(ST_GeomFromWKT('{parent_wkt}'), ST_GeomFromWKT('{child_wkt}')) as value"
    ).to_pandas()["value"][0]
    print(
        f"{predicate} duckdb={duckdb_value} sedona={sedona_value} shapely={shapely_value}"
    )
```

Output:

```
ST_Intersects duckdb=True sedona=True shapely=True
ST_Covers duckdb=False sedona=True shapely=False
ST_Contains duckdb=False sedona=True shapely=False
```

Postgis:

```
postgres=# SELECT
  ST_Intersects(parent, child) AS intersects,
  ST_Covers(parent, child) AS covers,
  ST_Contains(parent, child) AS contains
FROM (
  SELECT
    'POLYGON ((0 0, 3 0, 3 3, 0 3, 0 0), (1 1, 1 2, 2 1, 1 1))'::geometry AS parent,
    'POLYGON ((1 1, 1 2, 2 1, 1 1))'::geometry AS child
) t;
 intersects | covers | contains
------------+--------+----------
 t          | f      | f
(1 row)
```

`ST_Covers` and `ST_Contains` return true with SedonaDB but false with the other 3.

ajaypadwal73:
## Summary
- Fix tg polygon containment so ST_Covers/ST_Contains return false when the child polygon is identical to a parent polygon hole
- Add Rust and Python predicate regressions for the issue case

## Tests
- cargo test -p sedona-tg
- cargo fmt --check
- .venv/bin/python -m pytest -q python/sedonadb/tests/functions/test_predicates.py::test_st_contains python/sedonadb/tests/functions/test_predicates.py::test_st_covers

Fixes #709

### DIFF
--- a/c/sedona-tg/src/tg.rs
+++ b/c/sedona-tg/src/tg.rs
@@ -306,12 +306,22 @@ mod test {
 
         test_predicate::<Contains>("POLYGON ((0 0, 5 0, 0 5, 0 0))", "POINT (1 1)", true);
         test_predicate::<Contains>("POLYGON ((0 0, 5 0, 0 5, 0 0))", "POINT (-1 -1)", false);
+        test_predicate::<Contains>(
+            "POLYGON ((0 0, 3 0, 3 3, 0 3, 0 0), (1 1, 1 2, 2 1, 1 1))",
+            "POLYGON ((1 1, 1 2, 2 1, 1 1))",
+            false,
+        );
 
         test_predicate::<Within>("POINT (1 1)", "POLYGON ((0 0, 5 0, 0 5, 0 0))", true);
         test_predicate::<Within>("POINT (-1 -1)", "POLYGON ((0 0, 5 0, 0 5, 0 0))", false);
 
         test_predicate::<Covers>("POLYGON ((0 0, 5 0, 0 5, 0 0))", "POINT (1 1)", true);
         test_predicate::<Covers>("POLYGON ((0 0, 5 0, 0 5, 0 0))", "POINT (-1 -1)", false);
+        test_predicate::<Covers>(
+            "POLYGON ((0 0, 3 0, 3 3, 0 3, 0 0), (1 1, 1 2, 2 1, 1 1))",
+            "POLYGON ((1 1, 1 2, 2 1, 1 1))",
+            false,
+        );
 
         test_predicate::<CoveredBy>("POINT (1 1)", "POLYGON ((0 0, 5 0, 0 5, 0 0))", true);
         test_predicate::<CoveredBy>("POINT (-1 -1)", "POLYGON ((0 0, 5 0, 0 5, 0 0))", false);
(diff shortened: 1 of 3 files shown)


ajaypadwal73:
@paleolimbot can you please take a look at this?

ajaypadwal73:
@paleolimbot just following up.

ajaypadwal73:
@paleolimbot Ha, fair enough, quality over quantity on my end going forward! No rush at all, and thanks for the context on the perf angle, that makes total sense for a hot-path predicate. If it'd help, I'm happy to throw a benchmark on the predicate path so the perf impact is easy to eyeball. Otherwise I'll leave these in your capable hands and go bother some other part of the codebase. 🙂

paleolimbot (approved):
Thank you!

I did dig into this a little and I believe it's correct.

Did you want to fix the other issue while you're here? Contains(`GEOMETRYCOLLECTION (POINT (0 0), LINESTRING (0 0, 0 1))`, `POINT (0 0)`) should evaluate to false, but actually evaluates to true. We can also make a follow-up to track that since it looks like tidwall/tg isn't accepting issues or PRs at the moment.

PR #1025, opened 2026-07-04
dev/release: Simplify cargo publish

### LINKED ISSUE #702: dev/release: Simplify cargo publish

ajaypadwal73:
## Summary
- add a cargo dependency graph checker and derive publish order from cargo metadata
- split random data generation, base test helpers, and raster testing helpers out of sedona-testing
- move raster sd_format handling into sedona-raster-functions and register its kernel from sedona
- relocate pointcloud integration tests to the top-level sedona crate

Fixes #702

## Verification
- cargo fmt
- python3 dev/release/check-cargo-dependencies.py --print-publish-order
- cargo check/test sweeps run locally during review


paleolimbot (commented):
I will need a bit of time to look this over carefully, but in the meantime can you add a run of check-cargo-dependencies.py to one of the CI runs? Possibly the packaging run, or rust.yml with a new matrix name ("chekc-publish"?). Then when I review I'll know it works!

ajaypadwal73:
@paleolimbot Done, added a check-publish matrix job in rust.yml that runs dev/release/check-cargo-dependencies.py --print-publish-order. The new check is passing on this PR.

PR #1027, opened 2026-07-07
feat(sedona-raster-functions): add RS_Values

james-willis:
`RS_Values(raster, points [, band]) -> List<Double>` — the plural of `RS_Value`, sampling one pixel per point.

The points are carried as a **MultiPoint** (a bare Point is also accepted), diverging from Sedona's `RS_Values(raster, Array<Geometry>)`: `SedonaType` has no `List<geometry>` representation, so the array-of-geometry signature would need new type-system plumbing. This keeps the first cut self-contained; the canonical signature can follow.

Sampling, CRS handling, and pixel decoding are shared with `RS_Value` via a new `sampling` module rather than duplicated.

> [!IMPORTANT]
> **Behavior change to `RS_Value`:** the 2-argument form `RS_Value(raster, point)` used to silently sample band 1 on a multiband raster; it now raises `raster has N bands; specify which band to sample`. This diverges from Sedona Spark and PostGIS `ST_Value` (which default to band 1 unconditionally) in favor of the convention set by `RS_SetBandNoDataValue`'s 2-argument form: omitting the band is only unambiguous on a single-band raster. Queries hitting the new error should pass the band explicitly, e.g. `RS_Value(raster, point, 1)`.

### DIFF
--- a/python/sedonadb/tests/functions/test_raster_functions.py
+++ b/python/sedonadb/tests/functions/test_raster_functions.py
@@ -326,6 +326,48 @@ def test_rs_values_default_band_requires_single_band(con):
         ).to_arrow_table()
 
 
+def test_rs_values_ensureloaded_outdb(con, sedona_testing):
+    """RS_Values over an OutDb raster exercises the needs_pixels ->
+    RS_EnsureLoaded planner path end to end: the raster from RS_FromPath carries
+    no pixels, so the planner must materialise it before the kernel samples.
+
+    sentinel2.tif's top-left pixel holds 2324 (see test_rs_ensureloaded); its
+    world center is derived from the raster's own georeference so the test does
+    not hard-code the file's geotransform.
+    """
+    path = sedona_testing / "data/raster/sentinel2.tif"
+    t = con.sql("SELECT RS_FromPath($1) AS raster", params=(str(path),))
+    view = "test_rs_values_ensureloaded_outdb_raster"
+    t.to_view(view)
+    try:
+        meta = con.sql(
+            f"SELECT RS_GeoReference(raster) AS georef, RS_SRID(raster) AS srid FROM {view}"
+        ).to_arrow_table()
+        georef = [float(v) for v in meta["georef"][0].as_py().split()]
+        scale_x, skew_y, skew_x, scale_y, ul_x, ul_y = georef
+        srid = meta["srid"][0].as_py()
+        # World center of pixel (0, 0): upper-left corner + half a pixel step.
+        cx = ul_x + 0.5 * scale_x + 0.5 * skew_x
+        cy = ul_y + 0.5 * skew_y + 0.5 * scale_y
+
+        result = (
+            con.sql(
+                f"""
+            SELECT RS_Values(
+                raster,
+                ST_GeomFromText('MULTIPOINT ({cx} {cy})', 'EPSG:{srid}'),
+                1
+            ) AS v FROM {view}
+            """
+            )
+            .to_arrow_table()["v"][0]
+            .as_py()
+        )
+        assert result == [2324.0]
+    finally:
+        con.drop_view(view)
+
+
 def test_rs_values_matches_rasterio(con):
     """Cross-check RS_Values against rasterio on a random raster.
 
--- a/python/sedonadb/tests/functions/test_raster_functions.py
+++ b/python/sedonadb/tests/functions/test_raster_functions.py
@@ -108,33 +108,42 @@ def test_rs_ensureloaded(con, sedona_testing):
 # the top-left pixel which is set to the nodata value (127). (74.58, 110.57) is
 # the centroid of pixel (10, 10) (0-based) in the raster's OGC:CRS84 space; the
 # point and raster share a CRS so no reprojection happens. A point far outside
-# the footprint yields NULL. (The `needs_pixels` -> RS_EnsureLoaded planner path
-# is covered against a real OutDb raster by `test_rs_ensureloaded`.)
+# the footprint yields NULL. RS_Example is multiband, so the band is given
+# explicitly. (The `needs_pixels` -> RS_EnsureLoaded planner path is covered
+# against a real OutDb raster by `test_rs_ensureloaded`.)
 @pytest.mark.parametrize(
     ("expr", "expected"),
     [
         (
-            "RS_Value(RS_Example(), ST_SetCRS(ST_Point(74.58, 110.57), 'OGC:CRS84'))",
+            "RS_Value(RS_Example(), ST_Point(74.58, 110.57, 'OGC:CRS84'), 1)",
             1.0,
         ),
         (
-            "RS_Value(RS_Example(), ST_SetCRS(ST_Point(74.58, 110.57), 'OGC:CRS84'), 2)",
+            "RS_Value(RS_Example(), ST_Point(74.58, 110.57, 'OGC:CRS84'), 2)",
             2.0,
         ),
         (
-            "RS_Value(RS_Example(), ST_SetCRS(ST_Point(74.58, 110.57), 'OGC:CRS84'), 3)",
+            "RS_Value(RS_Example(), ST_Point(74.58, 110.57, 'OGC:CRS84'), 3)",
             3.0,
         ),
-        ("RS_Value(RS_Example(), ST_SetCRS(ST_Point(0.0, 0.0), 'OGC:CRS84'))", None),
+        ("RS_Value(RS_Example(), ST_Point(0.0, 0.0, 'OGC:CRS84'), 1)", None),
         # POINT EMPTY has no location to sample -> NULL (not an error).
         (
-            "RS_Value(RS_Example(), ST_SetCRS(ST_GeomFromText('POINT EMPTY'), 'OGC:CRS84'))",
+            "RS_Value(RS_Example(), ST_GeomFromText('POINT EMPTY', 'OGC:CRS84'), 1)",
             None,
         ),
     ],
 )
 def test_rs_value_point(expr, expected):
     SedonaDB().assert_query_result(f"SELECT {expr}", expected)
+
+
+def test_rs_value_default_band_requires_single_band(con):
+    # RS_Example has 3 bands, so omitting the band is ambiguous and errors.
+    with pytest.raises(Exception, match="specify which band"):
+        con.sql(
+            "SELECT RS_Value(RS_Example(), ST_Point(74.58, 110.57, 'OGC:CRS84'))"
+        ).to_arrow_table()
 
 
 def test_rs_value_matches_rasterio(con):
@@ -279,34 +288,42 @@ def test_rs_setbandnodatavalue_two_arg_requires_single_band():
 @pytest.mark.parametrize(
     ("expr", "expected"),
     [
-        # Default band: two in-bounds points + one outside.
+        # Two in-bounds points + one outside, on band 1.
         (
-            "RS_Values(RS_Example(), ST_SetCRS(ST_GeomFromText('MULTIPOINT (74.58 110.57, 74.58 110.57, 0 0)'), 'OGC:CRS84'))",
+            "RS_Values(RS_Example(), ST_GeomFromText('MULTIPOINT (74.58 110.57, 74.58 110.57, 0 0)', 'OGC:CRS84'), 1)",
             [1.0, 1.0, None],
         ),
         # Explicit band selects the plane; nodata corner and outside are NULL.
         (
-            "RS_Values(RS_Example(), ST_SetCRS(ST_GeomFromText('MULTIPOINT (74.58 110.57, 44.58 80.57, 0 0)'), 'OGC:CRS84'), 2)",
+            "RS_Values(RS_Example(), ST_GeomFromText('MULTIPOINT (74.58 110.57, 44.58 80.57, 0 0)', 'OGC:CRS84'), 2)",
             [2.0, None, None],
         ),
         (
-            "RS_Values(RS_Example(), ST_SetCRS(ST_GeomFromText('MULTIPOINT (74.58 110.57)'), 'OGC:CRS84'), 3)",
+            "RS_Values(RS_Example(), ST_GeomFromText('MULTIPOINT (74.58 110.57)', 'OGC:CRS84'), 3)",
             [3.0],
         ),
         # A bare Point is accepted and yields a one-element list.
         (
-            "RS_Values(RS_Example(), ST_SetCRS(ST_Point(74.58, 110.57), 'OGC:CRS84'))",
+            "RS_Values(RS_Example(), ST_Point(74.58, 110.57, 'OGC:CRS84'), 1)",
             [1.0],
         ),
         # An empty MultiPoint is an empty list (not NULL).
         (
-            "RS_Values(RS_Example(), ST_SetCRS(ST_GeomFromText('MULTIPOINT EMPTY'), 'OGC:CRS84'))",
+            "RS_Values(RS_Example(), ST_GeomFromText('MULTIPOINT EMPTY', 'OGC:CRS84'), 1)",
             [],
         ),
     ],
 )
 def test_rs_values_multipoint(expr, expected):
     SedonaDB().assert_query_result(f"SELECT {expr}", [(expected,)])
+
+
+def test_rs_values_default_band_requires_single_band(con):
+    # RS_Example has 3 bands, so omitting the band is ambiguous and errors.
+    with pytest.raises(Exception, match="specify which band"):
+        con.sql(
+            "SELECT RS_Values(RS_Example(), ST_GeomFromText('MULTIPOINT (74.58 110.57)', 'OGC:CRS84'))"
+        ).to_arrow_table()
 
 
 def test_rs_values_matches_rasterio(con):
--- a/python/sedonadb/tests/functions/test_raster_functions.py
+++ b/python/sedonadb/tests/functions/test_raster_functions.py
@@ -168,9 +168,17 @@ def test_rs_value_matches_rasterio(con):
     height, width = 7, 5
     data = rng.random((height, width)) * 1000.0
 
-    # GDAL-order geotransform: origin (100, 500), 2-wide pixels, -3 tall
-    # (north-up), no skew. Shared verbatim by both engines.
-    gdal_transform = (100.0, 2.0, 0.0, 500.0, 0.0, -3.0)
+    # Raster spans world bbox x[100, 110], y[479, 500] as 5 cols x 7 rows
+    # (2-wide, 3-tall north-up pixels, no skew). Shared verbatim by both engines.
+    xmin, ymin, xmax, ymax = 100.0, 479.0, 110.0, 500.0
+    gdal_transform = (
+        xmin,
+        (xmax - xmin) / width,
+        0.0,
+        ymax,
+        0.0,
+        -(ymax - ymin) / height,
+    )
     affine = Affine.from_gdal(*gdal_transform)
 
     # Sample points in pixel space (col_frac, row_frac).
@@ -229,7 +237,7 @@ def test_rs_value_matches_rasterio(con):
     finally:
         con.drop_view(view)
 
-    assert got == pytest.approx(expected)
+    assert got == expected
 
 
 def test_rs_setgeoreference_roundtrips_with_getter():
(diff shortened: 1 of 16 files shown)


python/sedonadb/tests/functions/test_raster_functions.py
```diff
@@ -268,3 +277,173 @@ def test_rs_setbandnodatavalue_two_arg_requires_single_band():
...
+                f"""
+            SELECT RS_Values(
+                raster,
+                ST_GeomFromText('MULTIPOINT ({cx} {cy})', 'EPSG:{srid}'),
+                1
+            ) AS v FROM {view}
+            """
+            )
+            .to_arrow_table()["v"][0]
+            .as_py()
```
paleolimbot: `Raster(t.to_arrow_table()["v"])` is zero copy, but the extraction `[0]` forces a copy in pyarrow. This is ok for testing but maybe better to set a good example for future changes that draw from this

rust/sedona-raster-functions/src/rs_value.rs
```diff
@@ -786,6 +658,74 @@ mod tests {
         assert_eq!(sample("POINT (100 100)"), None); // outside the footprint
     }
 
+    /// A two-band raster used to exercise the default-band ambiguity error.
+    fn two_band_raster() -> StructArray {
+        RasterSpec::d2(2, 2)
+            .band_values(&[1u8, 2, 3, 4])
+            .band_values(&[10u8, 20, 30, 40])
+            .transform([0.0, 1.0, 0.0, 10.0, 0.0, -1.0])
```
paleolimbot: Can be a follow-up, but using a bbox here would be a bit nicer (I believe that math got merged)

rust/sedona-raster-functions/src/rs_value.rs
```diff
@@ -786,6 +658,74 @@ mod tests {
...
+    #[test]
+    fn default_band_requires_single_band_raster_array_path() {
+        // No band argument + multiband raster is ambiguous -> error (general,
+        // array-raster path) rather than silently sampling band 1.
+        let udf: ScalarUDF = rs_value_udf().into();
+        let geom_type = SedonaType::Wkb(Edges::Planar, lnglat());
+        let tester = ScalarUdfTester::new(udf, vec![RASTER, geom_type.clone()]);
+        let geoms = create_geom_array(&[Some("POINT (0.5 9.5)")], &geom_type);
+        let err = tester
+            .invoke_arrays(vec![Arc::new(two_band_raster()), geoms])
```
paleolimbot: I think you added the requisite bits to the array tester to have this collapse to `.invoke_scalar_scalar(two_band_raster_spec(), "POINT (0.5, 9.5")`

rust/sedona-raster-functions/src/rs_value.rs
```diff
@@ -786,6 +658,74 @@ mod tests {
...
+        // Same ambiguity on the scalar-raster fast path.
+        let udf: ScalarUDF = rs_value_udf().into();
+        let geom_type = SedonaType::Wkb(Edges::Planar, lnglat());
+        let tester = ScalarUdfTester::new(udf, vec![RASTER, geom_type.clone()]);
+        let geoms = create_geom_array(&[Some("POINT (0.5 9.5)")], &geom_type);
+        let err = tester
+            .invoke(vec![
+                ColumnarValue::Scalar(ScalarValue::Struct(Arc::new(two_band_raster()))),
+                ColumnarValue::Array(geoms),
+            ])
```
paleolimbot: Also here (same invocation slightly differently phrased)

rust/sedona-raster-functions/src/rs_value.rs
```diff
@@ -786,6 +658,74 @@ mod tests {
...
+        let udf: ScalarUDF = rs_value_udf().into();
+        let geom_type = SedonaType::Wkb(Edges::Planar, lnglat());
+        let tester = ScalarUdfTester::new(udf, vec![RASTER, geom_type.clone()]);
+        let geoms = create_geom_array(&[None, None], &geom_type);
+        let result = tester
+            .invoke(vec![
+                ColumnarValue::Scalar(ScalarValue::Struct(Arc::new(two_band_raster()))),
+                ColumnarValue::Array(geoms),
+            ])
+            .unwrap();
```
paleolimbot, line 726: Also here

rust/sedona-raster-functions/src/rs_values.rs
```diff
@@ -0,0 +1,956 @@
...
+use crate::rs_ensure_loaded::NEEDS_PIXELS_METADATA_KEY;
+use crate::sampling::{
+    column_reproject_decision, default_band, int32_array_arg, next_band, read_pixel, reproject_wkb,
+    xy_to_pixel,
+};
+
+/// The `List<Float64>` output type, matching what a default
+/// `ListBuilder<Float64Builder>` produces (field "item", nullable).
+fn list_float64_type() -> DataType {
+    DataType::List(Arc::new(Field::new("item", DataType::Float64, true)))
```
paleolimbot: I think there's `DataType::new_list()` that avoids hard coding this

rust/sedona-raster-functions/src/rs_values.rs
```diff
@@ -0,0 +1,956 @@
...
+        let band_arr = if self.with_band {
+            Some(int32_array_arg(&args[2], num_iterations)?)
+        } else {
+            None
+        };
+        let band_array = band_arr.as_ref().map(|a| as_int32_array(a)).transpose()?;
+        let mut band_iter = band_array.map(|a| a.iter());
+
+        // Reprojecting the points into the raster CRS needs a PROJ engine.
+        with_global_proj_engine(|engine| {
```
paleolimbot: This can use the engine in the SedonaOptions now

rust/sedona-raster-functions/src/rs_values.rs
```diff
@@ -0,0 +1,956 @@
...
+                    engine,
+                )?;
+                let effective = reprojected.as_deref().unwrap_or(geom_wkb);
+
+                points_scratch.clear();
+                collect_point_xys(effective, &mut points_scratch)?;
+                for xy in &points_scratch {
+                    append_sample(*xy, &affine, &buffer, nodata, &mut list_builder)?;
+                }
+                list_builder.append(true);
```
paleolimbot: With a function like

```rust
fn visit_point_geometries(wkb: &Wkb, trans: &dyn CrsTransform,  visit: FnMut(x: f64 y: f64) -> Result<()>) { ... }
```

you could avoid the points scratch and the WKB copy (by transforming and sampling in place)

```rust
visit_point_geometries(geom_wkb, maybe_trans, |x, y| {
  append_sample((x, y), &affine, &buffer, nodata, &mut list_builder)
})
```

rust/sedona-raster-functions/src/rs_values.rs
```diff
@@ -0,0 +1,956 @@
...
+        raster_struct: &StructArray,
+    ) -> Result<ColumnarValue> {
+        let executor = RasterExecutor::new(arg_types, args);
+        let n = executor.num_iterations();
+
+        let all_null = |executor: &RasterExecutor| {
+            let mut builder = ListBuilder::new(Float64Builder::new());
+            for _ in 0..n {
+                builder.append_null();
+            }
```
paleolimbot: I think builders have `builder.append_nulls()` that do this?

rust/sedona-raster-functions/src/rs_values.rs
```diff
@@ -0,0 +1,956 @@
...
+        let mut geom = executor.make_geom_wkb_crs_accessor(1)?;
+
+        // Phase 1 — parse each row's Point/MultiPoint (reprojected into the
+        // raster CRS) into owned sub-point coordinates. All rows share one
+        // coordinate arena with per-row (start, len, band) spans; a `None` row
+        // is NULL output (NULL geometry or band element).
+        let mut coords: Vec<Option<(f64, f64)>> = Vec::new();
+        let mut rows: Vec<Option<(usize, usize, usize)>> = Vec::with_capacity(n);
+        let mut band_iter = band_array.map(|a| a.iter());
+        with_global_proj_engine(|engine| {
```
paleolimbot: This can use the engine from the options

rust/sedona-raster-functions/src/rs_values.rs
```diff
@@ -0,0 +1,956 @@
...
+    /// x∈[0,1), y∈[9,10) and holds 1; pixel (1, 1) holds 4.
+    fn raster_2x2() -> RasterSpec {
+        RasterSpec::d2(2, 2)
+            .band_values(&[1u8, 2, 3, 4])
+            .transform([0.0, 1.0, 0.0, 10.0, 0.0, -1.0])
+    }
+
+    /// Build MultiPoint WKB by hand: the test WKT parser cannot express an
+    /// EMPTY sub-point, which WKB encodes as a point with NaN ordinates.
+    fn multipoint_wkb(points: &[Option<(f64, f64)>]) -> Vec<u8> {
```
paleolimbot: I think we have a fixture for WKB with an empty point in sedona-testing

rust/sedona-raster-functions/src/rs_values.rs
```diff
@@ -0,0 +1,956 @@
...
+
+        // Pixel (0,0)=1, pixel (1,1)=4, and a point far outside -> NULL element.
+        let geoms = create_geom_array(
+            &[Some("MULTIPOINT (0.5 9.5, 1.5 8.5, 100 100)")],
+            &geom_type,
+        );
+        let result = tester
+            .invoke_arrays(vec![Arc::new(raster_2x2().build()), geoms])
+            .unwrap();
+        assert_eq!(row(&result, 0), vec![Some(1.0), Some(4.0), None]);
```
paleolimbot: If we implemented the scalar udf tester `Literal` for `Vec<Option<f64>>`, this might collapse to

```rust
let result = tester.invoke_scalar_scalar(raster_2x2(), "MULTIPOINT (...)").unwrap();
tester.assert_scalar_result(vec![Some(1.0), Some(4.0), None]);
```

rust/sedona-raster-functions/src/rs_values.rs
```diff
@@ -0,0 +1,956 @@
...
+            .return_type(&[RASTER, SedonaType::Wkb(Edges::Planar, lnglat())])
+            .unwrap();
+        assert_eq!(return_type, Some(SedonaType::Arrow(list_float64_type())));
+    }
+
+    #[test]
+    fn multipoint_samples_each_point_in_order() {
+        let udf: ScalarUDF = rs_values_udf().into();
+        let geom_type = SedonaType::Wkb(Edges::Planar, lnglat());
+        let tester = ScalarUdfTester::new(udf, vec![RASTER, geom_type.clone()]);
```
paleolimbot:
```suggestion 
        let tester = ScalarUdfTester::new(udf, vec![RASTER, WKB_GEOMETRY]);
```

rust/sedona-raster-functions/src/rs_values.rs
```diff
@@ -0,0 +1,956 @@
...
+            vec![
+                RASTER,
+                geom_type.clone(),
+                SedonaType::Arrow(DataType::Int32),
+            ],
+        );
+        let raster = RasterSpec::d2(2, 2)
+            .band_values(&[1u8, 2, 3, 4])
+            .band_values(&[10u8, 20, 30, 40])
+            .transform([0.0, 1.0, 0.0, 10.0, 0.0, -1.0])
```
paleolimbot: The bbox really is the right framing here...much easier to picture what you're testing

rust/sedona-raster-functions/src/rs_values.rs
```diff
@@ -0,0 +1,956 @@
...
+        // multiband raster returns NULL rather than erroring — matching the
+        // general path, which only resolves the band on rows with a geometry.
+        let udf: ScalarUDF = rs_values_udf().into();
+        let geom_type = SedonaType::Wkb(Edges::Planar, lnglat());
+        let tester = ScalarUdfTester::new(udf, vec![RASTER, geom_type.clone()]);
+        let raster = RasterSpec::d2(2, 2)
+            .band_values(&[1u8, 2, 3, 4])
+            .band_values(&[10u8, 20, 30, 40])
+            .transform([0.0, 1.0, 0.0, 10.0, 0.0, -1.0])
+            .build();
```
paleolimbot: almost all of these tests repeat this. Maybe create the scalar and the raster in a test function? Or wherever the 2x2 raster is defined can add this fixture?

rust/sedona-raster-functions/src/sampling.rs
```diff
@@ -0,0 +1,229 @@
...
+/// space.
+///
+/// `func` names the calling UDF for the error messages.
+pub(crate) fn reproject_wkb(
+    func: &str,
+    wkb: &[u8],
+    geom_crs: CrsRef<'_>,
+    raster_crs: CrsRef<'_>,
+    engine: &dyn sedona_geometry::transform::CrsEngine,
+) -> Result<Option<Vec<u8>>> {
```
paleolimbot: See the other comment...I think a vertex visitor or point geometry visitor would make more sense. You can add this to sedona-geometry.

rust/sedona-raster-functions/src/rs_value.rs
james-willis: done

rust/sedona-raster-functions/src/rs_value.rs
james-willis: done, calls invoke_scalar_scalar now

rust/sedona-raster-functions/src/rs_value.rs
james-willis: this test is intentionally testing the array array path, which is a distinct code path in this functions.

intentional

rust/sedona-raster-functions/src/rs_value.rs
james-willis: this one is also intentionally testing a distinct case, this time scalar raster + array points

rust/sedona-raster-functions/src/rs_values.rs
james-willis: done

rust/sedona-raster-functions/src/rs_values.rs
james-willis: done

rust/sedona-raster-functions/src/rs_values.rs
james-willis: done

rust/sedona-raster-functions/src/rs_values.rs
james-willis: done

rust/sedona-raster-functions/src/rs_values.rs
james-willis: done

rust/sedona-raster-functions/src/rs_values.rs
james-willis: added a `make_multipoint_wkb` in sedona-testing to support this case: mixed empty and non-empty points in the multipoint.

rust/sedona-raster-functions/src/rs_values.rs
james-willis: This would hit the orphan rule, so instead we use a function `list_f64`

rust/sedona-raster-functions/src/rs_values.rs
james-willis: I don't think WKB_GEOMETRY has a CRS, which would cause an error.

rust/sedona-raster-functions/src/rs_values.rs
james-willis: done

rust/sedona-raster-functions/src/rs_values.rs
james-willis: done

rust/sedona-raster-functions/src/sampling.rs
james-willis: done

python/sedonadb/tests/functions/test_raster_functions.py
```diff
@@ -283,6 +292,183 @@ def test_rs_setbandnodatavalue_two_arg_requires_single_band():
...
+
+def test_rs_values_matches_rasterio(con):
+    """Cross-check RS_Values against rasterio on a random raster.
+
+    The plural counterpart of `test_rs_value_matches_rasterio`: the same dense
+    set of sample points is passed as a single MultiPoint, so one `RS_Values`
+    call returns a list that must match rasterio's per-point reads element for
+    element, in order.
+    """
+    import numpy as np
```
paleolimbot: This can be module-level

rust/sedona-geometry/src/transform.rs
```diff
@@ -415,6 +415,57 @@ where
...
+        GeometryType::Point(point) => visit_one(point, trans, &mut visit),
+        GeometryType::MultiPoint(multi_point) => {
+            for point in multi_point.points() {
+                visit_one(&point, trans, &mut visit)?;
+            }
+            Ok(())
+        }
+        _ => Err(SedonaGeometryError::Invalid(
+            "expected a Point or MultiPoint geometry".to_string(),
+        )),
```
paleolimbot: I think you could reasonably also accept a GeometryCollection whose members are points or multipoints?

rust/sedona-raster-functions/src/crs_utils.rs
```diff
@@ -17,11 +17,32 @@
...
+use sedona_common::option::SedonaOptions;
 use sedona_geometry::transform::{transform, CrsEngine};
+use sedona_proj::transform::with_global_proj_engine;
 use sedona_schema::crs::{deserialize_crs, CoordinateReferenceSystem, Crs, CrsRef};
 use wkb::reader::read_wkb;
 
+/// Run `f` with the session's CRS engine: the [`SedonaOptions`] runtime engine
+/// when config options are available (the query path), falling back to the
+/// process-global PROJ engine otherwise (e.g. direct `invoke_batch` calls).
+pub fn with_crs_engine<T>(
```
paleolimbot, line 31: Same as the note on this in the RS_Clip PR...if the issue is the tester, we should fix the tester so it can be invoked and sees a reasonable CRS engine where it needs to be. I'd like to drop the sedona-proj dependency completely here and this workaround doesn't quite get us closer to that.

python/sedonadb/tests/functions/test_raster_functions.py
```diff
@@ -283,6 +292,183 @@ def test_rs_setbandnodatavalue_two_arg_requires_single_band():
...
+    got = (
+        con.sql(
+            "SELECT RS_Values($1, ST_GeomFromText($2)) AS v",
+            params=(raster, wkt),
+        )
+        .to_arrow_table()["v"]
+        .to_pylist()[0]
+    )
+
+    assert got == pytest.approx(expected)
```
paleolimbot: Does this need to be approximated or is the difference here an indicator of different behaviour?

python/sedonadb/tests/functions/test_raster_functions.py
```diff
@@ -283,6 +292,183 @@ def test_rs_setbandnodatavalue_two_arg_requires_single_band():
...
+
+    from sedonadb.raster import Raster
+
+    rng = np.random.default_rng(42)
+    height, width = 7, 5
+    data = rng.random((height, width)) * 1000.0
+
+    # GDAL-order geotransform: origin (100, 500), 2-wide pixels, -3 tall
+    # (north-up), no skew. Shared verbatim by both engines.
+    gdal_transform = (100.0, 2.0, 0.0, 500.0, 0.0, -3.0)
```
paleolimbot: Something bbox-parameterized here would be helpful as well

paleolimbot (approved):
Thank you! A few things possibly worth fixing now or following up with on the next PR/issue.

rust/sedona-geometry/src/transform.rs
james-willis: Geometry Collection of points sounds reasonable but a GeomCollection of multipoints or (especially) combo of points and multipoints sounds like the return type would be challenging

paleolimbot: For the purpose of this visitor I think it's reasonable...it's correct to visit any recursive point as long as it's a point geometry (of dimension 0)

james-willis: my mistake, thought this was a question of the input of the UDF

python/sedonadb/tests/functions/test_raster_functions.py
james-willis: done

python/sedonadb/tests/functions/test_raster_functions.py
james-willis: done

rust/sedona-geometry/src/transform.rs
james-willis: done

rust/sedona-raster-functions/src/crs_utils.rs
james-willis: ok removed all imports/mentions of proj here.

python/sedonadb/tests/functions/test_raster_functions.py
james-willis: removed. wasn't needed

python/sedonadb/tests/functions/test_raster_functions.py
james-willis: done

paleolimbot (approved):
Thank you!

PR #1028, opened 2026-07-07
feat(rust/sedona-raster-gdal): add RS_AsGeoTiff

james-willis:
`RS_AsGeoTiff(raster [, …]) -> binary` — encode a raster as a GeoTIFF and return the bytes, with optional compression and tiling overloads.

Split out of the GDAL raster draft #704, reusing that implementation. Tests were reworked onto the RasterSpec harness (no external `.tiff` fixtures, and the roundtrip check reopens the bytes via the merged GDAL reader instead of the not-yet-split RS_FromGDALRaster); adds a RasterSpec-based benchmark and SQL docs.

N-D handling is deferred (GeoTIFF is inherently 2-D/multiband) — export relies on the existing N-D→2-D plane decomposition and hasn't been given dedicated N-D semantics here.


docs/reference/sql/rs_asgeotiff.qmd
```diff
@@ -0,0 +1,99 @@
...
+        GeoTIFF is written tiled rather than stripped. TIFF tile dimensions
+        must be multiples of 16.
+    - name: tile_height
+      type: integer
+      description: Tile height in pixels; must be a multiple of 16.
+---
+
+## Description
+
+`RS_AsGeoTiff` encodes a raster as a GeoTIFF (via GDAL) and returns the file
```
paleolimbot, line 77: This is another one with a lot of options that we should mark as experimental so we can improve the options handling and break compatibility

python/sedonadb/tests/functions/test_raster_functions.py
```diff
@@ -268,3 +268,27 @@ def test_rs_setbandnodatavalue_two_arg_requires_single_band():
         SedonaDB().assert_query_result(
             "SELECT RS_SetBandNoDataValue(RS_Example(), 0)", None
         )
+
+
+# RS_AsGeoTiff smoke coverage: the GDAL-backed export is tested in depth in
+# Rust (rust/sedona-raster-gdal/src/rs_as_geotiff.rs); these only guard that
+# the function is registered in the Python build and returns GeoTIFF bytes.
+@pytest.mark.parametrize(
```
paleolimbot, line 286: You have rasterio here...you can also check that its contents is identical when decoded and parameterize on data type to make sure it works for all of them with the default parameters (e.g., with default compression heuristics)

rust/sedona-raster-gdal/src/rs_as_geotiff.rs
```diff
@@ -0,0 +1,870 @@
...
+        let (compression_array, quality_array, tile_width_array, tile_height_array) =
+            match self.variant {
+                Variant::Basic => {
+                    // No extra args → all null arrays
+                    let compression = ScalarValue::Utf8(None).to_array_of_size(num_iterations)?;
+                    let quality = ScalarValue::Float64(None).to_array_of_size(num_iterations)?;
+                    let tile_width = ScalarValue::UInt32(None).to_array_of_size(num_iterations)?;
+                    let tile_height = ScalarValue::UInt32(None).to_array_of_size(num_iterations)?;
+                    (compression, quality, tile_width, tile_height)
+                }
```
paleolimbot, line 376: Just putting it out there that with options as a single argument, there are only ever two kernels and `serde_json` takes care of deserializing the options for you

rust/sedona-raster-gdal/src/rs_as_geotiff.rs
```diff
@@ -0,0 +1,870 @@
...
+            .create_copy(&driver, &vsi_path, &options_refs)
+            .map_err(|e| exec_datafusion_err!("Failed to create GeoTiff: {}", e))?;
+
+        // Read bytes from the VSI memory file; the guard cleans up.
+        let bytes = gdal
+            .get_vsi_mem_file_bytes_owned(&vsi_path)
+            .map_err(|e| exec_datafusion_err!("Failed to read GeoTiff bytes: {}", e))?;
+
+        drop(guard);
+        Ok(bytes)
```
paleolimbot: This copy (and then the second copy when this is appended to the Binary output array) seems like something we could improve. Probably wrapping the vsimem file as a Bytes would put some pressure on the vsimem system that I'm not sure it's designed for, but we could at least get rid of the second copy by returning a binary view.

rust/sedona-raster-gdal/src/rs_as_geotiff.rs
```diff
@@ -0,0 +1,870 @@
...
+/// The fractional scale matches Apache Sedona (GeoTools' `setCompressionQuality`);
+/// a value outside the range errors rather than clamping — silently clamping
+/// would turn the most likely mistake (passing a 0–100 quality like `75`) into
+/// maximum quality with no warning.
+fn jpeg_quality_option(quality: f64) -> Result<i32> {
+    if !(0.0..=1.0).contains(&quality) {
+        return exec_err!(
+            "RS_AsGeoTiff: quality must be a fraction between 0.0 and 1.0 (got {quality}); \
+             e.g. use 0.75 for JPEG quality 75"
+        );
```
paleolimbot, line 287: Are all of these error pathways tested?

rust/sedona-raster-gdal/src/rs_as_geotiff.rs
james-willis: hmm this brings up the same question from RS_Value where we have to consider if we want to match sedona spark or not. 

https://sedona.apache.org/latest/api/sql/Raster-Output/RS_AsGeoTiff/

python/sedonadb/tests/functions/test_raster_functions.py
```diff
@@ -292,3 +292,38 @@ def test_rs_asgeotiff_out_of_range_quality_errors(con):
...
+# floating-point prediction for float bands).
+@pytest.mark.parametrize("dtype", ["uint8", "uint16", "int32", "float32", "float64"])
+@pytest.mark.parametrize("compression_args", ["", ", 'DEFLATE', 0.85", ", 'LZW', 0.85"])
+def test_rs_asgeotiff_roundtrips_contents(con, dtype, compression_args):
+    import numpy as np
+
+    pytest.importorskip("rasterio")
+    from rasterio.io import MemoryFile
+
+    from sedonadb.raster import Raster
```
paleolimbot: The non-rasterio imports here can move to the top level

rust/sedona-raster-gdal/src/rs_as_geotiff.rs
```diff
@@ -210,9 +211,13 @@ impl RsAsGeoTiff {
             .create_copy(&driver, &vsi_path, &options_refs)
             .map_err(|e| exec_datafusion_err!("Failed to create GeoTiff: {}", e))?;
 
-        // Read bytes from the VSI memory file; the guard cleans up.
+        // Seize the vsimem file's buffer without copying: `VSIBuffer` owns the
+        // GDAL allocation (freed on drop) and unlinks the file, so the only
+        // byte copy left is the append into the output builder. The guard's
```
paleolimbot, line 218: You can avoid the second copy by using a BinaryView for output (although may require a similar workaround to they way we build zero copy raster bands as well). In either case, using BinaryView for output is a good idea here for any sorry sod who tries to sort this later.

rust/sedona-raster-gdal/src/rs_as_geotiff.rs
paleolimbot: We want to make a high quality interface that people actually use, and provide a compatibility route to Sedona Spark (my preference would be to modify sedona spark to use a more pleasant interface as well, so everybody can benefit).

All can be future changes...just noting that arranging the functions in this way is causing a lot of verbosity and testing workarounds.

docs/reference/sql/rs_asgeotiff.qmd
james-willis: done

python/sedonadb/tests/functions/test_raster_functions.py
james-willis: done

rust/sedona-raster-gdal/src/rs_as_geotiff.rs
james-willis: There might be a middle ground with this PR as we have more parameters than spark sedona here. Should I drop everything after the 3rd arg?

rust/sedona-raster-gdal/src/rs_as_geotiff.rs
james-willis: First copy is gone — raster_to_geotiff now returns the owned VSIBuffer (GDAL allocation, freed on drop).

changed the return to a binaryView as well below

rust/sedona-raster-gdal/src/rs_as_geotiff.rs
james-willis: Quality-out-of-range (per codec) and CreateCopy failure were; added a UDF-level unknown-compression test ('GZIP' → "Unknown compression type")
 
 GDAL internal errors (e.g.failed to get GTiff driver") were not. i dont thin theres an obvious way to trigger that

python/sedonadb/tests/functions/test_raster_functions.py
james-willis: done

rust/sedona-raster-gdal/src/rs_as_geotiff.rs
james-willis: done

rust/sedona-raster-gdal/src/rs_as_geotiff.rs
james-willis: I should note we do copy into the binaryView. need to look at options to avoid that in a minute

james-willis: ok avoided the copy into the binaryView now

paleolimbot (approved):
Thank you!

rust/sedona-raster-gdal/src/rs_as_geotiff.rs
paleolimbot: You've implemented and tested them and marked the function as experimental so it can be changed later...I think it's a better contribution with them in.

PR #1029, opened 2026-07-07
test(rust/sedona-raster-gdal): make RS_Polygonize float assertion GDAL-version tolerant

james-willis:
RS_Polygonize's float-value assertion used an exact `assert_eq!` against an f32-widened literal, which fails on GDAL 3.13 (returns exact f64 `2.7`) even though it passed on older GDAL. Compare the polygon values with a small tolerance so the test is robust across GDAL versions.


paleolimbot (approved):
Thank you!

I added https://github.com/apache/sedona-db/issues/1031 for the CI failure

PR #1030, opened 2026-07-07
feat(rust/sedona-raster-gdal): add RS_FromGDALRaster

james-willis:
`RS_FromGDALRaster(binary) -> raster` — decode GDAL-readable raster bytes (e.g. a GeoTIFF) into an in-db raster. The inverse of `RS_AsGeoTiff` and the binary counterpart of `RS_FromPath` (which references a file path as out-db).

Split out of the GDAL raster draft #704, reusing that implementation, but adapted to main: decode via the public `open` + `append_as_indb_raster` path (dropping #704's test-only `load_as_indb_raster` helper and the `arrow` umbrella `concat`), accumulating rows into a single `RasterBuilder`. Tests were reworked onto the RasterSpec / `ScalarUdfTester` harness with GDAL-generated bytes — no `.tiff` fixtures. Adds a benchmark and SQL docs.

The doc example (round-tripping through `RS_AsGeoTiff`) is shown illustratively rather than executed, so the docs build stays independent of that PR.


paleolimbot (commented):
Thank you!

Should this be named something without GDAL in the name so it can be replicated in SedonaSpark if it's useful? FromBytes? FromContent?

Does it need a "format" parameter to give GDAL a hint or is GDAL's autodetection sufficient?

paleolimbot (approved):

james-willis:
> Should this be named something without GDAL in the name so it can be replicated in SedonaSpark if it's useful?

I think the utility of the function is that it can load anything GDAL can. So to signify that we need GDAL in the name. 

I think we're pretty screwed if we want to reproduce this in spark because of that! I think we need to consider this function outside the scope of sedona spark for the foreseeable future.

> Does it need a "format" parameter to give GDAL a hint or is GDAL's autodetection sufficient?

My understanding of GDAL is that its autodetection is sufficient; I dont see formats being passed in their tutorials.

PR #1034, opened 2026-07-08
feat(python/sedonadb): add read_csv and read_json

jiayuasu:
Adds toplevel readers for delimited text and newline-delimited JSON, filling a real gap — only `read_parquet` / `read_pyogrio` existed.

## API

```python
sd.read_csv(paths, *, has_header=True, delimiter=",")
sd.read_json(paths)          # newline-delimited JSON (NDJSON / JSON Lines)
```

- Both accept a `str` / `Path` or an iterable of paths (globs/directories per DataFusion), infer the schema, and are also available as `sd.read.csv` / `sd.read.json` alongside `sd.read.parquet` / `sd.read.pyogrio`.
- `read_json` reads **newline-delimited JSON** (one object per line), not a single JSON array — documented prominently.
- **Geometry is not inferred** (CSV and NDJSON carry no spatial metadata); parse WKT/WKB columns explicitly after reading (e.g. `ST_GeomFromText`).
- `read_csv` validates that `delimiter` is a single byte and raises a clear error otherwise.

## Implementation

| File | Change |
|---|---|
| `python/sedonadb/src/context.rs` | `InternalContext::read_csv` / `read_json` — thin wrappers over DataFusion's `SessionContext::read_csv` / `read_json` (`CsvReadOptions` has_header/delimiter; `NdJsonReadOptions`). |
| `python/sedonadb/python/sedonadb/read.py` | `Read.csv` / `Read.json` in the reader namespace. |
| `python/sedonadb/python/sedonadb/context.py` | `SedonaContext.read_csv` / `read_json` delegating to the namespace, mirroring `read_parquet`. |

## Test plan

`tests/io/test_read_csv_json.py` (tmp_path files):
- csv: basic, no-header, custom delimiter, multiple paths, bad-delimiter error, `read.csv` namespace form.
- json: ndjson basic, multiple paths, `read.json` namespace form.

Doctests on `read_csv`/`read_json` use a tempfile and show the inferred output.

Local: 9 unit + 17 doctests + `ruff` + `cargo fmt --check` + `clippy -Dwarnings` all clean.

### DIFF
--- a/python/sedonadb/python/sedonadb/read.py
+++ b/python/sedonadb/python/sedonadb/read.py
@@ -160,6 +160,19 @@ class Read:
                     None if partitioning is None else list(partitioning),
                 ),
             )
+        elif format == "csv":
+            options = options.copy()
+            has_header = options.pop("has_header", True)
+            delimiter = options.pop("delimiter", ",")
+            return DataFrame(
+                self._ctx,
+                self._ctx._impl.read_csv(table_paths, options, has_header, delimiter),
+            )
+        elif format == "json":
+            return DataFrame(
+                self._ctx,
+                self._ctx._impl.read_json(table_paths, options),
+            )
         else:
             raise ValueError(f"No format registered for extension '{format}'")
 
@@ -254,6 +267,85 @@ class Read:
         return self(
             table_paths, options=options, partitioning=partitioning, format="parquet"
         )
+
+    def csv(
+        self,
+        table_paths: Union[str, Path, Iterable[str]],
+        options: Optional[Dict[str, Any]] = None,
+        has_header: bool = True,
+        delimiter: str = ",",
+    ) -> DataFrame:
+        """Create a [DataFrame][sedonadb.dataframe.DataFrame] from one or more CSV files.
+
+        The schema is inferred from the file(s). Geometry is not inferred;
+        parse WKT/WKB columns explicitly (e.g. `ST_GeomFromText`) after reading.
+
+        Args:
+            table_paths: A str, Path, or iterable of paths/URLs to CSV files.
+            options: Optional dictionary of reader/object-store options. For
+                S3 access, use `{"aws.skip_signature": True, "aws.region": "us-west-2"}`
+                for anonymous access to public buckets.
+            has_header: Whether the first row is a header. Defaults to `True`.
+            delimiter: The field delimiter, as a single byte (i.e. a
+                one-character ASCII string). Defaults to `","`.
+
+        Examples:
+
+            >>> import tempfile
+            >>> from pathlib import Path
+            >>> sd = sedona.db.connect()
+            >>> with tempfile.TemporaryDirectory() as td:
+            ...     path = Path(td) / "t.csv"
+            ...     _ = path.write_text("a,b\\n1,x\\n2,y\\n")
+            ...     sd.read.csv(path).sort("a").show()
+            ┌───────┬──────┐
+            │   a   ┆   b  │
+            │ int64 ┆ utf8 │
+            ╞═══════╪══════╡
+            │     1 ┆ x    │
+            ├╌╌╌╌╌╌╌┼╌╌╌╌╌╌┤
+            │     2 ┆ y    │
+            └───────┴──────┘
+        """
+        options = (options or {}).copy()
+        options["has_header"] = has_header
+        options["delimiter"] = delimiter
+        return self(table_paths, options=options, format="csv")
+
+    def json(
+        self,
+        table_paths: Union[str, Path, Iterable[str]],
+        options: Optional[Dict[str, Any]] = None,
+    ) -> DataFrame:
+        """Create a [DataFrame][sedonadb.dataframe.DataFrame] from newline-delimited JSON.
+
+        Reads newline-delimited JSON (NDJSON / JSON Lines) — one JSON object
+        per line — not a single JSON array. The schema is inferred.
+
+        Args:
+            table_paths: A str, Path, or iterable of paths/URLs to NDJSON files.
+            options: Optional dictionary of reader/object-store options (e.g.
+                cloud credentials for `s3://` / `gs://` paths).
+
+        Examples:
+
+            >>> import tempfile
+            >>> from pathlib import Path
+            >>> sd = sedona.db.connect()
+            >>> with tempfile.TemporaryDirectory() as td:
+            ...     path = Path(td) / "t.json"
+            ...     _ = path.write_text('{"a": 1, "b": "x"}\\n{"a": 2, "b": "y"}\\n')
+            ...     sd.read.json(path).sort("a").show()
+            ┌───────┬──────┐
+            │   a   ┆   b  │
+            │ int64 ┆ utf8 │
+            ╞═══════╪══════╡
+            │     1 ┆ x    │
+            ├╌╌╌╌╌╌╌┼╌╌╌╌╌╌┤
+            │     2 ┆ y    │
+            └───────┴──────┘
+        """
+        return self(table_paths, options=options or {}, format="json")
 
     def pyogrio(
         self,
--- a/python/sedonadb/src/context.rs
+++ b/python/sedonadb/src/context.rs
@@ -39,6 +39,28 @@ pub struct InternalContext {
     pub runtime: Arc<Runtime>,
 }
 
+/// Convert a Python options dict to a string map for the reader/object-store
+/// layer, dropping `None` values (so callers can pass `None` to mean "unset").
+fn stringify_options(
+    py: Python<'_>,
+    options: HashMap<String, Py<PyAny>>,
+) -> HashMap<String, String> {
+    options
+        .into_iter()
+        .filter_map(|(k, v)| {
+            if v.is_none(py) {
+                None
+            } else {
+                v.bind(py)
+                    .str()
+                    .and_then(|s| s.extract())
+                    .map(|s: String| (k, s))
+                    .ok()
+            }
+        })
+        .collect()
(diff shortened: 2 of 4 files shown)


python/sedonadb/python/sedonadb/context.py
```diff
@@ -289,6 +289,68 @@ def read_parquet(
...
+            table_paths: A str, Path, or iterable of paths/URLs to CSV files.
+            has_header: Whether the first row is a header. Defaults to `True`.
+            delimiter: The single-character field delimiter. Defaults to `","`.
+
+        Examples:
+
+            >>> import tempfile, os
+            >>> sd = sedona.db.connect()
+            >>> path = os.path.join(tempfile.mkdtemp(), "t.csv")
+            >>> _ = open(path, "w").write("a,b\\n1,x\\n2,y\\n")
```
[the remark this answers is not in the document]
jiayuasu: Fixed in 0a78b34 — the doctests now use `Path(...).write_text(...)`, which flushes and closes deterministically instead of relying on the file handle being GC'd.

python/sedonadb/python/sedonadb/context.py
```diff
@@ -289,6 +289,68 @@ def read_parquet(
...
+
+        Args:
+            table_paths: A str, Path, or iterable of paths/URLs to NDJSON files.
+
+        Examples:
+
+            >>> import tempfile, os
+            >>> sd = sedona.db.connect()
+            >>> path = os.path.join(tempfile.mkdtemp(), "t.json")
+            >>> _ = open(path, "w").write('{"a": 1, "b": "x"}\\n{"a": 2, "b": "y"}\\n')
```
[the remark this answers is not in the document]
jiayuasu: Fixed in 0a78b34 — the doctests now use `Path(...).write_text(...)`, which flushes and closes deterministically instead of relying on the file handle being GC'd.

python/sedonadb/python/sedonadb/read.py
```diff
@@ -255,6 +255,45 @@ def parquet(
...
+    ) -> DataFrame:
+        """Create a [DataFrame][sedonadb.dataframe.DataFrame] from one or more CSV files.
+
+        The schema is inferred from the file(s). Geometry is not inferred;
+        parse WKT/WKB columns explicitly (e.g. `ST_GeomFromText`) after reading.
+
+        Args:
+            table_paths: A str, Path, or iterable of paths/URLs to CSV files.
+            has_header: Whether the first row is a header. Defaults to `True`.
+            delimiter: The single-character field delimiter. Defaults to `","`.
```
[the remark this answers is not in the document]
jiayuasu: Good catch — DataFusion's CSV delimiter is a single byte. Docstring now reads "the field delimiter, as a single byte (i.e. a one-character ASCII string)", so a multi-byte Unicode delimiter isn't implied. The Rust side already validates single-byte and raises otherwise (tested).

python/sedonadb/python/sedonadb/context.py
```diff
@@ -289,6 +289,68 @@ def read_parquet(
...
+    ) -> DataFrame:
+        """Create a [DataFrame][sedonadb.dataframe.DataFrame] from one or more CSV files.
+
+        The schema is inferred from the file(s). Geometry is not inferred;
+        parse WKT/WKB columns explicitly (e.g. `ST_GeomFromText`) after reading.
+
+        Args:
+            table_paths: A str, Path, or iterable of paths/URLs to CSV files.
+            has_header: Whether the first row is a header. Defaults to `True`.
+            delimiter: The single-character field delimiter. Defaults to `","`.
```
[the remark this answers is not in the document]
jiayuasu: Good catch — DataFusion's CSV delimiter is a single byte. Docstring now reads "the field delimiter, as a single byte (i.e. a one-character ASCII string)", so a multi-byte Unicode delimiter isn't implied. The Rust side already validates single-byte and raises otherwise (tested).

python/sedonadb/src/context.rs
```diff
@@ -165,6 +166,46 @@ impl InternalContext {
         Ok(InternalDataFrame::new(df, self.runtime.clone()))
     }
 
+    pub fn read_csv<'py>(
+        &self,
+        py: Python<'py>,
+        table_paths: Vec<String>,
+        has_header: bool,
+        delimiter: &str,
+    ) -> Result<InternalDataFrame, PySedonaError> {
```
paleolimbot, line 184: I think we will want to expose this more similarly to `read_parquet()`, which is implemented in rust/sedona and exposes its options as a `HashMap<String, PyAny>`. There are a few things this does...one of them is register any object stores that might be required, and allows cloud store options to be set. The options as a hash map at this level makes it easier to unify the Python and (future) R code paths...this might also automatically expose them from `read()` more naturally without us having to remember to update the options if new ones get added.

It would be nice to extract out any common read behaviour while we're here (e.g., maybe using `session.state().file_format_factory()` to resolve a file format) in rust/sedona so we can reuse it as needed.

python/sedonadb/src/context.rs
```diff
@@ -165,6 +166,46 @@ impl InternalContext {
...
+        )??;
+        Ok(InternalDataFrame::new(df, self.runtime.clone()))
+    }
+
+    pub fn read_json<'py>(
+        &self,
+        py: Python<'py>,
+        table_paths: Vec<String>,
+    ) -> Result<InternalDataFrame, PySedonaError> {
+        // DataFusion reads newline-delimited JSON (one object per line).
```
paleolimbot: This has the same issues as read_csv()

python/sedonadb/python/sedonadb/context.py
```diff
@@ -289,6 +289,71 @@ def read_parquet(
             partitioning=partitioning,
         )
 
+    def read_csv(
+        self,
+        table_paths: Union[str, Path, Iterable[str]],
+        *,
+        has_header: bool = True,
+        delimiter: str = ",",
+    ) -> DataFrame:
```
paleolimbot: Optional, but maybe we can avoid duplicating this from `.read.csv()` for now? I left `read_parquet()` and `read_pyogrio()` for now but I'd like to move docs and examples to use `sd.read(...)` since most of the time it can effectively guess the format.

python/sedonadb/tests/io/test_read_csv_json.py
```diff
@@ -0,0 +1,105 @@
...
+# under the License.
+
+import pandas as pd
+import pandas.testing as pdt
+import pytest
+
+
+def _write(path, text):
+    path.write_text(text)
+    return str(path)
```
paleolimbot: All of our other tests use `with tempfile.TemporaryDirectory() as td` for this, which cleans up the files (occasionally important on Windows)

python/sedonadb/src/context.rs
jiayuasu: Done in 79f62d7. Moved the implementation to `SedonaContext::read_csv` in rust/sedona, mirroring `read_parquet`: it registers the object store via `ensure_object_store_registered_with_options` before reading, so `s3://`/`gs://` paths with `aws.*`/`gcs.*` options now work (the previous version called DataFusion's context directly and skipped that). Options cross the PyO3 boundary as `HashMap<String, PyAny>`, shared with read_parquet via a `stringify_options` helper — so they unify with the future R path and auto-flow through `read()`. I kept it a dedicated `read_csv`/`read_json` for now rather than the generic `file_format_factory` resolution; happy to do that extraction as a follow-up if you'd like.

python/sedonadb/src/context.rs
jiayuasu: Same treatment — `SedonaContext::read_json` in rust/sedona registers the object store and takes the options map.

python/sedonadb/python/sedonadb/context.py
jiayuasu: Agreed — dropped the top-level `sd.read_csv`/`sd.read_json`. The entry points are now the generic `sd.read(...)` (guesses `.csv`/`.json` by extension) and the namespace `sd.read.csv` / `sd.read.json` for format-specific options.

python/sedonadb/tests/io/test_read_csv_json.py
jiayuasu: Switched the tests (and the read.csv/read.json doctests) to `with tempfile.TemporaryDirectory() as td` to match the rest of the suite.

python/sedonadb/tests/io/test_read_csv_json.py
```diff
@@ -0,0 +1,107 @@
...
+
+import pandas as pd
+import pandas.testing as pdt
+import pytest
+
+
+def _write(td, name, text):
+    path = Path(td) / name
+    path.write_text(text)
+    return str(path)
```
paleolimbot: I would love to remove this utility if possible. I think you can inline the entire contents into a single line anyway.

paleolimbot (approved):
Thank you!

Let's defer the FileFormatFactory-based approach to another PR since this is a standalone improvement. I general I would like to have everything go through the FileFormatFactory (which has a with_options that takes a String, String) and remove all special casing of formats (except for the Python-level wrappers that document the parameters nicely), which ensures we can align Python, R, and SQL. That may be more natural after we add point cloud readers (file formats in another Python package).

python/sedonadb/tests/io/test_read_csv_json.py
jiayuasu: Done in 0f57d56 — removed the `_write` helper. Each test now creates the `Path`, writes it, and passes the `Path` straight to the reader (the reader normalizes str/Path), so no helper needed.

PR #1043, opened 2026-07-11
feat(rust/sedona-geo): Add ST_ConvexHull_Agg

pranav-walimbe:
## Summary

- Adds `ST_ConvexHull_Agg`, a grouped aggregate that computes the convex hull of all geometries in a group directly instead of going through `ST_Area(ST_ConvexHull(ST_Collect_Agg(g)))`
- Accumulates raw  coords per group and computes the hull once at emit time using `geo`'s `quick_hull` to avoid intermediate WKB parsing current method requires
- Added a new `visit_xy_coords` helper in `sedona-geometry`, and refactored the existing `geo_traits_update_xy_bounds` (used by `ST_Envelope_Agg`) to avoid duplicated logic
- `benchmarks/test_aggregates.py` now runs perf comparison across SedonaDB, PostGIS, and DuckDB for this operation
- Added the SQL reference doc (`docs/reference/sql/st_convexhull_agg.qmd`)

## Perf

benchmark on 100k points grouped into 8/100/1000-point groups: ST_ConvexHull_Agg beats SedonaDB's own old chain by 4.0–7.8x, DuckDB's equivalent chain by 4.0–4.8x, and PostGIS's by 13.2–16.4x

## Testing

- [x] relevant tests pass across sedona-geo
- [x] `pytest benchmarks/test_aggregates.py`
- [x] added integration test `test_st_convexhull_agg_matches_collect_chain`


rust/sedona-geo/src/st_convexhull_agg.rs
```diff
@@ -0,0 +1,820 @@
...
+    }
+}
+
+fn check_dimension(state: Option<i32>) -> Result<()> {
+    if state == Some(-1) {
+        exec_err!("Can't ST_ConvexHull_Agg() mixed dimension geometries")
+    } else {
+        Ok(())
+    }
+}
```
paleolimbot: Why is this an error? I think we drop the Z and M coordinates anyway in this implementation (we could keep them in some future like PostGIS).

rust/sedona-geo/src/st_convexhull_agg.rs
```diff
@@ -0,0 +1,820 @@
...
+            Arc::new(WKB_GEOMETRY.to_storage_field("hull", true)?),
+            Arc::new(Field::new("dimension", DataType::Int32, true)),
+        ])
+    }
+}
+
+fn push_hull_coords(geom: impl GeometryTrait<T = f64>, out: &mut Vec<Coord>) -> Result<()> {
+    visit_xy_coords(geom, false, &mut |x, y| out.push((x, y).into()))
+        .map_err(|e| DataFusionError::Execution(format!("ST_ConvexHull_Agg(): {e}")))
+}
```
paleolimbot, line 85: Did you experiment with only pushing the convex hull of each item? My sense is that this would not be faster but might use less memory.

There also may be an opportunity to discard points that are definitely contained within the existing convex hull (but again, that requires some compute and might not be faster).

rust/sedona-geo/src/st_convexhull_agg.rs
```diff
@@ -0,0 +1,820 @@
...
+    }
+
+    fn emit_wkb_result(&mut self, emit_to: EmitTo) -> Result<(ArrayRef, ArrayRef)> {
+        let emit_size = match emit_to {
+            EmitTo::All => self.coords.len(),
+            EmitTo::First(n) => n,
+        };
+
+        let mut hull_builder =
+            BinaryBuilder::with_capacity(emit_size, emit_size * WKB_MIN_PROBABLE_BYTES);
```
paleolimbot: I am not sure it matters for performance, but you can probably calculate the exact output size here.

paleolimbot (commented):
Thank you for this!

I will take a closer look Monday, but I took a look through and have a few questions. In general this looks great!

One thing this does need is a Python integration test...I think you can reuse the benchmark, just make it smaller and put it in the Python tests so that it runs in CI.

rust/sedona-geo/src/st_convexhull_agg.rs
pranav-walimbe: good question, played around with this and it seems mem use does drop a ton but we deal with up to 5x decrease in speed from doing this so I think it makes more sense to not pursue this for now

rust/sedona-geo/src/st_convexhull_agg.rs
pranav-walimbe: good call, implementing this

rust/sedona-geo/src/st_convexhull_agg.rs
pranav-walimbe: fixed

jiayuasu:
@pranav-walimbe @paleolimbot is this PR ready to go? So excited to try this in SedonaDB!

python/sedonadb/tests/functions/test_aggregate.py
```diff
@@ -168,6 +168,32 @@ def test_st_envelope_nontrivial_input(eng, geoarrow_data):
...
+        SELECT bool_and(ST_Equals(hull_new, hull_old)) FROM (
+            SELECT
+                ST_ConvexHull_Agg(geometry) AS hull_new,
+                ST_ConvexHull(ST_Collect_Agg(geometry)) AS hull_old
+            FROM df_points
+            GROUP BY id % {num_groups}
+        )
+        """,
+        True,
+    )
```
paleolimbot, line 229: Optional, but this will generate more informative errors when it fails if you collect the first version to pandas, collect the second version to pandas, and use `geopandas.testing.assert_geodataframe_equal()`.

python/sedonadb/tests/functions/test_aggregate.py
```diff
@@ -168,6 +168,32 @@ def test_st_envelope_nontrivial_input(eng, geoarrow_data):
     geopandas.testing.assert_geodataframe_equal(df, expected, check_crs=False)
 
 
+@pytest.mark.parametrize("eng", [SedonaDB])
+def test_st_convexhull_agg_matches_collect_chain(eng, con):
```
paleolimbot: I think there are two more things we should test here:

- Parameterize this to check for a few numbers of groups including no groups (i.e., unlikely to go through the GroupsAccumulator)
- Add a test that aggregates zero rows
- Optional, but a test that confirms behaviour of a few known corner cases would help keep future PRs doing optimizations or refactoring honest. I think you can parameterize the like `(["POINT (0 1)", "POINT (0 1)"], "LINESTRING (0 0 1 1)")` and use `geoarrow.pyarrow.as_wkb([...values...])` + `pyarrow.table()` to generate the values (and `f"ST_ConvexHull(GEOMETRYCOLLECTION({', '.join(values_with_nulls_as_empties)}))"` to generate the expected value).

paleolimbot (approved):
Sorry I forgot to review this Monday! 😬 

This implementation looks great...reducing memory usage is a battle for another day (maybe DataFusion will make it easier to get a picture of current memory usage to see if the lower memory usage could help).

Just a few testing nits...thank you for working on this!

paleolimbot (approved):
Thank you!

PR #1044, opened 2026-07-12
feat(python/sedonadb): add DataFrame.to_csv and to_json

jiayuasu:
Adds tabular writers `DataFrame.to_csv` and `DataFrame.to_json`, the write-side complement to the `read_csv`/`read_json` readers from #1034.

- `to_csv(path, *, has_header=True, delimiter=",")`
- `to_json(path)` — newline-delimited JSON (NDJSON)

## Design notes

- These wrap DataFusion's native `write_csv`/`write_json` directly. Unlike GeoParquet (which needs the custom `write_geoparquet` trait method for geo-metadata handling), plain CSV/JSON need no `rust/sedona` layer — the asymmetry is intentional.
- **No `single_file_output` parameter.** `write_csv`/`write_json` pass only `partition_by` to `copy_to` and ignore `single_file_output`, so file-vs-directory is driven purely by the path: a `.csv`/`.json` path is written as a single file; any other path is treated as a directory and written as one file per partition. This behavior is documented rather than exposed as a no-op knob.
- **Not spatial-aware.** A geometry column is written as hex-encoded WKB (a legitimate interchange format — matches PostGIS CSV exports). Docstrings steer spatial users to `ST_AsText()` for WKT or `to_pyogrio()` for GeoJSON/FlatGeobuf/GeoPackage, keeping the mental model clean: spatial → `to_pyogrio`, tabular → `to_csv`/`to_json`, columnar → `to_parquet`.

## Tests

New `tests/io/test_write_csv_json.py` (8 tests): round-trip, no-header, custom delimiter, bad-delimiter error, directory output, geometry-as-WKB, and NDJSON format. Also adds doctests to both methods.

### DIFF
--- a/python/sedonadb/python/sedonadb/dataframe.py
+++ b/python/sedonadb/python/sedonadb/dataframe.py
@@ -1566,6 +1566,66 @@ class DataFrame:
             single_file_output,
         )
 
+    def to_csv(
+        self,
+        path: Union[str, Path],
+        *,
+        has_header: bool = True,
+        delimiter: str = ",",
+    ):
+        """Write this DataFrame to CSV
+
+        This is a plain tabular writer. CSV has no geometry representation, so a
+        geometry column (or one nested inside a struct/list) raises an error
+        rather than being written in some opaque encoding; convert it to text
+        first, e.g. `select(ST_AsText(col("geometry")).alias("geometry"))`. To
+        write geometry to a spatial format such as GeoJSON, FlatGeobuf, or
+        GeoPackage, use `to_pyogrio()` instead.
+
+        A path ending in `.csv` is written as a single file; any other path is
+        treated as a directory and written as one CSV file per partition.
+
+        Args:
+            path: A filename or directory to which CSV output should be written.
+            has_header: Whether to write the column names as the first row.
+            delimiter: The single-byte field delimiter to use between values.
+
+        Examples:
+
+            >>> import tempfile
+            >>> sd = sedona.db.connect()
+            >>> td = tempfile.TemporaryDirectory()
+            >>> sd.sql("SELECT 1 AS a, 'x' AS b").to_csv(f"{td.name}/tmp.csv")
+
+        """
+        self._impl.to_csv(str(path), has_header, delimiter)
+
+    def to_json(self, path: Union[str, Path]):
+        """Write this DataFrame to newline-delimited JSON
+
+        This is a plain tabular writer that emits one JSON object per row
+        (NDJSON); it does not produce GeoJSON. Like `to_csv()`, a geometry
+        column (or one nested inside a struct/list) raises an error rather than
+        being written in some opaque encoding; convert it to text first, e.g.
+        `select(ST_AsText(col("geometry")).alias("geometry"))`. To write
+        GeoJSON, use `to_pyogrio()` with a `.geojson` path instead.
+
+        A path ending in `.json` is written as a single file; any other path is
+        treated as a directory and written as one JSON file per partition.
+
+        Args:
+            path: A filename or directory to which JSON output should be written.
+
+        Examples:
+
+            >>> import tempfile
+            >>> sd = sedona.db.connect()
+            >>> td = tempfile.TemporaryDirectory()
+            >>> sd.sql("SELECT 1 AS a, 'x' AS b").to_json(f"{td.name}/tmp.json")
+
+        """
+        self._impl.to_json(str(path))
+
     def to_pyogrio(
         self,
         path: Union[str, Path, io.BytesIO],
--- a/python/sedonadb/tests/io/test_write_csv_json.py
+++ b/python/sedonadb/tests/io/test_write_csv_json.py
@@ -1,0 +1,128 @@
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+import json
+import tempfile
+from pathlib import Path
+
+import pandas as pd
+import pandas.testing as pdt
+import pytest
+
+from sedonadb._lib import SedonaError
+
+
+def test_to_csv_round_trip(con):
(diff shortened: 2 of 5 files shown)


python/sedonadb/python/sedonadb/dataframe.py
```diff
@@ -1566,6 +1566,64 @@ def to_parquet(
...
+
+        Examples:
+
+            >>> import tempfile
+            >>> sd = sedona.db.connect()
+            >>> td = tempfile.TemporaryDirectory()
+            >>> sd.sql("SELECT 1 AS a, 'x' AS b").to_csv(f"{td.name}/tmp.csv")
+
+        """
+        self._impl.to_csv(str(Path(path)), has_header, delimiter)
```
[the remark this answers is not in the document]
jiayuasu: Good catch — fixed. Switched to `str(path)` so URI-style paths (`s3://`, `gs://`) pass through unchanged; `Path()` was collapsing the `//`. This also matches the reader side, where `read.py` uses `str(path)`.

python/sedonadb/python/sedonadb/dataframe.py
```diff
@@ -1566,6 +1566,64 @@ def to_parquet(
...
+
+        Examples:
+
+            >>> import tempfile
+            >>> sd = sedona.db.connect()
+            >>> td = tempfile.TemporaryDirectory()
+            >>> sd.sql("SELECT 1 AS a, 'x' AS b").to_json(f"{td.name}/tmp.json")
+
+        """
+        self._impl.to_json(str(Path(path)))
```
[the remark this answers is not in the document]
jiayuasu: Fixed here too — `str(path)` instead of `str(Path(path))` to keep `s3://`/`gs://` URLs intact.

python/sedonadb/tests/io/test_write_csv_json.py
```diff
@@ -0,0 +1,97 @@
...
+        out = con.read.json(p).sort("a").to_pandas()
+    pdt.assert_frame_equal(out, pd.DataFrame({"a": [1, 2], "b": ["x", "y"]}))
+
+
+def test_to_json_ndjson_format(con):
+    # to_json emits one JSON object per line (NDJSON).
+    with tempfile.TemporaryDirectory() as td:
+        p = Path(td) / "out.json"
+        con.sql("SELECT 1 AS a, 'x' AS b").to_json(p)
+        assert p.read_text() == '{"a":1,"b":"x"}\n'
```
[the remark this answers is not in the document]
jiayuasu: Done — the NDJSON test now parses each line with `json.loads` and asserts on the resulting objects plus the line count, instead of matching the raw serialized bytes.

paleolimbot (commented):
For the behaviour here, I think we do want some geo-specific overrides for the output. I believe both GeoPandas and DuckDB via spatial will output WKT when writing CSV, which I think is a better default (I am not sure SedonaDB users will be expecting a PostGIS CSV writer default). There are a few alternatives for this:

- The "right" way to do this is to wrap the FileFormat for CSV/JSON (which is the only way to make this work in SQL and respect row ordering),
- Make an internal function `sd_prepare_write(x, format)` and add a projection automatically in the `SedonaDataFrame`'s `write_sedona_csv()`. This is similar to how formatting is implemented, but beware that adding a projection can add a repartition of the input, and can thus subtly mess with expected row order (I'm not sure the row order was guaranteed anyway, but this matters for display and simplifying arrow types for output, which is why it's implemented slightly differently there).

I am not sure what the expectation would be for LDJSON output that contains a geometry column (probably WKT makes more sense than whatever would happen by default). The geometry fragment is another option, but I am not sure that is possible to inline into the JSON output (e.g., the output of `ST_AsGeoJSON()` in DuckDB spatial https://duckdb.org/docs/current/core_extensions/spatial/functions#st_asgeojson ), and it wouldn't roundtrip without some modification of the JSON reader, too.

A way to circumvent all of this is to error when there is a geometry (or nested geometry) column with a message suggesting the required projection, like `geometry=sd.col("geometry").geo.as_text()`. That should also be in the `SedonaDataFrame` trait so we can wire it up in R too.

jiayuasu:
Went with the error approach (option 3). Added `write_sedona_csv`/`write_sedona_json` to the `SedonaDataFrame` trait so the guard lives in Rust and R can reuse it. A geometry/geography column — including one nested inside a struct/list (e.g. `ST_Dump` output) — now errors with a message pointing at the text projection, rather than silently writing hex WKB:

```
Cannot write geometry column(s) ["geometry"] to CSV: this format has no geometry representation. Convert them to text first (e.g. SELECT ST_AsText(geom) AS geom) before writing.
```

Same guard for LDJSON. I left the WKT-as-*default* (wrapping the CSV/JSON `FileFormat`) for the `FileFormatFactory` work as you suggested — this PR just stops the surprising hex-WKB default.

rust/sedona/src/context.rs
```diff
@@ -741,6 +759,80 @@ impl SedonaDataFrame for DataFrame {
...
+        self.write_json(path, DataFrameWriteOptions::new(), None::<JsonOptions>)
+            .await
+    }
+}
+
+/// Reject a schema that contains geometry/geography columns (including columns
+/// nested inside a struct, list, or map) for a text output format that has no
+/// geometry representation, with a message naming the columns and the required
+/// text projection. Lives here (not in the Python binding) so R can reuse it.
+fn reject_geometry_columns(schema: &Schema, format: &str) -> Result<()> {
```
paleolimbot, line 793: There's a SedonaSchema trait for schema-level geometry inspection where this could live, too (may have a higher chance of getting reused there)

python/sedonadb/tests/io/test_write_csv_json.py
```diff
@@ -0,0 +1,134 @@
...
+        p = Path(td) / "out.csv"
+        con.sql("SELECT 1 AS a, 'x' AS b").to_csv(p, delimiter=";")
+        assert p.read_text() == "a;b\n1;x\n"
+        # And it round-trips when read back with the same delimiter.
+        out = con.read.csv(p, delimiter=";").to_pandas()
+    pdt.assert_frame_equal(out, pd.DataFrame({"a": [1], "b": ["x"]}))
+
+
+def test_to_csv_bad_delimiter_raises(con):
+    from sedonadb._lib import SedonaError
```
paleolimbot: This is imported in a number of individual tests and can be hoisted to the module level

paleolimbot (approved):
Thank you!

rust/sedona/src/context.rs
jiayuasu: Moved it — added `geometry_column_indices_recursive()` as a default method on the `SedonaSchema` trait (backed by a nested-aware `sedona_type_contains_geometry` helper), and `reject_geometry_columns` now calls it. Kept it a default method so existing `SedonaSchema` impls are unaffected, and added a unit test covering the nested (`List<Struct<..., geom>>`) case.

python/sedonadb/tests/io/test_write_csv_json.py
jiayuasu: Done — hoisted `from sedonadb._lib import SedonaError` to a module-level import.

PR #1050, opened 2026-07-14
feat(python/sedonadb): add DataFrame.unnest

jiayuasu:
Adds `DataFrame.unnest(*columns)`, which expands list/array column(s) so each element becomes its own row — the relational form of pandas/GeoPandas `explode()`.

```python
df.unnest("vals")       # [10, 20, 30] -> 3 rows, other columns repeated
df.unnest("a", "b")     # multiple columns unnested in parallel (position-by-position)
df.unnest("parts")      # ST_Dump(geom) -> one row per part (spatial explode)
```

## Design

- Thin wrapper over DataFusion's `DataFrame::unnest_columns`; mirrors the existing `drop_columns` binding.
- **Engine-neutral naming** (`unnest`, matching Ibis / DuckDB SQL `UNNEST`) rather than the pandas-flavored `explode`, consistent with the relational DataFrame surface.
- Clear errors: no arguments -> `ValueError`; non-str argument -> `TypeError`; unknown column -> `SedonaError` ("No field named …"); a raw geometry (Binary) column -> `SedonaError` (unnest a multi-geometry via `ST_Dump` first).

## Tests

New `tests/expr/test_dataframe_unnest.py` (8 tests): list-to-rows expansion, parallel multi-column unnest, the `ST_Dump` + `unnest` multi-geometry explode, laziness, and the four error paths. Plus a doctest on the method.

### DIFF
--- a/python/sedonadb/python/sedonadb/dataframe.py
+++ b/python/sedonadb/python/sedonadb/dataframe.py
@@ -504,6 +504,65 @@ class DataFrame:
 
         return DataFrame(self._ctx, self._impl.drop_columns(list(cols)))
 
+    def unnest(self, *columns: str) -> "DataFrame":
+        """Expand list/array column(s) so each element becomes its own row.
+
+        This is the relational form of `pandas`/`GeoPandas` `explode()`: for
+        each input row, a list-typed column is expanded to one output row per
+        element, with the other columns repeated. When several columns are
+        given, they are unnested in parallel (position by position).
+
+        A common spatial use is flattening a multi-geometry: `ST_Dump(geom)`
+        returns a list of parts, and `unnest()` turns each part into its own
+        row.
+
+        Args:
+            *columns: One or more list/array column names to expand. At least
+                one is required.
+
+        Examples:
+
+            >>> sd = sedona.db.connect()
+            >>> df = sd.sql("SELECT 'a' AS label, [10, 20, 30] AS vals")
+            >>> df.unnest("vals").show()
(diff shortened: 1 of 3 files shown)


python/sedonadb/python/sedonadb/dataframe.py
```diff
@@ -504,6 +504,49 @@ def drop(self, *cols: str) -> "DataFrame":
...
+
+        This is the relational form of `pandas`/`GeoPandas` `explode()`: for
+        each input row, a list-typed column is expanded to one output row per
+        element, with the other columns repeated. When several columns are
+        given, they are unnested in parallel (position by position).
+
+        A common spatial use is flattening a multi-geometry: `ST_Dump(geom)`
+        returns a list of parts, and `unnest()` turns each part into its own
+        row.
+
```
paleolimbot, line 518: It would be nice to show an example of this (I think you have to select() or mutate() with `df.geometry.geo.dump()` and then unnest and then `col("geometry")["parts"]` to get the struct field).

paleolimbot (approved):
Thank you!

python/sedonadb/python/sedonadb/dataframe.py
jiayuasu: Good call — added an example. It walks through `ST_Dump` → `unnest` → pulling the geometry out of the `{path, geom}` struct (`dumped["dump"]["geom"]`), rendering the two points. Used the `.geo.dump()` accessor as you suggested.

Issue #1048, opened 2026-07-13
Python Wheel tests hang after slicing integration tests


paleolimbot:
In recent python-wheels runs, we see a hang after the Raster slicing integration tests start to run. For example:

https://github.com/apache/sedona-db/actions/runs/29224578250/job/86736087794

```
2026-07-13T05:12:46.7647357Z ../../../project/python/sedonadb/tests/functions/test_raster_functions.py::test_rs_setbandnodatavalue_two_arg_requires_single_band PASSED [ 59%]
2026-07-13T05:12:46.7716041Z ../../../project/python/sedonadb/tests/functions/test_raster_nd.py::test_num_dimensions PASSED [ 59%]
2026-07-13T05:12:46.7801636Z ../../../project/python/sedonadb/tests/functions/test_raster_nd.py::test_num_dimensions_with_band PASSED [ 59%]
2026-07-13T05:12:46.7869106Z ../../../project/python/sedonadb/tests/functions/test_raster_nd.py::test_dim_names PASSED [ 59%]
2026-07-13T05:12:46.7981534Z ../../../project/python/sedonadb/tests/functions/test_raster_nd.py::test_dim_size PASSED [ 59%]
2026-07-13T05:12:46.8048134Z ../../../project/python/sedonadb/tests/functions/test_raster_nd.py::test_shape PASSED [ 59%]
2026-07-13T05:12:46.8128752Z ../../../project/python/sedonadb/tests/functions/test_raster_nd.py::test_slice_drops_axis PASSED [ 59%]
2026-07-13T05:12:46.8182089Z ../../../project/python/sedonadb/tests/functions/test_raster_nd.py::test_slice_range_keeps_axis PASSED [ 59%]
2026-07-13T10:42:54.6824308Z ##[error]The operation was canceled.
2026-07-13T10:42:54.6923444Z Post job cleanup.
2026-07-13T10:42:54.6985143Z Post job cleanup.
2026-07-13T10:42:54.7659502Z [command]/usr/bin/git version
2026-07-13T10:42:54.7694858Z git version 2.54.0
```

This may not specifically relate to that test but may relate to something like loading or using GDAL functions for the first time, or using rasterio in conjunction with our functions when they are using the same GDAL shared object, or using pyogrio and rasterio's GDAL together (pyogrio would be used for reading gpkg and fgb files).

cc @james-willis

paleolimbot:
[...]

Issue #924, opened 2026-06-08
sedona-python: Fatal Python error: Aborted encountered in GitHub Action run


Kontinuation:
We hit a Python test failure in https://github.com/apache/sedona-db/actions/runs/27146906441/job/80127066116?pr=833. Not sure if it is a bug in sedona-db or pyogrio.

```
sedonadb/tests/io/test_pyogrio.py::test_read_ogr_path_suffix PASSED      [ 96%]
sedonadb/tests/io/test_pyogrio.py::test_read_ogr_file_not_found PASSED   [ 96%]
sedonadb/tests/io/test_pyogrio.py::test_write_ogr PASSED                 [ 96%]
sedonadb/tests/io/test_pyogrio.py::test_write_ogr_buffer PASSED          [ 96%]
sedonadb/tests/io/test_pyogrio.py::test_write_ogr_no_geometry PASSED     [ 96%]
sedonadb/tests/io/test_pyogrio.py::test_write_ogr_many_batches PASSED    [ 96%]
sedonadb/tests/io/test_pyogrio.py::test_write_ogr_from_view_types PASSED [ 96%]
Fatal Python error: Aborted

Thread 0x00007f4f9a3fd6c0 (most recent call first):
  <no Python frame>

Thread 0x00007f4f9a7ff6c0 (most recent call first):
  File "/opt/hostedtoolcache/Python/3.13.13/x64/lib/python3.13/contextlib.py", line 141 in __enter__
  File "/home/runner/work/sedona-db/sedona-db/python/sedonadb/python/sedonadb/datasource.py", line 208 in __init__
  File "__init__", line ??? in __init__
  <invalid frame>

Thread 0x00007f501482fb80 (most recent call first):
  File "/home/runner/work/sedona-db/sedona-db/python/sedonadb/python/sedonadb/dataframe.py", line 885 in to_arrow_table
  File "/home/runner/work/sedona-db/sedona-db/python/sedonadb/python/sedonadb/dataframe.py", line 911 in to_pandas
  File "/home/runner/work/sedona-db/sedona-db/python/sedonadb/tests/io/test_pyogrio.py", line 364 in test_read_ogr_partitioned
  File "/opt/hostedtoolcache/Python/3.13.13/x64/lib/python3.13/site-packages/_pytest/python.py", line 166 in pytest_pyfunc_call
[...]

PR #1051, opened 2026-07-14
ci(python-wheels): diagnose test hangs, serialize pyogrio open/close

james-willis:
For #1048; also addresses the pyogrio crash tracked in #924.

## Root cause

The wheel-test hang and the #924 SIGABRT are the same bug: a race between concurrent pyogrio dataset opens/closes. `pyogrio.raw.open_arrow` releases the GIL around the GDAL open, so a multi-file scan opens datasets from several DataFusion worker threads at once, while `PyogrioReaderShelter.__del__` closes datasets from whichever thread garbage collection happens to run on. Depending on platform and timing this crashes or deadlocks the process.

Main's wheel runs caught it twice this weekend, and pytest's built-in faulthandler captured the crash site on the abort ([run](https://github.com/apache/sedona-db/actions/runs/29274039339), linux-x86_64, **cp39** — so this is not specific to CPython 3.14 or arm64 as #1048 first guessed; the arm64 hangs are the same race manifesting differently):

```
Fatal Python error: Aborted

Thread 0x00007f855be77700 (most recent call first):
  File ".../contextlib.py", line 119 in __enter__
  File ".../sedonadb/datasource.py", line 208 in __init__     # PyogrioReaderShelter: pyogrio open_arrow.__enter__
  File ".../sedonadb/datasource.py", line 189 in open_reader  # on a worker thread, during test_read_ogr_multi_file
```

In the same run, macOS-arm64 hung until GitHub's 6-hour job kill; the [next main run](https://github.com/apache/sedona-db/actions/runs/29356925862) wedged both macOS-arm64 and linux-x86_64 for 4+ hours until superseded.

## Changes

- `datasource.py`: a module-level reentrant lock serializes pyogrio dataset open/close. Reentrant because a garbage-collected shelter's `__del__` can run on a thread already holding the lock in `__init__`; the `_entered` flag keeps `__del__` from closing a context that never opened. Only open/close are serialized — reads still run concurrently.
- `python-wheels.yml`: `-o faulthandler_timeout=600` on every `CIBW_TEST_COMMAND`, so any future test stuck for 10 minutes dumps all thread stacks to the CI log instead of wedging silently, and `timeout-minutes: 150` on the wheel jobs so a wedged run fails in 2.5 hours instead of occupying a runner for 6.

The full wheel matrix on this PR [passed in 2h08m](https://github.com/apache/sedona-db/actions/runs/29358220457) while the two uninstrumented main runs above aborted/hung.

The lock is the tactical fix; the durable one is reading OGR through a Rust `ArrowArrayStream` (`OGR_L_GetArrowStream` in c/sedona-gdal) so no Python runs on worker threads at all — tracked in #924.

### DIFF
--- a/.github/workflows/python-wheels.yml
+++ b/.github/workflows/python-wheels.yml
@@ -46,6 +46,7 @@ env:
 jobs:
   windows-x86_64:
     runs-on: windows-2022
+    timeout-minutes: 150
 
     steps:
       - uses: actions/checkout@v7
@@ -75,7 +76,7 @@ jobs:
           CMAKE_TOOLCHAIN_FILE: ${{ github.workspace }}/vcpkg/scripts/buildsystems/vcpkg.cmake
           CIBW_BUILD: "*-win_amd64"
           CIBW_TEST_REQUIRES: pytest adbc_driver_manager geoarrow-pyarrow geopandas
-          CIBW_TEST_COMMAND: pytest {package}/tests -vv
+          CIBW_TEST_COMMAND: pytest {package}/tests -vv -o faulthandler_timeout=600
 
       - uses: actions/upload-artifact@v7
         with:
@@ -84,6 +85,7 @@ jobs:
 
   macOS-arm64:
     runs-on: macOS-latest
+    timeout-minutes: 150
 
     steps:
       - uses: actions/checkout@v7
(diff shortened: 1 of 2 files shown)


paleolimbot (approved):
Thank you!

PR #1052, opened 2026-07-15
feat(python/sedonadb-geopandas): GeoPandas-compatible API package (experimental)

jiayuasu:
An **experimental** GeoPandas-compatible API on top of SedonaDB, as a new pure-Python package `sedonadb-geopandas` (structured like `sedonadb-expr`). It provides `GeoDataFrame` / `GeoSeries` wrappers whose methods mirror GeoPandas but delegate to lazy SedonaDB expressions via the `.geo` accessor.

```python
import geopandas
import sedonadb_geopandas as sgpd

gdf = sgpd.from_geopandas(geopandas.read_file("cities.geojson"))
big = gdf[gdf["pop"] > 1_000_000]      # boolean-mask filter
buf = gdf.geometry.buffer(0.5).area    # element-wise .geo operations
web = gdf.to_crs("EPSG:3857")          # reproject (CRS tracked through)
out = web.to_geopandas()               # escape hatch -> real GeoDataFrame
```

## Why this exists

Two motivations, neither of which is "GeoPandas compatibility as a headline feature":

**1. Consistency across Apache Sedona engines.** Sedona already ships a GeoPandas API for Spark (`sedona.spark.geopandas`, with 39 `GeoSeries` and 10 `GeoDataFrame` functions). Someone moving between SedonaSpark and SedonaDB should meet a recognizably similar API — the same motivation as aligning Python, R, and SQL elsewhere in the project. This package deliberately mirrors the Spark side's conventions:

- the `sgpd` import alias (parallel to GeoPandas' `gpd`),
- `GeoDataFrame` / `GeoSeries` naming,
- geometry operations implemented on `GeoSeries` and delegated from the active geometry column,
- `to_geopandas()` for conversion back,

and later slices will match its signatures (for example `sjoin(other, predicate=..., distance=...)`, plus `bounds`, `envelope`, `boundary`, `geom_type`, `is_valid`, `to_wkt()` / `to_wkb()`).

It is deliberately *similar*, not identical: the Spark implementation extends `pyspark.pandas.DataFrame` and so inherits a pandas index and a much wider surface, whereas this wraps a lazy relational SedonaDB frame with no row index. The Spark contributor guide's lazy-evaluation guidance applies here too — eagerly resolving the CRS on every `GeoSeries` construction would make otherwise-lazy calls eager, so `.crs` is read from the schema and `repr()` does not execute.

**2. Exercising the core DataFrame API.** Building a real API surface on top of SedonaDB's DataFrame and expression layer puts genuine load on it and surfaces gaps. It already has:

- `DataFrame.unnest` was missing, needed for `explode` (#1050, merged),
- `ST_Union_Agg` returns NULL for point inputs, which `dissolve` depends on (#1093),
- a predicate join over two frames that both name their geometry column `geometry` requires an explicit `.alias()` per side.

## Packaging

A standalone, opt-in package kept **separate from the core `sedonadb` API** (consistent with the direction in #791), co-located in this monorepo alongside `sedonadb` and `sedonadb-expr` so it shares CI and releases and can track the core API closely. It is not a dependency of `sedonadb`.

## Dependencies

Hard dependencies are `sedonadb`, `sedonadb-expr` (for the `.geo` accessor), and `pyarrow` (for literals). **`geopandas` is an optional extra**, needed only for the `from_geopandas()` / `to_geopandas()` interop helpers, so installing this package does not pull in the full GeoPandas dependency footprint.

## Scope (v1 "core slice")

- `from_geopandas()` + a lazy default context (no explicit `connect()`), and `to_geopandas()`.
- `GeoDataFrame`: `__getitem__` (boolean-mask filter, column access, column subset), `.geometry`, `.crs`, `.columns`, `.to_crs`, `len`.
- `Series` / `GeoSeries`: comparison / `& | ~` mask building; `GeoSeries.buffer()`, `.centroid`, `.area`, `.length`.

Deferred to later slices: `sjoin` / `dissolve`, `__setitem__` (rebind), reductions, and broader `.geo` method coverage — converging on the `sedona.spark.geopandas` signatures listed above.

## Intentional differences from GeoPandas (documented in the README)

- **Lazy, not eager** — operations build a query; data materializes on `to_geopandas()`. `repr()` is deliberately cheap and does not execute (IDEs call it frequently); Jupyter gets a small `_repr_html_` preview.
- **No row index / alignment** — filters and (eventual) joins are relational, not index-aligned. Comparing against a pandas/numpy array-like raises rather than silently misaligning.
- **Immutable under the hood** — "in-place" style operations return a new frame.
- **Plotting / arbitrary `apply`** — use the `to_geopandas()` escape hatch.

A companion "Migrating from GeoPandas" cookbook documents the relational model behind each mapping.

## Tests

`pip install -e` + 14 tests comparing behavior against GeoPandas: filter and boolean composition, buffer/area, centroid, `to_crs`, roundtrip, column types and subsets, lazy `repr`, array-like rejection, geometry-column validation, and frames with no geometry. Includes an `assert_geopandas_expr_equal(gdf, op)` harness that applies the same operation to a GeoPandas frame and to the wrapper and compares results, so a broader corpus of GeoPandas expressions can be added cheaply. `ruff format` and `ruff check` clean.

### DIFF
--- a/python/sedonadb-geopandas/pyproject.toml
+++ b/python/sedonadb-geopandas/pyproject.toml
@@ -1,0 +1,55 @@
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+[build-system]
+requires = ["hatchling"]
+build-backend = "hatchling.build"
+
+[project]
+name = "sedonadb-geopandas"
+version = "0.4.0"
+description = "GeoPandas-compatible API on top of SedonaDB"
+readme = "README.md"
+requires-python = ">=3.9"
+dependencies = [
+    "sedonadb",
+    "sedonadb-expr",
+    "pyarrow",
+]
+classifiers = [
+    "Programming Language :: Python :: 3",
+    "Programming Language :: Python :: 3.9",
+    "Programming Language :: Python :: 3.10",
+    "Programming Language :: Python :: 3.11",
+    "Programming Language :: Python :: 3.12",
+    "Programming Language :: Python :: 3.13",
+    "License :: OSI Approved :: Apache Software License",
+    "Operating System :: OS Independent",
+]
+
+[project.optional-dependencies]
+# geopandas is only needed for the from_geopandas / to_geopandas interop
+# helpers, so it is an opt-in extra rather than a hard dependency.
+geopandas = ["geopandas"]
+test = [
+    "pytest",
+    "pandas",
+    "geopandas",
+]
+
+[tool.hatch.build.targets.wheel]
+packages = ["python/sedonadb_geopandas"]
--- a/python/sedonadb-geopandas/python/sedonadb_geopandas/__init__.py
+++ b/python/sedonadb-geopandas/python/sedonadb_geopandas/__init__.py
@@ -1,0 +1,54 @@
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+"""GeoPandas-compatible API on top of SedonaDB.
+
+**EXPERIMENTAL.** This package provides `GeoDataFrame` / `GeoSeries` wrappers
+whose methods mirror GeoPandas but delegate to a lazy SedonaDB engine. It is a
+compatibility layer, not a drop-in replacement: see the package README for the
+intentional differences (laziness, no row index, immutability). The API may
+change without notice.
+"""
+
+from sedonadb_geopandas._context import default_context
+from sedonadb_geopandas._frame import GeoDataFrame
+from sedonadb_geopandas._series import GeoSeries, Series
+
+__all__ = ["GeoDataFrame", "GeoSeries", "Series", "from_geopandas"]
+
+
+def from_geopandas(data, *, context=None, geometry=None):
+    """Load a `geopandas.GeoDataFrame` into a SedonaDB-backed `GeoDataFrame`.
+
+    **EXPERIMENTAL.**
+
+    Args:
+        data: A `geopandas.GeoDataFrame` (or any object accepted by
+            `SedonaContext.create_data_frame`, which surfaces an error for
+            anything that cannot be turned into a DataFrame).
+        context: An optional SedonaDB context. Defaults to a shared,
+            lazily-created one.
+        geometry: The active geometry column name. Defaults to SedonaDB's
+            primary-geometry heuristic (the same one `to_geopandas` uses).
+
+    Returns:
+        A `GeoDataFrame`.
+    """
+    ctx = context or default_context()
+    df = ctx.create_data_frame(data)
+    if geometry is None:
+        return GeoDataFrame(df)
+    return GeoDataFrame(df, geometry=geometry)
--- a/python/sedonadb-geopandas/python/sedonadb_geopandas/_frame.py
+++ b/python/sedonadb-geopandas/python/sedonadb_geopandas/_frame.py
@@ -1,0 +1,174 @@
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+"""GeoPandas-style GeoDataFrame backed by a lazy SedonaDB frame."""
+
+from sedonadb_geopandas._series import GeoSeries, Series
+
+# Rows to collect for the Jupyter rich-text (`_repr_html_`) preview.
+_REPR_HTML_ROWS = 10
+
+# Default for the `geometry` argument, distinguishing "not specified, apply the
+# heuristic" from an explicit `None` meaning "this frame has no active geometry".
+_DERIVE = object()
+
+
+def _geometry_column_names(df):
+    names = df.schema.names
+    return {names[i] for i in df.schema.geometry_column_indices}
+
+
+class GeoDataFrame:
+    """A lazy SedonaDB frame in the shape of a `geopandas.GeoDataFrame`.
+
+    **EXPERIMENTAL.** Wraps a SedonaDB `DataFrame` and tracks the active
+    geometry column. Row selection, column access, and geometry operations
+    mirror GeoPandas but build a query rather than computing eagerly; call
+    `to_geopandas()` to materialize.
+    """
+
+    def __init__(self, df, geometry=_DERIVE):
+        self._df = df
+        if geometry is _DERIVE:
+            # Fall back to SedonaDB's primary-geometry heuristic (same one
+            # `to_geopandas` uses); `None` when the frame has no geometry.
+            geometry = df._impl.primary_geometry_column()
+        elif geometry is not None and geometry not in _geometry_column_names(df):
+            if geometry not in df.schema.names:
+                raise KeyError(
+                    f"Geometry column {geometry!r} not found; columns: "
+                    f"{df.schema.names}"
+                )
+            raise ValueError(f"Column {geometry!r} is not a geometry column")
+        self._geometry_name = geometry
+
+    @property
+    def geometry(self):
+        """The active geometry column as a `GeoSeries`."""
+        if self._geometry_name is None:
+            raise AttributeError("This GeoDataFrame has no active geometry column")
+        return GeoSeries(self._df, self._df[self._geometry_name], self._geometry_name)
+
+    @property
+    def crs(self):
+        """The CRS of the active geometry column, or `None` if there is none."""
+        if self._geometry_name is None:
+            return None
+        return self._df.schema.field(self._geometry_name).type.crs
+
+    @property
+    def columns(self):
+        """Column names, mirroring `GeoDataFrame.columns`."""
+        return list(self._df.schema.names)
+
+    def __getitem__(self, key):
+        # Boolean mask -> row filter (gdf[gdf["pop"] > 1000]).
+        if isinstance(key, Series):
+            return GeoDataFrame(self._df.filter(key._expr), self._geometry_name)
+
+        # Column subset -> GeoDataFrame. Matching GeoPandas, the active geometry
+        # column is persisted when it survives the subset (rather than being
+        # re-derived, which could silently pick a different geometry column) and
+        # is dropped when it does not. GeoPandas returns a plain DataFrame in
+        # that case; here the result keeps its type but has no active geometry,
+        # so `.geometry` raises just as it does there.
+        if isinstance(key, list):
+            geometry = self._geometry_name if self._geometry_name in key else None
+            return GeoDataFrame(self._df.select(*key), geometry)
+
+        # Single column -> (Geo)Series.
+        if isinstance(key, str):
+            expr = self._df[key]
+            if key == self._geometry_name:
+                return GeoSeries(self._df, expr, key)
+            return Series(self._df, expr, key)
+
+        if isinstance(key, slice):
+            raise TypeError(
+                "Positional row slicing isn't supported: this frame has no row "
+                "index, and row order isn't guaranteed. Use head(n) for a "
+                "bounded number of rows, or filter on a column."
+            )
+
+        if isinstance(key, int):
+            # Matches GeoPandas/pandas, where an integer key is a column label.
+            raise KeyError(
+                f"Column {key!r} not found (an integer key is a column label, "
+                f"not a row position). Columns: {self.columns}"
+            )
+
+        raise TypeError(
+            f"GeoDataFrame indices must be a column name, list of names, or "
+            f"boolean mask, not {type(key).__name__}"
+        )
+
+    def head(self, n=5):
+        """Return a `GeoDataFrame` of at most `n` rows.
+
+        Note that this applies a limit without an ordering, so *which* rows come
+        back isn't guaranteed — the frame has no inherent row order.
+        """
+        return GeoDataFrame(self._df.limit(n), self._geometry_name)
+
+    def to_crs(self, crs):
+        """Reproject the geometry column to `crs` (`ST_Transform`)."""
+        if self._geometry_name is None:
+            raise ValueError("to_crs() requires an active geometry column")
+        from sedonadb.expr import lit
+
+        transformed = self._df[self._geometry_name].geo.transform(lit(crs))
+        new_df = self._df.mutate(**{self._geometry_name: transformed})
+        return GeoDataFrame(new_df, self._geometry_name)
+
+    def to_geopandas(self):
+        """Execute and return a `geopandas.GeoDataFrame` (or plain DataFrame).
+
+        The active geometry column is carried over, so a frame whose geometry
+        column is not the one SedonaDB's own heuristic would pick (for example a
+        column named `geom` alongside one named `geometry`) still comes back with
+        the expected column active.
+        """
+        result = self._df.to_pandas()
+        if self._geometry_name is not None and hasattr(result, "set_geometry"):
+            try:
+                active = result.geometry.name
+            except Exception:
+                active = None
+            if active != self._geometry_name:
+                result = result.set_geometry(self._geometry_name)
+        return result
+
+    # Alias: results carry geometry, so this returns a GeoDataFrame too.
+    to_pandas = to_geopandas
+
+    def __len__(self):
+        return self._df.count()
+
+    def __repr__(self):
+        # Cheap: no execution. IDEs/consoles call repr frequently.
+        return f"GeoDataFrame(columns={self.columns}, geometry={self._geometry_name!r})"
+
+    def _repr_html_(self):
+        # Rich Jupyter display: collect only a small preview.
+        try:
+            preview = self._df.limit(_REPR_HTML_ROWS).to_pandas()
+            table = preview._repr_html_()
+        except Exception:
+            return None  # fall back to __repr__
+        return (
+            f"<div><b>GeoDataFrame</b> (preview of up to "
+            f"{_REPR_HTML_ROWS} rows)</div>{table}"
+        )
--- a/python/sedonadb-geopandas/python/sedonadb_geopandas/_series.py
+++ b/python/sedonadb-geopandas/python/sedonadb_geopandas/_series.py
@@ -1,0 +1,139 @@
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+"""pandas/GeoPandas-style Series backed by a SedonaDB expression."""
+
+
+def _operand(df, other):
+    """Coerce the right-hand side of an operator into something usable.
+
+    A `Series` is unwrapped to its expression, but only if it came from the same
+    source frame as `df`: combining columns from two different frames has no
+    defined meaning here (there is no row alignment) and would otherwise build a
+    plan that silently returns wrong rows. A raw SedonaDB `Expr` or `Literal` is
+    passed through as-is (`lit()` is a useful escape hatch for specifying a
+    literal that carries a CRS); other scalars pass through unchanged. A
+    pandas/numpy array-like is rejected with a clear message, since it would
+    otherwise fail obscurely as a multi-element literal.
+    """
+    from sedonadb.expr import Expr, Literal
+
+    if isinstance(other, Series):
+        if other._df is not df:
+            raise ValueError(
+                "Cannot combine Series that come from different DataFrames: "
+                "there is no row alignment, so the result would be silently "
+                "wrong. Reference columns of a single frame, or join the two "
+                "frames first."
+            )
+        return other._expr
+    if isinstance(other, (Expr, Literal)):
+        return other
+    if hasattr(other, "__array__"):
+        raise TypeError(
+            "Operating against a pandas/numpy array-like isn't supported "
+            "(there is no row alignment). Operate within this frame, or collect "
+            "with to_pandas() first."
+        )
+    return other
+
+
+class Series:
+    """A single column of a lazy SedonaDB frame, in the shape of a pandas Series.
+
+    **EXPERIMENTAL.** A `Series` pairs a source SedonaDB `DataFrame` with an
+    expression over its columns. Comparisons produce a boolean `Series` usable
+    as a filter mask (`gdf[gdf["pop"] > 1000]`). Nothing is computed until
+    `to_pandas()`.
+    """
+
+    def __init__(self, df, expr, name):
+        self._df = df
+        self._expr = expr
+        self._name = name
+
+    # -- element-wise comparisons -> boolean mask --------------------------
+    def __gt__(self, other):
+        return Series(self._df, self._expr > _operand(self._df, other), self._name)
+
+    def __ge__(self, other):
+        return Series(self._df, self._expr >= _operand(self._df, other), self._name)
+
+    def __lt__(self, other):
+        return Series(self._df, self._expr < _operand(self._df, other), self._name)
+
+    def __le__(self, other):
+        return Series(self._df, self._expr <= _operand(self._df, other), self._name)
+
+    def __eq__(self, other):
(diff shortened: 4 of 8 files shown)


python/sedonadb-geopandas/python/sedonadb_geopandas/_frame.py
```diff
@@ -0,0 +1,90 @@
...
+        return self._df.to_pandas()
+
+    # Alias: results carry geometry, so this returns a GeoDataFrame too.
+    to_pandas = to_geopandas
+
+    def __len__(self):
+        return self._df.count()
+
+    def __repr__(self):
+        return repr(self.to_geopandas())
```
paleolimbot: You probably want to make this a bit cheaper since it gets called by IDEs and such (even possibly omitting any collect whatsoever). The jupyter rich-text version of this could probably do a limit + collect to GeoPandas.

python/sedonadb-geopandas/python/sedonadb_geopandas/_series.py
```diff
@@ -0,0 +1,109 @@
...
+
+    __hash__ = None
+
+    # -- materialization ---------------------------------------------------
+    def to_pandas(self):
+        """Execute and return this column as a pandas (or GeoPandas) Series."""
+        return self._df.select(self._expr.alias(self._name)).to_pandas()[self._name]
+
+    def __repr__(self):
+        return repr(self.to_pandas())
```
paleolimbot: You probably want to leave this as spitting out the expression (perhaps with a suggestion to collect and then extract the series)

python/sedonadb-geopandas/python/sedonadb_geopandas/_series.py
```diff
@@ -0,0 +1,109 @@
...
+    A ``Series`` pairs a source SedonaDB ``DataFrame`` with an expression over
+    its columns. Comparisons produce boolean ``Series`` usable as a filter mask
+    (``gdf[gdf["pop"] > 1000]``). Nothing is computed until ``to_pandas()`` /
+    display.
+    """
+
+    def __init__(self, df, expr, name):
+        self._df = df
+        self._expr = expr
+        self._name = name
```
paleolimbot, line 66: I haven't spent enough time with Pandas knockoffs to know if this is sufficient to represent a series. (Seems reasonable to me but I haven't spent much time here!)

python/sedonadb-geopandas/python/sedonadb_geopandas/_frame.py
```diff
@@ -0,0 +1,90 @@
...
+        raise TypeError(
+            f"GeoDataFrame indices must be a column name, list of names, or "
+            f"boolean mask, not {type(key).__name__}"
+        )
+
+    def to_crs(self, crs):
+        """Reproject the geometry column to ``crs`` (``ST_Transform``)."""
+        from sedonadb.expr import lit
+
+        transformed = self._df[self._geometry_name].geo.transform(lit(str(crs)))
```
paleolimbot:
```suggestion
        from sedonadb.expr import lit

        transformed = self._df[self._geometry_name].geo.transform(lit(crs))
```

(in theory lit already knows about CRS objects)

python/sedonadb-geopandas/python/sedonadb_geopandas/_frame.py
```diff
@@ -0,0 +1,90 @@
...
+
+    @property
+    def geometry(self):
+        """The active geometry column as a :class:`GeoSeries`."""
+        return GeoSeries(self._df, self._df[self._geometry_name], self._geometry_name)
+
+    @property
+    def crs(self):
+        """The CRS of the active geometry column (via a zero-row materialization)."""
+        return self._df.limit(0).to_pandas().crs
```
paleolimbot: I think you are better off with `self._df.schema.field(self._geometry_name).type.crs` here

python/sedonadb-geopandas/python/sedonadb_geopandas/_frame.py
```diff
@@ -0,0 +1,90 @@
...
+
+    Wraps a SedonaDB ``DataFrame`` and tracks the active geometry column.
+    Row selection, column access, and geometry operations mirror GeoPandas but
+    build a query rather than computing eagerly; call ``to_geopandas()`` to
+    materialize.
+    """
+
+    def __init__(self, df, geometry="geometry"):
+        self._df = df
+        self._geometry_name = geometry
```
paleolimbot: We have an internal heuristic for picking the geometry column name that we should use for the default here (see `.to_geopandas()`).

For a non-default value, probably worth making sure it's actually a geometry type. I'm not sure about geography (probably would be fine to leave it since most of the functionality will still work)

Should this use `default_context().create_data_frame(df)` here? That will surface an error for something not dataframe-able.

python/sedonadb-geopandas/python/sedonadb_geopandas/_series.py
```diff
@@ -0,0 +1,109 @@
...
+
+    Element-wise geometry operations (``buffer``, ``centroid``, …) return a new
+    ``GeoSeries``; measures (``area``, ``length``) return a numeric ``Series``.
+    Each delegates to the corresponding ``ST_*`` function via SedonaDB's ``.geo``
+    accessor.
+    """
+
+    def buffer(self, distance):
+        """Buffer each geometry by ``distance`` (``ST_Buffer``)."""
+        return GeoSeries(self._df, self._expr.geo.buffer(distance), self._name)
```
paleolimbot, line 120: I am not sure how well the IDE complete will work if you do this, but you could consider trying to assign `.__doc__` (or generating it) from the geo accessor documentation to avoid repeating it (incompletely) here. (I'm trying to get as much use as possible out of our markdown docs, and as much incentive as possible to make them great)

python/sedonadb-geopandas/python/sedonadb_geopandas/__init__.py
```diff
@@ -0,0 +1,51 @@
...
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+"""GeoPandas-compatible API on top of SedonaDB.
+
+This package provides :class:`GeoDataFrame` / :class:`GeoSeries` wrappers whose
+methods mirror GeoPandas but delegate to a lazy SedonaDB engine. It is a
+compatibility layer, not a drop-in replacement: see the package README for the
+intentional differences (laziness, no row index, immutability).
+"""
```
paleolimbot, line 24: Worth adding EXPERIMENTAL to this (and all class docs) for merging the first version of this.

python/sedonadb-geopandas/tests/test_geopandas_compat.py
```diff
@@ -0,0 +1,97 @@
...
+
+@pytest.fixture
+def cities():
+    return gpd.GeoDataFrame(
+        {"name": ["A", "B", "C"], "pop": [100, 200, 300]},
+        geometry=gpd.GeoSeries.from_wkt(["POINT (0 0)", "POINT (1 1)", "POINT (5 5)"]),
+        crs="EPSG:4326",
+    )
+
+
```
paleolimbot, line 33: These tests are great for now, but you will probably get some mileage out of a testing framework like

```python
def assert_geopandas_expr_equal(gdf, op):
    sgdf = sgpd.from_geopandas(gdf)
    gdf_result = op(gdf)
    sgdf_result = op(sgdf)
    geopandas.assert_geodataframe_equal(sgdf_result, gdf_result)
```

...and throw a corpus of derived GeoPandas code at it. That will probably work better once you have a few more components in place.

python/sedonadb-geopandas/python/sedonadb_geopandas/_series.py
```diff
@@ -0,0 +1,109 @@
...
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+"""pandas/GeoPandas-style Series backed by a SedonaDB expression."""
+
+
+def _operand(other):
+    """Unwrap a Series to its expression; pass scalars through unchanged."""
+    return other._expr if isinstance(other, Series) else other
```
paleolimbot: A few gotchas could happen here worth erroring for or catching somehow

- An actual Pandas `Series` or numpy array. These will probably fail with no change because SedonaDB usually rejects literals that are len != 1 I think, but the error would be clearer if you check for these in advance. It's natural somebody might want to interact with a regular GeoPandas dataframe and this dataframe in the same code.
- A SedonaDB expression you might want to handle as is here

python/sedonadb-geopandas/python/sedonadb_geopandas/__init__.py
```diff
@@ -0,0 +1,51 @@
...
+from sedonadb_geopandas._series import GeoSeries, Series
+
+__all__ = ["GeoDataFrame", "GeoSeries", "Series", "from_geopandas"]
+
+
+def from_geopandas(data, *, context=None, geometry=None):
+    """Load a ``geopandas.GeoDataFrame`` into a SedonaDB-backed ``GeoDataFrame``.
+
+    Args:
+        data: A ``geopandas.GeoDataFrame`` (or any object accepted by
```
paleolimbot: The double quotes are rEST here but we're using markdown (also for other docs)

python/sedonadb-geopandas/python/sedonadb_geopandas/_frame.py
```diff
@@ -0,0 +1,90 @@
...
+        transformed = self._df[self._geometry_name].geo.transform(lit(str(crs)))
+        new_df = self._df.mutate(**{self._geometry_name: transformed})
+        return GeoDataFrame(new_df, self._geometry_name)
+
+    def to_geopandas(self):
+        """Execute and return a ``geopandas.GeoDataFrame``."""
+        return self._df.to_pandas()
+
+    # Alias: results carry geometry, so this returns a GeoDataFrame too.
+    to_pandas = to_geopandas
```
paleolimbot, line 155: You probably want to handle the case where there's no geometry, also? Many pipelines pass through a stage where geometry is dropped and re-added (or dropped completely).

python/sedonadb-geopandas/pyproject.toml
```diff
@@ -0,0 +1,50 @@
...
+requires = ["hatchling"]
+build-backend = "hatchling.build"
+
+[project]
+name = "sedonadb-geopandas"
+version = "0.4.0"
+description = "GeoPandas-compatible API on top of SedonaDB"
+readme = "README.md"
+requires-python = ">=3.9"
+dependencies = [
```
paleolimbot, line 28: Do you need `geopandas` as a hard dependency here? (Doing so increases the dependency footprint by quite a bit)

Realistically this needs `pyarrow` (for literals) and `sedonadb-expr` (for the `.geo` accessor)

paleolimbot (commented):
No major concerns from my end...I don't generally think we gain a lot as a project by spending a lot of effort emulating the GeoPandas API or publicizing the fact that we did (many of the performance gains we're able to get are from the rethinking of pandas pipelines as general sequences of dataframe transformations, and I am not sure the much loved GeoPandas project sees attempts to duplicate its API kindly, although I haven't asked).

That said, this is not conceptually complicated and doesn't add complexity to the core of SedonaDB. If nothing else, developing it will help expose bugs and friction points by actual real live use of our DataFrame API. It needs some polishing but it's definitely worth merging as experimental.

jiayuasu:
Thanks for the thorough review, and for the green light to land this as experimental. Each comment has an inline reply with the specific change; the conclusions and decisions are here.

**Framing.** The PR description is rewritten around two motivations rather than GeoPandas compatibility as an end in itself.

The one I had underweighted is **cross-engine consistency**: Sedona already ships `sedona.spark.geopandas`, so someone moving between SedonaSpark and SedonaDB meeting a recognizably similar API seems worth something, in the same spirit as aligning Python, R, and SQL. The conventions converge more than expected — that implementation also uses the `sgpd` alias, puts geometry operations on `GeoSeries` and delegates from the active geometry column, and its contributor guide makes exactly the lazy-evaluation point about not eagerly resolving the CRS, which is what motivated the `.crs` and `repr` changes here. Later slices will match its signatures (`sjoin(other, predicate=..., distance=...)`, `bounds`, `envelope`, `geom_type`, `to_wkt()`, and so on) rather than inventing new ones.

The second motivation is the one named in the review, and it is holding up: this exercise has already produced #1050 (`unnest` was genuinely missing, needed for `explode`) and #1093 (`ST_Union_Agg` returns NULL for point inputs, which `dissolve` depends on), plus the smaller friction that a predicate join between two frames that both name their geometry column `geometry` needs an explicit `.alias()` per side.

**Deferred:** generating method docstrings from the `.geo` accessor documentation — reasoning in that thread; worth a follow-up.

**Open question, probably separate from this PR:** for exact symmetry with `sedona.spark.geopandas`, the natural namespace would be `sedona.db.geopandas`. That lives in the `apache-sedona` package rather than here, so the import stays `sedonadb_geopandas` for now.

Tests went 9 → 14 with the new harness; `ruff format` and `ruff check` are clean.

python/sedonadb-geopandas/python/sedonadb_geopandas/_frame.py
jiayuasu: Fixed — `__repr__` no longer collects at all; it reports the columns and the active geometry column. Added a `_repr_html_` for Jupyter that collects a bounded preview (up to 10 rows) via `limit()`, falling back to `__repr__` if that fails. A test asserts the repr does not execute.

python/sedonadb-geopandas/python/sedonadb_geopandas/_series.py
jiayuasu: Fixed — `Series.__repr__` now shows the underlying expression rather than executing, with a hint to collect: `<GeoSeries Expr(st_centroid(geometry)) (lazy; call .to_pandas())>`.

python/sedonadb-geopandas/python/sedonadb_geopandas/_series.py
jiayuasu: Noted — keeping it minimal for now: a source frame plus an expression, with comparisons and `& | ~` composition so it can act as a filter mask. Reductions and the rest of the pandas `Series` surface are deferred to a later slice, where the `sedona.spark.geopandas` `GeoSeries` is a useful reference for which parts actually get used.

python/sedonadb-geopandas/python/sedonadb_geopandas/_frame.py
jiayuasu: Applied, thanks. Confirmed `lit` handles a plain string, a `pyproj.CRS` object, and an EPSG int, so the `str()` was both unnecessary and lossy for CRS objects.

python/sedonadb-geopandas/python/sedonadb_geopandas/_frame.py
jiayuasu: Applied — `.crs` is now `self._df.schema.field(self._geometry_name).type.crs`, so it no longer executes anything. Returns `None` when the frame has no active geometry column.

python/sedonadb-geopandas/python/sedonadb_geopandas/_frame.py
jiayuasu: All three applied:

- The default is now the internal heuristic (`primary_geometry_column()`), the same one `to_geopandas()` uses, rather than hard-coding `"geometry"`.
- A non-default value is validated against the schema geometry columns: `KeyError` if the column does not exist, `ValueError` if it exists but is not a geometry. Geography is left permitted, as suggested.
- `from_geopandas` now routes through `default_context().create_data_frame(...)`, so anything not dataframe-able surfaces that error.

python/sedonadb-geopandas/python/sedonadb_geopandas/_series.py
jiayuasu: Deferred for now, but agreed on the direction. Two things I want to check first: whether assigning `__doc__` on a property (rather than the underlying function) survives IDE hover and `help()`, and how the `.geo` text reads out of context where the wrapper signature differs (for example `area` is a property here but a function on the accessor). Worth doing as a follow-up so the markdown docs stay the single source.

python/sedonadb-geopandas/python/sedonadb_geopandas/__init__.py
jiayuasu: Added — on the module docstring, `from_geopandas`, and the `GeoDataFrame` / `Series` / `GeoSeries` class docstrings, each noting the API may change without notice.

python/sedonadb-geopandas/tests/test_geopandas_compat.py
jiayuasu: Added, close to your sketch:

```python
def assert_geopandas_expr_equal(gdf, op, *, sort_by):
    expected = op(gdf).sort_values(sort_by).reset_index(drop=True)
    got = op(sgpd.from_geopandas(gdf)).to_geopandas().sort_values(sort_by).reset_index(drop=True)
    assert_geodataframe_equal(got, expected, check_like=True, check_crs=False)
```

Sorting and resetting the index is needed because neither row order nor the index is preserved, and `check_crs=False` tolerates the EPSG:4326 to CRS84 normalization. Using it for the filter test now; agreed it earns its keep once there are more components to throw a corpus at.

python/sedonadb-geopandas/python/sedonadb_geopandas/_series.py
jiayuasu: Both handled. `_operand` now passes a SedonaDB `Expr` through unchanged, and rejects anything array-like (checked via `__array__`) up front with a clear message: there is no row alignment, so it suggests either staying within the frame or collecting with `to_pandas()` first. A test covers comparing against a real pandas `Series`.

python/sedonadb-geopandas/python/sedonadb_geopandas/__init__.py
jiayuasu: Fixed — converted the docstrings across the package from reST (double backticks, `:class:` roles) to markdown.

python/sedonadb-geopandas/python/sedonadb_geopandas/_frame.py
jiayuasu: Handled. The active geometry column is `None` when the frame has none, `.crs` returns `None`, `.geometry` raises a clear `AttributeError`, and a column subset re-derives the geometry column instead of carrying a stale name — so a pipeline that drops geometry and later re-adds it behaves. A test covers the dropped-geometry case.

python/sedonadb-geopandas/pyproject.toml
jiayuasu: Fixed — hard dependencies are now `sedonadb`, `sedonadb-expr` (for the `.geo` accessor), and `pyarrow` (for literals). `geopandas` moved to an optional extra, since it is only needed for the `from_geopandas` / `to_geopandas` interop helpers.

python/sedonadb-geopandas/tests/test_geopandas_compat.py
paleolimbot: You'll want to ensure you have `check_crs=True` (you definitely want to make sure this is one of the things that propagates). To avoid that normalization issue just avoid EPSG 4326 in your test data.

python/sedonadb-geopandas/python/sedonadb_geopandas/_series.py
paleolimbot: I forgot about `Literal` which you may want to explicitly accept (allows `lit()` escape hatch for one-row-one-col GeoDataFrames, which is actually a quite useful way to specify a literal with a CRS)

python/sedonadb-geopandas/tests/test_geopandas_compat.py
jiayuasu: Fixed — the harness now uses `check_crs=True`; agreed that CRS propagation is exactly one of the things worth asserting rather than waving through.

One finding while verifying: `check_crs=True` passes with EPSG:4326 data too, including propagation through `to_crs`. The normalization I had worked around only shows up via SQL (`ST_SetSRID(..., 4326)` comes back as OGC:CRS84), not through the `from_geopandas` path, so the fixture did not actually need to change. To keep that honest either way I added a parametrized test over both a geographic (EPSG:4326) and a projected (EPSG:32633) source, plus a harness-driven `to_crs("EPSG:3857")` case that asserts propagation directly. Happy to move the fixtures off 4326 as well if relying on that seems fragile.

python/sedonadb-geopandas/python/sedonadb_geopandas/_series.py
jiayuasu: Good catch — added. `Literal` is not an `Expr` subclass, so it was only surviving by falling through the final catch-all, which is exactly the kind of accident worth making explicit. `_operand` now accepts `(Expr, Literal)` up front, with a test using `lit()` as an operand, and the docstring notes the `lit()` escape hatch for carrying a CRS.

python/sedonadb-geopandas/python/sedonadb_geopandas/_frame.py
```diff
@@ -0,0 +1,129 @@
...
+
+    def __getitem__(self, key):
+        # Boolean mask -> row filter (gdf[gdf["pop"] > 1000]).
+        if isinstance(key, Series):
+            return GeoDataFrame(self._df.filter(key._expr), self._geometry_name)
+
+        # Column subset -> GeoDataFrame; re-derive the geometry column since the
+        # subset may have dropped it.
+        if isinstance(key, list):
+            return GeoDataFrame(self._df.select(*key))
```
paleolimbot: It's worth checking what GeoPandas does for the case where the subset drops the geometry and make sure you match it here. It's either that the geometry column is preserved (sticky even if not included in the column subset), that the active geometry column is persisted if it still exists (or a regular pandas DataFrame is returned if it's dropped). I would be surprised if the geometry column is rederived (but you can test and find out).

python/sedonadb-geopandas/python/sedonadb_geopandas/_frame.py
```diff
@@ -0,0 +1,129 @@
...
+
+        # Single column -> (Geo)Series.
+        if isinstance(key, str):
+            expr = self._df[key]
+            if key == self._geometry_name:
+                return GeoSeries(self._df, expr, key)
+            return Series(self._df, expr, key)
```
paleolimbot, line 97: Do integers or slices work here?

python/sedonadb-geopandas/python/sedonadb_geopandas/_frame.py
```diff
@@ -0,0 +1,129 @@
...
+            raise ValueError("to_crs() requires an active geometry column")
+        from sedonadb.expr import lit
+
+        transformed = self._df[self._geometry_name].geo.transform(lit(crs))
+        new_df = self._df.mutate(**{self._geometry_name: transformed})
+        return GeoDataFrame(new_df, self._geometry_name)
+
+    def to_geopandas(self):
+        """Execute and return a `geopandas.GeoDataFrame` (or plain DataFrame)."""
+        return self._df.to_pandas()
```
paleolimbot: Persist the custom geometry column name here?

python/sedonadb-geopandas/python/sedonadb_geopandas/_series.py
```diff
@@ -0,0 +1,131 @@
...
+    def __or__(self, other):
+        return Series(self._df, self._expr | _operand(other), self._name)
+
+    def __invert__(self):
+        return Series(self._df, ~self._expr, self._name)
+
+    __hash__ = None
+
+    # -- materialization ---------------------------------------------------
+    # -- materialization ---------------------------------------------------
```
paleolimbot:
```suggestion
    # -- materialization ---------------------------------------------------
```

python/sedonadb-geopandas/python/sedonadb_geopandas/_series.py
```diff
@@ -0,0 +1,131 @@
...
+    `Literal` is passed through as-is (`lit()` is a useful escape hatch for
+    specifying a literal that carries a CRS); other scalars pass through
+    unchanged. A pandas/numpy array-like is rejected with a clear message —
+    there is no row alignment, and it would otherwise fail obscurely as a
+    multi-element literal.
+    """
+    from sedonadb.expr import Expr, Literal
+
+    if isinstance(other, Series):
+        return other._expr
```
paleolimbot, line 42: Sorry, I keep thinking of other things that could happen here. Do we want to allow Series that derive from different DataFrames to be mixed with each other? Is it reasonable to error for that case or will that be annoying?

python/sedonadb-geopandas/tests/test_geopandas_compat.py
paleolimbot: I don't see this used in this PR but I may have missed it (no need to to include it here if it's not used yet)

paleolimbot (approved):
A few other things worth considering (or tracking) but this is looking good!

I think there a are some other more productive/maintenance reducing migration pathways into Spark but we can pursue those in parallel / guague interest in both as we go.

python/sedonadb-geopandas/python/sedonadb_geopandas/_frame.py
jiayuasu: Tested it, and you were right to be suspicious — re-deriving was wrong. GeoPandas behaves as follows:

- subset that keeps the geometry column: `GeoDataFrame`, with the active column persisted (including a custom name like `geom`),
- subset that drops it: a plain pandas `DataFrame`, whose `.geometry` raises `AttributeError`.

Now matched. The active column is persisted when it survives the subset and set to `None` when it does not, rather than re-derived — which mattered concretely: with an active `geom` alongside a `geometry` column, re-deriving silently switched the active column to `geometry`. There is no separate non-geo frame type here, so the dropped case returns a `GeoDataFrame` with no active geometry, where `.geometry` raises `AttributeError` just as it does in GeoPandas. Tests cover both, including the custom-name case.

python/sedonadb-geopandas/python/sedonadb_geopandas/_frame.py
jiayuasu: Neither did anything useful — both fell through to a generic `TypeError`. Now handled explicitly, following GeoPandas where it is well defined:

- an integer key raises `KeyError`, since in GeoPandas/pandas an integer is a column label rather than a row position (`g[0]` raises `KeyError: 0`),
- a slice raises `TypeError` explaining that positional row slicing is not supported: there is no row index and no guaranteed row order, so `g[0:2]` cannot be answered faithfully.

For the bounded-rows case I added `head(n)`, which is well defined as a limit and matches the method that already exists on the Spark side, with a docstring noting that *which* rows come back is not guaranteed without an ordering. Happy to make slices work as a best-effort limit instead if that trade seems better, but silently returning arbitrary rows for `[0:2]` felt like the wrong default.

python/sedonadb-geopandas/python/sedonadb_geopandas/_frame.py
jiayuasu: Good catch, this was a real bug. With an active `geom` column alongside a `geometry` column, `to_geopandas()` returned a frame whose active geometry was `geometry`, because the conversion applies its own primary-geometry heuristic. It now carries the active column over explicitly (via `set_geometry` when it differs), with a test using exactly that two-geometry-column setup.

python/sedonadb-geopandas/python/sedonadb_geopandas/_series.py
jiayuasu: Fixed — that was a duplicated line I introduced when adding the section comment. Thanks for catching it.

python/sedonadb-geopandas/python/sedonadb_geopandas/_series.py
jiayuasu: No apology needed, this one was worth catching: it was silently wrong rather than annoying. Comparing Series from two different frames built a plan that returned zero rows instead of failing, so it now raises `ValueError` explaining that there is no row alignment and suggesting either referencing a single frame or joining first.

I do not think it will be annoying in practice, since a mask is almost always derived from the frame being filtered, and same-frame column-vs-column comparisons (`gdf["pop"] > gdf["other"]`) still work. Test added.

python/sedonadb-geopandas/tests/test_geopandas_compat.py
jiayuasu: It is used — three call sites, though they are further down the file so easy to miss: `test_filter_matches_geopandas`, `test_to_crs_matches_geopandas` (which is what exercises CRS propagation now that `check_crs=True`), and a parametrized `test_crs_propagates_from_projected_and_geographic` over geographic and projected sources.

PR #1053, opened 2026-07-15
test(python/sedonadb): add cross-engine raster parity engines and RS_* parity tests

james-willis:
Cross-dialect divergences found while stamping out the parity tests. SedonaDB is the **subject** of every comparison (Rasterio and Sedona Spark are comparators — comparator↔comparator agreement is never asserted), and each divergence below is a `Deviation` ledger entry enforced as a strict expected failure: the full case matrix runs, and the ledger entry itself fails the suite the day the engines converge.

1. **Rasterization centroid rule on diagonal edges** — GDAL (SedonaDB, rasterio) burns every pixel whose center is inside the geometry; Sedona Spark's hand-rolled scanline rasterizer mis-places x-intercepts on non-square pixels (a pixel-space Δy is divided by a world-space slope) and drops some center-inside pixels along diagonal edges (on a small test triangle, 4 of the 16 center-inside pixels — up to 0.68 pixel-widths from the boundary — are not selected). Affects `RS_AsRaster` and `RS_ZonalStats` selection. Filed as apache/sedona#3111. The diagonal-centroid rows run against every comparator and carry ledger xfails for Sedona Spark.
2. **RS_AsRaster outside-pixel fill** — SedonaDB initializes the output grid with the nodata value; Sedona Spark always burns into zeros and records nodata as band metadata only. Filed as apache/sedona#3112. The rasterio comparator fills with the subject's policy; the nonzero-nodata rows (where outside pixels exist in the output) carry ledger xfails for Sedona Spark.
3. **CRS-less GeoTIFFs** — geotools reads them but its GeoTIFF writer refuses to encode one (SedonaDB's GDAL writer doesn't care). The Spark engine transports results out of the JVM as `RS_AsGeoTiff(RS_SetSRID(r, 3857))` — a transport-only stamp — and the `RS_AsGeoTiff` round-trip module uses EPSG:3857 fixtures.
4. **SRID-0 geometries against a CRS-bearing raster** — Sedona Spark treats a geometry without an SRID as EPSG:4326 and reprojects it into the raster's CRS; SedonaDB does not reproject. Geometry-taking fixtures stay CRS-less so nothing reprojects on either side.
5. **NULL optional arguments** — in Spark a SQL NULL argument nulls the whole expression (`InferredExpression`), so optional arguments must select the overload arity; SedonaDB kernels accept NULL argument columns (e.g. NULL `nodata_value` means "unset"). Consequence: Spark cannot express `useGeometryExtent=false` (or `crop=false`) without an explicit nodata, because those parameters sit after `noDataValue` in the overload ladder.
6. **Point collections and indexing bases** — `RS_Values` takes a MULTIPOINT in SedonaDB but an `ARRAY[Geometry]` in Spark (absorbed by the operation seam). Spark's grid-coordinate `RS_Value(raster, colX, rowY, band)` overload is 0-based while `RS_PixelAs*` is 1-based in both dialects.
7. **Interpolating resample at borders** — geotools bilinear/bicubic produce NaN outside the interpolation support where GDAL clamps to the edge, so only nearest-neighbor resampling is compared.
8. **Zonal statistic definitions** — Spark's `stddev`/`variance` are the sample (ddof=1) statistics, and engines reduce in different orders, so `sum`/`mean`/`stddev`/`variance` compare at rel 1e-9 (`count`/`min`/`max`/`median` match exactly).
9. **Out-of-bounds pixel accessors** — SedonaDB extrapolates out-of-bounds pixel coordinates through the affine transform (pinned against the rasterio comparator, as are Spark's extrapolating `RS_PixelAsCentroid`/`RS_PixelAsPolygon`); Spark's `RS_PixelAsPoint` raises there instead, recorded as a skip-kind ledger entry.


python/sedonadb/tests/functions/conftest.py
```diff
@@ -0,0 +1,43 @@
...
+
+
+@pytest.fixture()
+def subject(con):
+    """The engine under test, on the test session's connection."""
+    return SedonaDB(con)
+
+
+@pytest.fixture(params=COMPARATORS, ids=lambda engine: engine.name())
+def comparator(request):
```
paleolimbot, line 40: If this fixture magic makes your life easier, go for it...I have always liked that our existing engine-based parameterized tests are mostly self-contained (you can tell by reading a single test what it asserts and which engines pass). I get that raster has a lot more going on.

paleolimbot (approved):
I have to admit that I don't quite follow the full flow of the testing here, but I trust you to own this and ensure it's useful for your development of these functions. Thank you!

Our integration tests are now sufficiently complex that they should probably have their own workflow and/or Python folder...when we has 20 functions lumping them in with `pytest sedonadb` was easy, but now we're up to almost 200 and they all run with all the wheel tests. Whether it's `python/sedonadb-testing` or `integration/` (to match `benchmarks/`) it is probably time soon to split these out. (No need to do it here if you'd prefer your next step to be in a different direction).

python/sedonadb/tests/functions/conftest.py
james-willis: im going to merge this PR and consider this comment in the follow PR. 

having this merged will make it much easier to do the remaining UDF work + additional parity harness enhancements here.

PR #1055, opened 2026-07-17
perf(rust/sedona-raster-gdal): RS_Clip follow-ups from review — zero-copy band output, rotated-raster tests, tester CRS engine

james-willis:
clean up comments from dewey in #1000 

also drive by clean up some tests to use the raster test harness methods

### DIFF
--- a/python/sedonadb/tests/functions/test_rs_clip.py
+++ b/python/sedonadb/tests/functions/test_rs_clip.py
@@ -30,6 +30,8 @@ RS_Clip's documented behavior.
 import numpy as np
 import pyarrow as pa
 import pytest
+import shapely
+from rasterio.transform import Affine
 
 from sedonadb.expr import lit
 from sedonadb.raster_testing import (
@@ -40,10 +42,6 @@ from sedonadb.raster_testing import (
     random_raster_data,
     write_geotiff,
 )
-
-pytest.importorskip("rasterio")
-shapely = pytest.importorskip("shapely")
-Affine = pytest.importorskip("rasterio.transform").Affine
 
 
 # GDAL-order geotransform: origin (100, 500), 2-wide by 3-tall north-up pixels.
--- a/python/sedonadb/tests/functions/test_rs_clip.py
+++ b/python/sedonadb/tests/functions/test_rs_clip.py
@@ -42,7 +42,8 @@ from sedonadb.raster_testing import (
 )
 
 pytest.importorskip("rasterio")
-pytest.importorskip("shapely")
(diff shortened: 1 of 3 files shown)


python/sedonadb/tests/functions/test_rs_clip.py
```diff
@@ -42,7 +42,8 @@
 )
 
 pytest.importorskip("rasterio")
-pytest.importorskip("shapely")
+shapely = pytest.importorskip("shapely")
```
paleolimbot: shapely is one of our test dependencies (maybe not explicitly but could be since geopandas depends on it)...a regular import should be fine here

python/sedonadb/tests/functions/test_rs_clip.py
```diff
@@ -42,7 +42,8 @@
 )
 
 pytest.importorskip("rasterio")
```
paleolimbot: Should this importorskip be scoped to a test rather than a module?

rust/sedona-raster-gdal/src/rs_clip.rs
```diff
@@ -1432,4 +1456,105 @@ mod tests {
...
+        // The clipped band bytes move into the output as a zero-copy view
+        // block; pin that so a refactor can't silently reintroduce the copy.
+        // (Views at or under the inline threshold store their bytes inline,
+        // so the band must be bigger than that.)
+        let values: Vec<u8> = (0..32).collect();
+        let array = raster_array([Some(RasterSpec::d2(8, 4).crs(None).band_values(&values))]);
+
+        let kernel = RsClip { arg_count: 3 };
+        let result = kernel
+            .invoke_batch(
```
paleolimbot: I think the tester would work here unless there's something specific to what you're testing (would collapse some of these literal values into more compact representations like `1` and `"the wkt"`)

paleolimbot (approved):
Thank you!

rust/sedona-raster-gdal/src/rs_clip.rs
james-willis: done

python/sedonadb/tests/functions/test_rs_clip.py
james-willis: done

python/sedonadb/tests/functions/test_rs_clip.py
james-willis: rasterio is in the test dep group so im just going to do a top level import

PR #1056, opened 2026-07-17
test(rust/sedona-raster-functions): assert raster results with the raster_spec helpers

james-willis:


rust/sedona-raster-functions/src/rs_example.rs
```diff
@@ -116,23 +114,19 @@ mod tests {
...
-            assert_eq!(band_metadata.storage_type().unwrap(), StorageType::InDb);
-        } else {
+        let ColumnarValue::Scalar(scalar) = result else {
             panic!("Expected scalar struct result");
+        };
+
+        // RS_Example builds a 64x32, 3-band UInt8 raster with origin
+        // (43.08, 79.07), scale 2, skew 1, nodata 127. Every pixel of band N
+        // is N except the top-left corner, which is set to the nodata value.
+        let mut expected = RasterSpec::d2(64, 32).transform([43.08, 2.0, 1.0, 79.07, 1.0, 2.0]);
```
paleolimbot, line 124: Can this use the bbox setter or is this transform value significant?

rust/sedona-raster-functions/src/rs_setsrid.rs
```diff
@@ -395,9 +395,33 @@ mod tests {
...
+        let pixels: Vec<u16> = (0..(width * height) as u16).collect();
+        RasterSpec::d2(width, height)
+            .transform([
+                i as f64 + 1.0,
+                i.max(1) as f64 * 0.1,
+                i as f64 * 0.03,
+                i as f64 + 2.0,
+                i as f64 * 0.04,
+                i.max(1) as f64 * -0.2,
+            ])
```
paleolimbot: Can this now use the bbox setter?

rust/sedona-raster-functions/src/rs_setsrid.rs
```diff
@@ -395,9 +395,33 @@ mod tests {
...
     use sedona_testing::rasters::generate_test_rasters;
     use sedona_testing::testers::ScalarUdfTester;
 
+    /// The non-null raster that [`generate_test_rasters`] produces at index
+    /// `i`, expressed as a declarative spec: the same per-index geotransform
+    /// arithmetic and sequential UInt16 pixels with nodata 0. Used as the
+    /// expected side after an RS_SetSRID/RS_SetCRS, which only swaps the CRS
+    /// and preserves everything else — so callers append `.crs(...)` with the
+    /// CRS they expect.
+    fn generated_raster_spec(i: usize) -> RasterSpec {
```
paleolimbot: We have some other raster example generators (maybe in sedona-testing somewhere)...is this generator unique to set_srid or can/should it be reused?

paleolimbot (approved):
A few things maybe worth considering now, but also this looks great. Thank you!

rust/sedona-raster-functions/src/rs_setsrid.rs
james-willis: arguably. Two reasons we dont:

* bbox always has 0 skew so we wouldn't test that skew values are preserved
* we are testing the geotransform is round tripped so I dont think abstracting it to a bbox is important here. The actually spatial extent of the bbox/transform is not important, just the fact it survives the function call cleanly.


unresolve if you are not satisfied

rust/sedona-raster-functions/src/rs_setsrid.rs
james-willis: the RasterSpec is different than a Raster, its used to assert the interface of a raster is what is expected. 

Migrated this to a util in sedona-testing so we can reuse it with generate_test_rasters.

rust/sedona-raster-functions/src/rs_example.rs
james-willis: I added a bbox method - it didnt actually exist. 

but this is important here, two reasons:

1. we are asserting that RS_Example has stable behavior - ie it has the same geotransform that we expect. So why add indirection to that with a bbox
2. RS_Example has a geotransform with skew so we CANNOT use a bbox to represent it.

PR #1057, opened 2026-07-17
test(python/sedonadb): guard rasterio behind importorskip in raster tests

james-willis:
The wheel-build smoke tests install only the sedonadb wheel plus a small fixed set of test deps — rasterio isn't among them, and it has no cp39 wheel for linux aarch64 at all — so a module-level rasterio import breaks test collection there. Restoring the `importorskip` guards lets the rasterio cross-check tests skip cleanly in that environment while still running in the full test lane, where rasterio and an aligned GDAL are present.

### DIFF
--- a/python/sedonadb/tests/functions/test_rs_clip.py
+++ b/python/sedonadb/tests/functions/test_rs_clip.py
@@ -31,7 +31,6 @@ import numpy as np
 import pyarrow as pa
 import pytest
 import shapely
-from rasterio.transform import Affine
 
 from sedonadb.expr import lit
 from sedonadb.raster_testing import (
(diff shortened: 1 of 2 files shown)


python/sedonadb/tests/functions/test_rs_clip.py
```diff
@@ -43,6 +42,8 @@
     write_geotiff,
 )
 
+pytest.importorskip("rasterio")
```
paleolimbot: Is there a way to avoid this at the module level and move it into the test instead? (At the module level it prevents the tests from being collected which makes it trickier to see which tests were skipped)

paleolimbot (approved):
Thank you!

python/sedonadb/tests/functions/test_rs_clip.py
james-willis: will do. thanks

paleolimbot (approved):

PR #1063, opened 2026-07-20
feat: add experimental RS_MapAlgebra raster function

james-willis:
Adds an experimental `RS_MapAlgebra(raster, expr [, options])` that evaluates a per-pixel expression over a raster's bands.


james-willis:
Closing: we'd prefer raster UDFs over introducing a new per-pixel scripting language (Jiffle/evalexpr) for map algebra.

james-willis:
Maybe we should create a gh issue were we can decide what syntax we want to use and if we want a MapAlgebra function at all. the GIL point is a good one.

PR #1065, opened 2026-07-20
feat: add RS_Tile raster tiling function

james-willis:
Splits a raster into a grid of fixed-size tiles, returning one list item per tile with its grid position. Experimental.

### DIFF
--- a/docs/reference/sql/rs_tile.qmd
+++ b/docs/reference/sql/rs_tile.qmd
@@ -71,10 +71,8 @@ kernels:
 
 ## Description
 
-`RS_Tile` cuts a raster into a grid of `width` x `height` tiles,
-returning a list with one item per tile. Because SedonaDB has no per-row
-generator, the tiles are returned as a list that you `UNNEST` to get one row
-per tile.
+`RS_Tile` cuts a raster into a grid of `width` x `height` tiles, returning a
+list with one item per tile that you `UNNEST` to get one row per tile.
 
 It has two overloads (`padWithNoData` and `noDataVal` are trailing
 optionals):
--- a/docs/reference/sql/rs_tile.qmd
+++ b/docs/reference/sql/rs_tile.qmd
@@ -36,7 +36,7 @@ kernels:
         Optional (default false). When true, pad the last partial row/column of
         tiles to the full tile size with a nodata fill; when false the smaller
         edge tile is emitted as-is.
-    - name: no_data_val
+    - name: no_data_value
       type: double
       description: >
         Optional. The value written to padded pixels (only valid with
@@ -53,16 +53,12 @@ kernels:
         given order.
     - name: width
       type: integer
-      description: Tile width in pixels. Must be >= 1.
     - name: height
       type: integer
-      description: Tile height in pixels. Must be >= 1.
     - name: pad_with_no_data
       type: boolean
-      description: Optional (default false). See the no-band overload.
-    - name: no_data_val
+    - name: no_data_value
       type: double
-      description: Optional. See the no-band overload.
 ---
 
 ::: callout-warning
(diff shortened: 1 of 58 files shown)


docs/reference/sql/rs_tile.qmd
```diff
@@ -0,0 +1,130 @@
...
+      description: Tile width in pixels. Must be >= 1.
+    - name: height
+      type: integer
+      description: Tile height in pixels. Must be >= 1.
+    - name: pad_with_no_data
+      type: boolean
+      description: Optional (default false). See the no-band overload.
+    - name: no_data_val
+      type: double
+      description: Optional. See the no-band overload.
```
paleolimbot: These should get copied over automatically if you don't define the description here

docs/reference/sql/rs_tile.qmd
```diff
@@ -0,0 +1,130 @@
...
+    - name: height
+      type: integer
+      description: Tile height in pixels. Must be >= 1.
+    - name: pad_with_no_data
+      type: boolean
+      description: >
+        Optional (default false). When true, pad the last partial row/column of
+        tiles to the full tile size with a nodata fill; when false the smaller
+        edge tile is emitted as-is.
+    - name: no_data_val
```
paleolimbot: Do we have consistent naming for `no_data_val` vs `nodata_value` vs `nodata_val`?

docs/reference/sql/rs_tile.qmd
```diff
@@ -0,0 +1,130 @@
...
+
+`RS_Tile` cuts a raster into a grid of `width` x `height` tiles, returning a
+list with one item per tile that you `UNNEST` to get one row per tile.
+
+It has two overloads (`padWithNoData` and `noDataVal` are trailing
+optionals):
+
+- `RS_Tile(raster, width, height, padWithNoData = false, noDataVal = null)`
+- `RS_Tile(raster, bandIndices, width, height, padWithNoData = false, noDataVal = null)`
+
```
paleolimbot:
```suggestion
```

Not sure this is more informative than the rendered signatures

docs/reference/sql/rs_tile.qmd
```diff
@@ -0,0 +1,130 @@
...
+as-is, and with `padWithNoData = true` those tiles are padded to the full tile
+size with a nodata fill. When the band(s) have no nodata value, `noDataVal`
+supplies one.
+
+::: callout-note
+`RS_Tile` **copies the raster**: every tile carries its own pixels, so the
+returned list materializes a full copy of the raster spread across the tiles and
+held in memory as one list value. A small tile size produces many tiles (a
+`1 x 1` tile size yields one tile per pixel).
+:::
```
paleolimbot, line 94: This is probably the best reason to implement views. Tiling an in-memory raster is super useful to parallelize a zonal stats calculation from a numpy array.

rust/sedona-schema/src/matchers.rs
```diff
@@ -196,6 +196,13 @@ impl ArgMatcher {
         Arc::new(IsInteger {})
     }
 
+    /// Matches a list of integers (e.g. Spark's `Array[Int]`), covering
+    /// `List`/`LargeList`/`ListView`/`LargeListView`/`FixedSizeList` whose
+    /// element type is an integer.
+    pub fn is_integer_list() -> Arc<dyn TypeMatcher + Send + Sync> {
+        Arc::new(IsIntegerList {})
+    }
```
paleolimbot: This may be more useful to future kernels as `is_list_of(item_matcher: Arc<dyn TypeMatcher + Send + Sync>)`

paleolimbot (approved):
All minor issues from me for the current implementation of this, thank you!

I will point out that this is a footgun and will probably result in running out of memory even for trivailly tile-able inputs: it's a scalar function AND it requires bytes, so you'll need at least 2x the full raster memory in order to call it. This is exactly what the View-based approach is designed to handle...you tile the outDB/indb raster in place, unnest it (metadata only), rebatch and/or repartition according to future loaded byte size, and THEN do whatever you're about to do.

Alternatively you could implement something like Postgres' "set returning function" (duckDB's "table in-out function"), but that still requires loading all of the parent raster's bytes (even if you can stream the tiles and hope that the downstream operator drops them one batch at a time).

docs/reference/sql/rs_tile.qmd
james-willis: ok fixed

docs/reference/sql/rs_tile.qmd
james-willis: not quite.  but this was going to be a 3rd version so we can avoid that.

docs/reference/sql/rs_tile.qmd
james-willis: removed

docs/reference/sql/rs_tile.qmd
james-willis: yeah.

rust/sedona-schema/src/matchers.rs
james-willis: done. good idea. This will help with RS_Values as well.

PR #1066, opened 2026-07-20
feat(rust/sedona-raster-gdal): RS_ZonalStats and RS_ZonalStatsAll

james-willis:
Experimental `RS_ZonalStats` and `RS_ZonalStatsAll` (ported from #704), matching Sedona Spark's positional overload ladder.


docs/reference/sql/rs_zonalstats.qmd
```diff
@@ -0,0 +1,136 @@
...
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+title: RS_ZonalStats
+description: >
+  Computes a single summary statistic of the raster pixels covered by a roi
+  geometry.
```
paleolimbot:
```suggestion
  Computes a single summary statistic of the raster pixels covered by a region of interest
  geometry.
```

docs/reference/sql/rs_zonalstats.qmd
```diff
@@ -0,0 +1,136 @@
...
+---
+
+::: callout-warning
+**Experimental.** This function is experimental; its behavior may change without notice.
+:::
+
+## Description
+
+`RS_ZonalStats` returns one summary statistic of the pixels of a single band
+that a roi geometry covers. A pixel
```
paleolimbot:
```suggestion
`RS_ZonalStats` returns one summary statistic of the pixels of a single band
that a region of interest (ROI) geometry covers. A pixel
```

docs/reference/sql/rs_zonalstats.qmd
```diff
@@ -0,0 +1,136 @@
...
+is included when its center falls inside the roi (or, with `all_touched`, when
+the roi touches it at all). By default the band's nodata pixels are excluded.
+
+The statistic is one of `count`, `sum`, `mean`, `median`, `mode`, `stddev`,
+`variance`, `min`, or `max`. `count` is returned as a whole number; all other
+statistics are floating point. Variance and standard deviation are the sample
+(n-1) values, and `mode` breaks ties toward the larger value.
+
+When the roi overlaps the raster but selects no pixel, `count` is 0 and every
```
paleolimbot:
```suggestion
When the ROI overlaps the raster but selects no pixel, `count` is 0 and every
```

docs/reference/sql/rs_zonalstats.qmd
```diff
@@ -0,0 +1,136 @@
...
+other statistic is NULL. When the roi does not intersect the raster at all, the
```
paleolimbot:
```suggestion
other statistic is NULL. When the ROI does not intersect the raster at all, the
```

docs/reference/sql/rs_zonalstatsall.qmd
```diff
@@ -0,0 +1,116 @@
...
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+title: RS_ZonalStatsAll
+description: >
+  Computes every summary statistic of the raster pixels covered by a roi
```
paleolimbot:
```suggestion
  Computes every summary statistic of the raster pixels covered by a region of interest
```

python/sedonadb/tests/functions/test_rs_zonalstats.py
```diff
@@ -15,195 +15,290 @@
...
 pytest.importorskip("rasterio")
-pytest.importorskip("shapely")
 
-pytestmark = pytest.mark.skipif(
-    not SedonaDB.implements("zonal_stats"),
-    reason="RS_ZonalStats is not implemented in SedonaDB (the parity subject)",
+from sedonadb.raster_testing import (  # noqa: E402
+    random_raster_data,
+    write_geotiff,
 )
```
paleolimbot:
```suggestion
from sedonadb.raster_testing import (
    random_raster_data,
    write_geotiff,
)
```

python/sedonadb/tests/functions/test_rs_zonalstats.py
```diff
@@ -15,195 +15,290 @@
...
+
+
+def numpy_reference(band, wkt, *, all_touched, exclude_no_data):
+    """Reference statistics over the pixels the roi selects, via rasterio+numpy.
+
+    Returns a dict of every statistic, or the sentinel string ``"empty"`` when
+    the selection is empty (the caller maps that to count 0 / NULLs).
+    """
+    import rasterio.features
+    import shapely
```
paleolimbot: The shapely import can be at the top level

rust/sedona-raster-gdal/src/rs_zonal_stats.rs
```diff
@@ -0,0 +1,1609 @@
...
+    count.append_null();
+    for i in 1..=8 {
+        let Some(field) = builder.field_builder::<Float64Builder>(i) else {
+            return sedona_internal_err!("RS_ZonalStats: stat field {i} is not a Float64 builder");
+        };
+        field.append_null();
+    }
+    builder.append(false);
+    Ok(())
+}
```
paleolimbot: Downcasting in a loop can be a performance footgun (although the zonal stats computation itself is sufficiently expensive that it would only make a difference here in a 100% null case). The pattern we use elsewhere is to build the individual arrays and outer nullsbuffer and assemble them at the end (no downcasting per item)

rust/sedona-raster-gdal/src/rs_zonal_stats.rs
```diff
@@ -0,0 +1,1609 @@
...
+        .geometry_from_wkb(geom_wkb)
+        .map_err(|e| exec_datafusion_err!("RS_ZonalStats: failed to parse geometry: {e}"))?;
+    let Some(window) = envelope_window(&geometry, &transform, width, height)? else {
+        scratch.clear();
+        return Ok(Some(compute_statistics(scratch)));
+    };
+
+    // Rasterize the roi into a window-sized 0/1 mask (moves `geometry`, whose
+    // only remaining use is the burn).
+    let mask = rasterize_geometry_mask(gdal, geometry, &transform, &window, params.all_touched)?;
```
paleolimbot: Is this mask heap allocated / would it benefit from scratch space in a loop since it's temporary?

rust/sedona-raster-gdal/src/rs_zonal_stats.rs
```diff
@@ -0,0 +1,1609 @@
...
+        }
+    }
+}
+
+/// The raster's 6-coefficient GDAL geotransform as a fixed array.
+fn raster_transform(raster: &RasterRefImpl<'_>) -> Result<GeoTransform> {
+    let t = raster.transform();
+    <[f64; 6]>::try_from(t)
+        .map_err(|_| exec_datafusion_err!("RS_ZonalStats: expected a 6-element geotransform"))
+}
```
paleolimbot: I feel as though I have seen this function before. Is there a more reusable place to put it?

rust/sedona-raster-gdal/src/rs_zonal_stats.rs
```diff
@@ -0,0 +1,1609 @@
...
+        *counts.entry(v.to_bits()).or_insert(0) += 1;
+    }
+    let best_count = counts.values().copied().max().unwrap_or(0);
+    counts
+        .into_iter()
+        .filter(|&(_, c)| c == best_count)
+        .map(|(bits, _)| f64::from_bits(bits))
+        .fold(f64::NEG_INFINITY, f64::max)
+}
+
```
paleolimbot: A hash map seems like it may not be the most efficient way to compute this mode (BTreeMap may be faster because the comparison function is cheap). If this is the last thing you do, sorting in place first might also be a good way to compute this without a map at all (that may be a good idea anyway to avoid underflow when computing sums/means).

I think you may want to use the typed values rather than the unpacked bytes (particularly for floating point values). It is probably faster to compute and accumulate statistics in their native type anyway.

rust/sedona-raster-gdal/src/rs_zonal_stats.rs
```diff
@@ -0,0 +1,1609 @@
...
+/// An empty slice yields `count = 0` and NULL for the rest (Sedona Spark's
+/// empty-roi shortcut). Variance is the sample (n-1) variance, matching Spark;
+/// for a single pixel it is 0. Median is the linear-interpolated 50th
+/// percentile, which for the median reduces to the middle element (odd n) or
+/// the mean of the two central elements (even n). Mode is the most frequent
+/// value, breaking ties toward the larger value.
+///
+/// `values` is sorted in place (for the median); the caller owns it as reusable
+/// scratch.
+fn compute_statistics(values: &mut [f64]) -> ZonalStatistics {
```
paleolimbot: This should almost certainly be generic over `T` and use and perhaps use (or test against) a statistics crate if there is one. Computing summary statistics is surprisingly complicated in the presence of NAN, infinity, underflow, overflow.

I don't see references to GDAL semantics here...have we checked their handling of these cases? I would be more inclined to trust the level of scrutiny/usage for numpy and/or GDAL than Sedona Spark's here.

rust/sedona-raster-gdal/src/rs_zonal_stats.rs
```diff
@@ -0,0 +1,1609 @@
...
+                    match compute_zonal_stats(gdal, raster, &geom_wkb, &params, &mut scratch)? {
+                        Some(stats) => match stats.get(stat_type) {
+                            Some(value) => builder.append_value(value),
+                            None => builder.append_null(),
+                        },
+                        // The roi does not intersect the raster: NULL when
+                        // lenient (the default), an error otherwise.
+                        None if params.lenient => builder.append_null(),
+                        None => return no_intersection_err(),
+                    }
```
paleolimbot: Does this compute all statistics even if only a simple one (e.g., count) is requested?

paleolimbot (commented):
I think some thought about the corner cases of stats handling (and how this is done in non-Sedona Spark places) should be considered here before this merges, but in general this looks great!

rust/sedona-raster-gdal/src/rs_zonal_stats.rs
james-willis: ZonalStats always returns f64 anyway so I don't think templating over the numeric type buys us anything. 

A quick scan of statistics packages doesn't really show one thatll give us what we want without bringing in some heavy dependencies.

I'm inclined to follow the spark behavior here since we want sedona spark and db to provide the same results where possible.

python/sedonadb/tests/functions/test_rs_zonalstats.py
james-willis: done

rust/sedona-raster-gdal/src/rs_zonal_stats.rs
james-willis: fixed

rust/sedona-raster-gdal/src/rs_zonal_stats.rs
james-willis: yeah, deduped this

rust/sedona-raster-gdal/src/rs_zonal_stats.rs
james-willis: turns out this list is already sorted so we can just look for longest chain of the same value. no map needed.

rust/sedona-raster-gdal/src/rs_zonal_stats.rs
james-willis: fixed.

rust/sedona-raster-gdal/src/rs_zonal_stats.rs
paleolimbot: It does because comparisons and arithmetic are cheaper in integral types than floating point ones (plus no NaN or Inf). You can cast to f64 after you've done the computation.

james-willis: sure but the runtime of the loop body will be dominated by the masking and byte value sorting rather than the arithmetic. I think the branch elimination and faster arithmetic wont substantially move the needle on perf.

In the 64 bit int cases you'll get some benefits of result accuracy due to error accumulation in intermediate results but I think this benefit is not worth the amount of additional code complexity that type templating would bring in.

rust/sedona-raster-gdal/src/rs_zonal_stats.rs
```diff
@@ -0,0 +1,1749 @@
...
+/// The sample (n-1) variance of `values`, matching Sedona Spark; 0 for a single
+/// value. The two-pass form (mean first, then squared deviations) avoids the
+/// catastrophic cancellation of the naive sum-of-squares formula. `values` must
+/// be non-empty and NaN-free.
+fn sample_variance(values: &[f64]) -> f64 {
+    let n = values.len();
+    if n <= 1 {
+        return 0.0;
+    }
+    let mean = values.iter().sum::<f64>() / n as f64;
```
paleolimbot: Can we use a previously computed sum or mean here?

rust/sedona-raster-gdal/src/rs_zonal_stats.rs
```diff
@@ -0,0 +1,1749 @@
...
+    let mean = sum / count as f64;
+    let min = values.iter().copied().fold(f64::INFINITY, f64::min);
+    let max = values.iter().copied().fold(f64::NEG_INFINITY, f64::max);
+    let variance = sample_variance(values);
+    let stddev = variance.sqrt();
+
+    // Median and mode both read the values in sorted order; sort once in place.
+    sort_values(values);
+    let median = median_of_sorted(values);
+    let mode = mode_of_sorted(values);
```
paleolimbot: Since you're computing all of these, you can use your cumulative calculations to avoid a few passes over the data: min is the first of the sorted values; max is the last of the sorted values variance can use the cached sum with one additional pass. If you wire your sort so that `NaN`s are first, you can also avoid the pass over all the values to check for NaN.

You can then eliminate the single statistic path (just return early if there is any work that can be avoided, like returning before the sort or returning before the variance calculation)

paleolimbot (approved):
Thank you!

Float64 as a working type works for me (it's what GDAL does), although it would be good to check at least one adversarial example where the input is all INT64_MAX (if there isn't one already that I missed) to make sure we don't get a panic anywhere. I suggested a minor tweak to avoid a few passes over the data that shouldn't be too hard.

rust/sedona-raster-gdal/src/rs_zonal_stats.rs
james-willis: I think avoiding extra passes is good but im not sure about e.g. having to sort to get just max.

In isolation, max can be accomplished with a fold, `O(n)`. but a sort is of course `O(nlg(n))`, I worry that eliminating the single stats path in this way will be a performance regression on the 7 of 9 stats that don't need a sort.

This is what I am thinking:

For the all stats case we sort eagerly and thatll speed things up

For all single stats, we leave them as is to avoid a regression.

james-willis: done what is described above.

rust/sedona-raster-gdal/src/rs_zonal_stats.rs
james-willis: done

PR #1067, opened 2026-07-21
fix(python/sedonadb): free-threading-safe Tokio runtime teardown; re-enable cp314t macOS wheels

james-willis:
The free-threaded (cp\*t) macOS wheel test hung because the per-connect Tokio runtime's default `Drop` blocking-joins its worker threads while attached to the interpreter, which stalls CPython's stop-the-world under free-threading. The runtime now shuts down in the background on teardown instead, so the free-threaded wheels build and run the full test suite again (the macOS `CIBW_SKIP` is removed).

### DIFF
--- a/c/sedona-extension/src/runtime.rs
+++ b/c/sedona-extension/src/runtime.rs
@@ -1,0 +1,102 @@
+// Licensed to the Apache Software Foundation (ASF) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership.  The ASF licenses this file
+// to you under the Apache License, Version 2.0 (the
+// "License"); you may not use this file except in compliance
+// with the License.  You may obtain a copy of the License at
+//
+//   http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+//! A Tokio runtime owner that shuts down in the background on drop.
+
+use std::mem::ManuallyDrop;
+use std::ops::Deref;
+
+use tokio::runtime::Runtime;
+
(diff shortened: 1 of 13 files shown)


paleolimbot (commented):
We do in general want wheels for free-threaded Python and to test them. I am guessing this has to do with a violation of threading constraints somewhere between PyO3, Cython, and GDAL between RasterIO, pyogrio, and possibly system GDAL.

Maybe our function integration testing shouldn't be running on wheel tests / should be moved to a separate Python package?

james-willis:
@paleolimbot I went back and tried to build a real fix here.

I haven't taken a shot at windows; I don't really have a windows machine

c/sedona-extension/src/runtime.rs
```diff
@@ -0,0 +1,102 @@
...
+        &self.runtime
+    }
+}
+
+impl Drop for RuntimeHandle {
+    fn drop(&mut self) {
+        // SAFETY: `runtime` is initialized in `new` and taken exactly once,
+        // here in `drop`; it is never accessed again afterward.
+        let runtime = unsafe { ManuallyDrop::take(&mut self.runtime) };
+        runtime.shutdown_background();
```
paleolimbot: I believe `shutdown_background()` can leak resources...while I'm sure that ongoing tasks *should* have finished by the time this happens, I'm not sure that is always true.

How about a OnceLock "trash can" ...in this drop implementation? We launch a (non tokio) thread whose job is to drop the runtime (gracefully), and the trash can keeps track of the task. When a new runtimehandle is created, we check the trash and empty it of completed tasks. This approach gives us some visibility if we see consistently non-empty trash cans.

paleolimbot (approved):
Thank you for diagnosing this! I'll leave a note on the issue about static global runtimes and why that happened, but in the meantime this is a definite improvement. One idea about a slightly safer way to go about this if it's easy to do and maybe we can also reduce the number of runtimes we create to minimize these interactions.

c/sedona-extension/src/runtime.rs
james-willis: great idea. doing it now.

paleolimbot (approved):
Than you!

PR #1068, opened 2026-07-21
feat: add RS_ReprojectMatch

james-willis:
Adds `RS_ReprojectMatch(raster, reference[, algorithm])`, reprojecting a raster onto a reference raster's CRS, grid, and envelope (rioxarray `reproject_match` semantics). Positional overloads match Sedona Spark verbatim.

### DIFF
--- a/c/sedona-gdal/src/warp.rs
+++ b/c/sedona-gdal/src/warp.rs
@@ -45,11 +45,15 @@ use crate::spatial_ref::SpatialRef;
 /// left **untouched**: `GDALReprojectImage` with no warp options writes only
 /// covered pixels. Callers that grow the extent must therefore pre-fill `dst`'s
 /// band buffers with the desired background/nodata value before warping.
+///
+/// `warp_memory_limit_bytes` is GDAL's working-memory cache size for the warp
+/// (its `dfWarpMemoryLimit`); pass `0.0` to use GDAL's own default.
 pub fn reproject_image(
     api: &'static GdalApi,
     src: &Dataset,
     dst: &Dataset,
     alg: ResampleAlg,
+    warp_memory_limit_bytes: f64,
 ) -> Result<()> {
     let gra = alg.to_gdal_warp().ok_or_else(|| {
         GdalError::BadArgument(format!(
@@ -66,7 +70,7 @@ pub fn reproject_image(
             dst.c_dataset(),
             null(), // dst WKT: NULL means "use the destination dataset's own SRS"
             gra,
-            0.0,        // warp memory limit (0 = GDAL default)
+            warp_memory_limit_bytes, // warp memory limit (0.0 = GDAL default)
             0.0,        // max error in pixels (0 = exact transformer, no approximation)
             None,       // progress callback
             null_mut(), // progress arg
@@ -194,7 +198,7 @@ mod tests {
                 .unwrap();
             dst.set_projection("EPSG:4326").unwrap();
 
-            gdal.reproject_image(&src, &dst, ResampleAlg::NearestNeighbour)
+            gdal.reproject_image(&src, &dst, ResampleAlg::NearestNeighbour, 0.0)
                 .unwrap();
 
             let out = dst
--- a/c/sedona-gdal/src/warp.rs
+++ b/c/sedona-gdal/src/warp.rs
@@ -1,0 +1,241 @@
+// Licensed to the Apache Software Foundation (ASF) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership.  The ASF licenses this file
+// to you under the Apache License, Version 2.0 (the
+// "License"); you may not use this file except in compliance
+// with the License.  You may obtain a copy of the License at
+//
+//   http://www.apache.org/licenses/LICENSE-2.0
+//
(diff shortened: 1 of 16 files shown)


c/sedona-gdal/src/warp.rs
```diff
@@ -0,0 +1,241 @@
...
+    let rv = unsafe {
+        call_gdal_api!(
+            api,
+            GDALReprojectImage,
+            src.c_dataset(),
+            null(), // src WKT: NULL means "use the source dataset's own SRS"
+            dst.c_dataset(),
+            null(), // dst WKT: NULL means "use the destination dataset's own SRS"
+            gra,
+            0.0,        // warp memory limit (0 = GDAL default)
```
james-willis: im not sure if theres some value we should be passing here.

paleolimbot: I believe there are `GdalOptions` as a member of the ConfigOptions (`sedona.gdal.xxx`)...exposing this value there seems like a good idea for now. Setting it automatically would be tricky (DataFusion does not have great memory accounting, and especially does not have any accounting for the memory used by scalar functions).

python/sedonadb/tests/functions/test_rs_reprojectmatch.py
```diff
@@ -0,0 +1,299 @@
...
+
+from sedonadb.raster_testing import (
+    Rasterio,
+    SedonaDB,
+    decode_raster,
+    random_raster_data,
+    write_geotiff,
+)
+
+DTYPES = ["uint8", "uint16", "int16", "int32", "float32", "float64"]
```
paleolimbot, line 53: Are `int64` and `uint64` supported for warping?

python/sedonadb/tests/functions/test_rs_reprojectmatch.py
```diff
@@ -0,0 +1,299 @@
...
+
+
+@pytest.fixture()
+def sedona(con):
+    return SedonaDB(con)
+
+
+@pytest.fixture()
+def reference():
+    return Rasterio.create_or_skip()
```
paleolimbot: Can we inline these instead of use magic fixtures?

python/sedonadb/tests/functions/test_rs_reprojectmatch.py
```diff
@@ -0,0 +1,299 @@
...
+    path = tmp_path / f"reprojectmatch_{name}_{width}x{height}.tif"
+    write_geotiff(path, data, gdal_transform=transform, crs=crs)
+    return path
+
+
+def _assert_transform_and_nodata(got, expected):
+    assert got.gdal_transform == pytest.approx(
+        expected.gdal_transform, rel=1e-12, abs=1e-12
+    )
+    assert got.nodata == expected.nodata
```
paleolimbot: Are these utilities that already exist or should be added to raster_testing so they can be reused?

rust/sedona-raster-gdal/src/rs_reproject_match.rs
```diff
@@ -0,0 +1,538 @@
...
+    ) -> Result<ColumnarValue> {
+        let num_iterations = RasterExecutor::num_iterations_over(args);
+
+        // Algorithm string at index 2 (when arg_count == 3); otherwise the
+        // Spark default `NearestNeighbor`. Expand to an array so a per-row
+        // column and a scalar are handled identically.
+        let algorithm_array = if self.arg_count >= 3 {
+            args[2]
+                .clone()
+                .cast_to(&DataType::Utf8, None)?
```
paleolimbot: Using a utf8view here should be slightly better for the scalar-to-array case

paleolimbot (approved):
Thank you! All just details from me.

I think of this as "warp" (maybe I've just always done this with GDAL)...ReprojectMatch is a funny name to me but it sounds like there is precedent.

c/sedona-gdal/src/warp.rs
james-willis: ok done. added `sedona.gdal.warp_memory_limit_mb`

python/sedonadb/tests/functions/test_rs_reprojectmatch.py
james-willis: done

python/sedonadb/tests/functions/test_rs_reprojectmatch.py
james-willis: `approx_geotransform` already existed.

The other 3 were moved into the raster_testing file

rust/sedona-raster-gdal/src/rs_reproject_match.rs
james-willis: done

python/sedonadb/tests/functions/test_rs_reprojectmatch.py
james-willis: ok got them supported but 2 caveats:

1. as usual we reject no data values above 2^53
2. we dont allow arithmetic sample methods, only nearest and mode. I checked that mode will behave correctly according to GDAL docs

james-willis: never mind, I removed support for int64 and uint64 because they dont return correct results for the type. the only exception is Mode in GDAL 3.13

PR #1069, opened 2026-07-21
test(python/sedonadb): arity-aware raster parity harness for RS_* overloads

james-willis:


james-willis:
Superseded by #1070, which establishes the SedonaDialectEngine parity pattern (centralized mechanism, per-function policy) with inline deviations and exact-vs-approx assertions. The verified rasterio references from this branch are reused there.

PR #1070, opened 2026-07-21
test(python/sedonadb): DialectEngine parity harness + inline-deviation RS_ZonalStats spike

james-willis:
Design spike: reworks the RS_ZonalStats / RS_ZonalStatsAll parity tests and their harness support to (1) declare known engine divergences inline on the parametrized case rather than in a separate deviations ledger, (2) build each RS_* call string once on a shared `DialectEngine` base so SedonaDB and Sedona Spark cannot silently emit different SQL, and (3) assert that shared surface directly with a cross-dialect contract test.

### DIFF
--- a/python/sedonadb/python/sedonadb/raster_testing.py
+++ b/python/sedonadb/python/sedonadb/raster_testing.py
@@ -361,10 +361,10 @@ class RasterEngine:
         self._not_implemented("from_binary")
 
 
-class DialectEngine(RasterEngine):
+class SedonaDialectEngine(RasterEngine):
     """A `RasterEngine` that runs the RS_* function under test as SQL.
 
-    SedonaDB and Sedona Spark are both dialect engines: an operation is one
+    SedonaDB and Sedona Spark both implement the Sedona SQL dialect: an operation is one
     RS_* call, and the compatibility contract the parity tests exist to protect
     is that the identical call runs on both ("Spark SQL runs unchanged in
     SedonaDB"). To make that contract structural rather than a coincidence, the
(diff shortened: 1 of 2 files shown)


python/sedonadb/python/sedonadb/raster_testing.py
```diff
@@ -330,7 +361,155 @@ def from_binary(self, data: bytes) -> DecodedRaster:
...
+    is that the identical call runs on both ("Spark SQL runs unchanged in
+    SedonaDB"). To make that contract structural rather than a coincidence, the
+    argument string is built in exactly one place — the `*_expr` generators
+    below — and both dialects execute whatever those return. A dialect subclass
+    supplies only its per-engine I/O: how a GeoTIFF path becomes a `rast` column
+    and how a scalar / struct result is decoded (`_run_scalar` / `_run_struct`).
+    The two dialects therefore cannot emit different SQL for the same operation.
+
+    Only RS_ZonalStats / RS_ZonalStatsAll are routed through this shared surface;
+    the other operations keep their existing per-dialect implementations.
```
james-willis, line 378: we will add more code paths here after we agree on and merge this PR.

paleolimbot (commented):
You've lost me on the test harness complexity here. These are SedonaDB tests and they should assert that running some operation produces an expected result, with the other engines as a sanity check for the expected value...I think this PR with the SQL parity has passed the limits of scope for `pytest sedonadb`.

james-willis:
My perspective here is that sedonadb and sedona spark should have the same syntax and behavior and this framework would help us keep the the two matching.

Do you diagree with us trying to be strict about maintaining syntax and beahvior between the two projects or is your objection more about where this kind of framework lives.

ETA: or some third thing like this is very over engineered or unneccessary.

james-willis:
Closing, going a different direction based on Dewey's comments

PR #1073, opened 2026-07-23
feat(rust/sedona-spatial-join-raster): raster–vector spatial join (RS_Intersects/Contains/Within)

james-willis:
Adds an index-optimized spatial join for `RS_Intersects`/`RS_Contains`/`RS_Within` between a raster and a geometry. A new `sedona-spatial-join-raster` crate plugs a raster-aware operand evaluator into the existing `SpatialJoinExec` — the raster is indexed and refined by the convex hull of its four world corners, so the R-tree and WKB refiner are unchanged.

CRS handling matches SedonaDB's `RS_` predicate kernels: same-CRS compares directly (exact); a one-sided/missing CRS is an error (not silently assumed to be WGS84); cross-CRS reprojects the raster footprint into the geometry operand's CRS, **densifying each edge (~10 points) before reprojection** so the footprint follows the projected curve rather than a straight chord. Reprojection routes through the `SedonaOptions`-injected `CrsEngine` (both the predicate kernels and the join use `with_crs_engine`, mirroring `RS_Clip`). A footprint with a non-finite reprojected point yields a null footprint (no match) instead of garbage.

Out of scope (tracked in #1076): antimeridian-crossing footprints, and geodesic / geography-tolerance matching.

### DIFF
--- a/rust/sedona-raster-functions/src/footprint.rs
+++ b/rust/sedona-raster-functions/src/footprint.rs
@@ -43,16 +43,28 @@ pub fn raster_footprint_corners(raster: &dyn RasterRef) -> [(f64, f64); 4] {
     ]
 }
 
-/// Write WKB for the convex-hull polygon of the raster footprint.
+/// Write WKB for the convex-hull polygon through four footprint `corners`.
 ///
-/// The ring is the four [`raster_footprint_corners`] closed back to the
-/// upper-left corner. This can be used to build Binary arrays, as the arrow-rs
-/// `BinaryBuilder` implements [`std::io::Write`].
-pub fn write_convexhull_wkb(raster: &dyn RasterRef, out: &mut impl std::io::Write) -> Result<()> {
-    let [ul, ur, lr, ll] = raster_footprint_corners(raster);
+/// `corners` are in ring order (upper-left, upper-right, lower-right,
+/// lower-left, as produced by [`raster_footprint_corners`]); the ring is closed
+/// back to the first corner. Shared by the native footprint (corners in the
+/// raster's own CRS) and the reprojected footprint (corners transformed into
+/// another CRS), so both paths emit byte-identical polygon WKB. This can be used
+/// to build Binary arrays, as the arrow-rs `BinaryBuilder` implements
+/// [`std::io::Write`].
+pub fn write_footprint_wkb(corners: [(f64, f64); 4], out: &mut impl std::io::Write) -> Result<()> {
+    let [ul, ur, lr, ll] = corners;
 
     write_wkb_polygon(out, [ul, ur, lr, ll, ul].into_iter())
         .map_err(|e| DataFusionError::External(e.into()))?;
 
     Ok(())
+}
+
+/// Write WKB for the convex-hull polygon of the raster footprint.
+///
+/// The ring is the four [`raster_footprint_corners`] closed back to the
+/// upper-left corner.
+pub fn write_convexhull_wkb(raster: &dyn RasterRef, out: &mut impl std::io::Write) -> Result<()> {
+    write_footprint_wkb(raster_footprint_corners(raster), out)
 }
--- a/rust/sedona-raster-functions/src/footprint.rs
+++ b/rust/sedona-raster-functions/src/footprint.rs
@@ -23,9 +23,21 @@
 //! corner is computed individually rather than assumed axis-aligned.
 
 use datafusion_common::{DataFusionError, Result};
+use sedona_geometry::interpolate::densify_segment;
 use sedona_geometry::wkb_factory::write_wkb_polygon;
 use sedona_raster::affine_transformation::to_world_coordinate;
 use sedona_raster::traits::RasterRef;
+
+/// Number of interior points interpolated along each footprint edge when a
+/// footprint is densified before reprojection.
+///
+/// Reprojecting a footprint bends its straight edges into curves, so a footprint
+/// described only by its four corners chords across each curve and under-/over-
+/// covers the true extent. Densifying each edge with a handful of interior points
+/// (~10 per edge is a common, cheap choice) and reprojecting every point makes
+/// the reprojected footprint follow the curve. A footprint left in its own CRS
+/// has exact straight edges and needs no densification.
+pub const FOOTPRINT_POINTS_PER_EDGE: usize = 10;
 
 /// The four corners of a raster's footprint in world coordinates.
 ///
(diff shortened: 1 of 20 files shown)


rust/sedona-spatial-join-raster/src/join_provider.rs
```diff
@@ -0,0 +1,474 @@
...
+        let num_rows = rasters.len();
+        let mut builder = BinaryBuilder::with_capacity(num_rows, num_rows * 96);
+        let mut rects = Vec::with_capacity(num_rows);
+
+        // A missing CRS on either side is assumed to be WGS84 (matching the `RS_*`
+        // kernel and Sedona Spark); resolve it once rather than per row.
+        let lnglat_crs = lnglat();
+        let wgs84 = lnglat_crs.as_deref().expect("lnglat() is always Some");
+
+        with_global_proj_engine(|engine| {
```
paleolimbot: I believe we have access to the SedonaOptions when we create the array factory. If we do we should use that CrsEngine.

rust/sedona-spatial-join-raster/src/join_provider.rs
```diff
@@ -0,0 +1,474 @@
...
+        engine: &dyn CrsEngine,
+        builder: &mut BinaryBuilder,
+    ) -> Result<Bounds2D> {
+        let raster_crs = resolve_crs(raster.crs())?;
+        let corners = raster_footprint_corners(raster);
+
+        // Assume WGS84 for whichever side lacks a CRS, then compare in the target
+        // frame. Both-absent and same-CRS collapse to a direct comparison.
+        let raster_crs = raster_crs.as_deref().unwrap_or(wgs84);
+        let target_crs = self.target_crs.as_deref().unwrap_or(wgs84);
```
paleolimbot: If this is the current behaviour for any version of an RS predicate, we should change it. I think this is probably not as relevant in Sedona Spark since the GeoParquet reader properly attached CRSes (when there would have been a lot of crs-less values floating around that actually were WGS84).

rust/sedona-raster-functions/src/footprint.rs
```diff
@@ -0,0 +1,58 @@
...
+        to_world_coordinate(raster, 0, height),
+    ]
+}
+
+/// Write WKB for the convex-hull polygon of the raster footprint.
+///
+/// The ring is the four [`raster_footprint_corners`] closed back to the
+/// upper-left corner. This can be used to build Binary arrays, as the arrow-rs
+/// `BinaryBuilder` implements [`std::io::Write`].
+pub fn write_convexhull_wkb(raster: &dyn RasterRef, out: &mut impl std::io::Write) -> Result<()> {
```
paleolimbot, line 80: This should densify the edges at least a little bit (around 10 points per edge is common) to ensure that when projected into another geometry's space, the calculated intersection is more likely to be accurate (many transformations turn straight lines into curved ones). The logic to do this is already in the repo for `st_segmentize`...you can move any utilities to sedona-geometry that are needed here.

A more robustly parameterized method would transform the convex hull and the geometry into a geography (with a tolerance in meters), then compare it. If somebody needs that level of accuracy they can write a join that does that fairly easily.

rust/sedona-raster-functions/src/rs_spatial_predicates.rs
```diff
@@ -318,24 +316,54 @@ fn evaluate_predicate_with_crs<Op: tg::BinaryPredicate>(
...
+            compare_in_crs::<Op>(wkb_a, crs_a, wkb_b, crs_b, from_a_to_b, engine)
+        }
+        // Neither side has a CRS: both are in the same (unknown) frame, so compare
+        // directly without transformation.
+        (None, None) => evaluate_predicate::<Op>(wkb_a, wkb_b),
+        // Exactly one side has a CRS: assume the missing side is WGS84 (matching
+        // Sedona Spark) and compare in the resulting CRS pair. Only these arms
+        // materialize the lnglat CRS, so the common paths avoid the allocation.
+        (Some(crs_a), None) => {
+            let lnglat_crs = lnglat().expect("lnglat() should always return Some");
```
paleolimbot: The assumption of a missing side having a WGS84 CRS is a footgun and the current behaviour is better. We can fix this in Sedona Spark now that the GeoParquet reader attaches CRSes.

rust/sedona-spatial-join-raster/src/physical_planner.rs
```diff
@@ -0,0 +1,322 @@
...
+        let (physical_left, physical_right) = if args.join_options.repartition_probe_side {
+            repartition_probe_side(
+                args.physical_left.clone(),
+                args.physical_right.clone(),
+                args.spatial_predicate,
+                should_swap,
+            )?
+        } else {
+            (args.physical_left.clone(), args.physical_right.clone())
+        };
```
paleolimbot, line 93: Do we want to do anything here given some knowledge of which side is the raster side and which side is the vector side? The raster side will almost always be smaller and have fewer vertices. I would guess that it would always be better to index the geometry side but I don't know this.

rust/sedona-spatial-join-raster/src/join_provider.rs
```diff
@@ -0,0 +1,474 @@
...
+            .transform_coord(corner)
+            .map_err(|e| exec_datafusion_err!("Transform error: {e}"))?;
+    }
+
+    // Closed ring: the four reprojected corners plus the first repeated.
+    write_wkb_polygon(
+        builder,
+        [hull[0], hull[1], hull[2], hull[3], hull[0]].into_iter(),
+    )
+    .map_err(|e| exec_datafusion_err!("Failed to write footprint WKB: {e}"))?;
```
paleolimbot: Is there another utility in this PR that does this? There may also be a utility that does this already in wkb_factory.

rust/sedona-spatial-join-raster/src/join_provider.rs
```diff
@@ -0,0 +1,474 @@
...
+    builder.append_value([]);
+    Ok(bounds_from_coords(&hull))
+}
+
+/// Bounding rectangle of a set of coordinates. [`Bounds2D::new`] enlarges the
+/// f32 bounds outward so the rectangle conservatively contains every input
+/// coordinate.
+fn bounds_from_coords(coords: &[(f64, f64)]) -> Bounds2D {
```
paleolimbot: Unless this is a benchmarked performance optimization, using the existing utilities (Interval, WkbBounder2D, other lower level bounding utilities) would be better for this

rust/sedona-spatial-join-raster/src/join_provider.rs
```diff
@@ -0,0 +1,474 @@
...
+    use sedona_raster::traits::{BandMetadata, RasterMetadata};
+    use sedona_schema::crs::{deserialize_crs, lnglat};
+    use sedona_schema::datatypes::RASTER;
+    use sedona_schema::raster::{BandDataType, StorageType};
+    use sedona_testing::rasters::generate_test_rasters;
+
+    /// A 1x1 raster over world coords (0,0)-(1,1), with `crs` (or none).
+    fn build_unit_raster(crs: Option<&str>) -> arrow_array::StructArray {
+        let mut builder = RasterBuilder::new(1);
+        let metadata = RasterMetadata {
```
paleolimbot, line 384: Does the RasterSpec help this at all?

paleolimbot (commented):
All just details from me...thank you!

I get why, but it's a bummer we can't just optimize the join condition into `ST_Intersects(RS_ConvexHull(a), b) && RS_Intersects(a, b)` and let the internals deal with the rest.

rust/sedona-spatial-join-raster/src/join_provider.rs
james-willis: The only reason I didn't do this is because I didnt want it to have divergent behavior with the UDFs and other joins (im pretty sure the other joins havent adopted yet?)

paleolimbot: I don't think there are any other joins that require a CrsEngine?

rust/sedona-spatial-join-raster/src/physical_planner.rs
james-willis: There's a PR stacked on this one that has some idea here. #1074 

Would be great to get your input.

rust/sedona-spatial-join-raster/src/join_provider.rs
james-willis: oh that makes sense. Anyway I'll stack converting the UDFs + join on this PR, will provide before I merge this one.

rust/sedona-spatial-join-raster/src/join_provider.rs
james-willis: https://github.com/apache/sedona/issues/3153
https://github.com/apache/sedona-db/issues/1075

IMO we should keep the behavior in sync. I filed matching bugs in each project

rust/sedona-raster-functions/src/footprint.rs
james-willis: https://github.com/apache/sedona/issues/3154
https://github.com/apache/sedona-db/issues/1076

paleolimbot: You're here now and turning this off is easier than turning it on...this should get resolved here.

rust/sedona-spatial-join-raster/src/physical_planner.rs
paleolimbot: Perhaps put a comment here with a forward link/some ideas about considering the raster vs non raster side in case you aren't able to circle back soon?

rust/sedona-spatial-join-raster/src/join_provider.rs
paleolimbot: We should keep behaviour correct while we're here and have eyes to review. The current behaviour is correct.

james-willis: fair enough. done

rust/sedona-raster-functions/src/rs_spatial_predicates.rs
james-willis: ok changed back

rust/sedona-spatial-join-raster/src/physical_planner.rs
james-willis: done. I think you previously asked me not to inline comment links to PRs and issues, so I just added a gh issue to the PR: #1078

rust/sedona-raster-functions/src/footprint.rs
james-willis: ok done. Can revisit this and sedona spark's behavior as needed.

rust/sedona-spatial-join-raster/src/join_provider.rs
james-willis: I moved the RS_ functions and the join over but not `ST_Transform` and `sd_order` because `with_crs_engine` lives in `sedona-raster-functions`. 

I might be missing something about this new pattern that you know and I dont. Filed #1082 to for you to provide guidance on. Will pick that up after you bless the idea.

rust/sedona-spatial-join-raster/src/join_provider.rs
james-willis: yes, done

rust/sedona-spatial-join-raster/src/join_provider.rs
james-willis: done

rust/sedona-spatial-join-raster/src/join_provider.rs
james-willis: Yes adopted.

rust/sedona-spatial-join-raster/src/join_provider.rs
```diff
@@ -45,22 +51,33 @@ use sedona_spatial_join::{
...
     }
 }
 
+/// Pin the join's refiner to `tg`, the engine the `RS_*` predicate kernel uses,
+/// so the accelerated join and the kernel resolve boundary/touching cases
+/// identically regardless of the session's `spatial_join.spatial_library`.
+fn pin_refiner_to_tg(mut options: SpatialJoinOptions) -> SpatialJoinOptions {
+    options.spatial_library = SpatialLibrary::Tg;
+    options
+}
```
james-willis, line 79: @paleolimbot lmk if this was a bad idea.

paleolimbot: Great idea 👍

PR #1079, opened 2026-07-24
refactor(rust): route raster CRS reprojection through the injected CrsEngine

james-willis:
Stacked on #1073. Routes all raster CRS reprojection through the SedonaOptions-injected CrsEngine (UDFs via `invoke_batch_from_args`/`with_crs_engine`, mirroring RS_Clip; the raster join captures the engine from ConfigOptions at plan time) instead of the process-global PROJ engine, so a session-injected engine is honored. Behavior-neutral by default (the injected engine defaults to LazyProjEngine).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6454,7 +6454,6 @@ dependencies = [
  "rstest",
  "sedona-common",
  "sedona-expr",
- "sedona-functions",
  "sedona-geometry",
  "sedona-proj",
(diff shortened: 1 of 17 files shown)


james-willis:
Folded into #1073 — that branch is fast-forwarded onto this commit, so the CrsEngine migration now lives in #1073. Closing to keep it a single PR.

PR #1094, opened 2026-07-27
feat(c/sedona-extension): Push Expr filters into the TableProvider across the FFI boundary 

paleolimbot:
This PR allows (logical) filter expressions to be pushed from a consumer into a TableProvider. This is currently limited to a logical expression via `TableProvider::scan()` but we can use the same mechanism to implement `try_pushdown_filters()` for the FFI exec plan (which is the mechanism, for example, that will get used to prune using statistics between spatial join sides).

```python
import pandas as pd
import sedona.db

sd_producer = sedona.db.connect()
sd_consumer = sedona.db.connect()

path = "submodules/geoarrow-data/ns-water/files/ns-water_water-point.parquet"

producer_sql = 'SELECT "OBJECTID", "FEAT_CODE" FROM water_point'
consumer_sql = '''
SELECT "FEAT_CODE", COUNT(*) as cnt FROM df_producer
WHERE "OBJECTID" BETWEEN 10 AND 100
GROUP BY "FEAT_CODE"
HAVING COUNT(*) > 1
ORDER BY cnt DESC
'''

sd_producer.read_parquet(path).to_view("water_point")
df_producer = sd_producer.sql(producer_sql)

# Run the query on the producer (no FFI)
df_producer.to_view("df_producer")
result_no_ffi = sd_producer.sql(consumer_sql).to_pandas()

# Run the query on the consumer (separated via FFI)
sd_consumer.create_data_frame(df_producer).to_view("df_producer")
df_over_ffi = sd_consumer.sql(consumer_sql)
result_over_ffi =df_over_ffi.to_pandas()

# Same results!
pd.testing.assert_frame_equal(result_no_ffi, result_over_ffi)

# ...except our ImportedSedonaCExec wraps the inital plan. Projection and filter pushed down!
explain_pd = df_over_ffi.explain().to_pandas()
print(explain_pd[explain_pd["plan_type"] == "physical_plan"]["plan"].iloc[0])
# SortPreservingMergeExec: [cnt@1 DESC]
#   SortExec: expr=[cnt@1 DESC], preserve_partitioning=[true]
#     ProjectionExec: expr=[FEAT_CODE@0 as FEAT_CODE, count(Int64(1))@1 as cnt]
#       FilterExec: count(Int64(1))@1 > 1
#         AggregateExec: mode=FinalPartitioned, gby=[FEAT_CODE@0 as FEAT_CODE], aggr=[count(Int64(1))]
#           RepartitionExec: partitioning=Hash([FEAT_CODE@0], 12), input_partitions=12
#             AggregateExec: mode=Partial, gby=[FEAT_CODE@0 as FEAT_CODE], aggr=[count(Int64(1))]
#               CooperativeExec
#                 ImportedSedonaCExec: FilterExec: OBJECTID@0 >= 10 AND OBJECTID@0 <= 100, projection=[FEAT_CODE@1]
```

### DIFF
--- a/c/sedona-extension/src/expr.rs
+++ b/c/sedona-extension/src/expr.rs
@@ -20,7 +20,6 @@ use std::{
     fmt::{Debug, Display},
     os::raw::{c_char, c_void},
     ptr::null_mut,
-    sync::Arc,
 };
 
 use arrow_array::{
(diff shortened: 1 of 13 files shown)


c/sedona-extension/src/expr.rs
```diff
@@ -0,0 +1,451 @@
...
+    ///
+    /// The returned [SedonaCExprView] is valid only as long as this ExportedExprView
+    /// remains alive. The caller must ensure they do not use the FFI view after
+    /// this object is dropped.
+    pub fn as_ffi_view(&self) -> SedonaCExprView {
+        SedonaCExprView {
+            get_property_schema: Some(c_expr_get_property_schema),
+            get_property: Some(c_expr_get_property),
+            reserved: null_mut(),
+            private_data: self as *const ExportedExprView as *const c_void,
```
zhangfengcdt, line 81: Maybe point to Expr itself (self.expr) instead the wrapper object?

zhangfengcdt (approved):
minor comment, otherwise, lgtm

c/sedona-extension/src/expr.rs
paleolimbot: That's a great point and we could definitely do that...the wrapper object is purely organizational. I'm going to leave this for now to keep the pattern consistent with the other FFI exporters/importers but we can revisit if this becomes more confusing than helpful!

PR #1099, opened 2026-07-30
Add RS_Min/RS_Max and RS_Min_Agg/RS_Max_Agg raster functions

james-willis:
## Summary
**Experimental** — signatures/semantics may still change.

- `RS_Min_Agg`/`RS_Max_Agg`: aggregate UDFs computing the pixel-wise min/max of **one band** across a group of rasters. Optional `band_index` argument (1-based, default 1, same convention as `RS_BandPixelType`/`RS_BandNoDataValue` — an out-of-range index is skipped rather than erroring).
- `RS_Min`/`RS_Max`: the pairwise (non-aggregate) equivalent — `LEAST`/`GREATEST` for one band of two rasters. Same optional `band_index` argument.
- All four share one combine implementation (`raster_min_max.rs`), so the aggregate and scalar surfaces can't drift on what "min/max of two rasters" means.
- Nodata-aware: a pixel equal to the band's nodata sentinel is treated as absent — the other side wins outright, and a pixel that's nodata on both sides stays nodata.
- Requires the selected band to match dimension names, shape, data type, and nodata value across the two (or group of) rasters being combined — mismatches are a clean execution error, not a panic or silently wrong result.
- For the aggregate: state is the output raster type itself (min/max combine is associative/idempotent, so the running combined raster is valid partial state), letting `update_batch`/`merge_batch` share one implementation.

**v1 scope**, called out in the code:
- Plain `Accumulator` only for the aggregates — no `GroupsAccumulator` yet (DataFusion falls back to one accumulator per group, correct but not the fastest path for many small groups).
- Docs added at `docs/reference/sql/{rs_min,rs_max,rs_min_agg,rs_max_agg}.qmd`, each cross-linking its pairwise/aggregate counterpart.

This is the first of a small family of raster statistics aggregates (mean/sum/std/var/median tracked separately); min/max are the easiest to make exactly and cheaply distributable.

## Test plan
- [x] `cargo test -p sedona-raster-functions` — 269 passed, no regressions
- [x] `cargo build -p sedona-raster-functions`
- [x] `cargo fmt -p sedona-raster-functions`
- [x] `cargo clippy -p sedona-raster-functions --all-targets` — clean


james-willis:
I think I might be missing something important here. I think this method should return a raster collapse along an array of provided dimensions:

```sql
RS_Min_Agg(raster, ["x", "y",], 1) // assume some 3rd and 4th dim z and t
```

This will return a 2d non-spatial raster (tensor?) of all averages at each z/t combo.

james-willis:
going a different direction on this.

PR #1100, opened 2026-07-31
ci(python-wheels): upload nightly wheels best-effort when a platform fails

james-willis:
Nightly wheels are gated all-or-nothing on all four platform build jobs, so one failing platform blocks the entire nightly — no set has published since 2026-07-10 (recently the Windows job has been red). This lets `upload_nightly` run best-effort and publish whatever `release-wheels-*` artifacts exist; a build/test job that fails never reaches its upload-artifact step, so only wheels that built and passed their tests are present. It does not fix the underlying Windows failure — that still needs root-causing separately.

I'm not sure if we should fail the workflow after uploading.

### DIFF
--- a/.github/workflows/python-wheels.yml
+++ b/.github/workflows/python-wheels.yml
@@ -204,6 +204,12 @@ jobs:
 
   upload_nightly:
     needs: ["wheels-linux", "macOS-arm64", "macOS-amd64", "windows-x86_64"]
+    # Publish whatever wheels built, even when some platform jobs fail. A job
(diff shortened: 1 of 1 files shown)


paleolimbot (approved):
Thanks, this is better than our current setup (and I haven't had a chance to fix the tests yet). It would also be great to separate the testing from the building/uploading but cibuildwheel isn't really set up to make that easy.

james-willis:
I think one reason we havent done that is because the wheel build is where we actually test multiple platforms? not sure though, have never given this piece of the codebase a lot of thought

PR #1101, opened 2026-07-31
refactor(rust/sedona-raster): retire the pre-N-D metadata shim

james-willis:
Migrates every raster caller off the pre-N-D metadata shim (`RasterMetadata` / `BandMetadata` / `MetadataRef` / `StorageType` + the `start_raster`/`start_band` builder shims) onto the direct N-D trait accessors (`width()?`/`height()?`/`transform()`, `data_type()`/`nodata()`/`is_indb()`/`outdb_uri()`, `start_raster_2d`/`start_band_2d`/`start_band_nd`), then deletes the shim. Behavior-preserving. Two commits — migrate, then delete — so the deletion reviews as a self-contained "the shim is dead" diff.

### DIFF
--- a/rust/sedona-raster-functions/src/executor.rs
+++ b/rust/sedona-raster-functions/src/executor.rs
@@ -741,7 +741,7 @@ mod tests {
                 match raster_opt {
                     None => builder.append_null(),
                     Some(raster) => {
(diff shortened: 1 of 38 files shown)


paleolimbot (approved):
Thank you!

Issue #1082, opened 2026-07-24
Route remaining geometry CRS functions (ST_Transform, sd_order) through the injected CrsEngine


james-willis:
`RS_Intersects`/`RS_Contains`/`RS_Within` and the raster–vector spatial join now route CRS reprojection through the `SedonaOptions`-injected `CrsEngine` (via `crs_utils::with_crs_engine` / an `invoke_batch_from_args` override, mirroring `RS_Clip`) — landed in #1073.

The remaining production uses of the process-global `with_global_proj_engine` are the geometry functions in `c/sedona-proj`:

- `ST_Transform` (`c/sedona-proj/src/st_transform.rs`)
- `sd_order` (`c/sedona-proj/src/sd_order_lnglat.rs`)

These can't reuse `crs_utils::with_crs_engine` because that helper lives in `sedona-raster-functions`, which is above `sedona-proj` in the crate graph. Routing them through the injected engine needs a **lower-crate helper**: move the `SedonaOptions` → `runtime.crs_engine()` extraction (with the config-less fallback to `LazyProjEngine`) down into `sedona-common` or `sedona-proj` so `sedona-proj` functions can call it.

Notes:

- `ST_Transform` already has `config_options` plumbed through `invoke_batch_from_args` (currently ignored) — it just needs to route through the lower-crate helper.
- `sd_order` has no `invoke_batch_from_args` override yet — one needs adding.
- Behavior-neutral in the common case (the injected engine defaults to `LazyProjEngine`, which delegates to the global one).

Done when `git grep with_global_proj_engine` shows only the definition in `sedona-proj`, `LazyProjEngine`'s internal delegation, and test-only uses.

paleolimbot:
Closed by #1102.

PR #1102, opened 2026-07-31
Route ST_Transform and sd_order through the injected CRS engine

james-willis:
Migrates the two geometry CRS functions (`ST_Transform`, `sd_order`) off the process-global PROJ engine and onto the `SedonaOptions`-injected `CrsEngine`, mirroring the raster functions. The `SedonaOptions` -> runtime engine extraction now lives in a `with_crs_engine` helper in `sedona-proj` (with a config-less fallback to `LazyProjEngine`) so it sits below `sedona-proj` in the crate graph. Part of #1082.


paleolimbot (approved):
Looks like the benchmark needs config options for the tester as well, but when CI is green this looks great.

No need to do it here if you'd rather not, but I believe that these two functions can also move to sedona-functions now (and we can drop the datafusion dependencies of sedona-proj.

james-willis:
I moved those functions. 

@paleolimbot asked for a rereview because those changes are substantial.

c/sedona-proj/Cargo.toml
```diff
@@ -43,20 +40,10 @@ proj-sys = [ "dep:proj-sys" ]
...
-arrow-array = { workspace = true }
 datafusion-common = { workspace = true }
-datafusion-expr = { workspace = true }
 geo-traits = { workspace = true }
 proj-sys = { version = "0.26.0", optional = true }
 sedona-common = { workspace = true }
-sedona-expr = { workspace = true }
-sedona-functions = { workspace = true }
 sedona-geometry = { workspace = true }
 sedona-schema = { workspace = true }
```
paleolimbot, line 48: 🎉 !

I think this one can be removed too

```suggestion
```

paleolimbot (approved):
Thank you!

c/sedona-proj/Cargo.toml
james-willis: handled in #1104

PR #1103, opened 2026-08-01
Move ST_Transform and sd_order out of sedona-proj into sedona-functions

james-willis:
Stacks on #1102 (until it merges, the diff below includes #1102's commits).

Moves the geometry CRS functions `ST_Transform` and `sd_order` (`OrderLngLat`) from `sedona-proj` into `sedona-functions`, and relocates the `SedonaOptions` -> runtime CRS-engine resolution (`with_crs_engine`) into `sedona-common`, where the config-less fallback becomes the erroring `DefaultCrsEngine` instead of the global PROJ engine. Behavior-neutral: the session registers `LazyProjEngine` as the default engine, so real invocations are unchanged; reprojecting tests and the benchmark inject `LazyProjEngine`. The concrete PROJ engine stays in `sedona-proj`.

`sedona-proj` no longer hosts DataFusion UDFs and drops its `datafusion-expr`, `sedona-functions`, `sedona-expr`, and `arrow-*` dependencies. It still depends on `datafusion-common` because `configure_global_proj_engine`/`with_global_proj_engine` return DataFusion `Result`/`DataFusionError` (consumed by the Python/R bindings and ~40 reprojecting test closures across the raster crates); converting those off DataFusion is left as a separable follow-up.


james-willis:
merging this change into #1102

PR #1104, opened 2026-08-02
Drop datafusion-common from sedona-proj

james-willis:
Makes `sedona-proj` a pure PROJ-engine crate with no DataFusion dependency. The global-engine helpers now return the crate-local `SedonaProjError`; callers convert at the boundary.

Stacks on #1102 — the diff includes #1102's commits until it merges.


paleolimbot (approved):
This works too if you just want to merge this one on green CI...exactly the state I was hoping to get sedona-proj to. Thank you!

PR #1105, opened 2026-08-02
refactor(rust/sedona-raster): validate outdb #band=N in split_outdb_band_fragment

james-willis:
Deduplicates the two `#band=N` outdb-URI parsers into one validated parser in `sedona-raster`. `split_outdb_band_fragment` now returns `Result` and rejects a `#band=` fragment whose value is not a positive `u32` (previously it silently defaulted to band 1). The GDAL loader's out-db reads route through it, and the duplicate `parse_outdb_source` (`source_uri.rs`) is removed with its validation tests moved to `traits.rs`.


paleolimbot (approved):
Thank you!

PR #1108, opened 2026-08-03
fix: SRID extraction returns NULL for a valid authority-less CRS

james-willis:
A valid CRS with no top-level authority code (e.g. a custom polar-stereographic WKT from CF/rioxarray) is fully usable but has no EPSG/SRID. `RS_SRID`/`ST_SRID` on such a CRS now return NULL instead of erroring; a genuinely malformed CRS string still errors, and an authority-coded CRS still returns its integer SRID.

### DIFF
--- a/python/sedonadb/tests/functions/test_raster_functions.py
+++ b/python/sedonadb/tests/functions/test_raster_functions.py
@@ -83,12 +83,13 @@ def test_rs_srid_from_wkt():
     )
 
 
-def test_rs_srid_from_authorityless_wkt_errors(con):
-    """A WKT with no authority code anywhere has no SRID to extract."""
-    with pytest.raises(Exception, match="SRID"):
-        con.sql(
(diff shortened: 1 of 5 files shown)


paleolimbot (commented):
No notes on the implementation, but do we want this? We could also update the error to  say "Use `ST_Crs()` instead of `ST_SRID()` to support CRSes without authority/code or with a non-EPSG authority"?

james-willis:
I think this is the right thing. 2 reasons:

1. I believe null for a SRID is the correct value where the CRS does not have a SRID.
2. Users cannot catch a per-row error so we should be extremely cautious about letting a single row fail a whole query.

james-willis:
I think the round trip example is contrived and I'm not worried about it.

Maybe ST_SetSRID or ST_Intersects should be the one that rejects null CRS/SRID instead.

I'm imagining users who want to filter on the presence of a CRS with a SRID, so they can get rid of SRID-less rows (CRSes with no SRID)

PR #1109, opened 2026-08-03
refactor(sedona-raster,sedona-raster-gdal): Remove the Bands 1-based band-access wrapper

james-willis:
Removes the `Bands<'a>` band-access view and the `RasterRef::bands()` trait method, migrating every caller to the trait's 0-based `band()`/`num_bands()`.

The 1-based band convention that RS_Value/RS_Values expose to users moves into a `sampling::resolve_band` helper; GDAL loops that also index GDAL's own 1-based rasterband keep their 1-based loop variable and subtract 1 only for the raster accessor.


paleolimbot (approved):
Thank you!

PR #1112, opened 2026-08-03
refactor(rust/sedona-raster): lift GeoTransform/GeoTransformEx into sedona-raster

james-willis:
Lifts the six-coefficient `GeoTransform` type and `GeoTransformEx` trait (apply/invert) out of `c/sedona-gdal` into a new `sedona-raster::geo_transform` module so GDAL, Zarr, and the RS_ functions share one geo-transform home.

`AffineMatrix` delegates `transform`/`inv_transform` to the single `GeoTransformEx` implementation, so there is exactly one apply/invert. The free `rotation()` helper now builds on `AffineMatrix::rotation`, and `rs_georeference` reads `AffineMatrix` named fields instead of raw geotransform indices.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5943,6 +5943,7 @@ version = "0.4.0"
 dependencies = [
  "gdal-sys",
  "libloading 0.9.0",
+ "sedona-raster",
  "sedona-testing",
  "thiserror 2.0.17",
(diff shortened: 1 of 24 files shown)


paleolimbot (approved):
Thank you!

PR #1113, opened 2026-08-03
feat(rust/sedona-raster): persist and read non-identity band views

james-willis:
makes the raster type/trait/impl capable of creating, storing, and reading non-identity band views (slices, broadcasts, axis permutations, reverse/negative-step).

- **Store** (`builder.rs`): `start_band` persists a non-identity view (supplied via `StartBandArgs.view`) into the existing `band_view_*` columns; identity stays the canonical null sentinel. `with_view` composes a view delta onto an existing band.
- **Read** (`array.rs`): `band()` decodes the stored view and composes it onto the source's C-order byte strides (visible shape, byte strides, offset), with overflow + data-buffer bounds checks. `as_contiguous()` still borrows a packed layout and errors on a strided one.
- **Trait** (`traits.rs`): `copy_into` now persists a composed non-identity view instead of rejecting it.

No schema change was needed (the `band_view_*` columns already exist). No RS_ UDF creates a view here — `RS_EnsureContiguous`, the optimizer rule that injects it, and migrating RS_ functions to produce lazy views are follow-ups.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6211,6 +6211,7 @@ dependencies = [
  "approx",
  "arrow-array",
  "arrow-buffer",
+ "arrow-data",
  "arrow-ipc",
  "arrow-schema",
  "async-trait",
@@ -6450,7 +6451,7 @@ dependencies = [
 
 [[package]]
 name = "sedona-spatial-join-raster"
-version = "0.4.0"
+version = "0.5.0"
 dependencies = [
  "arrow-array",
  "arrow-schema",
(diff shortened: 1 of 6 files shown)


paleolimbot (approved):
Sorry for taking so long to review this...thank you!

I would feel better about the details with some Python tests of a numpy array that is sliced being able to roundtrip without copying memory / compose slices and logically resolve into the correct output; however, I think you're two PRs away from that (one for handling strided PyBuffer inputs/outputs and one for slicing using views) and so I think that is a battle for a different time.

james-willis:
Thanks @paleolimbot ! Very understandable - I had an onslaught of PRs for you last week.

are you saying that at a high level we need to make sure that our np roundtrip layer properly handles views?

PR #1116, opened 2026-08-04
fix(rust/sedona-geoparquet): Handle intermediary columns in projection expressions in GeoParquet reader

### LINKED ISSUE #1115: Panic "index out of bounds" in schema.rs during Arrow export when 2+ columns accompany an ST_* accessor on scanned geometry
## Describe the bug

Any query that combines **2 or more non-geometry columns** with an **`ST_*` accessor function** (e.g. `ST_IsEmpty`, `ST_IsValid`, `ST_XMin`, `ST_NPoints`) applied to a geometry column **sourced from a file scan** (e.g. `read_parquet`) panics on every Arrow-materializing export path — `to_arrow_reader()`, `to_arrow_table()`, `to_pandas()`, `to_parquet()`. `count()` and `show()` are unaffected, since they don't go through Arrow export.

```
thread '<unnamed>' panicked at .../arrow-schema-57.3.0/src/schema.rs:375:10:
index out of bounds: the len is 2 but the index is 2
```

The index value scales with the number of extra columns (reproduced with 1, 2, and 3 — 1 never panics, 2+ always does), which points at a fixed-size metadata array (likely CRS/extension-type metadata for the geometry column) being indexed by column position without accounting for the other columns in the schema.

This affects both the sync (`sedonadb`) and `sedona.db` Python entry points and reproduces on a fresh `sedona.db.connect()` with no options set. It is not related to column ordering, nullability, or the specific `ST_*` function — every accessor function tried reproduces it identically.

## To Reproduce

```python
import pyarrow as pa
import pyarrow.parquet as pq
import shapely
import sedona.db

path = "/tmp/repro.parquet"
geom = pa.array(
    [shapely.to_wkb(shapely.geometry.Point(1, 2)), shapely.to_wkb(shapely.geometry.Point(3, 4))],
    type=pa.binary(),
)
pq.write_table(pa.table({"geometry": geom}), path)

sd = sedona.db.connect()
sd.read_parquet(path).to_view("raw", overwrite=True)

# 2 extra non-geometry columns ('c1', 'c2') + one ST_* accessor on the scanned geometry
sd.sql("""
    CREATE OR REPLACE VIEW g1 AS
    SELECT 'a' AS c1, 'b' AS c2, ST_SetSRID(ST_GeomFromWKB(geometry), 4326) AS geometry
    FROM raw
""")
sd.sql("CREATE OR REPLACE VIEW t AS SELECT *, ST_IsEmpty(geometry) AS _e FROM g1")

for batch in sd.sql("SELECT * FROM t").to_arrow_reader():
    pass  # panics here
```

Reducing to a single extra column (`c1` only, dropping `c2`) does **not** panic. Reducing to zero `ST_*` calls does not panic. Using a literal geometry (`ST_Point(1, 2)`) instead of one read from a file scan does not panic.

## Expected behavior

The query executes and exports without panicking, as it did in `sedonadb` 0.3.0 (confirmed no reproduction there).

## Environment

- sedonadb: 0.4.0 (conda-forge, `py314h9a0e2ed_0` build, linux-64 and osx-arm64 both reproduce)
- Python: 3.14
- OS: reproduced on both macOS (osx-arm64) and Linux (linux-64) conda-forge builds

paleolimbot:
This PR fixes a workaround we have for metadata getting dropped in the Parquet implementation's try_pushdown_projection. Because we advertize the file schema with different field metadata than the inner opener, we need to wrap the expressions (or UDF calls fail because there is no type metadata).

I had thought the projection expressions we see had already been normalized to the file schema; however, it appears that these schemas are stacked (we might get more than one projection pushed down, and while the underlying implementation handles the accumulated schemas, we can't look to see what that is. There is a chance this approach will accidentally add metadata to a field whose metadata had been stripped and given the same name.

This PR applies our metadata wrapping once when constructing the file source, then lets the inner implementation handle the accumulating of projections.

Closes #1115.


python/sedonadb-geopandas/tests/__init__.py
paleolimbot, line 1: This is a drive by fix...pytest objects to this file and won't collect this test in VSCode if it exists

rust/sedona-geoparquet/src/format.rs
```diff
@@ -391,7 +391,34 @@ impl FileFormat for GeoParquetFormat {
...
+            initial_projection,
+            source.table_schema().table_schema(),
+        )
+        .map(|projection_with_types| source.try_pushdown_projection(&projection_with_types));
+
+        // These are both failable but we can't fail here, so fall back to the original
+        // source.
+        match projection_with_types {
+            Ok(Ok(Some(modified))) => modified,
+            _ => Arc::new(source),
```
james-willis, line 424: Is it ok to silently not wrap? When do we expect this case to hit?

rust/sedona-geoparquet/src/format.rs
```diff
@@ -391,7 +391,34 @@ impl FileFormat for GeoParquetFormat {
...
+        // This isn't a bug in DataFusion because we're the ones that advertised the table
+        // schema as having metadata'd expressions in the first place.
+        //
+        // We fix this by wrapping Column expressions with MetadataPreservingColumn,
+        // which stores the correct field from the file schema and returns it from
+        // return_field() regardless of the input schema.
+        let initial_projection = source.projection().cloned().unwrap_or_else(|| {
+            let indices: Vec<usize> =
+                (0..source.table_schema().table_schema().fields().len()).collect();
+            ProjectionExprs::from_indices(&indices, source.table_schema().table_schema())
```
james-willis, line 411: Isn't this just a copy of what is already in `ParquetSource.new`? When would we hit this case?

https://github.com/apache/datafusion/blob/branch-52/datafusion/datasource-parquet/src/source.rs#L308

rust/sedona-geoparquet/src/format.rs
```diff
@@ -391,7 +391,34 @@ impl FileFormat for GeoParquetFormat {
...
+
+        // DataFusion 52 has an issue where field metadata (like ARROW:extension:name)
+        // is stripped when evaluating embedded projections in ParquetOpener. This is
+        // because the batch schema comes from the parquet reader (which doesn't have
+        // extension metadata), and Column::return_field() looks up fields from that schema.
+        // This isn't a bug in DataFusion because we're the ones that advertised the table
+        // schema as having metadata'd expressions in the first place.
+        //
+        // We fix this by wrapping Column expressions with MetadataPreservingColumn,
+        // which stores the correct field from the file schema and returns it from
```
james-willis, line 406: field from table schema right? thats kind of the idea here that the table might have type metadata that isnt correctly reflected in the file?

rust/sedona-geoparquet/src/format.rs
```diff
@@ -691,8 +702,15 @@ fn wrap_expr_columns(
     expr.transform_down(|node| {
         if let Some(column) = node.as_any().downcast_ref::<Column>() {
             let index = column.index();
+
+            if index >= file_schema.fields().len() {
+                return sedona_internal_err!(
+                    "Unexpected projection expression in GeoParquet source: index {index} out of bounds"
+                );
```
james-willis, line 738: This would get swallowed by the match at  line 418.

Really wouldn't expect this to even happen since schema and expr come from the same place

rust/sedona-geoparquet/src/format.rs
```diff
@@ -1081,4 +1099,96 @@ mod test {
...
+                example
+            ))
+            .await
+            .unwrap();
+
+        // This should not panic - the issue was "index out of bounds" when
+        // projection pushdown tried to wrap columns with indices beyond the
+        // file schema's bounds
+        let batches = df.collect().await.unwrap();
+        assert!(!batches.is_empty());
```
james-willis, line 1220: should we assert the metadata survives as well?

james-willis (approved):

rust/sedona-geoparquet/src/format.rs
paleolimbot: That's a great point...it's not expected to fail but I added a deferred error mechanism to make sure we know if it does

rust/sedona-geoparquet/src/format.rs
paleolimbot: We shouldn't hit this case, but if that gets removed on upgrade and set to None at least it's still correct.

rust/sedona-geoparquet/src/format.rs
paleolimbot: There's a difference between the table schema that we have access to and the table schema the Parquet opener uses to evaluate projections / filters. This is not my favourite workaround...the idea is to "just wrap" the Parquet DataFusion implementation but it's accumulated quite a bit of workarounds and it's not as clean as I would like it to be.

All of this was rewritten on the Parquet end for each of the 53, 54, and 55 releases so we will have to redo these workarounds a few times in our future 😮‍💨

rust/sedona-geoparquet/src/format.rs
paleolimbot: I added the deferred error mechanism to make sure this gets caught

rust/sedona-geoparquet/src/format.rs
paleolimbot: I added this to the test!

Issue #1072, opened 2026-07-23
Python wheel builds are failing with handler timeout


paleolimbot:
After https://github.com/apache/sedona-db/pull/1067 and https://github.com/apache/sedona-db/pull/1051 we have better hygene with our context management and Python GDAL usage; however, we still have failing tests:

Windows fails after the Python UDF tests, and Linux still intermittently fails after the GDAL/OGR read via Python.

https://github.com/apache/sedona-db/actions/runs/29952109746/job/89051129963#step:5:63075

https://github.com/apache/sedona-db/actions/runs/29952109746/job/89032091791#step:4:79348

It's very possible that #1071 can fix this (fewer or static lifetimes), plus moving away from pyogrio as the source of OGR reads.

paleolimbot:
For Windows it's coming back to me that I once had a hang in the UDF tests where the read path went through pyogrio. In this case there was too much Python GIL interaction (I think) or some other lock and there was a deadlock where the query wouldn't finish.

PR #1117, opened 2026-08-04
ci: trigger the zarr wheel build on sedona-raster-zarr / sedonadb-zarr changes

james-willis:
The python-wheels PR path filter listed `wheels-zarr.yml` but not the zarr crate/package, so a change to `sedona-raster-zarr`/`sedonadb-zarr` skipped the zarr wheel build (and its Windows/MSVC + blosc link) on the PR that changed it. Add those paths so blosc-on-Windows stays validated pre-merge.

### DIFF
--- a/.github/workflows/python-wheels.yml
+++ b/.github/workflows/python-wheels.yml
@@ -25,6 +25,12 @@ on:
       - '.github/workflows/python-wheels.yml'
       - '.github/workflows/wheels-expr.yml'
       - '.github/workflows/wheels-zarr.yml'
+      # The zarr extension wheel — including the Windows/MSVC build that links
+      # blosc — only builds when this workflow runs. Trigger it on changes to
+      # the zarr crate/package too, so a change there can't skip wheel
+      # validation (blosc on Windows in particular) until the next main push.
+      - 'rust/sedona-raster-zarr/**'
+      - 'python/sedonadb-zarr/**'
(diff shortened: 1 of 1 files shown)


james-willis:
Wheel failures are expected preexisting issues on main. See:

linux: #1120
windows: #1072

note: 1120 is debugging a consistent failure on linux. 1072 also refers to a sporadic failure on linux

paleolimbot (approved):
It's ok with me if you want this, but this is a pretty time consuming/annoying workflow to have to wait for, are you sure you want to run it for all changes in these directories? (There are also other changes in other crate dependencies that could make these fail as well and you probably don't want it to run for all changes to all PRs).

If you're worried a particular change is going to break windows you can add a comment to this file to force the wheels to build when you want them.

james-willis:
I see your point. Theres probably a better way to run cross-plat test suites on PRs than this.

james-willis:
Closing per review feedback. Triggering the full zarr wheel build on every PR that touches these directories is slow to wait for and can go red for unrelated crate-dependency changes, so it adds false-red noise across unrelated PRs for limited value. The original Windows/blosc motivation is already handled by shipping zarr as the separate MSVC extension wheel; and if a specific change looks risky on a platform, forcing the wheel build on-demand (e.g. a marker comment in the workflow file) is the better mechanism than a blanket path-filter trigger.

PR #1119, opened 2026-08-04
feat(python/sedonadb): Raster.from_numpy(bbox=, registration=)

james-willis:
Wires `Raster.from_numpy` to accept `bbox=[xmin,ymin,xmax,ymax]` + `registration="pixel"|"node"`, building the geotransform via the Rust `geotransform_from_bbox_and_spatial_shape` (exposed through pyo3) — mutually exclusive with an explicit `transform`. Python test cross-checks the pixel case against `rasterio.transform.from_bounds`.


paleolimbot (approved):
Thank you!

PR #1121, opened 2026-08-04
refactor(rust/sedona-raster): consolidate start_band into a named-args constructor

james-willis:
Pure refactor, behavior-identical to main: `start_band_nd` + `start_band_with_view` collapse into one `start_band(StartBandArgs)` where `view: Option<&[ViewEntry]>` (`None` = identity). Non-identity views are still rejected (unchanged); `start_band_2d` keeps its 2-arg convenience signature. First of two PRs — the follow-up flips the single `start_band`'s non-identity arm from reject to persist.


paleolimbot (approved):
Thank you!

PR #1122, opened 2026-08-04
feat: RS_TileExplode streaming generator (SQL + df.tile_explode)

james-willis:
RS_TileExplode as a streaming generating operator. `SELECT RS_TileExplode(rast, w, h)` and `df.tile_explode(...)` both emit flattened top-level `(x, y, tile)` rows (Sedona Spark parity), via a marker UDF lifted into a custom plan node at plan-build time so `df.schema()` is honest. Experimental.


james-willis:
going to simplify this so it doesnt have a custom physical node. No need when we can just rewrite to `UNNEST(RS_Tile(...))`

paleolimbot (commented):
I took a close read through this and the implementation does what it says it will do on the tin!

Please correct me if I am misreading this, but it seems like Python `.tile_explode("raster", ...)` syntactic sugar to `.select(raster.rst.tile(...).unnest())`, without improving the streaming generation that is currently forcing two entire copies of every `raster` to be in memory at once.

I'm concerned that this is a lot of raster specific code being injected into a bunch of core non-raster pieces without actually improving the functionality (the streaming execution of `rs_tile()` vs. the whole-batch it-returns-a-list). If we are going going to go through with this we at least need a way to separate these things a bit (e.g., reduce the scope of the Python change to literally just calling `.select(self[col].fns.rs_tile(...).unnest()).select(x=, y=, raster=,)`; move the method into a `DataFrame`-level `rst` accessor whose functions live in a separate file; possibly generalize the pattern of a table in/out function).

It seems like we may need some other things (e.g., views, or an optimizer rule optimizing `unnest(RS_Tile))` into a streaming operation, table in/out functions) to land before we invoke the complexity here. Are we sure Sedona Spark needs the specialized operator (i.e., can we deprecate it there in favour of RS_Tile + Explode + Select and simplify both projects?)

james-willis:
>  Are we sure Sedona Spark needs the specialized operator (i.e., can we deprecate it there in favour of RS_Tile + Explode + Select and simplify both projects?)

No, I do not know the history or motivation for this. maybe @jiayuasu knows

> syntactic sugar ... without improving the streaming generation 

Yes. I had the AI prototype a version that has a physical node that does that kind of streaming. It was 4.5k LoC. kind of scared me off. For now im only targeting parity.

> currently forcing two entire copies of every raster to be in memory at once.

Yes. But to me the best way to solve this might be to just use lazy views that share the same backing buffer.

jiayuasu:
@james-willis @paleolimbot Glad that you pinged me here.

I am in favor of dropping the support of this function. Currently, this is a unique function in WherobotsDB but it is being phased out by the Wherobots native `raster` loader which does tiling on load and partitioning by default in a streaming fashion. What @paleolimbot suggests makes sense to me:

> .select(raster.rst.tile(...).unnest())

AND

> some other things (e.g., views, or an optimizer rule optimizing unnest(RS_Tile)) into a streaming operation

james-willis:
closing based on the above comment. This PR might be helpful in the future if we want that streaming behavior; the initial commit implements that

jiayuasu:
@james-willis Does our raster geotiff loader do tiling on read? If not, we need to keep this function

james-willis:
I don't think we have a geotiff loader in sedonaDB in the way you are thinking. 

The way to load geotiffs in sedonaDB today is to have a dataframe of URLs and call RS_FromPath on the url column.

jiayuasu:
.....OK. then there is a no way we can tile the geotiffs if we drop this feature. I suggest we reopen the PR and bring it back. I was thinking that in WDB we no longer need this function but apparently SedonaDB still needs it.

james-willis:
We already have the RS_Tile and Unnest functions. I don't see how there is no way to do this.

jiayuasu:
The streaming nature is important otherwise this will crash the engine easily

james-willis:
I feel that lazy cropping will obviate the need for a streaming operator in this case. maybe also a cache

jiayuasu:
Do we have lazy cropping in zonal stats or RS_Clip?

james-willis:
lazy cropping support is WIP, the PR for writing views into rasters is #1113 

Once this PR is merged we can migrates all cropping operations (e.g. Clip, Slice) to be lazy.

PR #1123, opened 2026-08-04
[DEBUG – do not merge] ST_Transform bisect: build #1104 (73828e3d5) under gdb

james-willis:
Throwaway CI-only run for DB-407: builds the arm64/cp39 wheel at #1104 and runs test_st_transform under gdb, to split whether the ST_Transform segfault came from #1104 or #1101. Not for review or merge; will be closed once the run completes. (Diff is nonsensical — it's an old commit vs main.)

### DIFF
--- a/.github/workflows/python-wheels.yml
+++ b/.github/workflows/python-wheels.yml
@@ -163,7 +163,7 @@ jobs:
     strategy:
       matrix:
         config:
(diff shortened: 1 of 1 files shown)


james-willis:
Bisect done: #1104-alone (73828e3d5) passes test_st_transform, so the ST_Transform regression is a latent CRS-engine UB (from #1102) exposed later, not #1104. Recorded in DB-407. Closing this throwaway.

PR #1124, opened 2026-08-05
Auto-register common OGR formats so file URLs work as tables

james-willis:
Auto-registers the common single-file OGR formats (`fgb`, `gpkg`, `shp`, `geojson`) at context creation so `SELECT * FROM 'file:///path/to/data.fgb'` works zero-config, the way GeoParquet already does. No-op when pyogrio is not installed.

### DIFF
--- a/python/sedonadb/python/sedonadb/context.py
+++ b/python/sedonadb/python/sedonadb/context.py
@@ -15,7 +15,6 @@
 # specific language governing permissions and limitations
 # under the License.
 
-import importlib.util
 import os
 import sys
 from functools import cached_property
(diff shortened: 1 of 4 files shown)


python/sedonadb/python/sedonadb/context.py
```diff
@@ -121,8 +128,31 @@ def _impl(self):
...
+        except (ImportError, ValueError):
+            pyogrio_available = False
+
+        if not pyogrio_available:
+            return
+
+        from sedonadb.datasource import PyogrioFormatSpec
+
+        for extension in _DEFAULT_PYOGRIO_EXTENSIONS:
+            self.register(PyogrioFormatSpec(extension))
```
paleolimbot, line 145: I *think* that pyogrio isn't imported until you actually open a file (so you should be able to just go ahead and register).

paleolimbot (approved):
Thank you!

python/sedonadb/python/sedonadb/context.py
james-willis: theres no risk that pyorgio isnt installed in the python runtime?

I have a hard time remembering what is and is not expected to be installed.

paleolimbot: It may not exist, but creating a PyogrioFormatSpec won't import it. You could improve the error in `PyogrioFormatSpec.open_reader()` when the import fails ("pyogrio must be installed to read {format}").

james-willis: ok got it. I removed the guard and improved the error message instead.

PR #1125, opened 2026-08-05
feat(sedona-datasource): resolve directory-shaped formats as URL tables

james-willis:
`SELECT * FROM '<url>'` for a directory-shaped external format (e.g. zarr) previously failed: DataFusion's URL-table resolver lists the directory, keys on the first inner object's extension, and parses it as the wrong format. This adds an experimental `UrlTableFactory` that routes a URL whose extension resolves to a registered `list_single_object` format through the single-object table path, and delegates every other URL to DataFusion's default resolver unchanged.


python/sedonadb-zarr/tests/test_zarr.py
```diff
@@ -65,6 +65,66 @@ def test_format_spec_via_read(zarr_group):
...
+        assert table.schema.field("raster").type == expected.schema.field("raster").type
+        got = [table["raster"][i].as_py() for i in range(table.num_rows)]
+        exp = [expected["raster"][i].as_py() for i in range(expected.num_rows)]
+        assert all(isinstance(r, Raster) for r in got)
+        # Same georeferencing and OutDb chunk anchors as the explicit read
+        # proves the URL resolved to this group, not a listing over its chunks.
+        assert [r.transform for r in got] == [r.transform for r in exp]
+        assert [r.bands[0].outdb_uri for r in got] == [
+            r.bands[0].outdb_uri for r in exp
+        ]
```
paleolimbot: Probably just checking the number of rows and the fact that it executed is sufficient here

rust/sedona-datasource/src/url_table.rs
```diff
@@ -0,0 +1,218 @@
...
+            .get_session()
+            .upgrade()
+            .and_then(|session| {
+                session
+                    .read()
+                    .as_any()
+                    .downcast_ref::<SessionState>()
+                    .cloned()
+            })
+    }
```
paleolimbot: It seems like this is handled here, but note that anywhere the `Session` is downcast into a `SessionState` could fail (usually the culprit is the FFI_Session or a testing session).

rust/sedona-datasource/src/url_table.rs
```diff
@@ -0,0 +1,218 @@
...
+/// Drop-in replacement for DataFusion's
+/// [`SessionContext::enable_url_table`] that additionally routes
+/// directory-shaped external formats through the single-object table
+/// path. Mirrors `enable_url_table`'s wiring: it wraps the current catalog
+/// list in a [`DynamicFileCatalog`] backed by a [`SedonaUrlTableFactory`],
+/// then points the factory's session store at the (unchanged) session
+/// state so it can resolve registered file formats at query time.
+///
+/// **Experimental.**
+pub fn enable_sedona_url_table(ctx: SessionContext) -> SessionContext {
```
paleolimbot: This file should probably go in the `sedona` crate, where we have other catalog/url stuff. I haven't been perfect about this but I would like to remove the `datafusion` dependency here (a lot of upstream work has been done to make datafusion-datasource *not* depend on `datafusion` to reduce the size of the binaries and compilation time for extension formats).

docs/reference/python-zarr.md
```diff
@@ -31,6 +31,18 @@ sd.register(sedonadb_zarr.ZarrExtension())
 sd.read("file:///path/to/foo.zarr").show()
 ```
 
+Once the extension is registered, you can also query a Zarr group directly in
+SQL by its URL, without an explicit format:
+
+```python
+sd.sql("SELECT * FROM 'file:///path/to/foo.zarr'").show()
+```
```
paleolimbot, line 39: Should this highlight `'path/to/foo.zarr'`, since that's what you'd normally type?

paleolimbot (approved):
Thank you!

docs/reference/python-zarr.md
james-willis: done

python/sedonadb-zarr/tests/test_zarr.py
james-willis: done

rust/sedona-datasource/src/url_table.rs
james-willis: yes, get_session will return None in this case.

rust/sedona-datasource/src/url_table.rs
james-willis: done

PR #1126, opened 2026-08-05
docs: add Coordinate Reference Systems concept page

james-willis:
Adds a concept page describing SedonaDB's CRS model (the three CRS forms, verbatim preservation, and the authority-code-first equality rule), wires it into the docs nav, and links to it from the CRS-related SQL function reference pages.

Thius PR adds a whole new section for Concepts. Not sure if this is where we want this to live

### DIFF
--- a/docs/concepts/crs.md
+++ b/docs/concepts/crs.md
@@ -131,13 +131,21 @@ same: normalize both sides to the same definition with `ST_Transform` or
 ### Raster joins are an exception
 
 A spatial join involving a raster — raster-to-raster, or a raster against a
-geometry — tests the raster's **envelope** (its footprint), not its pixels. So
-SedonaDB reprojects that envelope to a common CRS on the fly and the join works
-across mismatched CRSes rather than erroring. Reprojecting a footprint is a
-cheap, exact operation on a handful of corner coordinates, whereas reprojecting
-the pixel grid itself would require resampling — so the join aligns CRSes
-without touching the pixels. The error-on-mismatch rule above therefore applies
-to the vector/geometry path; raster joins handle the CRS mismatch for you.
+geometry — tests the raster's **footprint** (the polygon through its corners),
+not its pixels. So SedonaDB reprojects that footprint to a common CRS and the
+join works across mismatched CRSes rather than erroring — reprojecting a
+footprint is far cheaper and safer than resampling the pixel grid.
+
+It is a close approximation rather than an exact boundary, though: reprojection
+is nonlinear, so the footprint's straight edges bow into curves in the target
+CRS, and reprojecting only the four corners would under-cover the true extent.
+SedonaDB handles this by **densifying** each edge — adding interior points in the
+raster's own CRS, where the edges are still straight — then reprojecting all of
+them, so the reprojected footprint follows the curve. (A same-CRS raster join
+skips this and uses the exact four-corner footprint.)
+
+So the error-on-mismatch rule above applies to the vector/geometry path; raster
+joins reproject the densified footprint for you.
 
 ## SRID vs CRS
 
--- a/docs/concepts/crs.md
+++ b/docs/concepts/crs.md
@@ -137,8 +137,9 @@ join works across mismatched CRSes rather than erroring — reprojecting a
 footprint is far cheaper and safer than resampling the pixel grid.
 
 It is a close approximation rather than an exact boundary, though: reprojection
-is nonlinear, so the footprint's straight edges bow into curves in the target
-CRS, and reprojecting only the four corners would under-cover the true extent.
+is sometimes non-linear, so the footprint's straight edges may bow into curves
+in the target CRS, and reprojecting only the four corners can under-cover the
+true extent.
 SedonaDB handles this by **densifying** each edge — adding interior points in the
 raster's own CRS, where the edges are still straight — then reprojecting all of
 them, so the reprojected footprint follows the curve. (A same-CRS raster join
--- a/docs/concepts/crs.md
+++ b/docs/concepts/crs.md
@@ -103,11 +103,11 @@ PROJ using an approximate, confidence-based similarity threshold.
 
 ### What this means for you
 
-Operations that require their inputs to share a CRS — spatial joins,
+**Vector** operations that require their inputs to share a CRS — spatial joins,
 `ST_Intersects`, distance predicates, and so on — **raise an error on a
 mismatch** rather than silently reprojecting one side to match the other.
-Silent reprojection is exactly the kind of hidden, easy-to-miss behavior that
-produces wrong answers, so SedonaDB refuses to guess:
+Silently reprojecting geometry coordinates is exactly the kind of hidden,
+easy-to-miss behavior that produces wrong answers, so SedonaDB refuses to guess:
 
 ```
 Error during planning: Mismatched CRS arguments: epsg:3857 vs epsg:4326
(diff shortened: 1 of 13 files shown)


docs/concepts/crs.md
```diff
@@ -0,0 +1,186 @@
+<!---
```
jiayuasu, line 1: Can you also explain the row-level CRS and column level CRS?

docs/concepts/crs.md
```diff
@@ -0,0 +1,186 @@
...
+- the **round-trippable / PROJ-consumable** form (what `ST_CRS` and `RS_CRS`
+  return, and what is handed to PROJ at transform time), and
+- the **field-metadata** form (what is embedded in GeoArrow / GeoParquet
+  metadata).
+
+Both preserve the original authority code, PROJJSON, or WKT as given. The
+**only** canonicalization SedonaDB applies is for the longitude/latitude WGS84
+aliases: `EPSG:4326` and `OGC:CRS84` describe the same datum but imply different
+axis orders, so in the field-metadata form both are written as `OGC:CRS84` to
+keep the axis order (longitude, latitude) explicit for downstream readers. The
```
jiayuasu: Can you also make it clear that SedonaDB always assumes lon/lat axis order regardless what the CRS says

docs/concepts/crs.md
```diff
@@ -0,0 +1,186 @@
...
+two definitions as different when a full geodetic analysis would call them
+equivalent), but it will never *false-positive* (it never claims two genuinely
+different CRSes are the same). SedonaDB keeps the PROJ library out of the core
+schema layer entirely, so there is no semantic-equivalence check here. This is a
+different trade-off from, for example, GeoPandas, which compares CRSes through
+PROJ using an approximate, confidence-based similarity threshold.
+
+### What this means for you
+
+**Vector** operations that require their inputs to share a CRS — spatial joins,
```
jiayuasu, line 137: Can you elaborate how the planner checks the row-level CRS? This validation will run into the same problem as I mentioned here: https://github.com/apache/sedona/issues/3153

> My concern is specifically about enforcing a pairwise one-sided-CRS error when RS_Intersects, RS_Contains, or RS_Within is used as an optimized spatial join condition.

This should apply to vector geometry join too. We should not solely rely on pair-wise CRS mismatch check if we allow row-level CRSs differ across the column.

jiayuasu (changes requested):

docs/concepts/crs.md
james-willis: done

docs/concepts/crs.md
james-willis: done

docs/concepts/crs.md
james-willis: i updated the doc to include the details of how sedonadb works today on this issue

docs/concepts/crs.md
```diff
@@ -0,0 +1,225 @@
...
+  definitions — matching PROJJSON objects structurally, or comparing WKT strings
+  directly.
+
+This comparison is intentionally **lenient**: it can *false-negative* (report
+two definitions as different when a full geodetic analysis would call them
+equivalent), but it will never *false-positive* (it never claims two genuinely
+different CRSes are the same). SedonaDB keeps the PROJ library out of the core
+schema layer entirely, so there is no semantic-equivalence check here. This is a
+different trade-off from, for example, GeoPandas, which compares CRSes through
+PROJ using an approximate, confidence-based similarity threshold.
```
paleolimbot, line 133: For what it's worth we could sort of do this too (add `crses_compare_equal()` to the CrsEngine trait and wire up the PROJ function; however, the ConfigOptions are annoyingly *not* passed to `return_field()` the last time I checked, so we wouldn't be able to enforce it at planning time.

docs/concepts/crs.md
```diff
@@ -0,0 +1,225 @@
...
+  right one with `ST_SetCRS` / `ST_SetSRID` (metadata only — no coordinates
+  change).
+
+Because equality can false-negative, two definitions that *are* equivalent but
+are spelled differently (say, an authority code on one side and an
+authority-less WKT on the other) may be reported as mismatched. The fix is the
+same: normalize both sides to the same definition with `ST_Transform` or
+`ST_SetCRS`.
+
+### Raster joins are an exception
```
paleolimbot:
```suggestion
### `RS_Intersects` has different behaviour than `ST_Intersects`
```

docs/concepts/crs.md
```diff
@@ -0,0 +1,225 @@
...
+into curves in the target CRS, and reprojecting only the four corners can
+under-cover the true extent. SedonaDB handles this by **densifying** each edge —
+adding interior points in the raster's own CRS, where the edges are still
+straight — then reprojecting all of them, so the reprojected footprint follows
+the curve.
+
+One performance note: an accelerated spatial join relies on a spatial index
+built over a single, column-level CRS. A geometry or raster column that carries
+a *row-level* CRS can't use that index, so the join falls back to a slower
+per-row evaluation — prefer a column-level CRS for join-heavy work.
```
paleolimbot:
```suggestion
a *row-level* CRS can't use that index, so the join falls back to a slower
nested loop join evaluation — prefer a column-level CRS for join-heavy work.
```

docs/concepts/crs.md
```diff
@@ -0,0 +1,225 @@
...
+
+## SRID vs CRS
+
+An **SRID** (Spatial Reference Identifier) is just the numeric-code view of a
+CRS. `ST_SRID` / `RS_SRID` return the integer, and `ST_SetSRID` / `RS_SetSRID`
+set a CRS from one. SedonaDB maps between the two the obvious way: SRID `0`
```
paleolimbot:
```suggestion
set a CRS from one. SedonaDB maps between the two following the majority of SRID usage in the wild: SRID `0`
```

paleolimbot (approved):
Thank you!

docs/concepts/crs.md
```diff
@@ -0,0 +1,225 @@
...
+  than resampling the pixel grid.
+- **Only one side has a CRS** — this is an error, just as on the vector path:
+  SedonaDB will not reproject into or out of an unknown frame.
+
+The reprojected case is a close approximation rather than an exact boundary:
+reprojection is sometimes non-linear, so the footprint's straight edges may bow
+into curves in the target CRS, and reprojecting only the four corners can
+under-cover the true extent. SedonaDB handles this by **densifying** each edge —
+adding interior points in the raster's own CRS, where the edges are still
+straight — then reprojecting all of them, so the reprojected footprint follows
```
mdsumner, line 182: Extent reprojection requires internal points across the whole domain, not just the boundary. I can demonstrate if desired, my approach is in reproj::reproj_extent

james-willis: yes I would be interested.  I would say I'm not particularly talented with or knowledgable about projections.

The current densification logic is only intended to be better-than-nothing here.

mdsumner: very quickly here's a polar stereographic extent on the pole using CRAN R packages (note I use xmin,xmax,ymin,ymax generally)  to longlat: https://gist.github.com/mdsumner/cd3669c6dbc3afe8aafa9607f7b6f2d5   (I can spend more time on this, I'd like to write up in more detail with examples and compare to the tools in GDAL and PROJ as well). In the general case it's necessary to densify the field not just the boundary, as afaik that's the simplest way to capture most cases (GDAL warp of course does all kinds of other tricks and optimizations for the actual reprojection task)

mdsumner: and, nice doc!!

PR #1128, opened 2026-08-06
fix(python/sedonadb): share one Tokio runtime across contexts

james-willis:
`main`'s `python-wheels` run has been red on macOS because the free-threaded CPython 3.14 (`cp314t`) wheel crashes during its test run — it panics with `OS can't spawn worker thread: Resource temporarily unavailable` (EAGAIN) partway through the suite. Only the free-threaded build is affected; every GIL build passes the same tests. The free-threaded macOS wheel has already been disabled and re-enabled once for stability reasons, so rather than skip the target again this fixes the underlying cause.

Each `sedonadb.connect()` builds its own multi-threaded Tokio runtime (num_cpus workers plus a `spawn_blocking` pool). On a GIL build the runtime's Python-touching workers are serialized, so runtimes drain before many accumulate; on a free-threaded build they run in parallel, so across many short-lived sessions the worker and blocking threads pile up and exhaust the OS per-process thread limit — and the next runtime's worker spawn fails. Instrumenting the existing `RuntimeHandle` teardown ruled out a hung or leaking janitor: every runtime drained cleanly, and the problem was purely the number of per-session runtimes alive at once.

This shares a single process-global Tokio runtime across all contexts, so the worker and blocking-thread pools are bounded per process rather than per session: opening N sessions no longer costs N runtimes' worth of threads, a query still gets the full num_cpus worker pool, and the runtime lives for the process so it is never torn down on an interpreter thread. Verified on a local free-threaded 3.14 build — the full suite went from crashing (EAGAIN, dozens of runtimes draining at once) to zero spawn failures with the single shared runtime.

### DIFF
--- a/python/sedonadb/src/context.rs
+++ b/python/sedonadb/src/context.rs
@@ -14,7 +14,10 @@
 // KIND, either express or implied.  See the License for the
 // specific language governing permissions and limitations
 // under the License.
-use std::{collections::HashMap, sync::Arc};
+use std::{
+    collections::HashMap,
+    sync::{Arc, OnceLock},
+};
 
 use arrow_schema::DataType;
 use pyo3::prelude::*;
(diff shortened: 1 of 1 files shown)


james-willis:
Verification note, since a `context.rs`-only change doesn't match the `python-wheels` `paths:` filter and so isn't exercised by this PR's own CI:

The crash was reproduced locally on a free-threaded CPython 3.14 build (`python-freethreading`) running the full test suite — it fails partway through with `OS can't spawn worker thread` (EAGAIN), matching the CI signature, and only on the free-threaded interpreter (every GIL build passes the same tests). Instrumenting the existing `RuntimeHandle` teardown ruled out a hung or leaking janitor: every runtime drained cleanly (`drops == drained`). The cause was the *number* of per-session runtimes alive at once — dozens draining concurrently, each with a ~`num_cpus` worker pool. With the single shared runtime the same suite goes from crashing to **zero spawn failures**, with at most one runtime ever outstanding.

A full `python-wheels` run confirms this end-to-end: the `macOS-arm64` job passes, with every Python — including free-threaded 3.14 — at `3224 passed` and zero thread-spawn failures ([run](https://github.com/james-willis/sedona-db/actions/runs/31130579776)).

paleolimbot (approved):
Thank you! (merging to kick off the wheel builds on main)

PR #1129, opened 2026-08-06
SedonaType::UnrecognizedExtension for user-defined types

james-willis:
## What
Add a `SedonaType::UnrecognizedExtension(ExtensionType)` entry to the type enum, a plain-data variant for any Arrow `ARROW:extension:name` this crate doesn't have built-in support for. The `(name, storage_type, metadata)` triple just rides along verbatim.

## Context

Came out of a discussion with Dewey about potentially having a Tensor type: the existing "adding a variant is a closed-enum core-crate change" framing was called out as a pragmatic historical choice, not a design constraint.

This isn't a new idea in this codebase — it's exactly how `Wkb`/`WkbView`/`Raster` already work today. `Wkb(Planar, crs_a)` and `Wkb(Spherical, crs_b)` have identical physical storage (`DataType::Binary`); what distinguishes them is the `ARROW:extension:name`/`ARROW:extension:metadata` tag on the `Field`, not the shape. `UnrecognizedExtension` is that same mechanism, open-ended.

## How to implement a UDT with this

No sedona-db changes needed — this is all in an out-of-tree crate depending on `sedona-schema`/`sedona-expr`:

1. **Tag it.** Pick a name (`myorg.custom_type`) and a physical storage `DataType`. Build the `SedonaType` as `SedonaType::UnrecognizedExtension(ExtensionType::new(name, storage_type, metadata))`, and build `Field`s for it via `ExtensionType::to_field(name, nullable)` — never a bare `Field::new`, or the tag is lost and the column round-trips as an untagged `Arrow(Struct(..))`.
2. **Read it back for free.** `SedonaType::from_storage_field` already degrades any unrecognized `extension_name` to `UnrecognizedExtension` rather than erroring — nothing to register.
3. **Discriminate on it.** `ArgMatcher::is_extension("myorg.custom_type")` matches only columns tagged with that exact name; a same-shaped but untagged (or differently-tagged) column correctly doesn't match. Use it in a kernel's signature exactly like `is_raster()`/`is_geometry()`.
4. **Get pretty printing for free.** `logical_type_name()`/`Display` for `UnrecognizedExtension` return the real extension name, so `DESCRIBE` and error messages show `myorg.custom_type` instead of a struct dump.

Prototyped this against a real out-of-tree crate (a `Tensor` type + a strict `TN_Add(Tensor, Tensor)` UDF) to confirm it needs zero sedona-db changes end to end.

## What's in it

- `SedonaType::UnrecognizedExtension(ExtensionType)`.
- `from_extension_type`'s fallback now returns `Ok(UnrecognizedExtension(extension))` instead of `sedona_internal_err!`. Also fixes a latent bug: Wkb-specific metadata parsing ran unconditionally before the name check, so a third-party extension with non-JSON metadata would have errored before ever reaching that fallback. It now only runs for `geoarrow.wkb`.
- `ArgMatcher::is_extension` takes `impl Into<String>` instead of `&'static str` (a UDT's name is a runtime value, not a compile-time constant); `SedonaType::extension_name()` relaxed the same way.
- Every exhaustive match on `SedonaType` updated: `storage_type`/`extension_name`/`extension_type`/`logical_type_name`/`match_signature`/`Display` in `sedona-schema`, two sites in `sedona-functions/sd_format.rs`, one in `sedona-geoarrow-c`.


rust/sedona-schema/src/extension_type_registry.rs
```diff
@@ -0,0 +1,271 @@
...
+}
+
+/// Reconstructs a concrete `Arc<dyn SedonaExtensionType>` from the
+/// `(extension_name, storage_type, extension_metadata)` triple recovered
+/// from an external `Field`. Registered once per `extension_name`.
+pub type ExtensionTypeFactory =
+    dyn Fn(&ExtensionType) -> Result<Arc<dyn SedonaExtensionType>> + Send + Sync;
+
+static EXTENSION_TYPE_REGISTRY: LazyLock<RwLock<HashMap<&'static str, Arc<ExtensionTypeFactory>>>> =
+    LazyLock::new(|| RwLock::new(HashMap::new()));
```
paleolimbot: The real crux of extension type handling is how to avoid a static registry...DataFusion put its registry in the Session, although it's of limited use unless it can reach some actual execution points. For us the ConfigOptions can work, as it reaches most places (although annoyingly not some of the ones we want, like return type calculations).

paleolimbot (approved):
Thank you! I think this is a better solution for `SedonaType`...a dynamic `DFExtensionType`-like solution still has merit but is hard to pipe through to all the places its needed (the hardest thing to support is the printing of the type name for error messages, since that often happens far removed from any ConfigOptions or Session).

No need to do it here, but maybe worth making sure that "no kernel for arguments" error messages with the extension type argument don't barf the whole metadata string at the user (which will be long for things like geoarrow.point).

PR #1131, opened 2026-08-07
ci(python-wheels): cache vcpkg binaries on Windows and macOS

james-willis:
Caches vcpkg's built binaries (`VCPKG_DEFAULT_BINARY_CACHE` + `actions/cache`) so `vcpkg install` restores instead of rebuilding GEOS/Abseil/OpenSSL/S2 from source each run.

This should cut out the first 40 minutes of  the Windows and macOS wheel jobs.

Keyed on OS/arch + `VCPKG_REF` + triplet hashes.
 
Linux builds vcpkg inside the manylinux container and needs the cache threaded through the mount; left as a follow-up.

### DIFF
--- a/.github/workflows/python-wheels.yml
+++ b/.github/workflows/python-wheels.yml
@@ -66,14 +66,27 @@ jobs:
           ref: ${{ env.VCPKG_REF }}
           path: vcpkg
 
+      # Persist vcpkg's built binaries across runs. The pinned VCPKG_REF and
+      # custom triplets fully determine what gets built, so a hit turns the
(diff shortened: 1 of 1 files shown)


paleolimbot (approved):
Thank you!

May be worth pushing a small other commit once these pass to check that they actually work, and also making sure that the cache is populated even for failing wheel builds if you can.

james-willis:
@paleolimbot I have some evidence this works as I'm using it in my fork to test windows wheel build issues but will do it anyway. Its a good idea to catch CI issues before we merge as we are finding right now!

PR #1132, opened 2026-08-07
ci(python-wheels): cache vcpkg binaries on Linux via container mount

james-willis:
Stacked on #1131 — extends the vcpkg binary cache to the manylinux wheel jobs.


paleolimbot (approved):
Cool.

PR #1135, opened 2026-08-07
fix(sedona-gdal): skip GDALClose during interpreter shutdown

james-willis:
The Windows wheel job passes every test, then aborts at process exit with `0xC0000409`. Root cause: thread-local GDAL dataset caches (`Rc<Dataset>`) run `GDALClose` from their destructors while GDAL's shared library is being unloaded, which corrupts teardown on Windows. A Python `atexit` callback now latches a shutdown flag; `Dataset::drop` then leaves the handle open (harmless — the OS reclaims it at process exit).

Localized and validated with a subprocess probe on cp313: `raster_only` went from 10/10 crashes to 0/10 (and the full suite 12/12 → 0/12), with all 3217 tests still passing.

This should make the wheels all green now 🥳

### DIFF
--- a/c/sedona-gdal/src/dataset.rs
+++ b/c/sedona-gdal/src/dataset.rs
@@ -46,7 +46,11 @@ unsafe impl Send for Dataset {}
 
 impl Drop for Dataset {
     fn drop(&mut self) {
-        if !self.c_dataset.is_null() {
+        // Once the process is shutting down, leave the handle open. On Windows,
+        // closing a dataset while GDAL's shared library is being unloaded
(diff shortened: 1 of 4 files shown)


paleolimbot (approved):
This is great! We had some issues like this in pyarrow related to various C++ APIs with shutdown behaviour.

In addition to this, we should also strive to better scope the cache / maintain it a bit better so that it less frequently tries to scramble and delete a whole lot of datasets at once right at process close. I am skeptical of most non-connection scoped caches (we have a few here) because of issues like this. (Not a battle for this PR.)

PR #1137, opened 2026-08-08
R: fix build on Windows ARM64

jeroen:
See failure on https://github.com/r-universe/apache/actions/runs/31226347531/job/93050529031

More information: https://contributor.r-project.org/windows-arm64/

### DIFF
--- a/r/sedonadb/configure.win
+++ b/r/sedonadb/configure.win
@@ -107,8 +107,22 @@ if [ -n "${SAVVY_FEATURES}" ]; then
   FEATURE_FLAGS="--features '${SAVVY_FEATURES}'"
 fi
 
+# Derive the Rust target from the C compiler architecture. On Windows arm64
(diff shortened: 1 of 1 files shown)


james-willis (approved):

james-willis:
@paleolimbot do we want to add any kind of CI target for this?

yutannihilation:
Looks great!

> any kind of CI target for this

As `r-lib/actions` already supports Windows ARM64, it should be easy to add ARM64 Windows runner on GitHub Actions.

e.g. https://github.com/yutannihilation/savvy/pull/457/changes#diff-9c940e8ad2b7bc4c26ec3da57b94bc00e73e2166cfed689da51a4c59bcc0a310

But, considering `r.yml` runs only one case for each OS (e.g. no Intel macOS), I think we don't want to add it in order to keep the CI setup simple.

Issue #1086, opened 2026-07-25
ST_KNN: add a geography kernel ("Can't execute ST_KNN() outside a spatial join" for geography args)


jiayuasu:
## Description

`ST_KNN` has no geography kernel. Calling it with geography arguments inside a join fails with:

```
Can't execute ST_KNN() outside a spatial join
```

even though the join *is* a spatial join — the geography arguments simply are not recognised by the KNN join planner, so the predicate falls through to a non-spatial join and then rejects itself.

## Reproduction

```python
import json, sedonadb

sd = sedonadb.connect()
opts = json.dumps({"geom_type": "Point", "bounds": [0, 0, 10, 10], "seed": 11})
sd.sql(f"SELECT id, ST_GeogFromWKB(ST_AsBinary(geometry)) AS geog, "
       f"ST_GeomFromWKB(ST_AsBinary(geometry)) AS geom "
       f"FROM sd_random_geometry('{opts}') LIMIT 50").to_view("t", overwrite=True)
sd.sql("SELECT id, geog, geom FROM t").to_view("u", overwrite=True)

# geometry: works
sd.sql("SELECT COUNT(*) FROM t a JOIN u b ON ST_KNN(a.geom, b.geom, 5, true)").to_pandas()
# -> 250

# geography: fails
sd.sql("SELECT COUNT(*) FROM t a JOIN u b ON ST_KNN(a.geog, b.geog, 5, false)").to_pandas()
# -> SedonaError: Can't execute ST_KNN() outside a spatial join
```

The two queries are identical apart from the column type.

## Requested behaviour

`ST_KNN(geogA, geogB, k)` ranking neighbours by geodesic distance, consistent with `ST_Distance(geography, geography)` returning metres. The `use_spheroid` flag is meaningful only for geometry — for geography, spherical ranking is the definition, so a 3-argument geography kernel would be the natural signature.

## Also missing: ST_Union_Agg

For completeness, `ST_Union_Agg` has no geography kernel either:

```sql
SELECT ST_Union_Agg(geog) FROM t;
-- st_union_agg(geography): No kernel matching arguments
```

Lower priority — the aggregate that SpatialBench needs is `ST_Collect_Agg`, which already works on geography.

## Motivation

This is the last function blocking a complete geography version of the SpatialBench query suite. Q12 (rank trip pickups by average distance to their 5 nearest buildings) is a KNN join, and it is the one query of the twelve that has no geography formulation today — the other eleven are expressible, see
[...]

Issue #783, opened 2026-04-23
rust/sedona-spatial-join-geography: Support KNN join in geography join


paleolimbot:
After https://github.com/apache/sedona-db/pull/775 we have a join for geometry! This doesn't support geography; however, although some infrastructure exists to customize a distance metric in the internals. In geo-index we're hoping to remove the need for the `DistanceMetric` (https://github.com/georust/geo-index/issues/159) at which point it might be worth getting the interface right so that we can use S2's optimized distance predicates (perhaps even leveraging prepared distance, which other libraries can't do).

PR #1141, opened 2026-08-10
docs: add missing geography kernels to SQL function reference pages

### LINKED ISSUE #1087: docs: st_collect_agg and st_makeline are missing their geography kernels in the SQL reference
## Description

`docs/reference/sql/st_collect_agg.qmd` and `docs/reference/sql/st_makeline.qmd` declare only a `geometry` kernel:

```yaml
# st_collect_agg.qmd
kernels:
  - returns: geometry
    args: [geometry]

# st_makeline.qmd
kernels:
  - returns: geometry
    args: [geometry, geometry]
```

But 0.4.0 accepts **geography** for both, and returns geography:

```python
import json, sedonadb
sd = sedonadb.connect()
opts = json.dumps({"geom_type": "Point", "bounds": [0, 0, 10, 10], "seed": 11})
sd.sql(f"SELECT id, ST_GeogFromWKB(ST_AsBinary(geometry)) AS geog, "
       f"ST_GeomFromWKB(ST_AsBinary(geometry)) AS geom "
       f"FROM sd_random_geometry('{opts}') LIMIT 50").to_view("t", overwrite=True)
sd.sql("SELECT id, geog, geom FROM t").to_view("u", overwrite=True)

sd.sql("SELECT ST_GeometryType(ST_Collect_Agg(geog)) FROM t").to_pandas()
# -> ST_MultiPoint

sd.sql("SELECT ST_Area(ST_ConvexHull(ST_Collect_Agg(geog))) FROM t").to_pandas()
# -> 970542623930.96      (m^2 -- the spherical path)

sd.sql("""SELECT ST_Length(ST_MakeLine(a.geog, b.geog))
          FROM t a, u b WHERE a.id = 1 AND b.id = 2""").to_pandas()
# -> 865168.17            (metres)

sd.sql("""SELECT ST_Length(ST_MakeLine(a.geom, b.geom))
          FROM t a, u b WHERE a.id = 1 AND b.id = 2""").to_pandas()
# -> 7.87                 (degrees -- same two points, geometry overload)
```

865 168 m vs 7.87 degrees for the same pair of points confirms these are really going through the geography path, not silently downcasting to geometry.

## Suggested fix

Add the geography overloads to the `kernels:` blocks:

```yaml
# st_collect_agg.qmd
kernels:
  - returns: geometry
    args: [geometry]
  - returns: geography
    args: [geography]

# st_makeline.qmd
kernels:
  - returns: geometry
    args: [geometry, geometry]
  - returns: geography
    args: [geography, geography]
```

It would also be worth auditing the other `.qmd` kernel lists against the registry, in case more geography overloads are undocumented — a generated check would keep them honest.

## Why it matters

The `kernels:` metadata is the natural place to look when working out what the geography type can do. Taking it at face value, I concluded that SpatialBench's Q5 (`ST_Collect_Agg` + `ST_ConvexHull`) and Q7 (`ST_MakeLine` + `ST_Length`) had no geography formulation, and shipped them as documented-but-unimplemented. Both in fact work on 0.4.0, so ten of the twelve SpatialBench queries have a working geography version rather than eight.

## Environment

- `sedonadb` 0.4.0 (PyPI wheel), Python 3.13, macOS arm64
- `sedonadb.__features__ == ['s2geography']`

jiayuasu:
## What changes are proposed in this pull request?

Adds the missing `geography` entries to the `kernels:` frontmatter of the SQL function reference, so the docs match what the function registry actually accepts.

The two pages reported in #1087:

- `st_collect_agg.qmd` — `geography -> geography`
- `st_makeline.qmd` — `(geography, geography) -> geography`

Following the issue's suggestion, I also audited every other `.qmd` kernel list against the registry and found 14 more pages with working-but-undocumented geography signatures:

- `ST_AsEWKB`, `ST_CRS`, `ST_Dump`, `ST_EndPoint`, `ST_StartPoint`, `ST_HasM`, `ST_HasZ`, `ST_M`, `ST_Z`, `ST_SRID`, `ST_Zmflag`, `ST_SetSRID`
- `ST_Transform` — 2- and 3-arg geography kernels; the description notes the target CRS must be geographic
- `ST_Envelope_Agg` — geography input, but the envelope is returned as **geometry** (planar edges), which the description now calls out

## How was the audit done?

For every documented geometry kernel I built the geography twin and checked it against the registry at plan time, on two builds (the 0.4.0 wheel with `s2geography` and a dev build for the core kernels), then executed each candidate signature and inspected the output type's `edges`/`crs` metadata to get the correct `returns:` value. Each addition was also cross-checked against the `ArgMatcher` definitions in the Rust source. Probes that fail on argument *values* rather than types (CRS strings, SRIDs) can hide real kernels, which is how `ST_SetSRID` and `ST_Transform` were initially missed and then confirmed with valid values.

Deliberately excluded: `ST_KNN` type-matches geography arguments but the KNN join rewrite has no geography path, so it always fails with "Can't execute ST_KNN() outside a spatial join" — that's tracked in #1086 (see also #783/#801), and documenting it here would repeat the inverse of this bug.

All edited frontmatter renders cleanly through `docs/reference/sql/_render_meta.py`, showing both overloads in the Usage section.

## Are these changes tested?

Docs-only. Every added kernel signature was verified by executing it against the registry as described above; the rendering pipeline was exercised on all 16 edited pages.

Yes — the SQL reference now documents the geography overloads for 16 functions.

Closes #1087

### DIFF
--- a/docs/reference/sql/st_envelope_agg.qmd
+++ b/docs/reference/sql/st_envelope_agg.qmd
@@ -21,11 +21,13 @@ description: An aggregate function that returns the collective bounding box (env
 kernels:
   - returns: geometry
     args: [geometry]
+  - returns: geometry
+    args: [geography]
 ---
 
 ## Description
 
-An aggregate function that computes the collective bounding box (envelope) of all geometries in a group.
+An aggregate function that computes the collective bounding box (envelope) of all geometries in a group. Geography input is also accepted; the envelope is always returned as a geometry.
 
 ## Examples
 
--- a/docs/reference/sql/st_envelope_agg.qmd
+++ b/docs/reference/sql/st_envelope_agg.qmd
@@ -27,7 +27,7 @@ kernels:
 
 ## Description
 
-An aggregate function that computes the collective bounding box (envelope) of all geometries in a group. Geography input is also accepted; the envelope is always returned as a geometry.
+An aggregate function that computes the collective bounding box (envelope) of all geometries in a group. Geography input is also accepted; the envelope is always returned as a geometry. For geographies that cross the antimeridian, a multi geometry is returned with one polygon each for the west and east component.
 
 ## Examples
 
--- a/docs/reference/sql/st_asewkb.qmd
+++ b/docs/reference/sql/st_asewkb.qmd
@@ -21,6 +21,8 @@ description: Returns the EWKB representation of a geometry or geography.
 kernels:
   - returns: binary
     args: [geometry]
+  - returns: binary
+    args: [geography]
 ---
 
 ## Description
--- a/docs/reference/sql/st_collect_agg.qmd
+++ b/docs/reference/sql/st_collect_agg.qmd
@@ -17,17 +17,19 @@
 # under the License.
 
 title: ST_Collect_Agg
-description: Combines multiple geometries from a set of rows into a single collection.
+description: Combines multiple geometries or geographies from a set of rows into a single collection.
 kernels:
   - returns: geometry
     args: [geometry]
+  - returns: geography
+    args: [geography]
 ---
 
 ## Description
 
-An aggregate function that collects multiple geometries into a single
-GeometryCollection or the appropriate Multi-type if all inputs share the same
-geometry type.
+An aggregate function that collects multiple geometries or geographies into a
+single GeometryCollection or the appropriate Multi-type if all inputs share the
+same geometry type. Collecting geographies returns a geography.
 
 ## Examples
 
--- a/docs/reference/sql/st_crs.qmd
+++ b/docs/reference/sql/st_crs.qmd
@@ -21,6 +21,8 @@ description: Returns the Coordinate Reference System (CRS) metadata associated w
 kernels:
   - returns: string
     args: [geometry]
+  - returns: string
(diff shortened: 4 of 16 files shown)


docs/reference/sql/st_envelope_agg.qmd
```diff
@@ -21,11 +21,13 @@ description: An aggregate function that returns the collective bounding box (env
...
   - returns: geometry
     args: [geometry]
+  - returns: geometry
+    args: [geography]
 ---
 
 ## Description
 
-An aggregate function that computes the collective bounding box (envelope) of all geometries in a group.
+An aggregate function that computes the collective bounding box (envelope) of all geometries in a group. Geography input is also accepted; the envelope is always returned as a geometry.
```
paleolimbot: While we're here, I never documented this:

```suggestion
An aggregate function that computes the collective bounding box (envelope) of all geometries in a group. Geography input is also accepted; the envelope is always returned as a geometry. For geographies that cross the antimeridian, a multi geometry is returned with one polygon each for the west and east component.
```

paleolimbot (approved):
Thank you!

PR #1144, opened 2026-08-10
refactor(rust/sedona-raster): add RasterError to replace ArrowError plumbing

james-willis:
Introduces a dedicated `RasterError` for the raster crates (mirroring `sedona-geometry`'s `SedonaGeometryError`) so error plumbing is `?`/`.context()`-based instead of manual `ArrowError` `map_err`.

### DIFF
--- a/rust/sedona-raster/src/error.rs
+++ b/rust/sedona-raster/src/error.rs
@@ -43,7 +43,7 @@
 //! directly — same as `sedona-geometry`, which has no internal variant.
 
 use arrow_schema::ArrowError;
-use datafusion_common::DataFusionError;
+use datafusion_common::{exec_datafusion_err, DataFusionError};
 use thiserror::Error;
 
 /// Error type for the raster crates.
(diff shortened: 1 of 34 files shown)


rust/sedona-raster/src/error.rs
```diff
@@ -0,0 +1,192 @@
...
+            RasterError::Unknown => DataFusionError::Execution("Unknown raster error".to_string()),
+        }
+    }
+}
+
+impl From<RasterError> for ArrowError {
+    fn from(value: RasterError) -> Self {
+        match value {
+            RasterError::Arrow(e) => e,
+            other => ArrowError::ExternalError(Box::new(other)),
```
jiayuasu, line 107: Would it make sense to preserve the existing Arrow variants at this boundary? For example, Grid::from_raster() still returns ArrowError, but an empty spatial_shape now changes from InvalidArgumentError to ExternalError through this catch-all. That may be intentional, though callers matching on the Arrow variant would see a behavior change. Maybe map Invalid back to InvalidArgumentError and unwrap External(e) here?

rust/sedona-raster/src/error.rs
```diff
@@ -0,0 +1,192 @@
...
+    #[error("Unknown raster error")]
+    Unknown,
+}
+
+impl From<RasterError> for DataFusionError {
+    fn from(value: RasterError) -> Self {
+        match value {
+            // Keep Arrow errors typed so DataFusion renders them as Arrow
+            // errors rather than an opaque external error.
+            RasterError::Arrow(e) => DataFusionError::ArrowError(Box::new(e), None),
```
jiayuasu: Was dropping the DataFusion backtrace intentional here? A path such as rs_dim_band now reaches this arm via ?, while the old arrow_datafusion_err! path stored Some(get_back_trace()). With the backtrace feature enabled this becomes None. Maybe this arm could use DataFusionError::from(e), with equivalent handling for the Execution arms below.

james-willis: no not intentional. good idea, will add

rust/sedona-raster/src/error.rs
james-willis: done

jiayuasu (approved):

PR #1146, opened 2026-08-10
feat(python/sedonadb): Add a native scalar UDF import path for plugins

james-willis:
## What

Adds a runtime import path for natively-compiled scalar UDFs: an out-of-tree plugin can now hand SedonaDB a real `SedonaScalarKernel` (real compiled Rust, no Python callback per invocation) via a `PyCapsule`, the same way `sedona-extension`'s `SedonaCScalarKernel` ABI is already used to statically link in kernels at build time (see `c/sedona-s2geography`) -- this is the runtime counterpart of that same mechanism.

Concretely: `SedonaContext.register()` gains a new protocol, `__sedonadb_native_scalar_udfs__(self) -> list[PyCapsule]`, alongside the existing `__sedonadb_internal_udf__`/`__sedonadb_internal_aggregate_udf__`/`__sedonadb_external_format__`/`__sedonadb_raster_loader__`.

## Why

Today, `SedonaContext.register()`'s only path for a *scalar function* is `arrow_udf`/`sedona_scalar_udf` -- a Python callable invoked once per batch. There's no way for an out-of-tree crate (depending on `sedona-schema`/`sedona-expr` directly, like a prototype extension building a real `SedonaScalarKernel` with its own `ArgMatcher`-based dispatch) to register that kernel and get real Rust dispatch from a Python `SedonaContext` -- only a Python-callback wrapper around it. This closes that gap using the ABI SedonaDB already has, rather than inventing a new one.

## What's in it

- `import_sedona_ffi_scalar_kernel` (`import_from.rs`): imports a `PyCapsule` wrapping a `SedonaCScalarKernel` into a real `ScalarKernelRef`, reading the kernel's own declared name. Mirrors `import_sedona_ffi_table_provider`'s existing pattern exactly, including the double-free-prevention `ptr::read` + `ptr::write_bytes` zeroing.
- `sedona_native_scalar_udf` (`udf.rs`): a standalone pyfunction building a `PySedonaScalarUdf` from one or more same-named kernel capsules (errors on a name mismatch rather than silently registering under the wrong name) -- usable directly, the native-kernel analog of the existing `sedona_scalar_udf`. Also the documented escape hatch for a kernel that needs `Volatile`/`Stable` (see below).
- `register_component()` (`context.rs`): the new `__sedonadb_native_scalar_udfs__` branch. A plugin's capsules can span multiple distinct function names in one call; grouped by each kernel's own declared name before registering.
- `context.py`: `__sedonadb_native_scalar_udfs__` added to `register()`'s `supported_interfaces` and docstring.

## Two real limitations, documented rather than hidden

- **Registration replaces, it doesn't accumulate.** Grouping by name is scoped to one `__sedonadb_native_scalar_udfs__()` call's own capsule list -- it does *not* accumulate across calls the way `SedonaContext::register_scalar_kernels` accumulates statically-linked kernels. Registering under a name already in use (a plugin's own, a different plugin's, or a built-in's) replaces it outright -- a plain `HashMap` insert, the same as the existing `__sedonadb_internal_udf__` path already does. Not a new risk this PR introduces, but worth being explicit about in the code rather than implying false parity with the accumulate-behavior of `register_scalar_kernels`.
- **Volatility is always `Immutable` through this protocol** -- a bare capsule has nowhere to carry a volatility value, and this is not because every native kernel is `Immutable` (`RS_FromPath` is `Volatile`). A plugin kernel that needs `Volatile`/`Stable` should call `sedona_native_scalar_udf(kernels, volatility=...)` directly and return the resulting `PySedonaScalarUdf` via the existing `__sedonadb_internal_udf__` protocol instead.

## What's NOT in it

- Aggregate UDFs. `SedonaCScalarKernel` has no aggregate equivalent yet -- an accumulator's stateful lifecycle (create/update/merge/evaluate/state/size, to participate correctly in DataFusion's own parallel aggregation) needs a materially larger C ABI than a stateless scalar kernel's. This PR is scoped to the scalar case; the aggregate ABI is a separate design question.
- A Python-exposed way to build an `ArgMatcher::is_extension(name)`-style matcher (`import_arg_matcher` still only recognizes the fixed literal set). Not a blocker here: a plugin's own `ArgMatcher` lives entirely inside its compiled kernel's `return_type()` and never crosses the Python boundary, so this only matters for someone trying to express that matcher from pure Python via `arrow_udf`.

## Verification

- 10 real, executed tests (`import_from.rs`, `udf.rs`): capsule import + functional kernel invocation, rejecting a wrong capsule name, rejecting a second import of an already-consumed capsule, rejecting non-capsule input, building a working UDF, grouping two *disjoint-type* kernels into one overloaded UDF and dispatching both correctly via real SQL (`probe_grouped(42)` and `probe_grouped(4.5)` each hitting their own kernel), rejecting mismatched kernel names, rejecting an empty kernel list, an explicit name override, and a full SQL query executed through a live `SedonaContext` against an imported native kernel.
- Getting real coverage needed one addition: `pyo3 = { workspace = true, features = ["auto-initialize"] }` under `[dev-dependencies]`. `extension-module` (needed for the real wheel build) is only ever added by `maturin`'s own build flags (`pyproject.toml`), never by this crate's `Cargo.toml` -- so it's never present during `cargo test`, including `cargo test --all-features` (CI's actual invocation). Confirmed directly that the two configurations don't collide: explicitly compiling with `--features pyo3/extension-module` does fail to link (undefined libpython symbols), but that's never what `cargo test` requests.
- `cargo check`/`clippy --all-targets --all-features -- -D warnings`/`fmt --all -- --check` for the whole workspace (excluding the environment-only `gdal-sys` bindgen issue on this machine, unrelated to this change) -- clean.
- Built the real wheel (`maturin develop --release`) and ran the full existing `python/sedonadb` pytest suite against it: 2543 passed, 1753 skipped (optional deps not installed in this verification venv) -- zero regressions from this purely additive change.

### DIFF
--- a/python/sedonadb/Cargo.toml
+++ b/python/sedonadb/Cargo.toml
@@ -64,3 +64,15 @@ thiserror = { workspace = true }
 tokio = { workspace = true }
 mimalloc = { workspace = true, optional = true }
 libmimalloc-sys = { workspace = true, optional = true }
+
+[dev-dependencies]
+# `extension-module` (needed for the real cdylib/wheel build) is only ever
+# added by maturin's own build flags (see pyproject.toml's [tool.maturin]),
+# never by this Cargo.toml -- so `cargo test` builds this crate without it,
+# and `auto-initialize` here is what lets a #[test] embed and drive a real
+# Python interpreter via Python::attach. Confirmed directly: explicitly
+# compiling with `--features pyo3/extension-module` fails to link (undefined
+# libpython symbols), but plain `cargo test`/`cargo test --all-features`
+# (CI's actual invocation) never requests that feature, so the two never
+# collide in practice.
+pyo3 = { workspace = true, features = ["auto-initialize"] }
--- a/python/sedonadb/python/sedonadb/context.py
+++ b/python/sedonadb/python/sedonadb/context.py
@@ -444,6 +444,17 @@ class SedonaContext:
         - An ExternalFormatSpec implementing a custom datasource type
         - An object implementing __sedonadb_extension__(ctx, **kwargs), which
           is called with this context and any keyword arguments passed.
+        - An object implementing __sedonadb_native_scalar_udfs__(self), which
+          returns a list of PyCapsule objects each wrapping a natively-
+          compiled SedonaCScalarKernel -- real compiled Rust runs per
+          invocation, not a Python callback. Kernels sharing a declared name
+          are grouped into one overloaded UDF (within this one call only --
+          registering under a name already in use, including a built-in's,
+          replaces it, the same as __sedonadb_internal_udf__ already does).
+          Volatility is always Immutable through this protocol; a kernel
+          needing Volatile or Stable should be registered via
+          __sedonadb_internal_udf__ instead, built with
+          sedonadb._lib.sedona_native_scalar_udf(kernels, volatility=...).
 
         The extension interface is experimental and may change.
 
@@ -490,6 +501,7 @@ class SedonaContext:
             "__sedonadb_internal_aggregate_udf__",
             "__sedonadb_external_format__",
             "__sedonadb_raster_loader__",
+            "__sedonadb_native_scalar_udfs__",
         )
         for interface in supported_interfaces:
             if hasattr(component, interface):
--- a/python/sedonadb/python/sedonadb/context.py
+++ b/python/sedonadb/python/sedonadb/context.py
@@ -444,14 +444,15 @@ class SedonaContext:
         - An ExternalFormatSpec implementing a custom datasource type
         - An object implementing __sedonadb_extension__(ctx, **kwargs), which
           is called with this context and any keyword arguments passed.
-        - An object implementing __sedonadb_native_scalar_udfs__(self), which
-          returns a list of PyCapsule objects each wrapping a natively-
-          compiled SedonaCScalarKernel -- real compiled Rust runs per
-          invocation, not a Python callback. Kernels sharing a declared name
-          are grouped into one overloaded UDF (within this one call only --
-          registering under a name already in use, including a built-in's,
-          replaces it, the same as __sedonadb_internal_udf__ already does).
-          Volatility is always Immutable through this protocol; a kernel
+        - A single function object implementing __sedonadb_scalar_udf__(self),
+          which returns a list of PyCapsule objects -- the overload kernels of
+          that one function, each wrapping a natively-compiled
+          SedonaCScalarKernel sharing the function's SQL name -- so real
+          compiled Rust runs per invocation, not a Python callback. Register
+          each function individually. The kernels are grouped into one
+          overloaded UDF; registering under a name already in use, including a
+          built-in's, replaces it, the same as __sedonadb_internal_udf__ already
+          does. Volatility is always Immutable through this protocol; a kernel
           needing Volatile or Stable should be registered via
           __sedonadb_internal_udf__ instead, built with
           sedonadb._lib.sedona_native_scalar_udf(kernels, volatility=...).
@@ -501,7 +502,7 @@ class SedonaContext:
             "__sedonadb_internal_aggregate_udf__",
             "__sedonadb_external_format__",
             "__sedonadb_raster_loader__",
-            "__sedonadb_native_scalar_udfs__",
+            "__sedonadb_scalar_udf__",
         )
         for interface in supported_interfaces:
             if hasattr(component, interface):
--- a/python/sedonadb/python/sedonadb/context.py
+++ b/python/sedonadb/python/sedonadb/context.py
@@ -449,13 +449,16 @@ class SedonaContext:
           that one function, each wrapping a natively-compiled
           SedonaCScalarKernel sharing the function's SQL name -- so real
           compiled Rust runs per invocation, not a Python callback. Register
-          each function individually. The kernels are grouped into one
-          overloaded UDF; registering under a name already in use, including a
-          built-in's, replaces it, the same as __sedonadb_internal_udf__ already
-          does. Volatility is always Immutable through this protocol; a kernel
-          needing Volatile or Stable should be registered via
-          __sedonadb_internal_udf__ instead, built with
-          sedonadb._lib.sedona_native_scalar_udf(kernels, volatility=...).
+          each function individually. The kernels are appended as overloads to
+          the function of that name (a new function is created when none exists
+          yet). A kernel whose signature matches an existing overload --
+          including a built-in's -- overrides that one signature (kernels
+          resolve newest-first) while the function's other overloads stay
+          reachable, so a plugin can add an overload to an existing function
+          without replacing it. Volatility is always Immutable through this
+          protocol; a kernel needing Volatile or Stable should be registered
+          via __sedonadb_internal_udf__ instead, built with
+          sedonadb.udf.sedona_native_scalar_udf(kernels, volatility=...).
 
         The extension interface is experimental and may change.
 
--- a/python/sedonadb/src/context.rs
+++ b/python/sedonadb/src/context.rs
@@ -20,17 +20,19 @@ use std::{
 };
 
 use arrow_schema::DataType;
+use datafusion_expr::Volatility;
 use pyo3::prelude::*;
 use sedona::context::SedonaContext;
 use sedona::context_builder::SedonaContextBuilder;
 use sedona_datasource::format::ExternalFormatFactory;
+use sedona_expr::scalar_udf::SedonaScalarUDF;
 use sedona_extension::runtime::RuntimeHandle;
 
 use crate::{
     dataframe::InternalDataFrame,
     datasource::PyExternalFormat,
     error::PySedonaError,
-    import_from::import_table_provider_from_any,
+    import_from::{import_sedona_ffi_scalar_kernel, import_table_provider_from_any},
     raster_loader::PyRasterLoaderWrapper,
     runtime::wait_for_future,
     udf::{PyAggregateUdf, PyScalarUdf, PySedonaAggregateUdf, PySedonaScalarUdf},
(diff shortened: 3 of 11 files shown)


python/sedonadb/src/context.rs
```diff
@@ -354,6 +356,47 @@ impl InternalContext {
...
+            // drops whatever was there before; this mirrors the existing,
+            // already-trusted __sedonadb_internal_udf__ path's behavior, not
+            // a new risk this protocol introduces.
+            //
+            // Volatility isn't plugin-configurable through this protocol,
+            // hardcoded to Immutable -- not because every native kernel is
+            // Immutable (RS_FromPath, for one, is Volatile), but because a
+            // bare capsule has nowhere to carry a volatility value. A plugin
+            // needing Volatile/Stable can call sedona_native_scalar_udf(...,
+            // volatility=...) directly and return the resulting
```
paleolimbot: We can add this later via a `get_property()` callback on the kernel

python/sedonadb/src/context.rs
```diff
@@ -354,6 +356,47 @@ impl InternalContext {
...
+            let mut kernels_by_name: HashMap<String, Vec<_>> = HashMap::new();
+            for capsule in &capsules {
+                let (name, kernel) = import_sedona_ffi_scalar_kernel(capsule)?;
+                kernels_by_name.entry(name).or_default().push(kernel);
+            }
+            for (name, kernels) in kernels_by_name {
+                let udf = SedonaScalarUDF::new(&name, kernels, Volatility::Immutable);
+                self.inner.register_sedona_scalar_udf(udf)?;
+            }
+            return Ok(());
```
paleolimbot: The FunctionSet for our functions already consolidates multiple kernels into one. Apologies if I missed it, but do we need to do this especially for the Python registration?

python/sedonadb/src/import_from.rs
```diff
@@ -211,3 +256,97 @@ pub fn check_pycapsule(obj: &Bound<PyAny>, name: &str) -> Result<*mut c_void, Py
...
+    /// A trivial real kernel (matches any single numeric arg, returns it
+    /// unchanged), exported to a `SedonaCScalarKernel` and wrapped in a real
+    /// `PyCapsule` -- the exact same export path `sedona-extension`'s own
+    /// `ffi_roundtrip`/`named_kernel` tests already prove correct end to
+    /// end. `#[dev-dependencies] pyo3 = { features = ["auto-initialize"] }`
+    /// is what makes `Python::attach` usable here at all: `extension-module`
+    /// (needed for the real wheel build) is only ever added by maturin's own
+    /// build flags, never by this crate's Cargo.toml, so it's never present
+    /// during `cargo test`.
+    fn capsule_with_named_kernel<'py>(
```
paleolimbot, line 276: This works too, but another test that would be good to implement is `__sedona_scalar_udf__` on our PyO3 function representation and check roundtrip behaviour.

python/sedonadb/src/udf.rs
```diff
@@ -207,6 +210,58 @@ pub fn sedona_scalar_udf<'py>(
     })
 }
 
+/// Build a [`PySedonaScalarUdf`] from one or more natively-compiled kernel
+/// capsules (see [`crate::import_from::import_sedona_ffi_scalar_kernel`]),
+/// instead of `sedona_scalar_udf`'s single Python-callable kernel. Real
+/// compiled Rust runs per invocation -- no GIL re-entry, no Python callback
```
paleolimbot, line 246: I am not sure we need this because of the `FunctionSet` registration behaviour that does this already (feel free to correct me if I missed that).

python/sedonadb/Cargo.toml
```diff
@@ -64,3 +64,15 @@ thiserror = { workspace = true }
...
+# `extension-module` (needed for the real cdylib/wheel build) is only ever
+# added by maturin's own build flags (see pyproject.toml's [tool.maturin]),
+# never by this Cargo.toml -- so `cargo test` builds this crate without it,
+# and `auto-initialize` here is what lets a #[test] embed and drive a real
+# Python interpreter via Python::attach. Confirmed directly: explicitly
+# compiling with `--features pyo3/extension-module` fails to link (undefined
+# libpython symbols), but plain `cargo test`/`cargo test --all-features`
+# (CI's actual invocation) never requests that feature, so the two never
+# collide in practice.
+pyo3 = { workspace = true, features = ["auto-initialize"] }
```
paleolimbot, line 78: As above, we should also make sure this is tested at the Python level by enabling these to get exported as well (more realistic)

python/sedonadb/python/sedonadb/context.py
```diff
@@ -490,6 +501,7 @@ def register(self, component: Any, **kwargs: Any) -> None:
             "__sedonadb_internal_aggregate_udf__",
             "__sedonadb_external_format__",
             "__sedonadb_raster_loader__",
+            "__sedonadb_native_scalar_udfs__",
```
paleolimbot: The individual function definition is probably a better candidate here (i.e., `__sedonadb_scalar_udf__` or `__sedonadb_native_scalar_udf__`), mostly because we already have a function object at the Python level (e.g., `sd.funcs.st_asbinary`) that we can implement it on to ensure it works.

paleolimbot (commented):
Thank you!

I think we should do individual instead of the list of functions and ensure we can export as well (which enables us to write a roundtrip test at the Python level), but this will be great!

python/sedonadb/src/udf.rs
james-willis: refactored so now `sedona_native_scalar_udf` is the place that kernels are all registered at once for a `SedonaScalarUDF`

python/sedonadb/src/context.rs
james-willis: Same response as below about refactoring it so all kernels register together.

I kept `register_sedona_scalar_udf` (replace) rather than `add_scalar_udf_impl` (append), to match `__sedonadb_internal_udf__` and keep registration predictable — append would silently merge a plugin's kernel into an existing same-name function.

python/sedonadb/src/import_from.rs
james-willis: added the test.

python/sedonadb/Cargo.toml
james-willis: done

python/sedonadb/python/sedonadb/context.py
james-willis: done

python/sedonadb/src/context.rs
paleolimbot: > append would silently merge a plugin's kernel into an existing same-name function.

In some cases this is what you want (e.g., add an implementation of `sd_format()` for your type to make it print nicely; add an overload of `st_intersects()` to handle your H3 type or whatever). Should there be an `overwrite=False` that defaults to the safe thing?

python/sedonadb/tests/test_udf.py
```diff
@@ -218,3 +218,37 @@ def questionable_udf(arg):
...
+def test_native_scalar_udf_export_import_roundtrip(con):
+    # A SedonaDB built-in scalar function exposes its native overload kernels
+    # via __sedonadb_scalar_udf__ (the export side of the plugin protocol).
+    # Exporting those capsules and rebuilding a UDF under a new name must
+    # produce a function whose output is identical to the built-in's, proving
+    # the native kernels survive a full export -> capsule -> import roundtrip.
+    from sedonadb._lib import sedona_native_scalar_udf
+
+    st_asbinary = con.funcs.st_asbinary._impl
+    assert hasattr(st_asbinary, "__sedonadb_scalar_udf__")
```
paleolimbot: Can you ensure this is also present on whatever object is returned by `con.funcs.st_asbinary` to avoid the `_impl` hop?

python/sedonadb/python/sedonadb/context.py
```diff
@@ -444,6 +444,18 @@ def register(self, component: Any, **kwargs: Any) -> None:
...
+          that one function, each wrapping a natively-compiled
+          SedonaCScalarKernel sharing the function's SQL name -- so real
+          compiled Rust runs per invocation, not a Python callback. Register
+          each function individually. The kernels are grouped into one
+          overloaded UDF; registering under a name already in use, including a
+          built-in's, replaces it, the same as __sedonadb_internal_udf__ already
+          does. Volatility is always Immutable through this protocol; a kernel
+          needing Volatile or Stable should be registered via
+          __sedonadb_internal_udf__ instead, built with
+          sedonadb._lib.sedona_native_scalar_udf(kernels, volatility=...).
```
paleolimbot: It's a little strange to reference a method from _lib here. Can you add this as a regular function in `udf.py`?

python/sedonadb/src/context.rs
```diff
@@ -354,6 +354,23 @@ impl InternalContext {
...
+            // One function's overload kernels, each a natively-compiled kernel
+            // capsule (see import_sedona_ffi_scalar_kernel) all sharing this
+            // function's SQL name. `sedona_native_scalar_udf` imports each
+            // capsule, checks the declared names agree, and groups them into
+            // one overloaded UDF. Registration replaces any existing UDF of
+            // that name outright, mirroring the existing __sedonadb_internal_udf__
+            // path (not an append). Volatility is Immutable via this path; a
+            // plugin needing Volatile/Stable builds a PySedonaScalarUdf with
+            // sedona_native_scalar_udf(..., volatility=...) and returns it
+            // through __sedonadb_internal_udf__ instead.
```
paleolimbot: An earlier comment suggested we don't clobber existing names. Does this comment need to be updated?

paleolimbot (commented):
Two loose ends worth wrapping up here...I think this is a good API (i.e., caller groups kernels instead of context) but there are references to internals/protected members that should get cleaned up first.

python/sedonadb/src/context.rs
james-willis: checked the code. append functions as a kernel-level overwrite anyway so my plan is to just adopt the "append" behavior. The only downside to this is that different implementors might have bizarrely divergent implementations but I suppose the user of the extensions should know that

paleolimbot: It's a good point that we can/should provide some safety around that at some point...the existing system was designed to optimize for flexibility (and explicitly support overlapping definitions to enable multiple implementations of st_intersects by various libraries while we benchmarked them). Extensions can check this themselves too (via `"foofy" in ctx.funcs` or similar).

python/sedonadb/src/context.rs
james-willis: updated

python/sedonadb/python/sedonadb/context.py
james-willis: done

python/sedonadb/tests/test_udf.py
james-willis: done

paleolimbot (approved):
Thank you!

PR #1155, opened 2026-08-12
fix(rust/sedona): Disable the default memory limit to avoid excessive spilling

### LINKED ISSUE #1040: SedonaDB 0.4.0 release breaks SpatialBench SF 1 Q12 and SF10 Q5 and Q12 due to No space left on device
See https://github.com/apache/sedona-spatialbench/actions/runs/28970392822?pr=123

* SF 1: Q12
* SF 10: Q5, Q12

This happens after 0.4.0 release.

```
Results saved to sedonadb_q12_results.json
  Running q12... ERROR: Io error: No space left on device (os error 28)
```

Is this caused by the default spilling index behavior? e.g., It spills too much unnecessarily and slows down the query as well.

jiayuasu:
## What changes are proposed in this pull request?

The 0.4.0 release enabled memory-limited execution by default: `SedonaContextBuilder::new()` set the memory limit to 75% of physical memory with a fair spill pool. Under this default, spilling operators can write large amounts of intermediate data to disk, which broke SpatialBench SF1 Q12 and SF10 Q5/Q12 with `No space left on device` errors and slowed down queries that previously ran fine.

This PR reverts the default to unlimited memory so spilling is opt-in:

- `SedonaContextBuilder::new()` no longer sets a memory limit by default; DataFusion's unbounded memory pool is used unless a limit is explicitly configured via the builder, the CLI `--memory-limit` flag, or context options. Setting a limit still enables the fair pool and spill-to-disk behavior exactly as before.
- Removed the now-unused `default_memory_limit()` helper and the `sysinfo` dependency.
- Updated the CLI help text, Python option docstrings, and the memory-management guide (notebook + generated markdown) to describe the opt-in behavior.

## How was this change tested?

- `cargo test -p sedona --lib context_builder` (19 tests, updated to assert the new default)
- `cargo test -p sedona --lib test_auto_configure_spilled_batch_threshold`
- `cargo test -p sedona --doc context_builder`
- `cargo check -p sedona-cli`
- pre-commit on all changed files

## Does this PR introduce any user-facing change?

Yes. Contexts no longer enforce a memory limit by default; memory-limited execution with spill-to-disk is now opt-in via `memory_limit`. This restores the pre-0.4.0 default behavior.

Closes #1040

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2651,16 +2651,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "dispatch2"
-version = "0.3.1"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "1e0e367e4e7da84520dedcac1901e4da967309406d1e51017ae1abfb97adbd38"
-dependencies = [
- "bitflags",
- "objc2",
-]
-
-[[package]]
 name = "displaydoc"
 version = "0.2.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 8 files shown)


paleolimbot (approved):
Thank you! I think this is a good move until DataFusion provides a few more primitives to handle managing spilling in complex workflows.

PR #1158, opened 2026-08-12
feat(rust/sedona-raster): trinary Override for BandOverrides; copy_into stops composing views

james-willis:


jiayuasu:
I think it makes sense to use a trinary representation for nodata, or an equivalent dedicated mechanism. However, converting name, outdb_uri, outdb_format, and view at the same time, and changing copy_into from visible-coordinate composition to raw-source replacement is not required for that fix.

Can we scope this PR down to trinary nodata handling?

james-willis:
I don't think that makes sense. The purpose of this PR is to clean up the internal mechanism for raster -> raster functions.

name, uri, format, and view are all fields where callers benefit from being able to use trinary operators to alter the new raster from the existing raster.

The implicity composing of views was a bad ux choice that can now be effectively fixed by the trinary operator so it makes sense to bundle it in here.

There's no benefit anyway to breaking these changes out into many PRs, so why create more developer churn

paleolimbot (approved):
Thank you!

Double checking that `NULL` -> "unset" for nodata is actually what we want. A sentinel extension type (e.g., `RS_SetBandNoDataValue(..., RS_NodataUnset())`) or separate function `RS_UnsetBandNoDataValue(...)` would be cleaner.

james-willis:
Null -> unset No Data Value iswhat I intended. I think its intuititve, round trips properly, and matches PostGis

james-willis:
im not sure about in DF/sedonadb but in a lot of sql engines (e.g. duckdb, spark) you can update the values of a field in a struct to NULL. Duckdb example:

```sql
SELECT struct_update(user_profile, age := NULL) AS updated_profile 
FROM my_table;
```

I'm not sure what RS_SetSRID would do.

In Sedona Spark, its _supposed_ (ie documented) to work this way but the function harness intercepts before we get to the function implementation and we return a null raster instead.

PR #1159, opened 2026-08-13
test(integration): add a SedonaDB vs Sedona Spark raster parity suite

james-willis:


python/sedonadb/tests/functions/test_rs_raster_out_spark_parity.py
```diff
@@ -0,0 +1,69 @@
...
+
+GDAL_TRANSFORM = (100.0, 2.0, 0.0, 500.0, 0.0, -3.0)
+BANDS, HEIGHT, WIDTH = 2, 6, 7
+
+# Each value is representable in its dtype, so it packs into the band exactly.
+BAND_NODATA = {"uint8": 200.0, "int32": -99999.0, "float64": -12345.5}
+
+
+@pytest.mark.parametrize("dtype", list(BAND_NODATA))
+def test_rs_setbandnodata_raster_out_spark_parity(dtype, tmp_path):
```
jiayuasu: Could we wire these new `*_spark_parity` tests into the dedicated CI selector? This currently reproduces as:

```bash
cd python/sedonadb
SEDONADB_RUN_SPARK_TESTS=true python -m pytest tests/functions -q -k "sedonaspark" -rX
# 3588 deselected; exit code 5
```

Since that step has `continue-on-error: true`, the job stays green without running any of these six cases. `-k "spark_parity"` or an explicit marker/module list would pick them up.

python/sedonadb/python/sedonadb/testing_spark.py
```diff
@@ -0,0 +1,176 @@
...
+        if env:
+            return env
+        import pyspark
+
+        major, minor = (int(part) for part in pyspark.__version__.split(".")[:2])
+        # Sedona publishes per-Spark-minor artifacts. A pyspark older than every
+        # published artifact is an error (a newer jar on an older runtime fails
+        # at class load); a pyspark newer than the newest artifact tries the
+        # newest, which usually loads — override the coordinates if it doesn't.
+        known = ("3.5", "4.0")
```
jiayuasu: Could we reject Spark 3.5 here? `_packages()` accepts it, but the scalar path later calls `DataFrame.toArrow()`, which was added in Spark 4.0.

```bash
python -m pip install "pyspark==3.5.*"
python - <<PY
import pyspark
from pyspark.sql import DataFrame
from sedonadb.testing_spark import SedonaSpark

print(pyspark.__version__)
print(SedonaSpark._packages())
print(hasattr(DataFrame, "toArrow"))
PY
# 3.5.x
# ...sedona-spark-shaded-3.5_2.12:1.9.0...
# False
```

So a 3.5 session can get through package selection and then fail in `result_to_table()`. Enforcing 4.0+ here, or adding a 3.5-compatible Arrow path, would keep the contract consistent.

python/sedonadb/tests/functions/test_rs_raster_out_spark_parity.py
james-willis: Intentionally this is NOT run in the CI. I think it could be a good idea to do so but perhaps running spark in the CI on each PR is a bit heavy. 

I do not have a strong opinion on this though; perhaps we can wait for Dewey to weigh in.

python/sedonadb/python/sedonadb/testing_spark.py
james-willis: done

.github/workflows/python.yml
```diff
@@ -150,42 +150,6 @@ jobs:
           cd python/sedonadb
           python -m pytest -vv
 
-      - uses: actions/setup-java@v5.7.0
```
james-willis: This got abandoned at some point. I removed it until we decide to run this harness as part of the CI

python/sedonadb/python/sedonadb/testing.py
```diff
@@ -375,6 +395,22 @@ def create_table_arrow(self, name, obj) -> "SedonaDB":
...
+
+    def decode_raster_result(self, sql):
+        from sedonadb.raster_testing import decode_raster
+
+        # RS_FromPath yields an OutDb (lazy) raster, so materialize the result
+        # before reading its pixels; Sedona Spark's RS_FromGeoTiff loads eagerly,
+        # so this keeps the two engines comparing the same bytes.
+        loaded = f"SELECT RS_EnsureLoaded(c) AS c FROM ({sql}) AS sub(c)"
+        table = self.execute_and_collect(loaded)
+        return decode_raster(table.column(0)[0])
```
paleolimbot, line 412: This seems a bit strange...don't you want the parity tests to extend to whether something stays an InDB raster or not?

paleolimbot (approved):
Thank you!

I'd slightly prefer the top level directory to be outside of `python/` (e.g, `integration/`, like `benchmarks/`). I can move the PostGIS integration tests there, too, so that `pytest` in `python/` is mostly testing Python code.

python/sedonadb/python/sedonadb/testing.py
james-willis: No I don't think it should. I consider that to be an internal detail of the system rather than a documented/stable behavior. similarly I don't think lazy cropping should be tested for parity.

PR #1161, opened 2026-08-13
refactor(rust/sedona-raster): extract BandArrayBuilder from RasterBuilder

james-willis:
## What

Extracts `RasterBuilder`'s per-band column state and writing methods into a standalone `BandArrayBuilder` (`band_builder.rs`) — one row per band, no raster envelope (`crs`, `transform`, `spatial_dims`/`spatial_shape`, per-raster band grouping). `RasterBuilder` now holds one internally and delegates, layering its per-raster bookkeeping on top.

`BandRef::copy_into`/`append_data_into` now target a new `BandWriter` trait object instead of `&mut RasterBuilder`. `RasterBuilder` implements it, so existing call sites are unaffected — but `copy_into` can now also target a bare `BandArrayBuilder`.

Pure internal refactor: no schema change, no new column, no behavior change. All of `RasterBuilder`'s public inherent signatures are unchanged.

## Why

A derived band with no enclosing raster can't currently reuse `copy_into`'s view-composition/override logic or the builder's column assembly without wrapping and unwrapping a throwaway single-band `Raster`. The read side (`ViewEntries::compose`, `BandRef` defaults) was already reusable; this closes the gap on the write side.

## Verification

- `sedona-raster`: 177 tests pass (175 existing unmodified, plus 2 new covering the bare-builder and `copy_into`-into-bare-builder paths).
- `sedona-raster-functions`, `-zarr`, `sedona-spatial-join{,-raster}`, `sedona-testing`: all pass, no regressions.
- `sedona-raster-gdal` unbuildable locally (pre-existing GDAL/`gdal-sys` mismatch, fails identically on clean `main`); its call sites use only unchanged inherent methods.
- clippy `-D warnings` + fmt clean.

### DIFF
--- a/rust/sedona-raster/src/array.rs
+++ b/rust/sedona-raster/src/array.rs
@@ -25,7 +25,7 @@ use datafusion_common::cast::{
     as_string_array, as_string_view_array, as_struct_array, as_uint32_array,
 };
 
-use crate::builder::RasterBuilder;
+use crate::band_builder::BandWriter;
 use crate::traits::{BandRef, NdBuffer, RasterRef};
 use crate::view_entries::{ViewEntries, ViewEntry};
 use sedona_schema::raster::{band_indices, band_view_indices, raster_indices, BandDataType};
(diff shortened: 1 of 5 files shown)


paleolimbot (approved):
Thank you!

PR #1162, opened 2026-08-13
feat(rust/sedona-functions): add the Sedona Spark ST_Force3DZ alias for st_force3d

james-willis:

Register `st_force3dz` as an alias of `st_force3d`. It is both the name Sedona Spark registers (`ST_Force3DZ` as an alias of `ST_Force3D`) and the canonical name in current PostGIS (`ST_Force3D` is the alias there), so SQL written against either resolves unchanged on sedona-db.

Per review, the `st_force_2d` alias originally in this PR was **dropped**: it is the pre-2.1 PostGIS name that PostGIS itself deprecated, and rather than porting it here it is now being deprecated in Sedona Spark with a warning giving the canonical name (apache/sedona#3264).

This follows the existing alias pattern used by `st_geomfromwkt` (`st_geomfromtext`, `st_geometryfromtext`, ...).

Spark SQL written against Sedona (e.g. pipelines using `ST_Force3DZ`) currently fails with an unknown-function error when run on sedona-db. The alias makes such SQL portable unchanged, and matches current PostGIS naming.

## How was this patch tested?

Extended the existing `udf_metadata` test in `st_force_dim.rs` to assert the alias; `cargo test -p sedona-functions st_force_dim` passes.

Yes — the alias is listed on the `st_force3d` SQL reference page (kept in sync with the registered aliases, as required by the docs check).

### DIFF
--- a/docs/reference/sql/st_force2d.qmd
+++ b/docs/reference/sql/st_force2d.qmd
@@ -29,6 +29,8 @@ kernels:
 
 Discards any Z and M values present (if any).
 
+Aliases: `ST_Force_2D`.
+
 ## Examples
 
 ```sql
--- a/docs/reference/sql/st_force2d.qmd
+++ b/docs/reference/sql/st_force2d.qmd
@@ -29,8 +29,6 @@ kernels:
 
 Discards any Z and M values present (if any).
 
-Aliases: `ST_Force_2D`.
-
 ## Examples
 
 ```sql
--- a/docs/reference/sql/st_force3d.qmd
+++ b/docs/reference/sql/st_force3d.qmd
@@ -39,6 +39,8 @@ kernels:
 
 If the geometry already has Z values they are preserved; otherwise the optional `z` argument (default 0) is used.
 
+Aliases: `ST_Force3DZ`.
+
 ## Examples
 
 ```sql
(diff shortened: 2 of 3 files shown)


james-willis:
I found these issues when reviewing a real production sedona application and trying to run it in sedonadb. 

So I would say not to remove them from sedona spark

james-willis:
Opened apache/sedona#3264 taking the deprecation route suggested here: both aliases now warn toward the canonical names on the Spark surface (one-time runtime warning for SQL users, `@deprecated` on the Scala DataFrame API, `DeprecationWarning` in Python, docs marked), with behavior unchanged.

One nuance flagged on that PR: `ST_Force3DZ` is actually the canonical name in current PostGIS (`ST_Force3D` is its alias there), unlike `ST_Force_2D` which PostGIS itself deprecated in 2.1. So the `ST_Force3DZ` half trades PostGIS name-compatibility for a single canonical name — open to dropping that half if PostGIS parity is preferred.

Should I modify this PR to only add ST_Force3DZ

james-willis:
Done — `st_force_2d` is dropped from this PR; only the `st_force3dz` alias remains (in both the code and the SQL reference page, keeping the docs check green). The branch is also rebased onto main. The Sedona-side deprecation of `ST_Force_2D` is up at apache/sedona#3264, now scoped to that name only since `ST_Force3DZ` is canonical PostGIS.

jiayuasu (approved):

PR #1163, opened 2026-08-13
feat(rust/sedona-functions): implement ST_GeoHash

james-willis:

Implement `ST_GeoHash(geometry, precision)` returning the base-32 geohash string of the geometry, matching Apache Sedona Spark semantics:

- Non-point geometries hash the **center of the bounding box** (GeometryGeoHashEncoder.java:40-41)
- Returns null when the bbox falls outside [-180,180]×[-90,90] (GeometryGeoHashEncoder.java:33-38)
- `precision <= 0` returns an empty string; precision caps at 20 (PointGeoHashEncoder.java:29-32)
- Standard lon-first bit interleaving with the `0123456789bcdefghjkmnpqrstuvwxyz` alphabet

The encoder is implemented by hand (~40 lines) — no new crate dependency.

Follows crate conventions: `SedonaScalarUDF::new` + `ItemCrsKernel::wrap_impl`, `WkbExecutor`, `ArgMatcher::is_geometry() + is_integer()` with per-row Int64-cast precision (modeled on `st_geometryn`), Utf8 return via `StringBuilder` (modeled on `st_geometrytype`).

Two deliberate deviations from Sedona Spark, documented in code:
1. **Empty geometries return null** — Sedona Java hits a JTS null-envelope artifact (`minx=0, maxx=-1`) and accidentally hashes empty geometries as (-0.5, -0.5); returning null is PostGIS-consistent.
2. **Geometry-only** (no geography kernel), matching Sedona Spark's signature; precedent: `st_xmin`/`st_ymin` are geometry-only for the same planar-bbox reason.

`ST_GeoHash` is used in Sedona Spark pipelines (e.g. for spatial bucketing/partitioning keys); sedona-db currently has no implementation, so such SQL fails with an unknown-function error.

## How was this patch tested?

8 new tests in `st_geohash.rs` with expected values taken from Sedona Spark's `TestStGeoHash.scala` (each vector cited in a comment), including WKB_VIEW_GEOMETRY and ITEM_CRS variants, precision capping, out-of-range nulls, and pole/antimeridian edge cases. `cargo test -p sedona-functions` passes (513 tests), clippy and fmt clean.

Function-level rustdoc included; happy to add a SQL reference doc page if maintainers point me at the right template.

### DIFF
--- a/python/sedonadb/tests/functions/test_functions.py
+++ b/python/sedonadb/tests/functions/test_functions.py
@@ -2032,6 +2032,61 @@ def test_st_geohash_accepts_wgs84_and_undeclared_crs(srid):
     eng.assert_query_result(sql, "s5x1g8cu2y")
 
 
+@pytest.mark.parametrize("lat", [100, 91, 270, -91, -100])
+def test_st_geohash_geography_rejects_out_of_range_latitude(lat):
+    # s2geography bounds a geography with an S2LatLngRect, whose latitudes are
+    # constrained to [-90, 90] by construction, so an out-of-range latitude is
+    # clamped inside S2 before the domain check can see it. Without the raw
+    # planar bound taken alongside it, every latitude here encoded as the pole:
+    # POINT (50 100), POINT (50 91) and POINT (50 270) all returned
+    # 'vpgxczbzuryp', which is the hash of the genuine POINT (50 90).
+    #
+    # Only reachable in a build with the s2geography feature; the Rust unit
+    # tests substitute the planar bounder, which does no clamping and so cannot
+    # exercise this at all.
+    if "s2geography" not in sedonadb.__features__:
+        pytest.skip("Geography bounds require a build with feature s2geography")
+
+    eng = SedonaDB.create_or_skip()
+    eng.assert_query_result(
+        f"SELECT ST_GeoHash(ST_GeogFromText('POINT (50 {lat})'), 12)", None
+    )
+
+
+def test_st_geohash_geography_keeps_the_poles_and_wrapping():
+    # The other side of the check above: latitudes that really are in range
+    # still encode, including the poles themselves, and longitude wrapping is
+    # untouched. Longitude deliberately keeps using the spherical bounds --
+    # S2 normalizes it into [-180, 180], which is the same wrapping applied to
+    # a geometry -- so 190 and -170 agree.
+    if "s2geography" not in sedonadb.__features__:
+        pytest.skip("Geography bounds require a build with feature s2geography")
+
+    eng = SedonaDB.create_or_skip()
+    eng.assert_query_result(
+        "SELECT ST_GeoHash(ST_GeogFromText('POINT (50 90)'), 12)", "vpgxczbzuryp"
+    )
+    eng.assert_query_result(
+        "SELECT ST_GeoHash(ST_GeogFromText('POINT (50 -90)'), 12)", "j0581b0bh2n0"
+    )
+    eng.assert_query_result(
+        "SELECT ST_GeoHash(ST_GeogFromText('POINT (190 50)'), 12)", "b0zh7w1z0gs3"
+    )
+    eng.assert_query_result(
+        "SELECT ST_GeoHash(ST_GeogFromText('POINT (-170 50)'), 12)", "b0zh7w1z0gs3"
+    )
+
+    # A non-point geography takes the same path through its bounding box: this
+    # linestring reaches latitude 100, so it is rejected rather than clamped.
+    eng.assert_query_result(
+        "SELECT ST_GeoHash(ST_GeogFromText('LINESTRING (50 80, 60 100)'), 12)", None
+    )
+    eng.assert_query_result(
+        "SELECT ST_GeoHash(ST_GeogFromText('LINESTRING (50 80, 60 85)'), 12)",
+        "vnxj7d9v2fsm",
+    )
+
+
 def _item_crs_table(eng):
     """A table whose geometry column carries a per-row (item-level) CRS
 
--- a/python/sedonadb/tests/functions/test_functions.py
+++ b/python/sedonadb/tests/functions/test_functions.py
@@ -1889,8 +1889,13 @@ def test_st_geohash_out_of_range_untagged_diverges_from_postgis(eng, geom):
     # calls lwerror() rather than returning NULL. Pinned on both sides so the
     # divergence is verified rather than merely asserted in a comment.
     #
-    # Tagging the same geometry with a lon/lat CRS changes the longitude cases;
-    # see test_st_geohash_wraps_longitude_for_lnglat_crs below.
+    # An undeclared CRS is accepted (see test_st_geohash_accepts_wgs84_and_
+    # undeclared_crs) but does not earn wrapping: accepting it is passive, while
+    # wrapping it would invent a location for coordinates of unknown provenance.
+    # It also keeps faith with Apache Sedona, which has no CRS concept at all, so
+    # every geometry migrated from Spark arrives here undeclared. Tagging the
+    # same geometry 4326 opts into wrapping; see
+    # test_st_geohash_wraps_longitude_for_lnglat_crs below.
     raises = eng == PostGIS
     sql = f"SELECT ST_GeoHash({geom_or_null(geom)}, 10)"
     eng = eng.create_or_skip()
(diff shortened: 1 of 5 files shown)


paleolimbot (commented):
Thank you!

Can you add integration tests to Python that verify this returns the same values as PostGIS?

rust/sedona-functions/src/st_geohash.rs
```diff
@@ -0,0 +1,360 @@
...
+        .map_err(|e| sedona_internal_datafusion_err!("Error computing bounds: {e}"))?;
+    let (x, y) = (bounds.x(), bounds.y());
+    if x.is_empty() || y.is_empty() {
+        return Ok(None);
+    }
+
+    // Longitude can take values in [-180, 180]; latitude can take values in [-90, 90]
+    if x.lo() < -180.0 || y.lo() < -90.0 || x.hi() > 180.0 || y.hi() > 90.0 {
+        return Ok(None);
+    }
```
paleolimbot: This would be better normalizing to the -180, 180 range and validating that we have lon/lat CRS (you could check what PostGIS does for this case...we have a "get geographic params" to check if this is a CRS we know is in longitude/latitude units, or maybe we only want this if it's bona-fide WGS84). This is easy for type level CRS but sort of a pain for item level CRS.

rust/sedona-functions/src/st_geohash.rs
```diff
@@ -0,0 +1,360 @@
...
+/// Follows Apache Sedona's GeometryGeoHashEncoder.calculate(): the point that
+/// is hashed is the center of the geometry's bounding box, and null is
+/// returned when the bounding box is not fully contained in
+/// [-180, 180] x [-90, 90]. Unlike Sedona's Java implementation (where an
+/// empty geometry yields JTS' "null envelope" and thus an accidental hash of
+/// (-0.5, -0.5)), empty geometries return null here.
+fn invoke_scalar(geom: impl GeometryTrait<T = f64>, precision: i64) -> Result<Option<String>> {
+    let bounds = geo_traits_bounds_xy(geom)
+        .map_err(|e| sedona_internal_datafusion_err!("Error computing bounds: {e}"))?;
+    let (x, y) = (bounds.x(), bounds.y());
```
paleolimbot: I believe you can use the bounder from the config options to support Geography here, too. BigQuery supports this but only for points...you can punt on the integration tests there since you need credentials to set them up.

rust/sedona-functions/src/st_geohash.rs
```diff
@@ -0,0 +1,360 @@
...
+    )
+}
+
+#[derive(Debug)]
+struct STGeoHash {}
+
+impl SedonaScalarKernel for STGeoHash {
+    fn return_type(&self, args: &[SedonaType]) -> Result<Option<SedonaType>> {
+        let matcher = ArgMatcher::new(
+            vec![ArgMatcher::is_geometry(), ArgMatcher::is_integer()],
```
paleolimbot: PostGIS and BigQuery also support the single arg versions of these (BigQuery only supports points; PostGIS returns the largest cell that completely contains the geometry or a level 20 cell for points). The non-point case with no specified level sounds hard so you could just error for it.

python/sedonadb/tests/functions/test_functions.py
```diff
@@ -1778,6 +1778,47 @@ def test_st_geometrytype(eng, geom, expected):
     eng.assert_query_result(f"SELECT ST_GeometryType({arg})", expected)
 
 
+@pytest.mark.parametrize("eng", [SedonaDB, PostGIS])
+@pytest.mark.parametrize(
+    ("geom", "precision", "expected"),
+    [
+        (None, 10, None),
```
paleolimbot, line 1785: I know it seems silly, but could you add all the empties (POINT EMPTY .... GEOMETRYCOLLECTION EMPTY) here?

rust/sedona-functions/src/st_geohash.rs
james-willis: postgis will error here, sedona spark wil return null. I kept the spark behavior here

I believe we already check for lon/lat CRS before this point

rust/sedona-functions/src/st_geohash.rs
james-willis: done

rust/sedona-functions/src/st_geohash.rs
james-willis: done

python/sedonadb/tests/functions/test_functions.py
james-willis: yes. done

rust/sedona-functions/src/st_geohash.rs
paleolimbot: If we've already checked for a lon/lat CRS, normalizing to the -180...180 range is less likely to introduce a dropped filter row by accident (while maintaining the property of leniency to avoid failures in the middle of large processing).

james-willis: I was wrong, we were not checking the CRS. fixed that + added the normalizing

rust/sedona-functions/src/st_geohash.rs
```diff
@@ -0,0 +1,1102 @@
...
+/// argument's edge type, so a geography is bounded on the sphere rather than
+/// in the plane.
+fn invoke_scalar(
+    geom: &Wkb,
+    precision: i64,
+    bounder: &mut dyn WkbBounder2D,
+    wrap: LongitudeWrap,
+) -> Result<Option<String>> {
+    bounder.clear();
+    bounder
```
jiayuasu, line 628: Could we validate the raw geography latitude before invoking the spherical bounder? I may be missing another check, but this reproduces for me with the S2-backed CLI:

```sql
SELECT
  ST_GeoHash(ST_GeogFromText('POINT (50 100)'), 12) AS invalid_lat,
  ST_GeoHash(ST_GeogFromText('POINT (50 90)'), 12) AS pole;
-- vpgxczbzuryp | vpgxczbzuryp
```

It looks like S2 clamps 100° to 90° before the check below, so the original out-of-range latitude is no longer visible. Based on the comment below, I expected `invalid_lat` to be `NULL`. A test with the real S2 bounder would probably catch this; the current helper substitutes the planar bounder.

rust/sedona-functions/src/st_geohash.rs
```diff
@@ -0,0 +1,1102 @@
...
+
+/// ST_GeoHash() scalar UDF
+///
+/// Native implementation to compute the geohash of a geometry or geography.
+/// The two-argument form hashes at the requested precision (number of base32
+/// characters); the one-argument form hashes a point at [MAX_PRECISION].
+pub fn st_geohash_udf() -> SedonaScalarUDF {
+    SedonaScalarUDF::new(
+        "st_geohash",
+        ItemCrsKernel::wrap_impl(vec![
```
jiayuasu: Could this preserve the per-row CRS when deciding whether to wrap longitude? This gives different results for the same EPSG:4326 value depending on how the SRID is represented:

```sql
WITH t(srid) AS (VALUES (CAST(4326 AS BIGINT)))
SELECT
  ST_SRID(ST_SetSRID(ST_GeomFromText('POINT (190 50)'), srid)) AS srid,
  ST_GeoHash(ST_SetSRID(ST_GeomFromText('POINT (190 50)'), srid), 12) AS item_crs,
  ST_GeoHash(ST_SetSRID(ST_GeomFromText('POINT (190 50)'), 4326), 12) AS type_crs
FROM t;
-- 4326 | NULL | b0zh7w1z0gs3
```

My guess is `ItemCrsKernel` strips the row CRS before `longitude_wrap_for_arg_type()` runs. Could we either inspect the CRS per row here or keep the wrapping policy consistent between the two representations?

rust/sedona-functions/src/st_geohash.rs
```diff
@@ -0,0 +1,1459 @@
...
+        // ST_GeoHash cannot use ItemCrsKernel::wrap_impl(): that wrapper strips
+        // the CRS before calling the inner kernel, which is what makes it free
+        // for functions like ST_Area() whose answer does not depend on the CRS.
+        // This one rejects a non-WGS84 CRS, so it needs to see it. The sibling
+        // kernels below are the same shape ST_SRID() and ST_CRS() use, one
+        // matching the plain types and one matching item_crs.
+        vec![
+            Arc::new(STGeoHashItemCrs {
+                matcher: ArgMatcher::new(
+                    vec![ArgMatcher::is_item_crs()],
```
jiayuasu, line 72: Could this matcher also check the type of the wrapped `item`? As written, any struct with fields named `item` and `crs` matches, even when `item` is not spatial. This reproduces for me:

```sql
SELECT ST_GeoHash(
  named_struct(
    'item', 1,
    'crs', arrow_cast('EPSG:4326', 'Utf8View')
  ),
  12
);
```

I get `SedonaDB internal error: Expected geometry or geography argument but got Arrow(Int64)`. I think this should fail during planning as a normal signature mismatch. Could `STGeoHashItemCrs::return_type()` unwrap the item-CRS type and return no match unless the item is geometry/geography? A regression test with this struct shape would cover it.

rust/sedona-functions/src/st_geohash.rs
james-willis: fixed

rust/sedona-functions/src/st_geohash.rs
james-willis: This is a structural issue of how `is_item_crs` is used in UDFs - ST_SRID, ST_CRS, and ST_EWKB use the same pattern.

I will raise a stacked PR on top of this one to fix. Will link when drafted

rust/sedona-functions/src/st_geohash.rs
james-willis: fixed + pinned with test

rust/sedona-functions/src/st_geohash.rs
james-willis: https://github.com/apache/sedona-db/pull/1194/changes/9239b9d02ce6d5855c11c0c186cfb74f93cf6402..3e5080ae368d2698fc55e289abd59958d32d3e58

rust/sedona-functions/src/st_geohash.rs
```diff
@@ -0,0 +1,1628 @@
...
+fn append_geohash_with_precision(
+    executor: &WkbExecutor<'_, '_>,
+    args: &[ColumnarValue],
+    bounder: &mut dyn WkbBounder2D,
+    raw_bounder: &mut Option<Box<dyn WkbBounder2D>>,
+    next_wrap: &mut dyn FnMut() -> Result<LongitudeWrap>,
+    builder: &mut StringBuilder,
+) -> Result<()> {
+    let precision_value = args[1]
+        .cast_to(&DataType::Int64, None)?
```
jiayuasu: Could we handle `UInt64` before this cast? I may be missing another coercion path, but this reproduces for me:

```sql
SELECT ST_GeoHash(
  ST_Point(1, 2),
  arrow_cast('18446744073709551615', 'UInt64')
);
```

I get `Can't cast value 18446744073709551615 to type Int64`. Since the signature accepts `UInt64` and precision above 20 is capped, I expected the same 20-character hash as precision 20. Maybe values could be clamped before narrowing, or `UInt64` could be excluded from the matcher if it is not meant to be accepted. A boundary test around `i64::MAX` would probably cover it.

james-willis: done

jiayuasu (approved):
LGTM!

PR #1167, opened 2026-08-14
fix(rust/sedona-spatial-join): Handle empty left join build side

### LINKED ISSUE #1166: Spatial LEFT join with an empty left side fails with an internal bitmap error
A spatial LEFT join whose left (preserved) side is empty fails with an internal error instead of returning zero rows.

## Reproduce

```python
import sedonadb
sd = sedonadb.connect()

left = sd.sql("SELECT ST_Point(0.0, 0.0) AS g, 1 AS x").alias("l")
left_empty = left.filter(left["x"] > 99)          # empty after filtering
right = sd.sql("SELECT ST_Point(0.0, 0.0) AS g2, 2 AS y").alias("r")

left_empty.join(right, on=left_empty["g"].geo.intersects(right["g2"]), how="left").count()
```

```
External error: SedonaDB internal error: The bitmap for visited left side is not created.
```

The expected result is an empty, schema-correct frame — which is what an ordinary (non-spatial) left join returns for an empty left side, and what GeoPandas' `sjoin(how="left")` returns.

The symmetric case works: a RIGHT join with an empty right side returns 0 rows without error. An INNER join with an empty side also works. So the failure is specific to the preserved-side bitmap not being initialized when the build side is empty in the spatial join executor.

## Impact

"Filter, then left-join" is a common pipeline shape, and whether it crashes depends on data — a filter that happens to match nothing turns a working query into an internal error. Because the message says "internal error", users have no indication their query was valid.

Found by GeoPandas-parity testing of `sedonadb-geopandas` (#1142); documented there as a known engine limitation pending this fix.

Version: `sedonadb` 0.5.0 (source build of current `main`).

jiayuasu:

Initialize the default spatial index's visited build-side bitmap when the build side produces no batches. The empty-index path now follows `GPUSpatialIndexBuilder`: outer joins receive a configured but empty bitmap list (`Some(empty)`), while join types that do not need final build-side output continue to use `None`.

The empty index constructor now accepts that bitmap state instead of hard-coding `None`.

This PR also adds:

- unit coverage verifying that an empty inner-join index has no visited bitmap and an empty left-join index has an empty visited bitmap
- an integration regression comparing an optimized spatial left join with a completely empty left side against `NestedLoopJoinExec`

When a spatial left join's preserved left side was empty, `DefaultSpatialIndexBuilder` returned through its empty-index path before creating the visited build-side bitmap. Outer-join finalization interpreted the resulting `None` as an unconfigured invariant and raised an internal error instead of returning an empty, schema-correct result.

This made filter-then-left-join pipelines fail whenever the filter happened to produce no rows.

Fixes #1166.

## How was this patch tested?

- `cargo fmt --all -- --check`
- `cargo test -p sedona-spatial-join`
  - 302 unit tests passed; 1 ignored
  - 180 integration tests passed
  - 3 doc tests passed

The new integration test exercises the complete empty-left relation and confirms that the optimized spatial join matches the nested-loop reference result.

No. This is a correctness fix for existing spatial left-join behavior and does not change the public API or documented semantics.

### DIFF
--- a/python/sedonadb/tests/test_sjoin.py
+++ b/python/sedonadb/tests/test_sjoin.py
@@ -68,6 +68,21 @@ def test_spatial_join(join_type, on):
         eng_postgis.assert_query_result(sql, sedonadb_results)
 
 
+def test_spatial_left_join_with_empty_left_side(con):
+    left = con.sql("SELECT ST_Point(0.0, 0.0) AS g, 1 AS x").alias("l")
+    left_empty = left.filter(left["x"] > 99)
+    right = con.sql("SELECT ST_Point(0.0, 0.0) AS g2, 2 AS y").alias("r")
+
+    result = left_empty.join(
+        right,
+        on=left_empty["g"].geo.intersects(right["g2"]),
+        how="left",
+    ).to_arrow_table()
+
+    assert result.num_rows == 0
+    assert result.schema.names == ["g", "x", "g2", "y"]
+
+
 def _plan_text(df):
     query_plan = df.to_pandas()
     return "\n".join(query_plan.iloc[:, 1].astype(str).tolist())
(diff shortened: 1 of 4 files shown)


paleolimbot (approved):
Thanks!

Did you want a higher level Python integration test in our test_sjoin.py for real world confirmation that an empty table on one side joined against a non-empty table does the right thing?

PR #1171, opened 2026-08-15
 feat(r/sedonadb): Add SQL translation for nested field 

### LINKED ISSUE #1169: r/sdplyr: translate nested fields in WHERE clause
I tried to reproduce this DuckDB SQL on Overture Maps with the sdplyr package. 

https://docs.overturemaps.org/getting-data/duckdb/

The below code works without errors, however, it seems `categories$primary == "pizza_restaurant"` doesn't work as expected (I hope this is translated as `categories.primary = 'pizza_restaurant'`); the result is 0-rows. I think this should at least raise an error if the expression is not supported.

```r
library(sedonadb)
library(sdplyr)

Sys.setenv(
  AWS_SKIP_SIGNATURE = "true",
  AWS_DEFAULT_REGION = "us-west-2"
)

places_url <- "s3://overturemaps-us-west-2/release/2026-07-22.0/theme=places/type=place/"
target_wkt <- wk::rct(
  xmin = -75,
  ymin = 40,
  xmax = -73,
  ymax = 41,
  wk::wk_crs_longlat()
)

places_df <- sd_read_parquet(places_url)

res <- places_df |>
  filter(
    categories$primary == "pizza_restaurant",
    sd_intersects(
      geometry,
      target_wkt
    )
  ) |> 
  collect()
```

yutannihilation:
Fix #1169 

Translate an R expression `a$b` as SQL `get_field(a, 'b')`. 

``` r
library(sdplyr)
#> ── Attaching sdplyr packages ───────────────────────────────────── 0.4.0.9000 ──
#> ✔ sedonadb  0.4.0.9000
#> ✔ sedonafns 0.4.0.9000
#> ✔ dplyr     1.2.0

df <- sedonadb::sd_sql(
  "SELECT named_struct('bar', named_struct('baz', 1)) AS foo"
)

df |> 
  filter(foo$bar$baz > 0) |>
  transmute(value = foo$bar$baz) |>
  collect()
#> # A tibble: 1 × 1
#>   value
#>   <dbl>
#> 1     1
```

<sup>Created on 2026-08-16 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

I was thinking the implementation would be simpler than this, but it seems `sd_eval_expr_inner()` needs to be executed recursively in order to support a mixed case like this.

```r
df |> 
  transmute(value = .fns$get_field(foo, 'bar')$baz)
```

### DIFF
--- a/r/sdplyr/tests/testthat/test-filter.R
+++ b/r/sdplyr/tests/testthat/test-filter.R
@@ -83,6 +83,20 @@ test_that("filter() handles NA values", {
   )
 })
 
+test_that("filter() supports nested struct fields", {
+  df <- sedonadb::sd_sql(
+    "SELECT named_struct('bar', named_struct('baz', 1)) AS foo"
+  )
+
+  expect_identical(
+    df |>
+      filter(foo$bar$baz > 0) |>
+      transmute(value = foo$bar$baz) |>
+      collect(),
+    tibble(value = 1)
+  )
+})
+
 test_that("filter(..., .by) is unsupported", {
   df <- tibble(x = 1:3, g = c("a", "a", "b"))
   expect_snapshot_error(
--- a/r/sedonadb/R/expression.R
+++ b/r/sedonadb/R/expression.R
@@ -110,6 +110,23 @@ sd_expr_scalar_function <- function(function_name, args, factory = sd_expr_facto
   factory$scalar_function(function_name, args_as_expr)
 }
 
+# Construct access to a named field of a nested expression.
+sd_expr_get_field <- function(expr, field_name, factory) {
+  if (!is_sd_expr(expr)) {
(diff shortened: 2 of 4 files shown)


paleolimbot (approved):
Thank you!

For what it's worth I did go with a "custom evaluation" approach which made this a little awkward, and possibly I should have started with a "manipulate the environment and mask and use regular evaluation" approach (like arrow and maybe others).

yutannihilation:
Thanks!

Btw, I was wondering if there's some reason not to build on dbplyr. Maybe something related to interoperability with Arrow...?

yutannihilation:
> Other than that, it seemed like a bit of a waste to compile expressions into SQL and then parse the SQL immediately. DataFusion has a great logical plan/expression API

Ah, thanks, this makes sense to me!

PR #1172, opened 2026-08-16
docs: Add R-multiverse status badge to the main README

yutannihilation:
I think it would be nice if the main README also has a badge to the R package. This pull request adds the same one as the sedonadb R package's README.


paleolimbot (approved):
Thank you!

PR #1182, opened 2026-08-22
docs(dev/release): Document crates.io, PyPI, and verification pitfalls

jiayuasu:
## What changes are included in this PR?

Release-manager notes from the 0.4.1 release, added to `dev/release/README.md`:

- Verification: the Python comparison tests connect to PostGIS on `localhost:5432`; start the pinned `docker compose` service and make sure no other PostgreSQL is listening there.
- GitHub release notes: use the previous release tag as the comparison base (the patch-release text referenced a `.dev` tag that does not exist).
- PyPI: pure-Python packages use Metadata-Version 2.5 and need `twine>=7`; `~/.pypirc` must list `pypi` in `index-servers`; use `--skip-existing` for resumable uploads.
- crates.io: fix the script name (`publish-crates.sh`), publish from the release tag with the s2geography submodules initialized, list the dev-dependency cycles that must be removed locally, the `GDAL_VERSION=3.12.0` workaround for GDAL 3.13, and adding the committers team as owner of new crates.

## Are these changes tested?

Documentation only; pre-commit passes.

No.


dev/release/README.md
```diff
@@ -74,6 +74,12 @@ enable runtime binding generation:
 export SEDONADB_CARGO_TEST_ARGS="--features bindgen"
 ```
 
+The Python test suite includes comparison tests that connect to a PostGIS instance at
+`postgresql://localhost:5432/postgres?user=postgres&password=password` when one is
+reachable. Start the pinned version with `docker compose up -d postgis` from the
+repository root before verifying; an unrelated local PostgreSQL listening on port 5432
+(e.g., an older PostGIS that lacks `ST_HasM`) will cause these tests to fail.
```
paleolimbot: Optional, just making sure it's clear it's not strictly needed:

```suggestion
The Python test suite includes comparison tests that connect to a PostGIS instance at
`postgresql://localhost:5432/postgres?user=postgres&password=password` when one is
reachable.  A live PostGIS on localhost is not needed for successful verification;
however, those who wish to run these extra tests can start the pinned version
with `docker compose up -d postgis` from the repository root before verifying; 
an unrelated local PostgreSQL listening on port 5432 (e.g., an older PostGIS that
lacks `ST_HasM`) will cause these tests to fail.
```

dev/release/README.md
```diff
@@ -242,26 +248,54 @@ popd
...
+If publishing stops partway, fix the problem and resume with
+`--start-from <crate-name>`.
+
+`cargo publish` resolves `[dev-dependencies]` that carry a version requirement, so
+dev-dependencies that form a cycle with a crate published later must be removed from
+the local checkout before publishing (do not commit these changes). As of 0.4.1 these are
+`sedona-testing` in `sedona-geometry`, `sedona-gdal`, `sedona-geo-generic-alg`,
+`sedona-raster`, `sedona-expr`, and `sedona-functions`; `sedona-proj` in
+`sedona-functions`; and `sedona` in `sedona-pointcloud`, `sedona-tg`, `sedona-geos`,
+`sedona-geoarrow-c`, and `sedona-s2geography`.
```
paleolimbot, line 295: This is a bit of a hack and we're actively working on fixing this (I should have put the issue reference to https://github.com/apache/sedona-db/pull/1025 and https://github.com/apache/sedona-db/issues/702 ). I think in general this is frowned upon (modifying sources after a release vote / tag).

paleolimbot (approved):
Thank you!

dev/release/README.md
jiayuasu: Thanks, applied. The suggestion picked up a trailing space that tripped the pre-commit check; fixed in c6c4458.

dev/release/README.md
jiayuasu: Agreed that it is a hack. I reworded the paragraph to reference #702 and #1025 and to frame this as a stopgap until the cycles are removed, with the note that the modifications are local-only and the published crate contents are otherwise identical to the voted release. Happy to drop the paragraph entirely once #1025 lands.

Issue #1165, opened 2026-08-14
ST_Touches and ST_Within misclassify boundary-only line configurations


jiayuasu:
Two boundary-only configurations produce wrong answers from the native topology predicates. Both disagree with Shapely/GEOS and PostGIS, and both are silent wrong rows rather than errors.

## 1. `ST_Touches` misses a line whose interior passes through a polygon corner

```sql
SELECT
  ST_Touches(
    ST_GeomFromText('POLYGON ((0 0, 2 0, 2 2, 0 2, 0 0))'),
    ST_GeomFromText('LINESTRING (1 3, 3 1)')) AS touches,
  ST_Intersects(
    ST_GeomFromText('POLYGON ((0 0, 2 0, 2 2, 0 2, 0 0))'),
    ST_GeomFromText('LINESTRING (1 3, 3 1)')) AS intersects;
```

```
touches    = false
intersects = true
```

The line passes through the polygon's corner `(2 2)` without entering the interior — the intersection is the single boundary point, so `touches` should be `true` (Shapely returns `true`). `ST_Intersects` correctly sees the intersection, so the pair is classified as intersecting-but-not-touching, which no DE-9IM class supports for this geometry pair.

Notably, a line whose *endpoint* meets the corner (`LINESTRING (2 2, 3 3)`) is handled correctly — the failure is specific to the intersection point lying in the line's interior.

## 2. `ST_Within` wrongly matches a line lying on a hole boundary

```sql
SELECT ST_Within(
  ST_GeomFromText('LINESTRING (3 3, 3 7)'),
  ST_GeomFromText('POLYGON ((0 0, 10 0, 10 10, 0 10, 0 0), (3 3, 7 3, 7 7, 3 7, 3 3))'));
-- true; Shapely and PostGIS: false
```

The line lies entirely on the boundary of the polygon's hole. `within` requires the interiors to intersect (DE-9IM `T*F**F***`), and a line contained in the boundary never intersects the interior, so the correct answer is `false`.

## Scope checked

These neighbouring configurations are all correct: line endpoint at a corner (`touches`), line along an outer edge (`touches`), polygons sharing a corner (`touches`), point at a corner (`touches`), point inside a hole, point on a hole boundary, line across a hole, and polygon inside a hole (`within`/`contains`/`covers`/`covered_by`). So the defects look like specific gaps in boundary-interior classification rather than a broad relate problem.

## Impact

[...]

Issue #1093, opened 2026-07-27
ST_Union_Agg returns NULL for point inputs


jiayuasu:
`ST_Union_Agg` returns `NULL` when aggregating point geometries, instead of the expected unioned `MULTIPOINT`. It works correctly for polygons, and `ST_Collect_Agg` works correctly for points — so the issue appears specific to `ST_Union_Agg` with point inputs.

## Reproduce

```sql
SELECT ST_AsText(ST_Union_Agg(g))
FROM (VALUES (ST_Point(0.0, 0.0)), (ST_Point(1.0, 1.0))) AS t(g);
-- actual:   NULL
-- expected: MULTIPOINT((0 0),(1 1))
```

## Works as expected (for contrast)

`ST_Collect_Agg` on the same points:

```sql
SELECT ST_AsText(ST_Collect_Agg(g))
FROM (VALUES (ST_Point(0.0, 0.0)), (ST_Point(1.0, 1.0))) AS t(g);
-- MULTIPOINT((0 0),(1 1))
```

`ST_Union_Agg` on polygons (merges correctly):

```sql
SELECT ST_AsText(ST_Union_Agg(g))
FROM (VALUES
  (ST_GeomFromText('POLYGON((0 0,2 0,2 2,0 2,0 0))')),
  (ST_GeomFromText('POLYGON((1 1,3 1,3 3,1 3,1 1))'))
) AS t(g);
-- MULTIPOLYGON(((0 2,0 0,2 0,2 1,3 1,3 3,1 3,1 2,0 2)))
```

Only point inputs to `ST_Union_Agg` return `NULL`.

Found while writing a "Migrating from GeoPandas" dissolve example (GeoPandas `dissolve` unions each group's geometry, which for points yields a `MULTIPOINT`).

paleolimbot:
The current implementation uses Geo and is only for polygons, I believe. I think we need to move to a GEOS-based implementation to properly handle all geometry types.

PR #1184, opened 2026-08-23
feat(python/sedonadb-geopandas): add column assignment, arithmetic, and dissolve

jiayuasu:
The unblocked half of #1142, split out so it can move forward while the spatial-join engine issues are being fixed. `sjoin` stays in #1142, which is blocked on #1165 (`ST_Touches`/`ST_Within` boundary misclassification) and #1166 (empty preserved side fails in the spatial join executor); nothing here touches those code paths.

```python
gdf = sgpd.from_geopandas(points)
gdf["density"] = gdf["pop"] / gdf["area"]   # assignment + true division
zones = gdf.dissolve(by="region")           # group and union geometry
```

## What is added

- **`__setitem__`** — assign a `Series` from the same frame, or a scalar (a geometry included). Frame provenance is enforced: a `Series` from another frame, a bare expression (which records no origin), and multi-element values are rejected with focused errors. Replacing the active geometry with a non-geometry clears it; assigning geometry to a frame without one activates it; a geometry scalar or missing sentinel (`None`, NaN, `pandas.NA`) keeps the column's type and CRS, and a literal carrying its own CRS keeps that rather than being relabeled.
- **Arithmetic on `Series`** — `+`, `-`, `*`, `/` and the reflected forms. `/` is true division as in pandas: integer-typed expressions are cast to double (decided from the projected schema, a plan build) so SQL integer division does not truncate, while decimals stay exact and durations stay durations. `//` is deliberately not implemented, since SQL division truncates toward zero where Python floors. `Series` opts out of NumPy ufunc dispatch so an array operand is rejected whole instead of being broadcast element-by-element into an object array.
- **`dissolve(by, aggfunc="first", dropna=True)`** — unions each group's geometry via collect + unary union, which handles every geometry type (`ST_Union_Agg` only initializes for polygonal input, #1093). `dropna` covers IEEE NaN as well as SQL null, including dictionary-encoded float keys. A group whose geometries are all null yields an empty geometry collection carrying the source CRS, matching GeoPandas.
- **Shared scalar machinery** — one classification used by operators and assignment so they cannot disagree, with normalization: 0-d NumPy arrays unwrap to their value, `pandas.NA` becomes SQL null, NaN stays a float value.
- **Stale-mask guard** — boolean indexing rejects a mask built from another frame, including a mask captured before an assignment rebound this one; previously that could silently filter against the wrong frame state.

## Documented differences from GeoPandas

- `aggfunc="first"` is an unordered aggregate: it returns some value from the group and does not skip missing values.
- Dissolving an empty frame without a key returns one row (empty geometry collection, null attributes) rather than zero rows — a grouping-free SQL aggregate's shape.
- A group mixing 2D and 3D geometries raises rather than being promoted to 3D.

## Tests

73 tests, asserted against GeoPandas where a comparison exists (dissolve geometry equality across point/line/polygon groups, aggregation values, CRS propagation, division exactness for decimals, and the full assignment matrix including every missing-value sentinel). `ruff format` and `ruff check` clean.

### DIFF
--- a/python/sedonadb-geopandas/README.md
+++ b/python/sedonadb-geopandas/README.md
@@ -67,12 +67,12 @@ deliberately *not* identical to GeoPandas:
 Division follows pandas rather than SQL: `/` is true division, so integer
 columns do not silently truncate. `//` is not implemented, since SQL division
 truncates toward zero where Python floors. Duration arithmetic whose row
-results overflow the 64-bit tick range produces missing values (`NaT`) rather
-than raising the way pandas does — a lazy expression cannot raise for
-individual rows. (pandas clamps finite positive float overflow to
-`Timedelta.max`, a casting artifact this layer treats as overflow instead. An
-integer operand that itself exceeds the 64-bit range raises `OverflowError`,
-as in pandas.)
+results overflow the 64-bit tick range produces missing values (`NaT`) — a
+lazy expression cannot raise for individual rows the way pandas 3 does, and
+pandas 2 silently wraps integer overflow instead. (pandas also clamps finite
+positive float overflow to `Timedelta.max`, a casting artifact this layer
+treats as overflow. An integer operand that itself exceeds the 64-bit range
+raises `OverflowError`, as in every pandas version.)
 
 `dissolve()` aggregates non-geometry columns with `"first"`, which is an
 unordered aggregate: it returns *some* value from the group rather than the one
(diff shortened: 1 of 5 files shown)


jiayuasu:
Closing this in favor of a split into focused PRs, since the combined change grew too large to review well. The first slice (column assignment) is #1195; dissolve, arithmetic, and temporal support follow as their predecessors merge.

Issue #1092, opened 2026-07-26
Implement GDAL reading using sedona-gdal bindings


paleolimbot:
Currently we implement GDAL reading via OGR using pyogrio's Python interface. This leads to some Python calls that would be nice to avoid...among other things, when we have Python in multiple places in the plan (e.g., udfs, writes that use Python) we can get a hang. Even though we can probably fix those hangs otherwise, moving GDAL reads and/or writes into Rust means we can get them for free in SQL and R.

Roughly, we need to implement the Arrow interface here:

https://github.com/apache/sedona-db/blob/e3f57ff4714706a5839cce33537d0deccbcf6e7d/c/sedona-gdal/src/vector/layer.rs#L30-L35

...and wire up constructing the dataset and layer like pyogrio does. This is pretty involved, and handles the predicate pushdown (bbox, "ignored fields", limit).

https://github.com/geopandas/pyogrio/blob/main/pyogrio/_io.pyx#L1908-L1929

We currently implement the ExternalFormatSpec via Python...we just need to implement the ExternalFormatSpec directly in Rust.

PR #1187, opened 2026-08-23
fix(python/sedonadb): serialize pyogrio reader lifecycles

### LINKED ISSUE #1072: Python wheel builds are failing with handler timeout

jiayuasu:
## What

- add an `ExternalFormatSpec` capability for formats that cannot safely consume multiple file readers concurrently
- disable concurrent file reads for pyogrio and serialize schema inference and the complete reader lifecycle within each physical scan
- drop the Arrow FFI stream before closing its owning Python context and releasing the scan permit
- add regression tests for scan-local serialization and deterministic Python-context cleanup

## Why

#1051 serialized pyogrio context entry and exit, but active Arrow streams could still overlap. Subsequent wheel runs continued to abort during multi-file and partitioned pyogrio reads, as tracked in #1072.

An unchanged CPython 3.12 comparison repeatedly ran the two affected tests in fresh processes and reproduced SIGABRT at iteration 72 during `test_read_ogr_partitioned`: [run 32625781579](https://github.com/jiayuasu/sedona-db/actions/runs/32625781579/job/97160805281).

This change keeps one scan-local permit from reader creation through consumption and native cleanup. Pyogrio multi-file scans therefore trade file-level parallelism for stability. Other external formats retain concurrent reads by default, and independent physical scans do not share a process-wide lock.

## Verification

- `cargo fmt --all -- --check`
- `cargo test -p sedona-datasource -p sedonadb --lib` (9 + 11 tests passed)
- `cargo clippy -p sedona-datasource -p sedonadb --lib --no-deps -- -D warnings`
- local pyogrio and external-read modules (22 tests passed)
- pinned CPython 3.12 stress run:
  - full pyogrio module: 18/18 passed
  - 500 same-process repetitions and 500 fresh-process repetitions
  - 2,002 executions of the two affected tests with no abort, hang, timeout, or test failure
  - [run 32628586406](https://github.com/jiayuasu/sedona-db/actions/runs/32628586406)
- all five core wheel platforms passed on the combined verification branch: [run 32627022432](https://github.com/jiayuasu/sedona-db/actions/runs/32627022432)

Closes #1072.

Follow-up to #924 and #1051. Replacing the Python reader with the native GDAL reader remains tracked in #1092.

### DIFF
--- a/python/sedonadb/src/datasource.rs
+++ b/python/sedonadb/src/datasource.rs
@@ -52,6 +52,9 @@ pub struct PyExternalFormat {
     /// `False`); we snapshot it once to avoid GIL traffic in
     /// `list_single_object()`, which is called on hot paths.
     list_single_object: bool,
+    /// Cached from the Python spec so physical scan planning does not need the
+    /// GIL merely to decide whether file readers may overlap.
+    supports_concurrent_file_reads: bool,
     py_spec: Py<PyAny>,
 }
 
@@ -60,6 +63,7 @@ impl Clone for PyExternalFormat {
         Python::attach(|py| Self {
             extension: self.extension.clone(),
             list_single_object: self.list_single_object,
+            supports_concurrent_file_reads: self.supports_concurrent_file_reads,
             py_spec: self.py_spec.clone_ref(py),
         })
     }
@@ -78,9 +82,12 @@ impl PyExternalFormat {
             .getattr(py, "extension")?
             .extract::<String>(py)?;
         let new_list_single_object = read_list_single_object(py, &new_py_spec)?;
+        let new_supports_concurrent_file_reads =
+            read_supports_concurrent_file_reads(py, &new_py_spec)?;
         Ok(Self {
             extension: new_extension,
             list_single_object: new_list_single_object,
+            supports_concurrent_file_reads: new_supports_concurrent_file_reads,
             py_spec: new_py_spec,
         })
     }
@@ -138,8 +145,10 @@ impl PyExternalFormat {
         )?;
 
         let reader = import_arrow_array_stream(py, reader_obj.bind(py), None)?;
+        let schema = reader.schema();
         let wrapped_reader = WrappedRecordBatchReader {
-            inner: reader,
+            inner: Some(reader),
+            schema,
             shelter: Some(reader_obj),
         };
         Ok(Box::new(wrapped_reader))
@@ -152,9 +161,11 @@ impl PyExternalFormat {
     fn new<'py>(py: Python<'py>, py_spec: Py<PyAny>) -> Result<Self, PySedonaError> {
         let extension = py_spec.getattr(py, "extension")?.extract::<String>(py)?;
         let list_single_object = read_list_single_object(py, &py_spec)?;
+        let supports_concurrent_file_reads = read_supports_concurrent_file_reads(py, &py_spec)?;
         Ok(Self {
             extension,
             list_single_object,
+            supports_concurrent_file_reads,
             py_spec,
         })
     }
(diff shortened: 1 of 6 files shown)


jiayuasu:
@paleolimbot Do you think this makes sense at all?

python/sedonadb/src/datasource.rs
```diff
@@ -400,25 +427,99 @@ impl PyProjectedRecordBatchReader {
...
+        let (shelter, shelter_class) = Python::attach(|py| {
+            let module = PyModule::from_code(
+                py,
+                cr#"
+class Shelter:
+    destroyed = 0
+
+    def __del__(self):
+        type(self).destroyed += 1
+"#,
```
paleolimbot, line 524: If there's any way to test this in Python instead of PyO3 I think it would be cleaner.

rust/sedona-datasource/src/format.rs
```diff
@@ -216,6 +226,10 @@ impl FileFormat for ExternalFileFormat {
 #[derive(Debug, Clone)]
 struct ExternalFileSource {
     spec: Arc<dyn ExternalFormatSpec>,
+    /// Shared by all openers cloned from this physical scan. Keeping the lock
+    /// here scopes serialization to one scan instead of coupling independent
+    /// user streams through a process-wide lock.
+    reader_lock: Option<Arc<Mutex<()>>>,
```
paleolimbot, line 232: For the pyogrio case, this may not be sufficient: the serialization probably has to occur process-wide (so two Python threads doing two scans may still crash?).

paleolimbot (approved):
Thank you!

We should really fix the underlying cause here (use our attached GDAL to do all of this directly without invoking Python), although we might need this anyway. We also need to revisit the ExternalFormatSpec to get it FFIed so we can add datasource packages.

I left a comment about this maybe not being sufficient for all cases (but if it solves the more common case of fixing current crashes, it's still worth it).

python/sedonadb/src/datasource.rs
jiayuasu: Good point. I added Python integration coverage for the observable behavior. It now covers serialization within one scan, concurrent multibatch scans, and paused reader progress for FlatGeoBuf and GeoPackage.

I kept the two narrow Rust tests because they verify an FFI ownership invariant that Python cannot reliably observe: the imported Arrow stream must be released before its Python shelter at EOF and early cancellation. The user visible behavior is covered in Python. The Rust tests only pin that cleanup order.

rust/sedona-datasource/src/format.rs
jiayuasu: Thanks, you were right that the previous coexistence test was not enough. The committed Python test uses two threads, two connections, and two distinct datasets. It rendezvouses immediately before every real pyogrio backed read_next_batch call, requires multiple successful pulls from both readers, and validates every row for FlatGeoBuf and GeoPackage. It runs in a spawned child process with a hard timeout and exit code reporting.

I initially asserted overlap using timestamps around each pull. A macOS arm64 wheel run showed why that is not a sound oracle: after the barrier releases, one very fast pull can finish before the scheduler runs the other thread. Both scans and all rows were correct. I removed only that timing assertion. The per batch rendezvous and real reads remain.

I also added a progress test. Reader A consumes one batch and remains alive. Reader B completes, then reader A resumes. A temporary process wide lifetime lock mutation makes this test time out, so holding one global mutex for the full lifetime of a lazy reader would deadlock valid usage. That is why this PR keeps lifecycle serialization scoped to one scan.

I also stress tested the original FlatGeoBuf path locally with open versus read and close versus read overlap. I left that timing harness out of the PR because it added substantial test complexity and its evidence is specific to the tested build.

The final full wheel run at 64c6d3fca passed all five core platforms and all five Zarr extension platforms: https://github.com/jiayuasu/sedona-db/actions/runs/33616709019

This evidence is limited to the tested drivers and builds. It is not a claim that every GDAL driver is universally safe. If a future wheel run reproduces a crash under coordinated independent reads, the next design should serialize individual native operations across the process rather than hold a lock for the reader lifetime.

paleolimbot (approved):
Thank you! I am going to try to work on datasources after the DataFusion 55 update and hopefully in the next release or two we can make the GDAL read more robust and also move off of it for the most common file types (fgb, gpkg, shapefile, geojson).

PR #1188, opened 2026-08-23
fix(ci): fail wheel builds immediately when tests fail

jiayuasu:
## What changes are included in this PR?

Wheel tests currently suppress pytest's exit code, continue building the remaining wheels, upload artifacts, and report the failure later through a `.test_failed` marker. This change makes test failures propagate directly from pytest to cibuildwheel on every platform.

Specifically, it:

- runs `python -m pytest` directly so a failed test stops cibuildwheel immediately
- explicitly exits with `$LASTEXITCODE` after the Windows build script so the cibuildwheel status is preserved across the PowerShell invocation
- removes marker creation, cleanup, and final marker checks
- retains the `always()` vcpkg cache-save steps, allowing binaries produced before a failure to be reused on a cache miss, while leaving wheel artifact uploads success-only
- restores the default success gate for the expr, zarr, geopandas, and nightly upload jobs so they cannot consume or publish a partial core wheel set

The marker approach deliberately returns success to cibuildwheel, so a failure can take tens of minutes to surface while later Python versions continue building. It also requires platform-specific coordination across host, container, and PowerShell boundaries. Returning pytest's real status is simpler and prevents failed or incomplete builds from producing wheel artifacts or starting dependent jobs.

## Are these changes tested?

- [Full fork matrix](https://github.com/jiayuasu/sedona-db/actions/runs/32626533696): all five core wheel jobs passed on Windows x86_64, macOS arm64/amd64, and Linux x86_64/arm64.
- [Intentional failure run](https://github.com/jiayuasu/sedona-db/actions/runs/32629325994): an injected pytest failure returned code 1 through cibuildwheel; the wheel artifact step was skipped, the run produced zero artifacts, and expr, zarr, geopandas, and nightly upload were all skipped.
- workflow YAML parse and `git diff --check`

No. This only changes wheel CI failure handling.

### DIFF
--- a/.github/workflows/python-wheels.yml
+++ b/.github/workflows/python-wheels.yml
@@ -84,10 +84,10 @@ jobs:
 
       - name: Build and test wheels (sedonadb)
         run: |
-          Remove-Item -Force .test_failed -ErrorAction SilentlyContinue
           New-Item -ItemType Directory -Force -Path "$env:VCPKG_DEFAULT_BINARY_CACHE" | Out-Null
           cd ci/scripts
           .\wheels-build-windows.ps1
+          if ($LASTEXITCODE -ne 0) { exit $LASTEXITCODE }
         env:
           VCPKG_ROOT: ${{ github.workspace }}/vcpkg
           VCPKG_DEFAULT_TRIPLET: x64-windows-dynamic-release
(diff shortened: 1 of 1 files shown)


jiayuasu:
Thanks, that context is helpful. I agree the lazy-failure behavior has intentional value: it preserves wheels for debugging and lets extension/nightly work continue when a core wheel flakes. I will keep #1188 as a draft rather than treating removal of that behavior as the long-term fix.

The main issue I traced is the pyogrio/GDAL concurrent-reader abort. #1187 is the proposed tactical fix: it serializes the full pyogrio reader lifecycle within each physical scan, and the original reproducer (which aborted at iteration 72) completed 2,002 affected test executions after the change. The native GDAL reader in #1092 remains the durable direction.

I am not yet certain that the scan-local locking approach is directionally what we want, so I would appreciate your take on #1187. If it is, I will focus on getting #1187 in instead of moving #1188 forward.

james-willis:
Making the wheel build step durable to test failures and build failures was important when we had a lot of xplat and flakey build issues a month or so ago. I think it is preferable that we publish nightlies that might be screwed up or missing some platforms rather than publish nothing.

We had some hard to debug failures in e.g. the linux wheel test & build that were blocking release of all nightlies. Without these changes there were no nightlies to work with until that fix was done. I think ultimately wheel builds were failing for more than a month and the thing that unblocked them was making these failures ignorable. It took another couple weeks of effort from me to get the builds back to all green on all platforms.

PR #1192, opened 2026-08-26
fix(rust/sedona-raster): carry band names through copy_raster_from

james-willis:
## Problem

`RS_SetGeoReference(raster, georef)` returns a raster whose band names are all NULL. User-visible for named bands — e.g. Zarr variable names like `temperature`.

Confirmed on `main` (8e0a26328):

```
SOURCE band_name = Some("temperature")
COPIED band_name = None
```

## Root cause

Band name lives on `RasterRef::band_name(index)`. **`BandRef` had no name accessor**, so `copy_into` could not inherit the name the way it inherits `nodata` / `outdb_uri` / `outdb_format` — every caller had to remember to thread the name through by index. `copy_raster_from` didn't, and `RS_SetGeoReference` is its only production caller.

## Two commits

**1. `fix:` carry band names through `copy_raster_from`** — pass `source.band_name(band_idx)` explicitly. Minimal, backportable.

**2. `refactor:` add `BandRef::name()` so `copy_into` inherits it** — removes the cause rather than patching the one call site. `copy_into` falls back to `self.name()`, which lets `copy_raster_from` go back to a plain `BandOverrides::default()`.

`name()` is required rather than defaulted: a backend silently returning `None` would drop names on every derived band, which is exactly the failure it exists to prevent. Only two implementors exist — the Arrow-backed `BandRefImpl` and a test stub.

## Why nothing caught it

Two gaps, both closed here:

1. `copy_raster_from_overrides_transform_and_preserves_bands` used an **unnamed** band — despite the name, it asserted nothing about band names.
2. **`assert_raster_arrays_equal` never compared band names at all.** It checks dim names, shape, dtype, nodata, pixels — but not the band name. That blind spot hid this suite-wide.

Added the missing assertion to the shared comparator. It passes across `sedona-raster`, `sedona-raster-functions`, and `sedona-raster-zarr` with no other fallout, which confirms this was the only band-name drop in the tree.

Then gave the affected fixtures real names. `preserves_per_band_names_across_multiple_bands` uses two differently-named bands so the per-index mapping is pinned, not just the presence of a name, and `band_ref_exposes_its_name` covers the new accessor directly (including that an unnamed band still reads back as `None`, and that it agrees with the raster-level fast path).

## Verification

Reverting *either* the explicit pass (commit 1) or the `or_else(|| self.name())` inheritance (commit 2) fails the same 8 tests with `Band names do not match` — 7 `RS_SetGeoReference` tests plus the `copy_raster_from` builder test.

```
sedona-raster            178 passed; 0 failed
sedona-raster-functions  255 passed; 0 failed
```

## Interaction with #1158

#1158 converts `BandOverrides` to a trinary `Override<T>`, and for `name` it currently collapses:

```rust
// A `BandRef` has no source name to inherit, so `Keep` and `Clear`
// both leave the derived band unnamed.
Override::Keep | Override::Clear => None,
```

`name` is the only field whose trinary degenerates, purely because of the missing accessor. Once this lands that can become `Keep => self.name(), Clear => None`, making `name` uniform with every other overridable field. Small follow-up on that branch; no conflict with this one.

### DIFF
--- a/rust/sedona-raster/src/traits.rs
+++ b/rust/sedona-raster/src/traits.rs
@@ -280,8 +280,8 @@ pub trait RasterRef {
 }
 
 /// Field overrides for [`BandRef::copy_into`]. Each field defaults to `None`,
-/// meaning "inherit from the source band". `name` has no source on a `BandRef`
-/// (band names live at the raster level), so it defaults to unnamed.
+/// meaning "inherit from the source band" — `name` included, sourced from
+/// [`BandRef::name`].
 #[derive(Default)]
 pub struct BandOverrides<'a> {
     /// Override the band name; `None` inherits the source's.
--- a/rust/sedona-raster/src/traits.rs
+++ b/rust/sedona-raster/src/traits.rs
@@ -284,7 +284,7 @@ pub trait RasterRef {
 /// (band names live at the raster level), so it defaults to unnamed.
 #[derive(Default)]
 pub struct BandOverrides<'a> {
-    /// Name for the derived band (the source has none to inherit).
+    /// Override the band name; `None` inherits the source's.
     pub name: Option<&'a str>,
     /// Override the dimension names; `None` inherits the source's.
     pub dim_names: Option<&'a [&'a str]>,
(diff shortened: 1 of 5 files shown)


jiayuasu (approved):

rust/sedona-raster/src/traits.rs
```diff
@@ -284,7 +284,7 @@ pub trait RasterRef {
 /// (band names live at the raster level), so it defaults to unnamed.
```
jiayuasu: Minor, non-blocking docs nit: this still describes the pre-refactor behavior. `BandRef` now exposes `name()`, and `copy_into` falls back to it when `name` is `None`, so the derived band keeps the source name. Could we update these two lines, plus the similar comment in `rs_set_georeference.rs`, to match?

james-willis: thanks, fixed. will merge on green CI

PR #1193, opened 2026-08-26
refactor(rust/sedona-raster-functions): adopt start_raster_from / copy_into in the structural raster functions

james-willis:
Closes DB-410.

## Header rebuild → `start_raster_from`

Five call sites each hand-rolled the same thing:

```rust
let t: [f64; 6] = raster.transform().try_into().map_err(|_| { ... })?;
let spatial_dims = raster.spatial_dims();
new_builder.start_raster_nd(&t, &spatial_dims, raster.spatial_shape(), raster.crs())?;
```

`RasterBuilder::start_raster_from` already does exactly this, so all five now call it: `RS_Slice`, `RS_SliceRange`, `RS_DimToBand`, `RS_BandToDim`, `RS_EnsureLoaded`.

In `RS_EnsureLoaded` this also retires a block of owned-metadata locals (`transform` / `spatial_dims_owned` / `spatial_shape` / `crs`) that existed only to keep borrows from spanning the per-band `await` points. The header is built synchronously before the band loop, so none of them are needed — verified nothing downstream reads them.

## Pass-through bands → `copy_into`

The pass-through paths in `RS_Slice`, `RS_SliceRange` and `RS_DimToBand` — bands that don't carry the named dimension and are emitted unchanged — did:

```rust
start_band(StartBandArgs { name, nodata, ..StartBandArgs::new(&dim_names, band.shape(), band.data_type()) })?;
let data = band.nd_buffer()?.as_contiguous()?;
band_data_writer().append_value(data);
```

That materialises a fresh identity band and re-copies the bytes. `copy_into` shares the source buffer zero-copy and carries `view()` forward instead of dropping it.

Two gates, both now open:
- **View persistence** — non-identity views weren't representable, so a view-preserving pass-through had nothing to land on. #1113 merged 2026-08-12.
- **Band-name inheritance** — `BandRef` had no name accessor, so `copy_into` couldn't inherit it. #1192 added `BandRef::name()` and has now merged, so these sites need only `BandOverrides::default()` rather than threading the name through by index.

## Test gap this exposed

The pass-through name preservation was **not covered anywhere**:

- `RS_Slice` / `RS_SliceRange` had mixed-dim pass-through tests, but with **unnamed** bands — so a dropped name was invisible.
- `RS_DimToBand` had **no pass-through test at all**; that branch was uncovered.

Named the existing `build_mixed_dim_raster` fixture and its expectations, and added `dimtoband_passes_through_bands_without_dim_preserving_names`, which also asserts the pass-through band stays 2-D rather than being expanded.

Reverting #1192's `or_else(|| self.name())` inheritance fails all three.

## Deliberately out of scope

`RS_EnsureLoaded`'s **band loop** is untouched. It's async with owned-data scoping around the `await`s, its InDb pass-through is already zero-copy via `append_band_data_from`, and it carries an explicit internal-error guard for non-identity views. Converting it is a restructure that also has to keep the OutDb loader path emitting identity bands — worth its own change, not a rider on a DRY pass. The ticket only calls for "revisit," so it stays a follow-up.

## Verification

```
sedona-raster            178 passed; 0 failed
sedona-raster-functions  256 passed; 0 failed
```

fmt and clippy clean.

### DIFF
--- a/rust/sedona-raster-functions/src/rs_dim_band.rs
+++ b/rust/sedona-raster-functions/src/rs_dim_band.rs
@@ -22,10 +22,9 @@ use datafusion_common::cast::as_string_view_array;
 use datafusion_common::error::Result;
 use datafusion_common::{arrow_datafusion_err, exec_err};
 use datafusion_expr::{ColumnarValue, Volatility};
-use sedona_common::sedona_internal_datafusion_err;
 use sedona_expr::scalar_udf::{SedonaScalarKernel, SedonaScalarUDF};
-use sedona_raster::builder::{RasterBuilder, StartBandArgs};
-use sedona_raster::traits::RasterRef;
+use sedona_raster::builder::{RasterBuilder, RasterOverrides, StartBandArgs};
+use sedona_raster::traits::{BandOverrides, RasterRef};
 use sedona_schema::datatypes::SedonaType;
 use sedona_schema::matchers::ArgMatcher;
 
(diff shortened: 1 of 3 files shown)


james-willis:
moved to draft, waiting for #1192 so we can use its revised solution for pulling band names.

jiayuasu (approved):

PR #1195, opened 2026-08-27
feat(python/sedonadb-geopandas): add column assignment

jiayuasu:
## What changes

This is the first slice of the column assignment / arithmetic / dissolve work drafted in #1184, which grew too large to review as one change. It is being split into four focused PRs — assignment, dissolve, arithmetic, temporal support — that build on each other, so each can be reviewed on its own terms.

This PR adds `gdf[key] = value` to the experimental `sedonadb-geopandas` package:

- **Same-frame `Series` assignment.** A `Series` read from the same frame can be assigned back (renamed or replacing a column). A `Series` from a different frame, a bare SedonaDB expression, and array-likes are rejected with actionable messages: none of them records a usable row provenance, and each previously failed obscurely or would silently write wrong values. Because assignment rebinds the frame, a `Series` read before an assignment is stale and is rejected the same way.
- **Scalar broadcasting.** Python scalars, NumPy scalars (including 0-d arrays), Arrow scalars, masked values, and `pandas.NA` broadcast to every row, via a shared scalar classifier/normalizer (`is_scalar` / `normalize_scalar`).
- **Geometry assignment with CRS bookkeeping.** Assigning a geometry value keeps its type and CRS; a destination CRS is stamped only onto CRS-less geometry. A newly assigned geometry column becomes the active geometry only when the frame had none, and replacing the active geometry column with a non-geometry value clears the active geometry, matching GeoPandas.
- **Temporal scalars.** NumPy `datetime64`/`timedelta64` values are converted at a lossless Arrow unit (ambiguous or sub-nanosecond units are rejected the way pandas rejects them, with an overflow check); pandas `Timestamp`/`Timedelta` keep their nanoseconds, zone-aware timestamps keep their zone, and `NaT` assigns as a datetime missing value.
- **Series stay usable across column-adding assignments.** A `Series` read from a frame remains valid while assignments only add columns, so one captured geometry can supply several derived columns; replacing a column or filtering invalidates earlier reads, which then raise instead of silently resolving to different values.
- Raises the `sedonadb` floor to 0.4.1, the first released version with `DataFrame.mutate`, which assignment is built on.
- Declares `shapely>=2` as a dependency (geometry scalars are Shapely objects; Shapely 2 dropped the sequence protocol on multipart geometries).

## Testing

118 tests (96 new, including direct tests of the temporal scalar normalization), run with warnings promoted to errors against both a source-built current main and the released sedonadb 0.4.1 in a clean environment (covering pandas 2.x and 3.x). Ruff lint/format clean.

### DIFF
--- a/python/sedonadb-geopandas/python/sedonadb_geopandas/_frame.py
+++ b/python/sedonadb-geopandas/python/sedonadb_geopandas/_frame.py
@@ -347,7 +347,18 @@ class GeoDataFrame:
             )
             scalar_crs = getattr(raw.type, "crs", None)
             if str(raw.type.extension_name) == "geoarrow.wkb":
-                expr = ctx.lit(pa.array([raw.as_py()], type=raw.type))
+                arr_type = raw.type
+                if pa.types.is_large_binary(arr_type.storage_type):
+                    # SedonaDB's WKB importer requires Binary storage; the
+                    # type is rebuilt on Binary with the same CRS and edge
+                    # metadata rather than passed through and rejected.
+                    import geoarrow.pyarrow as ga
+
+                    rebuilt = ga.wkb().with_edge_type(arr_type.edge_type)
+                    if arr_type.crs is not None:
+                        rebuilt = rebuilt.with_crs(arr_type.crs)
+                    arr_type = rebuilt
+                expr = ctx.lit(pa.array([raw.as_py()], type=arr_type))
             elif missing:
                 # Non-WKB storage cannot become a literal; a null of it is
                 # rebuilt from its own metadata. Kind and CRS survive; the
--- a/python/sedonadb-geopandas/python/sedonadb_geopandas/_frame.py
+++ b/python/sedonadb-geopandas/python/sedonadb_geopandas/_frame.py
@@ -16,7 +16,7 @@
 # under the License.
 """GeoPandas-style GeoDataFrame backed by a lazy SedonaDB frame."""
 
-from sedonadb_geopandas._series import GeoSeries, Series
+from sedonadb_geopandas._series import GeoSeries, Series, is_scalar
 
 # Rows to collect for the Jupyter rich-text (`_repr_html_`) preview.
 _REPR_HTML_ROWS = 10
@@ -29,6 +29,31 @@ _DERIVE = object()
 def _geometry_column_names(df):
     names = df.schema.names
     return {names[i] for i in df.schema.geometry_column_indices}
+
+
+def _expr_crs(df, expr):
+    """The CRS carried by `expr`, read from a projected schema (a plan build)."""
+    field = df.select(expr.alias("x")).schema.field("x")
+    return getattr(field.type, "crs", None)
+
+
+def _is_missing(value):
+    """Whether `value` is one of the missing-value sentinels.
+
+    `None`, NaN, and `pandas.NA` all mean "no value" to GeoPandas, and a geometry
+    column assigned any of them keeps its type and CRS rather than becoming an
+    ordinary column of nulls.
+    """
+    if value is None:
+        return True
+    try:
+        import pandas as pd
+
+        # Safe for scalars; geometries and numbers simply return False.
+        return bool(pd.isna(value))
+    except Exception:
+        # Without pandas, catch NaN via its self-inequality.
+        return isinstance(value, float) and value != value
 
 
 class GeoDataFrame:
@@ -77,6 +102,18 @@ class GeoDataFrame:
     def __getitem__(self, key):
         # Boolean mask -> row filter (gdf[gdf["pop"] > 1000]).
         if isinstance(key, Series):
+            if key._df is not self._df:
+                # Same check assignment makes. Without it a mask captured before an
+                # assignment is silently reused against the rebound frame: it
+                # happens to resolve while the referenced column still exists, and
+                # fails obscurely at collection when it does not.
+                raise ValueError(
+                    "Cannot filter with a mask built from a different DataFrame: "
+                    "there is no row alignment, so the result would be silently "
+                    "wrong. Note that assigning a column rebinds this frame, so a "
+                    "mask taken beforehand is stale; re-read it as gdf[...] > ... "
+                    "and try again."
+                )
             return GeoDataFrame(self._df.filter(key._expr), self._geometry_name)
 
         # Column subset -> GeoDataFrame. Matching GeoPandas, the active geometry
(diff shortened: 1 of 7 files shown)


python/sedonadb-geopandas/python/sedonadb_geopandas/_frame.py
```diff
@@ -115,6 +177,262 @@ def __getitem__(self, key):
...
+        than "no longer a geometry column", so neither should silently reset what
+        the frame already knew.
+
+        A `Literal` is unwrapped and rebuilt on this frame's context: a literal
+        constructed by the bare `lit()` has no context, so functions cannot be
+        applied to it, and passing it straight through would skip the CRS handling.
+        """
+        from sedonadb.expr import Literal, lit
+
+        from sedonadb_geopandas._series import normalize_scalar
```
paleolimbot: It probably makes more sense to have these be top level imports?

python/sedonadb-geopandas/python/sedonadb_geopandas/_frame.py
```diff
@@ -92,7 +151,10 @@ def __getitem__(self, key):
         # Single column -> (Geo)Series.
         if isinstance(key, str):
             expr = self._df[key]
-            if key == self._geometry_name:
+            # Any geometry-typed column reads back as a GeoSeries — not just
+            # the active one — so a freshly assigned geometry column supports
+            # .area and .buffer() immediately, as it does in GeoPandas.
+            if key == self._geometry_name or key in _geometry_column_names(self._df):
```
paleolimbot: Do you need to check these both or is the second one sufficient?

python/sedonadb-geopandas/python/sedonadb_geopandas/_frame.py
```diff
@@ -115,6 +177,262 @@ def __getitem__(self, key):
...
+
+        if isinstance(value, Series):
+            if value._df is not self._df:
+                raise ValueError(
+                    "Cannot assign a Series that comes from a different "
+                    "DataFrame: there is no row alignment, so the result would "
+                    "be silently wrong. Note that assigning to this frame "
+                    "rebinds it, so a Series read before an earlier assignment "
+                    "is already stale; re-read it as gdf[...] and try again."
+                )
```
paleolimbot, line 232: I can see how this might be annoying...you might want to assign a few different transformations of a geometry into the dataframe like perimeter, area, etc. I don't have a great solution...UUID that bumps on everything except a projection?

python/sedonadb-geopandas/python/sedonadb_geopandas/_frame.py
```diff
@@ -115,6 +177,262 @@ def __getitem__(self, key):
...
+        if geoarrow_typed:
+            scalar_spherical = (
+                "SPHERICAL" in str(getattr(raw.type, "edge_type", "")).upper()
+            )
+            scalar_crs = getattr(raw.type, "crs", None)
+            if str(raw.type.extension_name) == "geoarrow.wkb":
+                arr_type = raw.type
+                if pa.types.is_large_binary(arr_type.storage_type):
+                    # SedonaDB's WKB importer requires Binary storage; the
+                    # type is rebuilt on Binary with the same CRS and edge
```
paleolimbot, line 370: We should probably fix to also accept LargeBinary. I ran into this issue in Iceberg (LargeBinary is what they return BLOBs as).

In general it would be great to know what needs to change in geoarrow-pyarrow or sedonadb to make this easier (it feels like it should be easier than it is).

python/sedonadb-geopandas/python/sedonadb_geopandas/_frame.py
```diff
@@ -115,6 +177,262 @@ def __getitem__(self, key):
...
+        # projection shows no CRS. A value that carries its own CRS (a
+        # GeoSeries literal, say) keeps it: stamping the destination CRS over
+        # it would relabel the coordinates without transforming them, which
+        # is silently wrong data.
+        if crs is not None and (inherits_crs or _expr_crs(self._df, expr) is None):
+            expr = expr.funcs.st_setcrs(ctx.lit(crs.to_json()))
+        elif strips_crs and _expr_crs(self._df, expr) is not None:
+            # SRID 0 means "no CRS" and keeps the value; st_setcrs(NULL)
+            # would null-propagate and erase every row.
+            expr = expr.funcs.st_setsrid(ctx.lit(0))
```
paleolimbot, line 446: SetSRID to 0 is probably best (`ST_SetCRS("0")` is the totally ungraceful way to do this with setcrs)

python/sedonadb-geopandas/python/sedonadb_geopandas/_series.py
```diff
@@ -17,6 +17,191 @@
...
+        if isinstance(value, pa.Scalar):
+            return True
+    except ImportError:
+        pass
+    try:
+        from shapely.geometry.base import BaseGeometry
+
+        # Shapely 1.x multipart geometries implement __len__ and __iter__;
+        # they are still single values. (Shapely 2 removed the sequence
+        # protocol, but the package floor does not require Shapely 2.)
```
paleolimbot: You could probably require shapely 2 (I think it is fairly old at this point)

python/sedonadb-geopandas/python/sedonadb_geopandas/_series.py
```diff
@@ -17,6 +17,191 @@
...
+
+
+def normalize_scalar(value):
+    """Normalize an accepted scalar into something a literal can hold.
+
+    Passing the classifier is not the same as being constructible: a 0-d NumPy
+    array is a scalar but `lit()` cannot take it, and `pandas.NA` is a missing
+    sentinel `lit()` does not recognize. Unwrap the former to its Python value
+    and convert missing sentinels to `None` (SQL null). Callers apply this only
+    after `is_scalar` has accepted the value.
```
paleolimbot, line 57: Can we update `lit()` to handle some of these values to make it more useful for non-geopandas? I know that is slightly tricky because you can't import pandas there (but special casing the class names like we do for some other literal types might work).

python/sedonadb-geopandas/tests/test_geopandas_compat.py
```diff
@@ -242,3 +252,777 @@ def test_head(cities):
...
+        gdf["g"] = value
+        dtype = str(gdf._df.schema.field("g").type)
+        assert "geography" in dtype
+        assert "4267" in dtype
+
+
+def test_numpy_scalar_dtypes_are_preserved(points):
+    # .item() promoted np.int8/np.float32 to int64/float64 columns and
+    # overflowed np.uint64 past int64, which the engine supports natively.
+    import numpy as np
```
paleolimbot: These and other can be top level imports (we always test with numpy/pyarrow)

paleolimbot (commented):
A few things worth handling here, but in general this looks great!

Unless the timezone unit stuff is truly untenable, probably worth handling it while you're here.

python/sedonadb-geopandas/python/sedonadb_geopandas/_frame.py
jiayuasu: Done: `sedonadb.expr`, `pyarrow`, and the `_series` helpers are module-level now (`shapely` too, since it's a declared dependency as of this round), and the `ImportError` guards that only wrapped pyarrow are gone. numpy and pandas stay function-local because they're optional.

python/sedonadb-geopandas/python/sedonadb_geopandas/_frame.py
jiayuasu: The second one is sufficient. The active name is validated to be a geometry column at construction and cleared whenever the column stops being one, so it's just the schema check now.

python/sedonadb-geopandas/python/sedonadb_geopandas/_frame.py
jiayuasu: Agreed, and it turned out to be tractable without tracking column references inside expressions. A Series read from a frame now stays valid across assignments that only *add* columns: the rows and every column it could reference are unchanged, so its expression resolves by name to the same values. That covers `g = gdf.geometry; gdf['area'] = g.area; gdf['len'] = g.length; ...`. Replacing a column or filtering still invalidates earlier reads, since a Series captured before `gdf['v'] = ...` would otherwise resolve to the new `v`. It's implemented as a list of ancestor frames on the GeoDataFrame rather than a token on the Series, so `Series` didn't need to change. README updated to say this.

python/sedonadb-geopandas/python/sedonadb_geopandas/_frame.py
jiayuasu: Filed #1215 for the importer. For the broader question, #1214 collects everything the wrapper currently has to normalize before `lit()` can take it. The two that matter most here: `_lit_from_geoarrow_scalar` rebuilds the type as `ga.wkb().with_crs(crs)`, which drops the edge type (spherical scalars come back planar), and only `WkbScalar` is registered, so WKT and native-point scalars fail in the `pa.array([obj])` fallback. With those two fixed this whole GeoArrow branch collapses to `ctx.lit(raw)`. The rebuild stays for now so the package keeps working on 0.4.1.

python/sedonadb-geopandas/python/sedonadb_geopandas/_frame.py
jiayuasu: Yes, SetSRID(0) is what's used here. It's the value-preserving spelling; `ST_SetCRS(NULL)` null-propagated and erased the rows, which is how this branch came to exist.

python/sedonadb-geopandas/python/sedonadb_geopandas/_series.py
jiayuasu: Done: `shapely>=2` is now a declared dependency. It wasn't declared at all before and only came in through the geopandas extra. The 1.x multipart guard and its test are gone.

python/sedonadb-geopandas/python/sedonadb_geopandas/_series.py
jiayuasu: Agreed, that's #1214. It lists each case with today's behavior (`pd.NA` and `NaT` unrecognized, `Timestamp`/`Timedelta` silently truncated to microseconds, numpy day/week/year units rejected, `np.ma.masked`, 0-d arrays, structured voids, typed-null nested scalars) and notes that the class-name table already handles the pandas cases without importing pandas. Each one that lands upstream lets `normalize_scalar` shrink to a pass-through.

python/sedonadb-geopandas/tests/test_geopandas_compat.py
jiayuasu: Done: numpy, pyarrow, pandas, shapely, geoarrow and `lit` are module-level in the test file now.

jiayuasu:
Thanks for the review. Pushed `888a8dbd9` with everything addressed; the decisions in one place:

- **Temporal scalars are handled here rather than deferred.** A new `_temporal` module picks a lossless Arrow unit for `numpy.datetime64`/`timedelta64` (rejecting ambiguous or sub-nanosecond units the way pandas does, with an overflow check), keeps nanoseconds for `pandas.Timestamp`/`Timedelta` (including zone-aware timestamps), and assigns `NaT` as a datetime missing value. The "temporarily rejected" note is gone from the README. Duration arithmetic and temporal group keys come with the arithmetic and dissolve follow-ups.
- **A `Series` now survives assignments that only add columns**, so one captured geometry can supply several derived columns. Replacing a column or filtering still invalidates earlier reads.
- **`shapely>=2` is a declared dependency**; the Shapely 1.x compatibility code is gone.
- **Upstream:** #1214 (what `lit()` could resolve on its own: GeoArrow scalars with their edge type, and the missing/temporal scalar cases) and #1215 (WKB import rejecting LargeBinary storage). The wrapper keeps its workarounds until those land so it stays usable on 0.4.1.

92 tests, run with warnings promoted to errors against both current main and the released 0.4.1.

python/sedonadb-geopandas/tests/test_geopandas_compat.py
```diff
@@ -242,3 +262,848 @@ def test_head(cities):
...
+def test_explicitly_inactive_geometry_stays_inactive(points):
+    # geometry=None is a choice; a no-op reassignment of an existing geometry
+    # column must not silently reactivate it.
+    df = sgpd.default_context().create_data_frame(points)
+    gdf = GeoDataFrame(df, geometry=None)
+    gdf["geometry"] = gdf["geometry"]
+    assert gdf._geometry_name is None
+
+
+# -- regressions from review ------------------------------------------------
```
paleolimbot: I think you can remove these headers

python/sedonadb-geopandas/tests/test_geopandas_compat.py
```diff
@@ -31,6 +43,16 @@ def cities():
...
 
 
+@pytest.fixture
+def points():
+    """A small frame in a projected CRS, shared across the tests."""
+    return gpd.GeoDataFrame(
+        {"name": ["A", "B", "C"], "v": [1, 2, 3]},
+        geometry=gpd.GeoSeries.from_wkt(["POINT (0 0)", "POINT (5 5)", "POINT (9 9)"]),
+        crs="EPSG:3857",
+    )
```
paleolimbot: I know it sounds dumb, but these tests are better off inlining this (or whatever subset of it they need). Saves both humans and LLMs a lot of tokens and scrolling 🙂

paleolimbot (approved):
Thank you!

I didn't check test coverage here...the temporal stuff in general might benefit from some tests of its internals if they aren't already tested.

python/sedonadb-geopandas/tests/test_geopandas_compat.py
jiayuasu: Done, all three are gone.

python/sedonadb-geopandas/tests/test_geopandas_compat.py
jiayuasu: Not dumb at all. Done: each test that used it now builds the frame inline (the same three points, via `points_from_xy`, so nothing changed in what the tests assert). It also turned up one dissolve test that took the fixture and never used it. `cities` and `con_free_geom_frame` predate this PR, so I left those alone.

jiayuasu:
Thanks! On the temporal internals: the compatibility tests exercise them end to end (assignment, then materialization), but nothing pinned the helper itself, so `0b14e05a5` adds `tests/test_temporal.py` testing `normalize_temporal_scalar` and `nat_scalar` directly. It covers the unit table (Arrow-native units kept; week/day/hour/minute and datetime year/month converted exactly to seconds, including values outside the nanosecond range; exact sub-nanosecond values narrowed to nanoseconds), the zone-aware Timestamp path, every missing-value form (numpy NaT, `pd.NaT`, and the INT64_MIN sentinel as duration and timestamp), Arrow scalar pass-through, and the error classes for ambiguous units, lossy values, and overflow. That last one runs under both NumPy 2.3 and 2.5, which differ only in which layer raises the `OverflowError`.

PR #1196, opened 2026-08-27
refactor(rust/sedona-raster): make ViewEntries the interface for passing views

james-willis:
## The smell

```rust
ViewEntries::new(input.view().to_vec()).compose(&ViewEntries::new(view.to_vec()))?
```

Two heap allocations to read two slices. It looks like a workaround because it is one.

## Why it was there

`ViewEntries` existed as the container type, but nothing actually passed it around. `BandRef::view()` returned `&[ViewEntry]`, and every struct field carrying a view held a bare slice:

| | before |
|---|---|
| `BandRef::view()` | `&[ViewEntry]` |
| `StartBandArgs::view` | `Option<&[ViewEntry]>` |
| `WithViewArgs::view` | `&[ViewEntry]` |
| `BandOverrides::view` | `Option<&[ViewEntry]>` |
| `RasterLoadRequest::view` | `&[ViewEntry]` |
| `RasterLoadResult::view` | `Vec<ViewEntry>` |

So slices were the real currency and `ViewEntries` was a method bag. `compose` was the only operation demanding the owned type on *both* sides — which is why every caller had to allocate its way in.

Two supporting details:

- `BandRefImpl` **already stored** a `ViewEntries` and called `.as_slice()` on the way out. Returning the container removes a conversion rather than adding one.
- Both doc comments describing this API were **uncompilable as written** — they said `source.view().compose(&next)?`, but `view()` returned a slice, which has no `compose`. They described the API this PR builds.

## Changes

- `BandRef::view() -> &ViewEntries`
- `compose`'s `next` takes `impl AsRef<[ViewEntry]>`, so a `ViewEntries` or a bare slice both work — existing `a.compose(&b)` call sites are untouched
- The six fields above all carry `ViewEntries`

The motivating site is now:

```rust
let effective_view = match overrides.view {
    Some(v) => self.view().compose(v)?,
    None => self.view().clone(),
};
```

## A duplicate identity check, silently diverged

Chasing the same thread turned up `rs_ensure_loaded::view_is_identity`, a private copy of `ViewEntries::is_identity` — which had **drifted from it**:

| | empty view, `source_shape = [2, 3]` |
|---|---|
| `view_is_identity` | `true` (short-circuits on `is_empty()`) |
| `ViewEntries::is_identity` | `false` (length mismatch) |

An empty view is identity only for a 0-dimensional raster, which is what the method already implements — the local copy was wrong.

It was unreachable in production: `RasterLoadResult::unresolved` clones the request's view, the request's view is the band's view, and `start_band` rejects 0-dimensional bands outright. The only things exercising the divergence were loader tests passing an empty view alongside a 2-D or 3-D `source_shape` — a state that cannot occur. Those fixtures now pass a real identity view over their source shape, and the helper is deleted in favour of the method.

## No behaviour change

Composition, validation, and persistence are untouched. The only semantic change is that the identity check now has one definition instead of two that disagreed.

```
sedona-raster            183 passed; 0 failed
sedona-raster-functions  256 passed; 0 failed
```

fmt clean, clippy clean (no unused imports), `sedona-raster-gdal` builds.

## Note

#1158 is stacked on this branch — it converts `BandOverrides` to a trinary `Override<T>`, which composes cleanly with the type change here.

### DIFF
--- a/python/sedonadb-zarr/src/lib.rs
+++ b/python/sedonadb-zarr/src/lib.rs
@@ -30,7 +30,7 @@ use pyo3::exceptions::{PyRuntimeError, PyValueError};
 use pyo3::ffi::Py_buffer;
 use pyo3::prelude::*;
 use pyo3::types::PyCapsule;
-use sedona_raster::view_entries::ViewEntry;
+use sedona_raster::view_entries::{ViewEntries, ViewEntry};
 use sedona_raster_zarr::{
     object_store_for_uri, open_storage_from_uri, ZarrChunkReader, ZarrLoader,
 };
@@ -341,7 +341,7 @@ impl PyZarrRasterLoader {
                 uri,
                 dim_names,
                 source_shape,
-                view,
+                view: ViewEntries::new(view),
                 data_type,
             });
         }
@@ -386,7 +386,7 @@ impl PyZarrRasterLoader {
                     source_shape: r.source_shape,
                     view: r
                         .view
-                        .into_iter()
+                        .iter()
                         .map(|v| ZarrViewEntry {
                             source_axis: v.source_axis,
                             start: v.start,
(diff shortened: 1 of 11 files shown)


jiayuasu (approved):

jiayuasu:
LGTM!

PR #1207, opened 2026-09-01
feat(rust/sedona-expr): parse logical spatial filters

### LINKED ISSUE #1152: rust/sedona-expr: Allow `SpatialFilter` to be constructed from a logical `Expr`
Currently our SpatialFilter infrastructure can only be created from `Arc<dyn PhysicalExpr>`; however, when implementing `TableProvider` it's useful to be able to support the `filters` argument, which are logical `Expr`. We can export a logical expression parser that mirrors the physical one to make it easier to implement simple table providers.

paleolimbot:
Adds logical Expr parsing to SpatialFilterFactory while sharing the existing physical predicate parser. Supports spatial range and distance predicates, boolean AND/OR, boolean literals, literal metadata, and name-based logical columns.

This is done by converting to a physical expression and reusing that parsing. I think this is a good thing because the logic is complicated for the specific functions and the chance of an error in the dummy conversion is smaller than a chance of an error in the parsing / bounding logic.

Closes #1152

### DIFF
--- a/rust/sedona-expr/src/spatial_filter.rs
+++ b/rust/sedona-expr/src/spatial_filter.rs
@@ -21,7 +21,7 @@ use datafusion_common::config::ConfigOptions;
 use datafusion_common::{exec_datafusion_err, DataFusionError, Result, ScalarValue};
 use datafusion_expr::{Expr, Operator};
 use datafusion_physical_expr::{
-    expressions::{BinaryExpr, Column, Literal, NegativeExpr},
+    expressions::{BinaryExpr, Column, Literal},
     PhysicalExpr, ScalarFunctionExpr,
 };
 use geo_traits::Dimensions;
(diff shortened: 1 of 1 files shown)


james-willis (approved):

rust/sedona-expr/src/spatial_filter.rs
```diff
@@ -488,6 +499,83 @@ impl SpatialFilterFactory {
...
+                function.func.name(),
+                Arc::clone(&function.func),
+                args,
+                Arc::new(Field::new("", return_type, true)),
+                Arc::new(ConfigOptions::default()),
+            )))
+        }
+        // Preserve an unsupported node as a physical expression that the shared
+        // argument parser will classify as `Other`.
+        _ => Ok(Arc::new(UnknownSentinelExpr)),
```
james-willis, line 538: what actually happens in this case?

paleolimbot: This will translate into `SpatialFilter::Unknown` (which will prevent pruning, since Unknown could evaluate to True).

PR #1208, opened 2026-09-01
fix(rust/sedona-raster-gdal): make sedona-proj dev-only

### LINKED ISSUE #1036: rust/sedona-raster-functions: Remove sedona-proj dependency for raster functions
After https://github.com/apache/sedona-db/pull/1001 we have access to reprojection from the ConfigOptions so we no longer need to depend on sedona-proj and `with_global_proj_engine()` to do this from other crates. The sedona-raster-functions crate does this in a few places (e.g., RS_Value, RS_Intersects.

paleolimbot:
Remove sedona-proj from regular dependencies because all uses are test-only. The existing dev-dependency retains proj-sys for tests.

closes #1036

### DIFF
--- a/rust/sedona-raster-gdal/Cargo.toml
+++ b/rust/sedona-raster-gdal/Cargo.toml
@@ -42,7 +42,6 @@ sedona-common = { workspace = true }
 sedona-expr = { workspace = true }
 sedona-functions = { workspace = true }
 sedona-gdal = { workspace = true }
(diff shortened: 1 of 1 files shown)


james-willis (commented):
Yes. I think the whole SedonaDB is at this point only consuming sedona-proj through CrsEngine. 

I've already checked raster but not 100% sure about vector

PR #1212, opened 2026-09-01
test(integration): compare() accepts an optional expected anchor

james-willis:
**Stacked on #1211** (its branch lives on a fork, so the base can't be retargeted onto it — its commits are merged in instead; this diff deduplicates when #1211 lands. Merge #1211 first.)

Implements @paleolimbot's post-merge suggestion on #1203: an optional parameter on `compare()` that lets the reference value be specified, guarding the parity claim against vacuous agreement — every engine returning the same wrong thing passes a pure engine-vs-engine comparison.

### The API

```python
compare(sql, sedona, spark, expected=200.0)        # table path: anything assert_result accepts
compare(sql, sedona, spark, expected=[(None,)])    # e.g. anchoring a NULL
compare(sql, sedona, spark, expected=decoded)      # raster path: a DecodedRaster
```

`expected` is keyword-only and optional. The subject is asserted against the anchor first (clearest failure when all engines agree on a wrong value), then against each reference as before — so the references are anchored transitively. Parity-only remains the default for cases where the reference is hard to state.

### Coverage added

Anchors on every test whose expected value is trivially known:

- `test_rs_band_nodata`: the written nodata for both bands, and `[(None,)]` for the no-nodata fixture.
- All four passing setter tests (`test_rs_setbandnodata`, `_band2`, `_overwrite`, `_two_arg_single_band`): the full `DecodedRaster` — the exact seeded pixels (reconstructible because `random_raster_data` is seeded), the grid (stated explicitly so the anchor and the fixture can't drift apart), and the per-band nodata.

The anchored tests place their grid with `bbox=`, and the anchors state the same BBOX via `DecodedRaster`'s bbox construction (added to #1211) — the instance still carries the derived `gdal_transform`, which is what `assert_decoded_equal` compares.

The setter anchors close a real hole: with parity only, both engines no-opping identically (source nodata unchanged) would have passed every setter test.

### Verification

- Suite: `21 passed, 13 xfailed` locally (pyspark 4.0.4, Sedona 1.9.1), plus #1211's `test_raster_testing.py` (26 passed) in-branch.
- Negative-tested both paths: a wrong scalar anchor and a wrong raster-nodata anchor each raise `AssertionError`; the correct anchors pass.

This suite is not yet wired into CI (#1206 does that), so the counts above are from a local run.

### DIFF
--- a/integration/spark-parity/test_rs_raster_out.py
+++ b/integration/spark-parity/test_rs_raster_out.py
@@ -47,11 +47,9 @@ BAND_NODATA = {"uint8": 200.0, "int32": -99999.0, "float64": -12345.5}
 # The grid the anchored tests write and reconstruct. Anchors must state the
 # exact raster they expect, so the placement is passed to
 # create_random_raster_view explicitly rather than relying on its defaults
-# staying in sync. GDAL_TRANSFORM is BBOX on the WIDTH x HEIGHT grid — the
-# transform an anchored `DecodedRaster` must decode back to.
+# staying in sync — and the anchors state the same BBOX.
 BANDS, HEIGHT, WIDTH = 2, 6, 7
 BBOX = (100.0, 482.0, 114.0, 500.0)
-GDAL_TRANSFORM = (100.0, 2.0, 0.0, 500.0, 0.0, -3.0)
 
 
 def _anchor(dtype, nodata, *, bands=BANDS, plants=None):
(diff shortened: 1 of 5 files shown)


james-willis:
Re-raised from the fork as #1218 (contributor branches shouldn't live in the Apache repo). Same commits; the apache branch is deleted.

PR #1213, opened 2026-09-01
test(integration): RS_Resample parity between SedonaDB and Sedona Spark

james-willis:
**Stacked on #1211 → #1212.** Based on #1212's branch (`jw/compare-expected`), so this diff shows only the new resample module; merge those first.

The next step of the parity roadmap from #1203/#1212: RS_Resample, probed first on both engines, then locked in with anchored `compare()` calls where agreement holds and `xfail`-cataloged where it doesn't. RS_MapAlgebra is out of scope (SedonaDB doesn't implement it; the roadmap docstring is updated accordingly).

### What agrees (13 passing tests, all anchored)

The nearest-neighbour path is bit-for-bit identical across every overload:

- **Dimension mode**: 2x upsample (uint8 and float64-with-nodata), identity, non-integer-ratio downsample, integer SQL literals, lowercase algorithm spelling.
- **Scale mode**: exact tiling, and the grown-extent case — `ceil(extent/pixel)` output, border filled with band nodata *or 0 when absent* (both engines share the 0-fill convention).
- **Reference-raster overload** (4-arg, via CROSS JOIN) and the **grid-snap overload** (7-arg) with centres strictly inside the source.
- Both engines refuse zero width / zero scale (parity on refusal, per-engine `pytest.raises`).

Nearest resampling on the seeded grid is hand-computable (block replication up, centre-rule decimation down), so every passing test states its exact expected `DecodedRaster` — no vacuous agreement.

### What diverges (12 xfails)

- **Positive scaleY**: SedonaDB keeps the origin at the top edge and grids upward off the raster; Spark rebases to the bottom edge and covers the source south-up.
- **Bilinear** (uint8 + float64) and **Bicubic**: both engines really interpolate, but GDAL and JAI kernels differ (edge clamping vs nodata/zero fill, integer rounding).
- **Silent fallback family** (Cubic, CubicSpline, Lanczos, Average, Mode): Sedona Spark only implements NearestNeighbor/Bilinear/Bicubic and silently resamples nearest for every other name — SedonaDB runs the real kernel. Probed on a downsample, where Average/Mode can't agree by construction.
- **Unknown algorithm**: SedonaDB rejects `'sinc'`; Spark silently resamples nearest.
- **Fractional width**: SedonaDB rejects; Spark truncates 3.5 → 3 columns.
- **Grid-snap edge centres**: output centres landing exactly on the source's left edge — GDAL samples the edge pixel, JAI zero-fills.

### Verification

- Suite: `34 passed, 25 xfailed` locally (pyspark 4.0.4, Sedona 1.9.1) — the module adds 13 passed / 12 xfailed to #1212's 21/13, and every xfail was probe-confirmed to actually trip.
- Anchors were verified against both engines' observed output before being baked in.

### DIFF
--- a/integration/spark-parity/test_rs_raster_out.py
+++ b/integration/spark-parity/test_rs_raster_out.py
@@ -31,8 +31,8 @@ failure with a real traceback, not a skip. `SedonaSpark` caches its
 
 `RS_SetBandNoDataValue` is the first case on purpose: it is raster-in/raster-out
 but passes pixels through untouched, so a mismatch is a round-trip bug, not an
-operation divergence — it isolates the transport. Pixel-transforming ops
-(RS_Resample, RS_MapAlgebra) come next, with their known divergences as xfails.
+operation divergence — it isolates the transport. Pixel-transforming coverage
+continues in test_rs_resample.py.
 """
 
 import pytest
--- a/integration/spark-parity/test_rs_resample.py
+++ b/integration/spark-parity/test_rs_resample.py
@@ -1,0 +1,328 @@
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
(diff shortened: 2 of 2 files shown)


james-willis:
Re-raised from the fork as #1219 (contributor branches shouldn't live in the Apache repo). Same commits; the apache branch is deleted.

PR #1216, opened 2026-09-02
refactor(rust/sedona-functions): centralize bounding functions

### LINKED ISSUE #1037: rust/sedona-functions: Move geography bounding functions into sedona-functions
After https://github.com/apache/sedona-db/pull/1001 we have access to a bounder that can bound non-planar literals. Our bounding-related functions with a geography overload are currently in sedona-s2geography, but because we can access the bounders from the ConfigOptions, the functions can just live in sedona-functions. This allows us to add support for more edge types (e.g., Karney using sedona-proj).

paleolimbot:
This PR makes bounding functions for geography generic, implemented in sedona-functions using the bounder present in the config options. This lets geography bounding occur based on runtime injection of a bounder, rather than compile time features present in the sedona build.

I checked the bounding benchmarks for sedona-functions and didn't find any regressions.

Closes #1037.


rust/sedona-functions/src/executor.rs
```diff
@@ -19,14 +19,37 @@ use std::iter::zip;
...
+    arg_type: &SedonaType,
+    config_options: Option<&ConfigOptions>,
+    function_name: &str,
+) -> Result<Box<dyn WkbBounder2D>> {
+    let edges = match arg_type {
+        SedonaType::Wkb(edges, _) | SedonaType::WkbView(edges, _) => *edges,
+        SedonaType::Arrow(DataType::Null) => Edges::Planar,
+        _ => return sedona_internal_err!("Expected geometry or geography, got {arg_type:?}"),
+    };
+    let factory = config_options
```
jiayuasu, line 42: Could we register the bounder in `new_from_context()` too? With the `sedona/s2geography` feature enabled, this test now fails:

```rust
#[tokio::test]
async fn geography_bounds_default_context() {
    let ctx = sedona::context::SedonaContext::new();
    ctx.ctx
        .sql("SELECT ST_XMin(ST_GeogFromText('POINT (1 2)'))")
        .await.unwrap()
        .collect().await.unwrap();
}
```

The error is `ST_XMin() requires a bounder for Spherical edges, which is not registered in this session`. `new_from_context(SessionContext::new())` has the same problem. `new_local_interactive()` works because it installs the bounder.

rust/sedona-functions/src/st_envelope.rs
```diff
@@ -82,18 +80,29 @@ impl<T: WkbBounder2D + Default> SedonaScalarKernel for STEnvelope<T> {
...
+        config_options: Option<&ConfigOptions>,
     ) -> Result<ColumnarValue> {
         let executor = WkbBytesExecutor::new(arg_types, args);
         let mut builder = BinaryBuilder::with_capacity(
             executor.num_iterations(),
             WKB_MIN_PROBABLE_BYTES * executor.num_iterations(),
         );
 
-        let mut bounder = T::default();
+        let mut bounder = bounder_for_arg_type(&arg_types[0], config_options, "ST_Envelope")?;
```
jiayuasu, line 101: Is geography support meant to survive a native UDF export/import roundtrip? I reproduced a failure at the Rust FFI boundary; the equivalent Python example is:

```python
import sedonadb
from sedonadb.udf import sedona_native_scalar_udf

con = sedonadb.connect()
capsules = con.funcs.st_envelope.__sedonadb_scalar_udf__()
con.register(sedona_native_scalar_udf(capsules, name="rt_envelope"))

con.sql("SELECT ST_Envelope(ST_GeogFromText('POINT (1 2)'))").to_arrow_table()
con.sql("SELECT rt_envelope(ST_GeogFromText('POINT (1 2)'))").to_arrow_table()
```

The direct call works, but the imported call reports `ST_Envelope() requires a bounder for Spherical edges, which is not registered in this session`. The C wrapper invokes the exported kernel with `None` for `ConfigOptions`, so it cannot find the session's bounder. `ST_XMin` has the same issue; its previous S2 kernel worked through this roundtrip.

paleolimbot: I updated this to store a copy of the source config options, which I think is better for the future foisting of our functions on other engines (the existing case of foisting our own functions on our own engine from extension packages wasn't affected).

rust/sedona-functions/src/executor.rs
paleolimbot: Good catch! Updated.

PR #1219, opened 2026-09-02
test(integration): RS_Resample parity between SedonaDB and Sedona Spark

james-willis:
Two follow-ups ride along with the resample module:

- **`DecodedRaster.random` / `DecodedRaster.write_geotiff`** — addresses @paleolimbot's `_anchor` comment on #1218 by combining the random-raster constructors: one definition of the standard raster (seeded pixels on the `RANDOM_GRID_*` defaults), which `write_random_geotiff` and `create_random_raster_view` now delegate to. An `expected=` anchor is the same object the fixture registers, so they cannot drift; the duplicated grid constants and the module-local `_anchor` helper are gone. `write_geotiff` refuses per-band nodata (GeoTIFF nodata is file-wide), and a round-trip unit test pins the default grid to the historical transform.
- This PR's CI runs the parity suite for the first time (the #1217 lane is on this branch via main).

The next step of the parity roadmap from #1203/#1212: RS_Resample, probed first on both engines, then locked in with anchored `compare()` calls where agreement holds and `xfail`-cataloged where it doesn't. RS_MapAlgebra is out of scope (SedonaDB doesn't implement it; the roadmap docstring is updated accordingly).

### DIFF
--- a/.github/workflows/python.yml
+++ b/.github/workflows/python.yml
@@ -28,6 +28,7 @@ on:
       - 'rust/**'
       - 'c/**'
       - 'python/**'
+      - 'integration/spark-parity/**'
   push:
     branches:
       - main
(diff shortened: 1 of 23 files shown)


paleolimbot (approved):
Thank you!

PR #1223, opened 2026-09-02
fix(rust/sedona-raster): align RS_Rotation with Sedona Spark's formula

james-willis:
`RS_Rotation` parity coverage that turned into a kernel fix, in three commits:

1. **Probe-first coverage** of the divergence: SedonaDB computed `atan2(-skewX, scaleX)` where Sedona Spark computes `acos(scaleX / sqrt(scaleX² + skewY²))`, negated when `skewY > 0`. They coincide only while `|skewX| == |skewY|`.
2. **One parity module per RS_ function**: `test_rs_scalar.py` → `test_rs_bandnodatavalue.py`, `RS_Rotation` cases into `test_rs_rotation.py` (the suite's layout convention going forward).
3. **The fix**: `GeoTransformEx::rotation` now uses Spark's formula — the angle of the grid's actual column-axis direction `(scaleX, skewY)`.

### Why align rather than document the divergence

- The old formula didn't just diverge under shear: it misreported a **plain rigid rotation of non-square (2x3) pixels** as −0.7137 for a true −π/6 (the new formula recovers −π/6 exactly, for any pixel shape).
- `RS_GeoTransform` (#1221) ports Spark's decomposition, so after this change its `thetaI` agrees with `RS_Rotation` by construction — no internal inconsistency.
- `acos` incidentally fixes a zero-sign nit: the old `atan2` returned IEEE `-0.0` for axis-aligned rasters where Spark returns `+0.0`.

Blast radius: `GeoTransformEx::rotation` feeds only the `RS_Rotation` UDF; rust unit tests in `geo_transform.rs`, `affine_transformation.rs`, and `rs_geotransform.rs` are rewritten around derived cases (axis-aligned asserts the positive zero's sign; sheared expectation matches Spark bit-for-bit in probing).

### The parity tests (all passing, anchored)

- Axis-aligned (north-up, south-up) → exactly `0.0`, sign included.
- Rigid ±30° rotations, square and non-square pixels → `round(-theta, 12)`.
- Shear → `round(-acos(scaleX/hypot(scaleX, skewY)), 12)`.

Rotated/sheared queries compare through `ROUND(..., 12)`, with the evidence in a comment above the tests: neither runtime promises a bit-exact `acos` (Java `Math.acos` is specified to within 1 ulp of correctly rounded; Rust `f64::acos` is the platform libm with no accuracy contract), and a 72-angle sweep on macOS vs Temurin 17 measured 2 angles differing by exactly 1 ulp — π/6 among them — with the unlucky set being libm-dependent, so a raw comparison would be cross-platform flaky rather than strict. The anchored constants sit thousands of ulps from a 12-digit rounding boundary, so `ROUND` collapses 1-ulp pairs without ever splitting them. Axis-aligned cases compare raw and exact.

### Verification

`cargo test -p sedona-raster` and `-p sedona-raster-functions` green locally; `integration/spark-parity` at `28 passed, 14 xfailed` (RS_Rotation contributes 7 anchored passes, no xfails). This branch carries the #1217 lane so CI runs the parity suite.

Aside, noted while probing and out of scope: SedonaDB evaluates `-0.0 = 0.0` as `false` in SQL where Spark says `true`.

### DIFF
--- a/integration/spark-parity/test_rs_bandnodatavalue.py
+++ b/integration/spark-parity/test_rs_bandnodatavalue.py
@@ -16,7 +16,6 @@
 # under the License.
 """SedonaDB vs Sedona Spark parity for RS_BandNoDataValue.
 
-The parity suite keeps one module per RS_ function.
 
 Sedona Spark is the compatibility target, so each test runs one shared SQL
 string on both engines and asserts they agree with the harness-level
--- a/integration/spark-parity/test_rs_bandnodatavalue.py
+++ b/integration/spark-parity/test_rs_bandnodatavalue.py
@@ -14,7 +14,9 @@
 # KIND, either express or implied.  See the License for the
 # specific language governing permissions and limitations
 # under the License.
-"""SedonaDB vs Sedona Spark parity for the scalar raster readers.
+"""SedonaDB vs Sedona Spark parity for RS_BandNoDataValue.
+
+The parity suite keeps one module per RS_ function.
 
 Sedona Spark is the compatibility target, so each test runs one shared SQL
 string on both engines and asserts they agree with the harness-level
@@ -32,8 +34,6 @@ the case `xfail(reason=...)` so the suite doubles as a catalog of what to fix �
 it flips to xpass the day the fix lands. When today's divergence is that one
 engine raises, that error is what trips the xfail.
 """
-
-import math
 
 import pytest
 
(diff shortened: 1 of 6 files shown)


integration/spark-parity/test_rs_bandnodatavalue.py
```diff
@@ -14,7 +14,9 @@
 # KIND, either express or implied.  See the License for the
```
james-willis, line 14: drivey by: clean up the parity harness to more comprehensively model 1 file for 1 function

integration/spark-parity/test_rs_bandnodatavalue.py
```diff
@@ -14,7 +14,9 @@
...
 # specific language governing permissions and limitations
 # under the License.
-"""SedonaDB vs Sedona Spark parity for the scalar raster readers.
+"""SedonaDB vs Sedona Spark parity for RS_BandNoDataValue.
+
+The parity suite keeps one module per RS_ function.
```
james-willis:
```suggestion
```

paleolimbot (approved):
Thank you!

PR #1225, opened 2026-09-02
perf(rust/sedona-functions): replace ST_Collect_Agg's per-group hash sets with the type/dimension bitset

james-willis:
## What this PR does

`ST_Collect_Agg`'s `Accumulator::size()` under-reported memory: it counted `size_of::<CollectionAccumulator>() + item.capacity()` but omitted the heap of `unique_geometry_types: HashSet<GeometryTypeId>` and `unique_dimensions: HashSet<Dimensions>` (~100 bytes/group on first insert, measured). Since a UDAF accumulator is instantiated per group, a high-cardinality `GROUP BY` hides ~100 bytes/group from the memory pool — on SpatialBench Q5 at sf=10 (10M+ groups, 12 GiB container, 7.73 GB fair pool) the query peaked at 9.55 GB anon with **zero spills and no `ResourcesExhausted`**, so survival was luck rather than the pool's decision.

This replaces both HashSets with the existing `GeometryTypeAndDimensionsSet` u32 bitset (`rust/sedona-geometry/src/types.rs`):

- `size()` becomes exact with no extra terms;
- two heap allocations per group disappear (~170 bytes/group; multiple GB at sf=10 group counts);
- the per-row `convert_to_state` path (used when skip-partial-aggregation engages at high cardinality) sheds two malloc/free pairs per row.

## Notes

- **State**: the accumulator's intermediate state is the raw bitset `u32` (a single `Int64` column) and `merge_batch` is a bitwise OR. (Earlier revisions kept the pre-bitset JSON wire format; dropped per review, since aggregate intermediate state is ephemeral within a query execution and never persisted or merged across versions.)
- **Semantics**: uses `insert()`, not `insert_or_ignore()`, so a mixed/unknown-dimension group errors rather than silently emitting an XY header. A mixed-dimensions test is included.

## Testing

Full `sedona-functions` and `sedona-geometry` suites green. End-to-end memory validation on our deployment is in progress (with exact accounting, a ~500 MB collect under a 512 MiB pool must now spill or `ResourcesExhausted` instead of silently overrunning RSS); will report here.

## Follow-ups (not in this PR)

- DataFusion `GroupsAccumulatorAdapter`: per-group `indices` Vec capacity never enters `allocation_bytes`, plus a small inline double-count — to be proposed upstream separately.
- A native `GroupsAccumulator` for `ST_Collect_Agg` (shared buffer + per-group offsets) to remove the remaining per-group adapter overhead and re-enable partial aggregation.
- `ItemCrsAccumulator` has the same class of gap (per-group `crs: Option<String>` uncounted).

### DIFF
--- a/rust/sedona-functions/src/st_collect_agg.rs
+++ b/rust/sedona-functions/src/st_collect_agg.rs
@@ -22,7 +22,7 @@ use arrow_schema::{DataType, Field, FieldRef};
 use datafusion_common::{
     cast::{as_binary_array, as_int64_array, as_string_array},
     error::{DataFusionError, Result},
-    exec_err, HashSet, ScalarValue,
+    exec_err, ScalarValue,
 };
 use datafusion_expr::{Accumulator, ColumnarValue, Volatility};
 use geo_traits::Dimensions;
@@ -32,7 +32,7 @@ use sedona_expr::{
     item_crs::ItemCrsSedonaAccumulator,
 };
 use sedona_geometry::{
-    types::{GeometryTypeAndDimensions, GeometryTypeId},
+    types::{GeometryTypeAndDimensions, GeometryTypeAndDimensionsSet, GeometryTypeId},
     wkb_factory::{
         write_wkb_geometrycollection_header, write_wkb_multilinestring_header,
         write_wkb_multipoint_header, write_wkb_multipolygon_header,
@@ -97,8 +97,7 @@ impl SedonaAccumulator for STCollectAggr {
 #[derive(Debug)]
 struct CollectionAccumulator {
     input_type: SedonaType,
-    unique_geometry_types: HashSet<GeometryTypeId>,
-    unique_dimensions: HashSet<Dimensions>,
+    types_and_dims: GeometryTypeAndDimensionsSet,
     count: i64,
     item: Option<Vec<u8>>,
 }
@@ -116,8 +115,7 @@ impl CollectionAccumulator {
 
         Ok(Self {
             input_type,
-            unique_geometry_types: HashSet::new(),
-            unique_dimensions: HashSet::new(),
+            types_and_dims: GeometryTypeAndDimensionsSet::new(),
             count: 0,
             item: Some(item),
         })
@@ -134,13 +132,15 @@ impl CollectionAccumulator {
         let mut new_header = Vec::new();
         let count_usize = self.count.try_into().unwrap();
 
-        if self.unique_dimensions.len() != 1 {
+        let unique_dimensions = self.types_and_dims.dimensions();
+        if unique_dimensions.len() != 1 {
             return exec_err!("Can't ST_Collect_Agg() mixed dimension geometries");
         }
 
-        let dimensions = *self.unique_dimensions.iter().next().unwrap();
-        if self.unique_geometry_types.len() == 1 {
-            match self.unique_geometry_types.iter().next().unwrap() {
+        let dimensions = unique_dimensions[0];
+        let unique_geometry_types = self.types_and_dims.geometry_types();
+        if unique_geometry_types.len() == 1 {
+            match unique_geometry_types[0] {
                 GeometryTypeId::Point => {
                     write_wkb_multipoint_header(&mut new_header, dimensions, count_usize)
                         .map_err(|e| DataFusionError::External(Box::new(e)))?;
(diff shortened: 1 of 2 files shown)


rust/sedona-functions/src/st_collect_agg.rs
```diff
@@ -275,10 +281,24 @@ impl Accumulator for CollectionAccumulator {
...
+                    for geometry_type in &geometry_types {
+                        for dimensions in &dimensions {
+                            self.types_and_dims
+                                .insert(&GeometryTypeAndDimensions::new(
+                                    *geometry_type,
+                                    *dimensions,
+                                ))
+                                .map_err(|e| DataFusionError::External(Box::new(e)))?;
+                        }
+                    }
```
paleolimbot: How about just storing the 64-bit integer as the state? Otherwise, this comment needs to be rewritten for humans.

rust/sedona-functions/src/st_collect_agg.rs
```diff
@@ -205,14 +208,17 @@ impl Accumulator for CollectionAccumulator {
     }
 
     fn state(&mut self) -> Result<Vec<ScalarValue>> {
-        let geometry_types_value =
-            serde_json::to_string(&self.unique_geometry_types.iter().collect::<Vec<_>>())
-                .map_err(|e| DataFusionError::External(Box::new(e)))?;
+        // Both columns keep the exact pre-bitset wire format: a JSON list of
+        // geometry types, and a JSON list of dimensions wrapped as
+        // (Geometry, dimensions) pairs.
+        let geometry_types_value = serde_json::to_string(&self.types_and_dims.geometry_types())
```
paleolimbot: I don't think the wire format needs to be maintained?

rust/sedona-geometry/src/types.rs
```diff
@@ -429,6 +429,28 @@ impl GeometryTypeAndDimensionsSet {
...
+    /// The distinct dimensions in this set, in bitset iteration order.
+    pub fn dimensions(&self) -> Vec<Dimensions> {
+        let mut out = Vec::new();
+        for item in self.iter() {
+            if !out.contains(&item.dimensions()) {
+                out.push(item.dimensions());
+            }
+        }
+        out
+    }
```
paleolimbot: These seem like inefficient ways (particular vector contains) to do these...I think you can probably create bitmasks and & them against the current value if you truly need these.

rust/sedona-functions/src/st_collect_agg.rs
```diff
@@ -525,4 +545,61 @@ mod test {
             "CRS values not equal: ogc:crs84 vs epsg:3857"
         );
     }
+
+    /// The bitset swap must not change the serialized state wire format:
+    /// column 0 is a JSON list of geometry types and column 1 a JSON list of
+    /// dimensions wrapped as (Geometry, dims) pairs, exactly as the HashSet
+    /// implementation produced, so states merge across versions.
+    #[test]
+    fn state_wire_format_is_unchanged() {
```
paleolimbot: States merging across versions sounds fishy. Am I missing something about how this can be run?

paleolimbot (commented):
Cool!

rust/sedona-geometry/src/types.rs
james-willis: done. probably doesnt really matter.

I added them to the class because I assumed itd be useful for future users and doesn't really cost anything. seems like a better home than in the UDF impl

rust/sedona-functions/src/st_collect_agg.rs
james-willis: Yeah I was overly cautious trying to maintain the wire format. I can't think of a reason that it'd be neccessary.

Serializing to use the u32 simplifies all the code here.

rust/sedona-functions/src/st_collect_agg.rs
james-willis: agreed. this really simplifies the code here. thanks

rust/sedona-functions/src/st_collect_agg.rs
james-willis: dropped this whole story by just changing the wire format. Agree that theres no reason to be consistent across version about it. AFAIK all uses of it are ephemeral

rust/sedona-geometry/src/types.rs
```diff
@@ -429,6 +441,37 @@ impl GeometryTypeAndDimensionsSet {
...
+            (0x0000_00FF, Dimensions::Xy),
+            (0x0000_FF00, Dimensions::Xyz),
+            (0x00FF_0000, Dimensions::Xym),
+            (0xFF00_0000_u32, Dimensions::Xyzm),
+        ]
+        .into_iter()
+        .filter(|(mask, _)| self.types & mask != 0)
+        .map(|(_, dim)| dim)
+        .collect()
+    }
```
paleolimbot, line 474: If you are going to add these two functions you should check the matrix of geometry type by dimensions that each pair correctly returns the correct list-of-one GeometryType and Dimensions (I don't think there's anything making sure the bits are correct at the moment)

paleolimbot (approved):
Thank you!

Just the one test and then this is good to go.

rust/sedona-geometry/src/types.rs
james-willis: done

PR #1226, opened 2026-09-02
chore: upgrade DataFusion to 54.1.0

### LINKED ISSUE #1204: Update to DataFusion 54
I'd like to update the DataFusion version before the next release! We could do DataFusion 54 or 55; however, updating is a bit of a pain and there are some DataFusion 55 features (notably: arrow-rs that no longer borks GeoArrow types when reading/writing Parquet; finer grained parallelism in the Parquet reader; struct pushdown) that might be nice and there's a patch release that will likely be live before we're set to release our next version to fix any earlier discovered bugs.

paleolimbot:
Upgrades DataFusion from 52.5.0 to 54.1.0 with the corresponding Arrow, Parquet, object_store, GeoArrow, and Zarr versions. Includes initial trait, downcast, execution-plan, and metadata-cache API migrations.

A few larger updates:

- The Parquet reader now uses a "moreslizer" instead of an opener; however, conceptually this fits well with what we were doing already so the change is minimal.
- Logical serde was updated, so our placeholder approach had to be tweaked.
- A few metadata bugs were encountered, both worked around: a subquery getting optimized into a join borks on metadata (so I disabled/tracked) and one logical optimizer rule can't handle unnest + select (which is in our doctests for what you normally do with ST_Dump).

The big diff is mostly the lockfile for the examples.

Closes #1204.


zhangfengcdt (approved):

rust/sedona/src/context.rs
```diff
@@ -100,7 +103,11 @@ impl SedonaContext {
     pub fn new() -> Self {
         // This will panic only if the default build settings are
         // incorrect which we test!
-        Self::new_from_context(SessionContext::new()).unwrap()
+        let state_builder = SessionStateBuilder::new_with_default_features();
+        // DataFusion #22620 workaround tracked by
+        // https://github.com/apache/sedona-db/issues/1232.
+        let state_builder = register_vendored_optimizer_rules(state_builder).unwrap();
```
jiayuasu, line 109: Could this also be applied in `new_from_context()`? Passing a regular `SessionContext` still hits the Unnest bug. This test fails in `push_down_leaf_projections`; changing the constructor to `SedonaContext::new()` returns the expected two rows.

Repro: save as `rust/sedona/tests/supplied_context_dump.rs`, then run `cargo test -p sedona --test supplied_context_dump`.

```rust
#[tokio::test]
async fn supplied_context_dump() -> datafusion::error::Result<()> {
    use datafusion::{
        functions::core::expr_fn::get_field,
        prelude::{col, SessionContext},
    };
    use sedona::context::SedonaContext;

    let ctx = SedonaContext::new_from_context(SessionContext::new())?;
    let batches = ctx
        .sql("SELECT ST_Dump(ST_GeomFromText('MULTIPOINT (0 0, 1 1)')) AS dump")
        .await?
        .unnest_columns(&["dump"])?
        .select(vec![get_field(col("dump"), "geom").alias("geometry")])?
        .collect()
        .await?;
    assert_eq!(batches.iter().map(|b| b.num_rows()).sum::<usize>(), 2);
    Ok(())
}
```

rust/sedona-pointcloud/src/las/statistics.rs
```diff
@@ -33,7 +33,7 @@ use byteorder::{LittleEndian, ReadBytesExt};
 use datafusion_common::{arrow::compute::concat_batches, Column, DataFusionError, ScalarValue};
 use datafusion_pruning::PruningStatistics;
 use las::Header;
-use object_store::{path::Path, ObjectMeta, ObjectStore, PutPayload};
+use object_store::{path::Path, ObjectMeta, ObjectStore, ObjectStoreExt, PutPayload};
```
jiayuasu, line 36: Could we add the zero-point guard to `extract_chunk_stats()` too? An empty LAZ file using point format 6 has a `477..477` chunk. Normal reading now works, but `collect_statistics=true` still fails with `Requested range was invalid`. The old object_store 0.12.4 accepted that empty range.

Repro: save as `rust/sedona-pointcloud/tests/empty_laz_statistics.rs`, then run `cargo test -p sedona-pointcloud --test empty_laz_statistics`. The final call fails; setting `collect_statistics` to `false` succeeds.

```rust
#[tokio::test]
async fn empty_laz_statistics() {
    use las::{point::Format, Builder, Writer};
    use object_store::{local::LocalFileSystem, path::Path, ObjectStoreExt};
    use sedona_pointcloud::las::{metadata::LasMetadataReader, options::LasOptions};

    let dir = tempfile::tempdir().unwrap();
    let path = dir.path().join("empty.laz");
    let mut builder = Builder::from((1, 4));
    builder.point_format = Format::new(6).unwrap();
    builder.point_format.is_compressed = true;
    let mut writer = Writer::from_path(&path, builder.into_header().unwrap()).unwrap();
    writer.close().unwrap();

    let store = LocalFileSystem::new();
    let location = Path::from_filesystem_path(&path).unwrap();
    let object = store.head(&location).await.unwrap();
    let metadata = LasMetadataReader::new(&store, &object)
        .fetch_metadata().await.unwrap();
    assert_eq!(metadata.header.number_of_points(), 0);

    let mut options = LasOptions::default();
    options.collect_statistics = true;
    LasMetadataReader::new(&store, &object)
        .with_options(options)
        .fetch_metadata()
        .await
        .unwrap();
}
```

examples/sedonadb-rust/Cargo.lock
```diff
@@ -4414,26 +5390,28 @@ dependencies = [
 
 [[package]]
 name = "sedona"
-version = "0.4.0"
-source = "git+https://github.com/apache/sedona-db.git#9d6bcfa178a341e23f40d58a7ec6244f7b5dca25"
+version = "0.5.0"
+source = "git+https://github.com/apache/sedona-db.git#7b2864a996cd14edde6bb2c8d451ca5df52756fe"
```
jiayuasu, line 5394: This revision still uses DataFusion 52.5, while the example now depends on 54.1. Could we regenerate the lockfile against a Sedona revision containing the upgrade?

From this PR checkout:

```sh
cd examples/sedonadb-rust
cargo tree --locked --depth 1 -p sedonadb-rust-example
cargo tree --locked --depth 1 -p sedona
cargo check --locked
```

The first tree shows DataFusion 54.1; the second shows 52.5. The example then passes a 54.1 `Expr` from `col("name")` to `sort_by`, which expects a 52.5 `Expr` (`E0308`). The `?` calls also have incompatible `DataFusionError` types. CI rewrites the Sedona dependency to the PR head before building, so it does not test this committed lockfile.

rust/sedona/src/context.rs
```diff
@@ -255,31 +265,57 @@ impl SedonaContext {
...
+
+        Ok(out)
     }
 
     /// Creates a new context from a previously configured DataFusion context
     pub fn new_from_context(ctx: SessionContext) -> Result<Self> {
+        // DataFusion #22620 workaround tracked by
+        // https://github.com/apache/sedona-db/issues/1232.
+        let state_builder = register_vendored_optimizer_rules(ctx.into_state_builder())?;
+        let ctx = SessionContext::new_with_state(state_builder.build());
```
jiayuasu: Could we keep the existing session state here? Rebuilding it leaves the URL-table catalog pointing to the old state, which has been dropped. Moving the catalog setup fixes the interactive constructor, but a caller-supplied context still breaks.

Repro: save as `rust/sedona/tests/supplied_url_table.rs`, then run `cargo test -p sedona --test supplied_url_table`. The query succeeds before wrapping and fails afterward with `get current SessionStore error`.

```rust
#[tokio::test]
async fn supplied_url_table() -> datafusion::error::Result<()> {
    use datafusion::prelude::SessionContext;
    use sedona::context::SedonaContext;

    let dir = tempfile::tempdir()?;
    let path = dir.path().join("example.csv");
    std::fs::write(&path, "x\n1\n2\n")?;
    let sql = format!("SELECT * FROM '{}'", path.display());
    let ctx = SessionContext::new().enable_url_table();
    ctx.sql(&sql).await?.collect().await?; // succeeds

    let ctx = SedonaContext::new_from_context(ctx)?;
    ctx.ctx.sql(&sql).await?.collect().await?; // fails
    Ok(())
}
```

rust/sedona/src/context.rs
```diff
@@ -255,31 +265,57 @@ impl SedonaContext {
...
+            )));
+
+        Ok(out)
     }
 
     /// Creates a new context from a previously configured DataFusion context
     pub fn new_from_context(ctx: SessionContext) -> Result<Self> {
+        // DataFusion #22620 workaround tracked by
+        // https://github.com/apache/sedona-db/issues/1232.
+        let state_builder = register_vendored_optimizer_rules(ctx.into_state_builder())?;
```
jiayuasu: Could this accept a context where the caller has already disabled the rule? There is nothing to replace in that case, but construction now fails with `PushDownLeafProjections rule not found in default optimizer rules`.

Repro: save as `rust/sedona/tests/disabled_leaf_projection.rs`, then run `cargo test -p sedona --test disabled_leaf_projection`.

```rust
#[tokio::test]
async fn disabled_leaf_projection() -> datafusion::error::Result<()> {
    use datafusion::prelude::SessionContext;
    use sedona::context::SedonaContext;

    let ctx = SessionContext::new();
    assert!(ctx.remove_optimizer_rule("push_down_leaf_projections"));
    SedonaContext::new_from_context(ctx)?;
    Ok(())
}
```

rust/sedona-pointcloud/src/las/metadata.rs
```diff
@@ -379,6 +394,10 @@ async fn laz_chunk_table(
     let laz_vlr = header.laz_vlr()?;
 
     let num_points = header.number_of_points();
+    if num_points == 0 {
+        return Ok(Vec::new());
```
jiayuasu, line 398: Could we skip or regenerate the old statistics when the file has no points? The previous reader gave an empty format-0 LAZ one zero-point chunk and persisted one statistics row. Returning no chunks here makes `chunk_statistics()` panic when it reuses that sidecar: `statistics.rs:232`, `left: 1, right: 0`.

Repro: save this as `rust/sedona-pointcloud/tests/empty_laz_sidecar.rs` in both checkouts. On `69c8a4c0`, run:

```sh
export EMPTY_LAZ_PROBE="$(mktemp -d)/empty.laz"
cargo test -p sedona-pointcloud --test empty_laz_sidecar
```

Then run the same Cargo command on `10335895` in the same shell, keeping the LAZ file and its `.stats` sidecar. The old reader succeeds; the new reader panics.

```rust
#[tokio::test]
async fn empty_laz_statistics_reuse() {
    use las::{point::Format, Builder, Writer};
    use object_store::{local::LocalFileSystem, path::Path, ObjectStoreExt};
    use sedona_pointcloud::las::{metadata::LasMetadataReader, options::LasOptions};

    let path = std::path::PathBuf::from(std::env::var("EMPTY_LAZ_PROBE").unwrap());
    if !path.exists() {
        let mut builder = Builder::from((1, 4));
        builder.point_format = Format::new(0).unwrap();
        builder.point_format.is_compressed = true;
        let mut writer = Writer::from_path(&path, builder.into_header().unwrap()).unwrap();
        writer.close().unwrap();
    }

    let store = LocalFileSystem::new();
    let location = Path::from_filesystem_path(&path).unwrap();
    let object = store.head(&location).await.unwrap();
    let mut options = LasOptions::default();
    options.collect_statistics = true;
    options.persist_statistics = true;
    LasMetadataReader::new(&store, &object)
        .with_options(options)
        .fetch_metadata()
        .await
        .unwrap();
}
```

rust/sedona/src/context.rs
paleolimbot: I just made new_from_context private given the complexity of what we need to do to construct it at this point.

rust/sedona-pointcloud/src/las/metadata.rs
paleolimbot: I fixed this but also updated the arithmetic to use checked variants so the failure mode of a corrupted sidecar is not a panic!

PR #1228, opened 2026-09-03
chore(rust): upgrade geo to 0.33.1

### LINKED ISSUE #491: rust/sedona-geo: Upgrade to geo 0.32

paleolimbot:
Uses the dependency-free ranked-query callbacks from georust/geo-index#161, pinned to its current commit, to remove geo-index's geo 0.31 feature. Upgrades geo to 0.33.1, migrates concave-hull options, updates Triangle accessors, and preserves KNN behavior with conservative spherical pruning.

Closes #455
Closes #491
Replaces #566

### DIFF
--- a/rust/sedona-geo-generic-alg/benches/intersection.rs
+++ b/rust/sedona-geo-generic-alg/benches/intersection.rs
@@ -413,7 +413,7 @@ fn point_triangle_intersection(c: &mut Criterion) {
             }
 
             assert_eq!(intersects, 533);
-            assert_eq!(non_intersects, 5450151);
+            assert_eq!(non_intersects, 5484911);
         });
     });
 
--- a/rust/sedona-geo/src/st_concavehull.rs
+++ b/rust/sedona-geo/src/st_concavehull.rs
@@ -21,6 +21,7 @@ use arrow_array::builder::BinaryBuilder;
 use datafusion_common::error::Result;
 use datafusion_common::{cast::as_float64_array, DataFusionError};
 use datafusion_expr::ColumnarValue;
+use geo::concave_hull::ConcaveHullOptions;
 use geo::{ConcaveHull, CoordsIter, Geometry, GeometryCollection, Point, Polygon};
 use geo_traits::to_geo::{ToGeoGeometry, ToGeoPoint};
 use geo_traits::{GeometryCollectionTrait, GeometryTrait, MultiPointTrait, PointTrait};
@@ -41,6 +42,10 @@ use crate::to_geo::item_to_geometry;
 /// Geo returns a Polygon for every concave hull computation
 /// whereas the Geos implementation returns a MultiPolygon for
 /// certain geometries concave hull computation.
+///
+/// Note that this does *not* match the PostGIS implementation.
+/// In particular, the GEOS/PostGIS "pctconvex" parameter, which extends
+/// from 0..1 does not match this kernel's "concavity" (0..Infinity).
 pub fn st_concavehull_impl() -> Vec<ScalarKernelRef> {
     ItemCrsKernel::wrap_impl(STConcaveHull {})
 }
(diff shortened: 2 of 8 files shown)


rust/sedona-geo-generic-alg/benches/intersection.rs
```diff
@@ -413,7 +413,7 @@ fn point_triangle_intersection(c: &mut Criterion) {
             }
 
             assert_eq!(intersects, 533);
-            assert_eq!(non_intersects, 5450151);
+            assert_eq!(non_intersects, 5484911);
```
paleolimbot, line 416: This I believe is due to an update to the earcut algorithm, not to behaviour here.

rust/sedona-geo/src/st_concavehull.rs
```diff
@@ -327,11 +340,11 @@ mod tests {
             ),
             (
                 "MULTIPOINT ((0 0), (10 0), (0 10), (10 10), (5 5))",
-                "POLYGON ((10 0, 10 10, 0 10, 0 0, 5 5, 10 0))",
+                "POLYGON ((10 0, 10 10, 0 10, 0 0, 10 0))",
             ),
             (
                 "MULTILINESTRING ((10 10, 20 20, 10 40), (40 40, 30 30, 40 20, 30 10))",
-                "POLYGON ((20 20, 10 40, 30 30, 40 40, 40 20, 30 10, 10 10, 20 20))",
+                "POLYGON ((30 10, 40 20, 40 40, 10 40, 10 10, 30 10))",
```
paleolimbot, line 347: This is a change to output geometry for this kernel; however, this kernel is not used by default because it's parameterized differently and we can't get PostGIS behaviour alignment.

rust/sedona-spatial-join/src/index/default_spatial_index.rs
```diff
@@ -339,30 +336,32 @@ impl SpatialIndex for DefaultSpatialIndex {
...
+                    let bbox = Geometry::Rect(Rect::new(
+                        coord! { x: min_x as f64, y: min_y as f64 },
+                        coord! { x: max_x as f64, y: max_y as f64 },
+                    ));
+                    Euclidean.distance(&probe_geom, &bbox)
+                }
+            },
+            |item_index, _bbox| {
+                geometry_accessor.distance(&probe_geom, item_index as usize, use_spheroid)
+            },
```
paleolimbot, line 364: This is the main update (did not require any test changes)

rust/sedona-spatial-join/src/index/default_spatial_index.rs
```diff
@@ -339,30 +336,32 @@ impl SpatialIndex for DefaultSpatialIndex {
...
-        let initial_results = self.inner.rtree.neighbors_geometry(
-            &probe_geom,
-            Some(k as usize),
-            None, // no max_distance filter
-            distance_metric,
-            &geometry_accessor,
+        // Use dependency-free callbacks from geo-index. A zero lower bound is required for
+        // spherical distances because planar longitude/latitude boxes do not provide a valid
+        // Haversine lower bound. Planar queries use the exact geometry-to-rectangle distance.
+        let initial_results = self.inner.rtree.neighbors_with_callbacks(
```
zhangfengcdt, line 345: Understood returning 0 is safe and yes the old estimate is not a strictly valid spherical lower bound. 

But, we might consider using great-circle distance - it is never less than the latitude difference, so R * |lat_probe - clamp(lat_probe, min_lat, max_lat)| in radians is a valid non-zero bound for point probes. It will help for most of the pruning. 

see this for reference:
https://github.com/google/s2geometry/blob/079611b654ad89afd9c3c3a1796d64bdd6a6b340/src/s2/s2latlng_rect.cc#L571

zhangfengcdt (approved):

rust/sedona-spatial-join/src/index/default_spatial_index.rs
paleolimbot: I added https://github.com/apache/sedona-db/issues/1308 for this one...perhaps you could update that since I have no experience with that particular bit of code 😬

PR #1229, opened 2026-09-03
fix(python/sedonadb): resolve GeoArrow scalars with their edge type, and common scalar values

### LINKED ISSUE #1214: lit() should resolve GeoArrow scalars with their edge type, and common missing/temporal scalars
## Summary

`lit()` (`python/sedonadb/python/sedonadb/expr/literal.py`) resolves Python values through a class-name → handler table plus a `pa.array([obj])` fallback. While building the `sedonadb-geopandas` assignment surface (#1195) a number of common scalar values turned out to need normalization *before* they can reach `lit()`, and the workarounds currently live in that package. Most of them would be more useful in `lit()` itself, where every SedonaDB user benefits. Listing them here so they can be picked off independently.

## GeoArrow scalars

- `_lit_from_geoarrow_scalar` rebuilds the type as `ga.wkb().with_crs(crs)`, which **drops the edge type**: a spherical (geography) WKB scalar becomes planar geometry.
  ```python
  sph = ga.wkb().with_edge_type(ga.EdgeType.SPHERICAL).with_crs(crs.to_json())
  df.select(lit(pa.scalar(point.wkb, sph)))   # -> geometry<Wkb(epsg:4267)>, edge type lost
  ```
- Only `geoarrow.pyarrow._scalar.WkbScalar` is registered, so scalars of other GeoArrow types (`WktScalar`, native point/linestring/polygon scalars) fall through to `pa.array([obj])` and fail with `Can't create SedonaDB literal from object of type geoarrow.pyarrow._scalar.WktScalar`. They all expose `.wkb`, so the same WKB path could serve them.

## Missing-value sentinels and temporal scalars

| value | today |
|---|---|
| `pandas.NA` | `ValueError` (unrecognized type) |
| `pandas.NaT` | `ValueError` — it is neither a `Timestamp` nor a `Timedelta` instance |
| `pandas.Timestamp` / `pandas.Timedelta` | resolve at **microsecond** resolution; nanoseconds are silently dropped (`Timedelta(1)` becomes `0`) |
| `numpy.datetime64` / `numpy.timedelta64` in non-Arrow units (`D`, `W`, `Y`, …) | `ValueError` — a lossless conversion to `s` exists for these |
| `numpy.ma.masked` | `ValueError` — it means "missing" |
| 0-d `numpy.ndarray` | `ValueError` — it is one value |
| structured `numpy.void` | `ValueError` — representable as a typed Arrow struct |
| typed-null nested Arrow scalar (`pa.scalar(None, pa.list_(...))`) | `ValueError` — the one-element array spelling `pa.array([None], type)` works |

The pandas cases can be handled the way the table already handles pandas frames — by qualified class name — without importing pandas.

## Related

- #1195 carries the current workarounds in `sedonadb_geopandas/_series.py` (`normalize_scalar`) and `_temporal.py`; each item above that lands here lets that code shrink to a pass-through.
- LargeBinary-backed WKB is a separate engine-side issue (WKB import rejects LargeBinary storage) and is tracked on its own.

jiayuasu:
## What changes

Closes #1214.

`lit()` resolves Python values through a class-name → handler table with a `pa.array([obj])` fallback. Two gaps in the GeoArrow handling and a set of common scalar values that the fallback rejects (or handles lossily) came up while building the `sedonadb-geopandas` assignment surface (#1195), where they are currently worked around before the value reaches `lit()`. Handling them here benefits every SedonaDB user, and lets that wrapper code shrink to a pass-through later.

**GeoArrow scalars**

- The type was rebuilt from the CRS alone, so a spherical (geography) WKB scalar came back as planar geometry. The edge type now travels with the CRS.
- Only the WKB scalar class was registered; WKT and native-encoded (point, linestring, …) scalars fell through to `pa.array([obj])` and failed. Every GeoArrow scalar exposes its WKB, so one handler now serves them all, valid or null.

**Scalar values**

| value | before | after |
|---|---|---|
| `pandas.NA`, `numpy.ma.masked` | `ValueError` | NULL |
| `pandas.NaT` | `ValueError` | timestamp NULL |
| `pandas.Timestamp` / `Timedelta` | truncated to microseconds | own resolution, time zone kept |
| `numpy.datetime64` / `timedelta64` in non-Arrow units (`D`, `W`, `Y`, …) | `ValueError` | lossless Arrow resolution; ambiguous or lossy units rejected the way pandas rejects them, with an overflow check |
| 0-d `numpy.ndarray` | `ValueError` | its typed scalar (dtype kept) |
| structured `numpy.void` | `ValueError` | typed struct |
| null nested / extension Arrow scalar | `ValueError` | typed one-element array |

pandas 3 renamed these classes (`pandas.api.typing.NAType`, `pandas.Timestamp`, …), so both sets are registered, the way the table already handles `DataFrame` and `Series`. Existing paths (valid Arrow scalars, arrays with dimensions, shapely, GeoSeries, pandas frames) are unchanged.

## Testing

23 new tests in `tests/expr/test_literal.py`, run with warnings promoted to errors against pandas 2.3 / NumPy 2.3 and pandas 3.0 / NumPy 2.5. The full `expr/`, `test_dataframe.py`, and `test_context.py` suites pass unchanged.

### DIFF
--- a/python/sedonadb/python/sedonadb/expr/literal.py
+++ b/python/sedonadb/python/sedonadb/expr/literal.py
@@ -134,6 +134,11 @@ def _resolve_arrow_lit(obj: Any):
 
     import pyarrow as pa
 
+    # A null Arrow scalar of a nested or extension type is not accepted by
+    # pa.array([obj]); its one-element array spelling carries the same type.
+    if isinstance(obj, pa.Scalar) and not obj.is_valid:
+        return pa.array([None], type=obj.type)
+
     try:
         return pa.array([obj])
     except Exception as e:
@@ -143,8 +148,11 @@ def _resolve_arrow_lit(obj: Any):
 
 
 def _lit_from_geoarrow_scalar(obj):
+    # Every GeoArrow scalar (WKB, WKT, native point/linestring/...) exposes
+    # its WKB, so one path serves them all. The edge type travels with the
+    # CRS: a spherical (geography) scalar must not come back planar.
     wkb_value = None if obj.value is None else obj.wkb
-    return _lit_from_wkb_and_crs(wkb_value, obj.type.crs)
+    return _lit_from_wkb_and_crs(wkb_value, obj.type.crs, obj.type.edge_type)
 
 
 def _lit_from_dataframe(obj):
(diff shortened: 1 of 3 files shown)


python/sedonadb/python/sedonadb/expr/literal.py
```diff
@@ -191,15 +199,137 @@ def _lit_from_shapely(obj):
     return _lit_from_wkb_and_crs(obj.wkb, None)
 
 
-def _lit_from_wkb_and_crs(wkb, crs):
+def _lit_from_wkb_and_crs(wkb, crs, edge_type=None):
```
james-willis: function name probably needs an update

python/sedonadb/python/sedonadb/expr/literal.py
```diff
@@ -191,15 +199,137 @@ def _lit_from_shapely(obj):
...
+    return pa.array([None])
+
+
+def _lit_from_nat(obj):
+    # NaT is a datetime missing value in pandas (assigning it yields a
+    # datetime64 column), so it resolves to a typed timestamp null rather
+    # than an untyped NULL.
+    import pyarrow as pa
+
+    return pa.array([None], pa.timestamp("ns"))
```
james-willis, line 228: how do we know it should be ns? I supposed it doesn't matter since the system can cast it to other TS types as needed in the query?

james-willis (approved):

python/sedonadb/python/sedonadb/expr/literal.py
jiayuasu: Done, it's `_lit_from_wkb` now (named for what it builds from rather than its argument list, since the edge type joined the CRS).

python/sedonadb/python/sedonadb/expr/literal.py
jiayuasu: Right on both counts. `NaT` itself carries no unit (it's a unit-less singleton), so some unit has to be picked, and nanoseconds is where pandas stores it and the finest one, so it can never be a lossy choice. And it does coerce: `coalesce(us_col, lit(pd.NaT))` resolves to `timestamp(µs)` and `coalesce(s_tz_col, lit(pd.NaT))` to `timestamp(s, UTC)`, so the null takes the surrounding expression's unit and zone. Added that as a comment on the function.

PR #1230, opened 2026-09-03
test(python): skip raster helper tests without rasterio

jiayuasu:
## What

Skip the four raster helper tests that require optional rasterio when it is not installed. Keep test_decoded_raster_requires_nodata running because it does not use rasterio.

## Why

The python-wheels workflow installs a reduced dependency set without rasterio. The tests added in #1211 currently fail after wheel builds complete, causing all five core jobs to fail at the final test marker. Standard Python CI installs the test extra, so these tests still run there.

Failing run: https://github.com/apache/sedona-db/actions/runs/33675876705

## Verification

* Without rasterio: 1 passed, 4 skipped
* With rasterio: 5 passed
* Ruff check and format check
* git diff --check

### DIFF
--- a/python/sedonadb/tests/test_raster_testing.py
+++ b/python/sedonadb/tests/test_raster_testing.py
@@ -37,18 +37,21 @@ BBOX = (100.0, 482.0, 114.0, 500.0)
 
 
 def test_write_geotiff_bbox_places_the_grid(tmp_path):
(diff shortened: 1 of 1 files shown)


james-willis (approved):
Thanks Jia this was my mistake

All of `python/sedonadb/tests` should run without rasterio.

PR #1236, opened 2026-09-04
test(integration): metadata accessor parity, one module per RS_ function

james-willis:
Probe-first parity for the eleven metadata scalar accessors (RS_Width, RS_Height, RS_NumBands, RS_ScaleX/Y, RS_SkewX/Y, RS_UpperLeftX/Y, RS_SRID, RS_BandPixelType), one module per function, RS-only SQL. Every value probed bit-identical across engines — standard, sheared (raw-transform), bottom-up, and EPSG:3857 fixtures; pixel-type names identical for all six GeoTIFF dtypes; band addressing checked on band 2. `17 passed`, all anchored, no xfails or tolerances. Locally verified with pyspark 4.0.4 + Sedona 1.9.1; CI runs the parity lane.


paleolimbot (approved):
🚀

Issue #1130, opened 2026-08-07
Consider polar projections in Raster spatial join bounding box calculation


paleolimbot:
Filing this as an issue so we don't forget!

> Extent reprojection requires internal points across the whole domain, not just the boundary. I can demonstrate if desired, my approach is in reproj::reproj_extent  

 _Originally posted by @mdsumner in [#1126](https://github.com/apache/sedona-db/pull/1126/changes/13f834a1ee048e97f894ba016bb25e50a0e1c87b..497e36bdfa8b7f757fb3b49cf653006e81adb1f4#r3732660054)_

This is a great point...in addition to the example in the linked thread, any polar projection when converted to lat/lon falls in this category (e.g., the south pole is directly in the middle of some projections, so even densifying the edges is not going to work, and even including points in the middle is not going to necessarily catch the minimum). Some of the minimum values it covers might also not be coverable by the other projection (e.g., mercator projections with a sigularity at the south pole).

Probably a geography covering is more accurate than a bounding box.

mdsumner:
a coverage is a great idea ... nice convergence of problem spaces here. If we consider the polar case of mine as 1) then 

1) The pole inside a polar raster -> longlat
2) The antipode inside the source when targeting an azimuthal projection (the global EPSG:4326 boundary has nothing to do with the global Azimuthal Equal Area boundary), and many others
3) Antimeridian/seam crossing
4)  Boundary points that are invalid but the interior is fine: Geostationary full-disk (Himawari, GOES), Sinusoidal (MODIS), or any of Mollweide/Robinson/HEALPix/Goode Homolosine etc

Certainly interested to spin up a few as examples here.

paleolimbot:
> Certainly interested to spin up a few as examples here.

Examples are great! We can use them as test cases.

PR #1237, opened 2026-09-04
test(integration): raster-geometry predicate parity

james-willis:
Probe-first parity for RS_Intersects, RS_Contains, and RS_Within, one module per function. Geometry inputs use ST_GeomFromWKT as the merged zonal suite established; every output is a raw boolean, anchored — contained, overlapping, disjoint, and covering geometries all agree across engines. The geometry-returning accessors (RS_Envelope, RS_ConvexHull, RS_PixelAs*) are deliberately not here: raw geometry columns defeat the harness's tuple stringifier on both engines (SedonaDB geoarrow struct, Spark WKB), and wrapping RS outputs in ST_ machinery would make the tests depend on ST_ parity — they wait for a geometry-aware compare path. `3 passed`; CI runs the parity lane.

### Polar cross-CRS cases and the densification proof

Per review: an EPSG:3413 pole-square and an EPSG:4326 latitude-band raster, each queried from the other CRS with points that state their coordinates in both CRSs (so every expected answer is checkable against the native rectangle by hand). Truly-outside points and the polar→lat/lon-band direction agree exactly. The **densification discrepancy is proven** and xfail-cataloged with apache/sedona#3323: Sedona Spark reprojects the raster footprint corner-only, so the lon/lat chords between corners miss most of the polar cap — `POINT(45 82)` transforms to `(867972, 0)`, on the +x axis at 87% of the half-width, and still reads false — where SedonaDB densifies each footprint edge (`FOOTPRINT_POINTS_PER_EDGE`) and matches the exact point-in-square truth on every probed case. `RS_Contains` diverges identically. A pole-spanning `RS_Within` polygon is a separate xfail: flat lon/lat space degenerates there (SedonaDB answers false, Sedona Spark throws, and neither models the spherical truth).

Now `10 passed, 6 xfailed`.

### Inside-out reprojection: proving both engines wrong

Per review: cases where a region's *interior* in one CRS maps to the *edges* of another — the pole to the degenerate latitude-90 line, the antimeridian to the ±180 seam — so boundary-only reprojection turns the region inside out and **no densification can fix it**. These xfails anchor the spherical truth (`expected=True`), so they trip on the truth anchor and prove wrongness, not just mismatch:

- **Pole-spanning `RS_Within`** (the ±1e6 pole square within the latitude-75 cap): truth true; SedonaDB answers false (apache/sedona-db#1240), Sedona Spark throws (apache/sedona#3323) — *both engines wrong, in different ways*.
- **Antimeridian `RS_Within`** (a square on the longitude-180 direction within a two-lobe region): truth true; SedonaDB's densified footprint jumps the ±180 seam and reads false — the suite's first SedonaDB-wrong entry — while Sedona Spark transforms the geometry into the raster's CRS and answers true.
- `RS_Intersects` on the same antimeridian fixture agrees and anchors true for contrast.

Now `11 passed, 7 xfailed`.

### DIFF
--- a/integration/spark-parity/test_rs_within.py
+++ b/integration/spark-parity/test_rs_within.py
@@ -22,6 +22,9 @@ The geometries are placed against the standard grid (x in [100, 114],
 y in [482, 500]).
 """
 
+import pytest
+
+from sedonadb.raster_testing import write_random_geotiff
 from sedonadb.testing import SedonaDB, compare
 from sedonadb.testing_spark import SedonaSpark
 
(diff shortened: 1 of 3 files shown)


integration/spark-parity/test_rs_within.py
```diff
@@ -0,0 +1,132 @@
...
+    compare(sql, sedona, spark, expected=True)
+
+
+@pytest.mark.xfail(
+    reason="SedonaDB reprojects the densified footprint into the geometry's "
+    "CRS, and a footprint straddling the antimeridian turns inside out there "
+    "(longitudes jump across the ±180 seam), so it answers false for a "
+    "truly-within raster (apache/sedona-db#1240); Sedona Spark transforms "
+    "the geometry into the raster's CRS instead and answers true — the "
+    "suite's first case where SedonaDB is the wrong engine"
```
paleolimbot, line 104: 🤦

paleolimbot (approved):
Do these fixtures cover the cases in https://github.com/apache/sedona-db/issues/1130 where the middle of the bbox is definitely not covered in either engine's implementation of this? Or maybe we just check that it should intersect the north pole exactly?

PR #1246, opened 2026-09-09
feat(python/sedonadb-geopandas): add dissolve

jiayuasu:
## What changes

Second slice of the series that started with #1195 (column assignment). This one adds `GeoDataFrame.dissolve()` to the experimental `sedonadb-geopandas` package: group rows on one or more key columns (or on nothing, to dissolve everything into one row) and union each group's geometry.

- **Geometry aggregation** goes through `ST_Collect_Agg` followed by a `ST_UnaryUnion` in a separate projection. `ST_Union_Agg` only initializes for polygonal input (#1093), so a group of points or lines would dissolve to NULL; collect-then-union is geometry-general and yields the geometry types GeoPandas yields. The union sits in its own projection because a scalar function wrapped around an aggregate is not a valid aggregate expression. A group whose geometries are all null coalesces to an empty geometry collection re-typed back to geometry, matching GeoPandas.
- **Missing keys** follow pandas: IEEE NaN in a floating-point key is normalized to null before grouping, in both `dropna` modes, so a column holding both representations forms one missing group rather than two (and `dropna=True` drops both). Floating-ness is decided from the Arrow type, dictionary encoding unwrapped, rather than from the type's string form, so a `list<double>` column is not mistaken for something `isnan()` can be applied to.
- **Validation** matches GeoPandas: `by=[]` raises (unlike `by=None`), unknown keys raise `KeyError`, and only `aggfunc="first"` is supported for now with an error that points at the SedonaDB `DataFrame` API for anything else.
- **Temporal group keys are rejected for now.** Numpy-backed pandas delivers a missing timestamp or duration as an `INT64_MIN` tick, which must group (and `dropna`) as missing rather than as a value; that handling arrives with the temporal follow-up.

Documented differences from GeoPandas (also in the README): group keys stay ordinary columns rather than becoming the index; `"first"` is an unordered aggregate that does not skip missing values; dissolving an empty frame without keys yields one row rather than zero; a group mixing 2D and 3D geometries raises rather than being promoted to 3D; and grouping is observed-only for categorical keys, since the category domain does not survive a relational aggregation.

## Testing

18 new tests (136 in the package), including a corpus over every geometry type, GeoPandas parity for `dropna`, nested-float and dictionary-encoded keys, the all-null group, and the categorical case; run with warnings promoted to errors against both a source-built current main and the released sedonadb 0.4.1 in a clean environment (pandas 2 and 3). Ruff lint/format clean.

### DIFF
--- a/python/sedonadb-geopandas/README.md
+++ b/python/sedonadb-geopandas/README.md
@@ -34,7 +34,10 @@ big = gdf[gdf["pop"] > 1_000_000]          # boolean-mask filter
 gdf["surveyed"] = True                     # broadcast a scalar column
 gdf["centers"] = gdf.geometry.centroid     # assign a computed column
 web = gdf.to_crs("EPSG:3857")              # reproject (CRS tracked through)
-result = web.to_geopandas()                # back to a real GeoDataFrame
+
+zones = gdf.dissolve(by="region")          # group and union geometry
+
+result = zones.to_geopandas()               # back to a real GeoDataFrame
 ```
 
 ## Intentional differences from GeoPandas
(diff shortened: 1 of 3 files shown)


zhangfengcdt (approved):
LGTM

PR #1248, opened 2026-09-09
test(integration): RS_MetaData parity

james-willis:
Probe-first coverage for RS_MetaData, compared field by field (`m['field']` parses in both dialects): all twelve field names are shared and every value agrees exactly, anchored to the fixture's placement. The whole struct is not compared (harness stringifies scalars, not structs; field types differ — SedonaDB int64/unsigned vs Spark int32 — though values match). Parity-on-refusal for an unknown field; one xfail for the case-sensitivity divergence (SedonaDB's lookup is case-sensitive, Spark's is not). `13 passed, 1 xfailed`; CI runs the parity lane.

### DIFF
--- a/integration/spark-parity/test_rs_metadata.py
+++ b/integration/spark-parity/test_rs_metadata.py
@@ -1,0 +1,85 @@
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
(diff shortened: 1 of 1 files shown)


paleolimbot (approved):
This feels more like struct indexing parity rather than raster functionality. It would be nice to switch to `snake_case` since Python users tend to get motion sick reading too much `camelCase` but I get how that's a very annoying breaking change.

james-willis:
yes; its really just included for completeness. not every xfail will turn into some behavior change

Issue #1224, opened 2026-09-02
Add cache layer for raster loading


paleolimbot:
The current raster loader (or any of the implementations) make no attempt to cache multiple requests to the same object, or any version of a cache (except possibly a dataset cache in the GDAL layer, whose details I forget). In Sedona Spark I believe there are at least two levels of caching, although the per-row execution may be better suited to effectively using it (as opposed to the current execution that loops through potentially 8000 rasters before circling back to a new expression evaluation). We should implement some version of a cache before we get too far into evaluation to make sure the APIs between the loader and their implementations effectively capture an effective cache scheme.

Issue #1220, opened 2026-09-02
Raster execution with materialized values needs adaptive batch sizing to prevent out of memory


paleolimbot:
Some initial attempts at implementing a spatialbench Raster run have resulted in immediate OOMs for sf1, which is expected because we use the default batch size (8192) and materializing 8000 rasters is almost always a bad idea.

When we have an operation that will allocate a new raster for every row (e.g., ensure loaded or rasterize), we need to adaptively shrink the batch size based on some heuristic of number of bytes that will be allocated. I think this will be a physical optimizer rule inserted before a projection or filter, and in its initial form it could be crude (just make the batches defensively smaller when a projection contains a raster operation).

PR #1251, opened 2026-09-09
perf(raster): bundle RS_EnsureLoaded loader requests per batch; dedupe Zarr opens and fan out chunk reads

### LINKED ISSUE #1222: Issue load requests more than one at a time from the RasterLoader
In https://github.com/apache/sedona-db/pull/916, the raster loader is used to inject Zarr support into the engine for inline loading when required. In the PR, the interface allows for issuing multiple requests at once; however, some implementations still only issue one request at time.

This is inefficient because often load requests can be issued in batches and waited on in serial (although, there would be one serial loop in each partition). For Zarr in particular, it is particularly inefficient if multiple bands are getting fetched that happen to come from the same chunk. At the very least, issuing all band load requests at once for a single raster will probably help quite a bit and reduce the reliance on caches.

james-willis:
Closes #1222.

`AsyncRasterLoader::load` has taken a slice of requests since #916, and every backend honours it, but `RS_EnsureLoaded` still called `load(&[&req])` one band at a time. This PR groups requests to improve perf.

## Commit 1 — `source_shape` and `data` overrides for `BandRef::copy_into` (`sedona-raster`)
Some outstanding cleanup of the raster construction utilities. Simplies the band/raster creation logic later in the PR.

## Commit 2 — one `load()` per loader per batch (`sedona-raster-functions`)
Coalesces the loads to one per Loader. Measured speed up below. Peak memory is unchanged.

Driveby: add view support to this section of the code base that was rejecting non-identity views.

## Commit 3 — Zarr loader: dedupe opens, bounded fan-out (`sedona-raster-zarr`)
zarr now opens each store, array, and chunk only once. It also fans out requests in parallel inside the loader.

## Commit 4 — GDAL batch test (`sedona-raster-gdal`)
- `GdalLoader::load` already handled a slice, but every test passed one request. Adds a two-file, three-request call asserting one result per request in request order.

## Measurements

Local filesystem store, release build, 128 requests (4 arrays × 32 chunks), min of 5 runs:

| path | time |
|---|---|
| 128 × `load(&[&req])` (previous caller behaviour) | 7.8 ms |
| 1 × `load(&all)`, concurrency 1 | 3.8 ms |
| 1 × `load(&all)`, concurrency 8 | 1.8 ms |

Treat this as a lower bound: on a local FS the per-request cost being removed is a store build plus a `zarr.json` read. Against S3/HTTP each of those is a network round trip.

## Not in this PR

- Cross-call caching of stores / arrays / chunks (#1224).
- Credentialed object stores from the host `ObjectStoreRegistry`.
- Byte-bounded batches (#1220). `ideal_batch_size` (1024) now doubles as the bundle size and is left as is.

### DIFF
--- a/rust/sedona-raster-gdal/src/raster_loader.rs
+++ b/rust/sedona-raster-gdal/src/raster_loader.rs
@@ -647,4 +647,41 @@ mod tests {
         })
         .unwrap();
     }
+
+    #[tokio::test]
+    async fn gdal_loader_serves_a_batch_of_requests_in_request_order() {
+        // RS_EnsureLoaded issues every OutDb band of a batch in one call:
+        // results must come back one per request, in request order, and a
+        // file that appears twice is read twice without disturbing the
+        // others. Distinct shapes make any reordering observable.
+        let tmp = TempDir::new().unwrap();
+        let a = write_uint8_geotiff(&tmp, "a.tif"); // 2 × 3, pixels 0..6
+        let b = write_pattern_geotiff(&tmp, "b.tif", 4, 2); // 2 × 4, pixels 0..8
+        let a_uri = format!("{a}#band=1");
+        let b_uri = format!("{b}#band=1");
+        let view_a = ViewEntries::identity_for_shape(&[2, 3]);
+        let view_b = ViewEntries::identity_for_shape(&[2, 4]);
+        let req_a = RasterLoadRequest {
+            uri: &a_uri,
+            dim_names: &["y", "x"],
+            source_shape: &[2, 3],
+            view: &view_a,
+            data_type: BandDataType::UInt8,
+        };
+        let req_b = RasterLoadRequest {
+            uri: &b_uri,
+            dim_names: &["y", "x"],
+            source_shape: &[2, 4],
+            view: &view_b,
+            data_type: BandDataType::UInt8,
+        };
+
+        let loader = GdalLoader::new();
+        let results = loader.load(&[&req_a, &req_b, &req_a]).await.unwrap();
+        assert_eq!(results.len(), 3);
+        assert_eq!(results[0].bytes.as_slice(), &[0u8, 1, 2, 3, 4, 5]);
+        assert_eq!(results[1].bytes.as_slice(), &[0u8, 1, 2, 3, 4, 5, 6, 7]);
+        assert_eq!(results[2].bytes.as_slice(), &[0u8, 1, 2, 3, 4, 5]);
+        assert_eq!(results[1].source_shape, vec![2, 4]);
+    }
 }
--- a/rust/sedona-raster-zarr/src/raster_loader.rs
+++ b/rust/sedona-raster-zarr/src/raster_loader.rs
@@ -20,11 +20,16 @@
 //! Resolves a band's OutDb URI back into a Zarr chunk read: the URI is
 //! a chunk anchor of the form
 //! `<store_uri>#array=<array_path>&chunk=<i0>,<i1>,...` (see
-//! [`crate::source_uri::build_chunk_anchor`]). The loader parses the
-//! anchor, opens the Zarr store and array, and retrieves the named
-//! chunk's bytes via `zarrs` — all wrapped in
-//! `tokio::task::spawn_blocking` so the caller's async runtime is not
-//! stalled by Zarr's blocking decoder.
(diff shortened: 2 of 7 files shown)


rust/sedona-raster-zarr/src/raster_loader.rs
```diff
@@ -82,72 +177,120 @@ impl AsyncRasterLoader for ZarrLoader {
...
         }
-        Ok(results)
+
+        // Fetch each distinct chunk once, `concurrency` at a time. Results
+        // complete out of order, so carry the chunk index and slot them
+        // back afterwards. `try_collect` stops at the first error and drops
+        // whatever is still in flight.
+        // (A plain loop rather than `iter().map(|..| async move {..})`: the
+        // closure form trips rustc's "implementation of `FnOnce` is not
+        // general enough" check under async_trait's `Send` bound.)
```
james-willis, line 235: This might alleviate some of the need for a cache

rust/sedona-raster-zarr/src/raster_loader.rs
```diff
@@ -52,22 +62,107 @@ use crate::source_uri::{object_store_for_uri, open_storage_from_uri, parse_chunk
 /// (see `crate::loader`).
 pub const ZARR_FORMAT: &str = "zarr";
 
+/// Default cap on chunk reads in flight per `load` call. Every partition
+/// issues its own `load` concurrently, so this bounds the per-partition
+/// fan-out, not the process-wide one.
+pub const DEFAULT_LOAD_CONCURRENCY: usize = 8;
```
paleolimbot, line 68: I am not sure if we have the ability to wire this through (could be a follow up), but tweaking this at runtime is probably going to be needed, particularly because we probably want this to be a global IO concurrency budget rather than a per-partition one. SedonaOptions would be a good place for this config.

rust/sedona-raster-gdal/src/raster_loader.rs
```diff
@@ -647,4 +647,41 @@ mod tests {
         })
         .unwrap();
     }
+
+    #[tokio::test]
+    async fn gdal_loader_serves_a_batch_of_requests_in_request_order() {
+        // RS_EnsureLoaded issues every OutDb band of a batch in one call:
+        // results must come back one per request, in request order, and a
+        // file that appears twice is read twice without disturbing the
+        // others. Distinct shapes make any reordering observable.
```
paleolimbot, line 656: Is this true (a file that appears twice is read twice)? Or does it just appear twice in the output?

paleolimbot (approved):
Thank you!

There are a lot of bookkeeping details here...the included focused unit tests seem to do a great job covering the functionality, although some Python tests that try to do something with a few thousand tiny local multiband Zarrs and/or GeoTiffs in various randomized orders might help fuzz test this a little. No need to do that yet (would benefit all of the raster infrastructure not just this).

rust/sedona-raster-zarr/src/raster_loader.rs
james-willis: I wrote down in my backlog to look into this but im not sure if this being in an extension will be an impediment.

Ambivalent about whether it should be global vs local. hosts with more cores tend to have more network and thus maybe we automatically want those to have more requests in flight at a time?

rust/sedona-raster-gdal/src/raster_loader.rs
james-willis: load_all opens the dataset once per file via the thread-local cache, but each request runs its own read_band_blockwise into a fresh buffer, so the pixels are read twice into two buffers for two different rows.

james-willis:
added a task to my backlog for fuzzing tests

rust/sedona-raster-zarr/src/raster_loader.rs
paleolimbot: Most systems model this as global, but you could have both. Playing with this strategy at runtime is probably the best way to find out and ensure there's a workaround in case there are extremes (at 128 cores for compute you probably don't want 900 IO tasks). Very possible we need to expand the extension surface slightly to figure this out.

james-willis: yeah maybe the default has some logic about cores/ safe upper and lower bounds but users can then set explicit static values

PR #1252, opened 2026-09-10
feat(raster): byte-bounded batches for RS_EnsureLoaded (EnsureLoadedExec + sedona.raster.max_batch_bytes)

james-willis:
First step of the plan on #1220: bound the bytes `RS_EnsureLoaded` materializes per batch, sized dynamically per input batch from metadata, inside a Sedona-owned operator.

## Problem

`datafusion.execution.batch_size` counts rows, but a row holding a raster can carry megabytes of pixels, so the batch DataFusion hands `RS_EnsureLoaded` can be gigabytes once loaded — 8192 × 1 MiB spatialbench tiles is 8 GiB per partition, which is the sf1 OOM. The obvious knobs don't help: DataFusion's `AsyncFuncExec` re-coalesces its input to exactly `batch_size` rows before evaluating, and the UDF's `ideal_batch_size` only chunks the *invocation* — the per-chunk results are `concat`ed back into one array. Same in DataFusion 54.1.

## What this does

- **`sedona_raster::size`** — metadata-only estimates: `Σ_bands Π raw_source_shape × pixel bytes`. One formula for both storage kinds: for InDb bands it equals the buffer length, for OutDb bands it is the post-load size. Uses the raw source shape (not the visible one) so views don't distort it; shared buffers over-count, the safe direction. Never touches `data`, so it works on OutDb rows before anything is loaded.
- **`sedona.raster.max_batch_bytes`** — new `raster` namespace on `SedonaOptions`, `SET`-able. It is a **per-batch, per-partition** budget, not a global cap: every partition slices its own stream, and a pipeline holds two or three batches at once, so loaded pixels in flight are roughly `target_partitions × 3 × budget`. Default 256 MiB; when a memory limit is configured, `SedonaContext` derives `min(256 MiB, max(16 MiB, (memory_limit / target_partitions) / 8))` at session construction (16 GiB on 16 partitions → 128 MiB; 2 GiB on 16 → the 16 MiB floor; worked table on the option's docs). The rule itself never reads the memory limit — it uses whatever the session holds, including a later `SET`. `0` disables.
- **`RasterBatchBudgetRule` + `EnsureLoadedExec`** (`sedona-query-planner`) — a physical optimizer rule appended after DataFusion's own replaces every `AsyncFuncExec` carrying `rs_ensureloaded` with `EnsureLoadedExec`: same expressions, same schema and plan properties (including the `__async_fn_N` output columns the planner projects on top), but each input batch is sliced so the estimated bytes about to be materialized stay within the budget, and each slice is evaluated and emitted as its own batch. A row over budget goes alone; null rows cost nothing; empty batches flow through; other async expressions in the same node are evaluated per slice. Slices are zero-copy `RecordBatch::slice`s, so order is preserved.
- Wired into `SedonaContext::new_local_interactive_with_runtime_env` next to the existing planner rules (`SedonaContext::new()` wraps an external `SessionContext` and, as today, installs no rules).

Rebased on `main` with #1251 merged and DataFusion 54.1: each slice now reaches the loader as one bundled `load()` per loader, which the end-to-end test asserts.

## Overhead in the non-materialized case

`benches/raster_batch_budget.rs` in `sedona-query-planner` runs the same plan through DataFusion's stock `AsyncFuncExec` and through `EnsureLoadedExec`, with an identity async UDF standing in for `rs_ensureloaded` over rasters that are already in memory, so only the operator's own work is measured (the per-row estimate, the slicing, one invocation and one output batch per slice). 8192 in-db 16×16 UInt8 rasters, release:

| plan | time / batch |
|---|---|
| `AsyncFuncExec` (baseline) | 8.1 µs |
| `EnsureLoadedExec`, 1 slice | 44 µs |
| 8 slices | 54 µs |
| 64 slices | 130 µs |
| 512 slices | 753 µs |
| 8192 slices (one row each) | 12.2 ms |

So the fixed cost is ~35 µs per 8192-row batch (~4 ns per row, the metadata estimate) plus ~1.5 µs per slice. The one-row-per-slice case only arises when every row alone exceeds the budget, i.e. each row is at least `max_batch_bytes` of pixels, against which 1.5 µs is nothing. The estimator originally went through `RasterStructArray::get` and a boxed `BandRef` per band and cost 1.7 ms per batch; it now reads the flat shape and dtype columns directly, which is the `sedona-raster` commit's `estimated_row_bytes`.

## Measurements

2048 OutDb rasters of 1024 × 1024 UInt8 (1 MiB each) through `SELECT RS_EnsureLoaded(rast)`, mock loader returning committed pages, single partition, output streamed and dropped, debug build, macOS max RSS of the test process:

| `sedona.raster.max_batch_bytes` | output batches | max RSS |
|---|---|---|
| 0 (unbounded — today's behaviour) | 1 | 2.20 GB |
| 256 MiB (default) | 8 | 779 MB |
| 64 MiB | 32 | 591 MB |
| 16 MiB | 128 | 145 MB |

RSS is a high-water mark that includes allocator retention, so it tracks the budget loosely rather than matching it; the point is that peak memory now follows the budget instead of the row count.

## Not in this PR (later steps of the same plan)

- Stopping `FilterExec` / `CoalesceBatchesExec` / `RepartitionExec` from re-batching downstream. The end-to-end test pins `target_partitions = 1` so the round-robin repartition the planner puts above the exec doesn't obscure its output.
- Byte-bounded slicing of already in-db columns below raster-producing projections; scan-side batch sizing; memory-pool reservations with adaptive shrinking.
- Output-size estimation for kernels whose output exceeds their input (`RS_FromGDALRaster`, `RS_AsRaster`, upsampling).

## Testing

- `sedona-raster`: estimator tests — InDb equals buffer length, OutDb from metadata without loading, raw vs visible shape under a view, multi-band + null + zero-band rows, overflow rejected instead of wrapping, and the flat fast path agreeing with the per-row trait path.
- `sedona-query-planner`: `slice_ranges` unit test; exec tests with a mock `rs_ensureloaded` async UDF — uniform rasters sliced to the budget, oversized rows alone with nulls free, `0` disables, output schema and `__async_fn_0` column parity with `AsyncFuncExec`, empty batch, the rule ignores unrelated async UDFs, mixed async expressions sized by the ensure-loaded argument only.
- `sedona`: end-to-end SQL through `SedonaContext` with a mock loader and `SET sedona.raster.max_batch_bytes = 600`: `EXPLAIN` shows `EnsureLoadedExec` and no `AsyncFuncExec`; output batches are `[2, 2, 2]` and the loader receives exactly one bundled call per slice (`[2, 2, 2]` requests).
- Full suites locally on the rebased branch: `sedona-raster` 196, `sedona-common` 10, `sedona-query-planner` 90, `sedona` 101 (3 skipped for missing geoarrow-data assets). clippy and fmt clean.

## Round 3 (empty batches, single argument evaluation)

- **Empty input batches emit nothing.** `AsyncFuncExpr` chunks by `ideal_batch_size` (1024 for the real UDF) and concatenates the chunk results, so invoking it on zero rows fails on an empty `concat`. The stock exec never did that because its coalescer absorbs empty batches; this one produced one zero-row slice. Zero-row batches are now skipped in `execute`, `slice_ranges` yields no slices for empty input, the test mock carries the production `ideal_batch_size` so the chunked path is what the tests exercise, and the `sedona` end-to-end test feeds an empty batch ahead of the rows.
- **The raster argument is evaluated once per batch.** Sizing evaluated it, then `invoke_with_args` evaluated the expression's children again per chunk: for a nested argument like `RS_FromPath(path)` every file was opened for metadata twice, and a volatile argument could load a different raster from the one that was sized. The exec now evaluates a non-column argument once, appends it to the batch, invokes a rewritten expression that reads it back by column, and drops the column from the output. A plain column argument, the common case, is left untouched so its slices carry nothing extra; the bench numbers above are unchanged. Regression test: a volatile counted UDF wrapping the raster column is invoked once under both the stock exec and this one.


rust/sedona-query-planner/Cargo.toml
```diff
@@ -36,8 +37,10 @@ datafusion-expr = { workspace = true }
 datafusion-optimizer = { workspace = true }
 datafusion-physical-expr = { workspace = true }
 datafusion-physical-plan = { workspace = true }
+futures = { workspace = true }
 sedona-common = { workspace = true }
 sedona-expr = { workspace = true }
+sedona-raster = { workspace = true }
```
james-willis, line 43: flagging this new dependency here

rust/sedona-query-planner/src/raster_batch_budget.rs
```diff
@@ -0,0 +1,738 @@
...
+        if idx > start && acc.saturating_add(bytes) > budget {
+            ranges.push((start, idx));
+            start = idx;
+            acc = 0;
+        }
+        acc = acc.saturating_add(bytes);
+    }
+    ranges.push((start, estimates.len()));
+    ranges
+}
```
paleolimbot, line 158: Another strategy would be a binary search (cut each chunk in half and check recursively). My last adventure with this was here: https://github.com/apache/sedona-db/pull/1267

paleolimbot (commented):
I will take a closer look at details on Monday, but the only thing I am wondering is how the set budget interacts with the partition count (is it per partition or global?). Perhaps this just needs to be documented.

rust/sedona-query-planner/src/raster_batch_budget.rs
james-willis: I think this is the wrong link

james-willis: why do you suggest the binary search? do you think itll be faster? or is there another motivation that I am missing?

paleolimbot: I did think it might be faster but I am not sure (it may depend on whether the actual batch size ends up being closer to the default or closer to 1). RS_Value on a memory-derived raster might be a good benchmark because it's super cheap because there's no fetch and the value extraction is in theory super fast.

james-willis: I think the batch-building runtime is much much less than all the processing tasks so its worth picking the algorithm that produces the fewest, largest batches. 

That is the main preference for the greedy algo over the binary search

rust/sedona-common/src/option.rs
```diff
@@ -48,6 +48,30 @@ config_namespace! {
...
+        /// slices each input batch so that the estimated bytes of the rasters it
+        /// materializes (sum over bands of source shape × pixel size, taken from
+        /// band metadata) stay within this budget, emitting each slice as its own
+        /// batch instead of trusting the row-count `datafusion.execution.batch_size`.
+        /// A single row larger than the budget is still processed, on its own.
+        /// `0` disables the slicing. When a memory limit is configured
+        /// `SedonaContext` lowers the default to a fraction of the per-partition
+        /// limit.
+        pub max_batch_bytes: usize, default = DEFAULT_RASTER_MAX_BATCH_BYTES
     }
```
paleolimbot, line 75: Can you add an example of what this value might be for some reasonable values of a max memory limit (e.g., if I set it to 16 GB, what is the value of the fraction and what will I get here?).

rust/sedona-query-planner/Cargo.toml
paleolimbot: In a perfect world maybe the raster rules would be separated but that day is not this day

rust/sedona-query-planner/src/raster_batch_budget.rs
```diff
@@ -0,0 +1,740 @@
...
+        .extensions
+        .get::<SedonaOptions>()
+        .map(|opts| opts.raster.max_batch_bytes)
+        .unwrap_or(DEFAULT_RASTER_MAX_BATCH_BYTES);
+    if bytes == 0 {
+        u64::MAX
+    } else {
+        bytes as u64
+    }
+}
```
paleolimbot, line 176: I don't see a reference to the fraction of the total memory limit here like is documented?

paleolimbot (commented):
Worth a pass for documentation and a benchmark for the non-materialized case, but looks great!

rust/sedona-query-planner/src/raster_batch_budget.rs
james-willis: actual benchmarking bears out the story above.

rust/sedona-common/src/option.rs
james-willis: done

rust/sedona-query-planner/src/raster_batch_budget.rs
james-willis: yeah that lives in the SedonaContext. This just references the computed value. Updated the comment to make this clearer to the reviewer

rust/sedona-query-planner/src/raster_batch_budget.rs
```diff
@@ -0,0 +1,746 @@
...
+                .then(move |slice| {
+                    let async_exprs = Arc::clone(&async_exprs);
+                    let schema = Arc::clone(&schema);
+                    let config_options = Arc::clone(&config_options);
+                    let baseline = baseline.clone();
+                    async move {
+                        let mut columns = slice.columns().to_vec();
+                        for expr in async_exprs.iter() {
+                            let value = expr
+                                .invoke_with_args(&slice, Arc::clone(&config_options))
```
jiayuasu, line 532: Can we skip empty batches before invoking `RS_EnsureLoaded`? The real UDF sets `ideal_batch_size()` to `Some(1024)`. On a zero-row slice, DataFusion produces no chunks and then fails on `concat([])`.

To reproduce with the existing test:

1. Add the production batch-size override to `impl AsyncScalarUDFImpl for MockEnsureLoaded` in this file:

   ```rust
   fn ideal_batch_size(&self) -> Option<usize> {
       Some(1024)
   }
   ```

2. Run:

   ```sh
   cargo test -p sedona-query-planner --lib empty_batch_flows_through
   ```

It fails with `Compute error: concat requires input of at least one array`. The mock currently has no batch-size override, which is why the test passes.

I also checked `SELECT RS_EnsureLoaded(rast) FROM tiles` with an empty batch followed by one valid raster. Stock `AsyncFuncExec` returns one row; this operator fails with the same error.

rust/sedona-query-planner/src/raster_batch_budget.rs
```diff
@@ -0,0 +1,746 @@
...
+    /// metadata-only estimate of their raster argument.
+    fn estimate_rows(
+        async_exprs: &[Arc<AsyncFuncExpr>],
+        sized_by: &[usize],
+        batch: &RecordBatch,
+    ) -> Result<Vec<u64>> {
+        let mut totals = vec![0u64; batch.num_rows()];
+        for &idx in sized_by {
+            let array = raster_arg(&async_exprs[idx])?
+                .evaluate(batch)?
```
jiayuasu: Please reuse the raster argument evaluated here when loading each slice. This evaluates it for sizing, then `AsyncFuncExpr::invoke_with_args` evaluates it again. For `RS_EnsureLoaded(RS_FromPath(path))`, the physical plan keeps `RS_FromPath` nested, so every path is opened for metadata twice. A volatile argument can also return a different raster on the second call, so the loaded raster may differ from the one used to size the batch.

To reproduce at this head, add this test inside the existing `tests` module in `rust/sedona-query-planner/src/raster_batch_budget.rs`, then run:

```sh
cargo test -p sedona-query-planner raster_argument_is_evaluated_once -- --nocapture
```

The stock `AsyncFuncExec` case passes with one invocation. `EnsureLoadedExec` fails with `left: 2`, `right: 1`.

<details>
<summary>Regression test</summary>

```rust
#[tokio::test]
async fn raster_argument_is_evaluated_once() {
    use std::sync::atomic::{AtomicUsize, Ordering};

    let (schema, batch) = raster_batch(&[Some(4)]);
    let calls = Arc::new(AtomicUsize::new(0));
    let observed_calls = Arc::clone(&calls);
    let raster_type = schema.field(0).data_type().clone();
    let counted_udf = datafusion_expr::create_udf(
        "counted_raster",
        vec![raster_type.clone()],
        raster_type,
        Volatility::Volatile,
        Arc::new(move |args| {
            observed_calls.fetch_add(1, Ordering::SeqCst);
            Ok(args[0].clone())
        }),
    );
    let counted_arg = Arc::new(
        ScalarFunctionExpr::try_new(
            Arc::new(counted_udf),
            vec![col("rast", &schema).unwrap()],
            &schema,
            Arc::new(ConfigOptions::default()),
        )
        .unwrap(),
    );
    let expr = async_expr(
        Arc::new(MockEnsureLoaded::new(Arc::new(Mutex::new(Vec::new())))),
        "__async_fn_0",
        &schema,
    );
    let nested = Arc::clone(&expr.func)
        .with_new_children(vec![counted_arg])
        .unwrap();
    let expr = Arc::new(AsyncFuncExpr::try_new("__async_fn_0", nested, &schema).unwrap());
    let input = MemorySourceConfig::try_new_exec(&[vec![batch]], schema, None).unwrap();
    let stock: Arc<dyn ExecutionPlan> =
        Arc::new(AsyncFuncExec::try_new(vec![expr], input).unwrap());
    let bounded = RasterBatchBudgetRule
        .optimize(Arc::clone(&stock), &ConfigOptions::default())
        .unwrap();

    for (name, plan) in [("AsyncFuncExec", stock), ("EnsureLoadedExec", bounded)] {
        calls.store(0, Ordering::SeqCst);
        let batches = collect(plan.execute(0, task_context(1024)).unwrap())
            .await
            .unwrap();
        assert_eq!(row_counts(&batches), vec![1]);
        assert_eq!(calls.load(Ordering::SeqCst), 1, "{name}");
    }
}
```

</details>

rust/sedona-query-planner/src/raster_batch_budget.rs
james-willis: nice catch. fixed

rust/sedona-query-planner/src/raster_batch_budget.rs
james-willis: good callout. this should be opened only once now.

paleolimbot (approved):
No conceptual concerns here...it may be nice to get some more complex expressions running through here but we will get there with spatialbench. Thank you!

PR #1255, opened 2026-09-10
feat(rust/sedona-functions): add typed WKB constructors

### LINKED ISSUE #206: rust/sedona-functions: Add `ST_xxxFromWKB()` functions
As part of #174, there are a few functions that are the same as WKB + validation:

- ST_LinestringFromWKB (with alias: ST_LineFromWKB)
- ST_PointFromWKB

These are slightly different than our existing "FromWKB/Binary" because our existing fromwkb/binary is lazy and doesn't copy anything. We do have a validation loop and we can add an optional type validator:

https://github.com/apache/sedona-db/blob/95c156d507f9ce67dde1fd3ec6e1a196c70a39a6/rust/sedona-functions/src/st_geomfromwkb.rs#L111

...could be

```rust
if let Some(expected_geometry_type_id)  = self.expected_geometry_type {
  // Create a builder
  executor.execute_wkb_void(|_maybe_item| {
    // add them item or add null to the builder for _maybe_items whose type id doesn't match
  })?;
  executor.finish(builder.finish())
} else {
  executor.execute_wkb_void(|_maybe_item| Ok(()))?;
  args[0].cast_to(self.out_type.storage_type(), None)
}
```

The steps for adding a parameter to the Udf struct and creating the udf with documentation are very similar to https://github.com/apache/sedona-db/issues/205

paleolimbot:
Adds ST_PointFromWKB and ST_LineStringFromWKB, including the ST_LineFromWKB alias. Both constructors validate WKB, return NULL for another geometry type, and support optional SRID/CRS arguments. Adds Rust and Python parity tests plus SQL reference documentation.

closes #206


rust/sedona-functions/src/st_geomfromwkb.rs
```diff
@@ -105,6 +133,7 @@ impl SedonaScalarKernel for STGeomFromWKB {
                 SedonaType::Arrow(data_type) => match data_type {
                     DataType::Binary => WKB_GEOMETRY,
                     DataType::BinaryView => WKB_VIEW_GEOGRAPHY,
+                    DataType::Null => WKB_VIEW_GEOMETRY,
```
jiayuasu, line 136: Could we preserve `SedonaType::Arrow(DataType::Null)` here? For an all-null Arrow column, the array is still a `NullArray`; declaring it as WKB makes the executor try a BinaryView downcast.

This reproduces it in the existing test module:

```rust
#[rstest]
#[case(st_pointfromwkb_udf())]
#[case(st_linestringfromwkb_udf())]
fn typed_wkb_null_column(#[case] udf: SedonaScalarUDF) {
    let tester = ScalarUdfTester::new(
        udf.into(),
        vec![SedonaType::Arrow(DataType::Null)],
    );
    let result = tester
        .invoke_array(Arc::new(arrow_array::NullArray::new(3)))
        .unwrap();
    assert_eq!(result.null_count(), 3);
}
```

Run `cargo test -p sedona-functions typed_wkb_null_column`. Both cases fail with `could not cast array of type Null ...`, where I'd expect three NULL geometries. The executor already handles the Null type.

paleolimbot: I kept the typed return to minimize schema changes when something like `st_pointfromwkb($1)` happens (the `$1` first shows up as a `Null`, and then is eventually replaced), but added the test. Thanks!

Issue #258, opened 2025-10-29
Update Rust edition to 2024


paleolimbot:
From #256:

> Rust 1.86 means that SedonaDB can switch to Rust 2024 edition. It might be good idea to migrate sooner before SedonaDB's code grows more and more.

It looks like arrow-rs updated to 2024 but not DataFusion yet:

https://github.com/apache/datafusion/blob/0a650a0d27cab21591eccb85d048a1907fc75f14/Cargo.toml#L72-L82

Of course, we don't have to sync our update with DataFusion's exactly but it would probably be nice given that we do have some portions of copied code.

yutannihilation:
~~DataFusion migrated to the 2024 edition 🎉 Probably this change will be in version 52?~~ (Sorry, looking at this PR again, it seems not all the crates have migrated to the 2024 edition yet. But it began!)

https://github.com/apache/datafusion/pull/18915

PR #1260, opened 2026-09-10
chore(rust/sedona-schema): migrate to Rust 2024 edition

paleolimbot:
Begins #258 by migrating `sedona-schema` independently to Rust 2024 and applying the standard Rust 2024 formatter output for that crate.

Validated with `cargo fmt --all --check`, `cargo check -p sedona-schema --all-targets`, `cargo test -p sedona-schema`, and Clippy with warnings denied.


jiayuasu (approved):
LGTM

PR #1263, opened 2026-09-10
chore(rust): scope datafusion "sql" feature to the engine crate

james-willis:
## What
Scope DataFusion's `sql` feature to the `sedona` engine crate instead of enabling it workspace-wide.

## Why
`datafusion-expr`'s default features include `sql`, which pulls `sqlparser`, switches `RawBinaryExpr.op` / `RenameSelectItem` to sqlparser types, and gates `datafusion-functions-nested`'s sql planner. Only the `sedona` engine crate uses that surface (`RawBinaryExpr` / `sqlparser`); the function, spatial-join, query-planner and geoparquet crates never do. Enabling it workspace-wide (via `datafusion-expr`'s defaults) forces every downstream consumer onto the sql-gated `datafusion-expr`, which breaks non-sql embedders such as my Sail fork (whose resolver depends on the `not(sql)` forms and cannot compile the sql-gated nested planner).

## Change
- `datafusion-expr` → `default-features = false` in the workspace; the `sedona` engine crate opts into `features = ["sql"]`.
- Drop the unused `sql` feature from `sedona-functions`, `sedona-spatial-join`, `sedona-query-planner`, `sedona-geoparquet`.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -94,7 +94,7 @@ datafusion-proto = { version = "54.1.0", default-features = false }
 datafusion-datasource = { version = "54.1.0", default-features = false }
 datafusion-datasource-parquet = { version = "54.1.0" }
 datafusion-execution = { version = "54.1.0", default-features = false }
-datafusion-expr = { version = "54.1.0" }
+datafusion-expr = { version = "54.1.0", default-features = false }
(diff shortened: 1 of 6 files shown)


paleolimbot (approved):
Thank you!

Issue #1235, opened 2026-09-04
RS_RasterToWorldCoord*/RS_WorldToRasterCoord* are 0-based; Sedona Spark, PostGIS, and RS_PixelAs* are 1-based


james-willis:
`RS_RasterToWorldCoord`/`X`/`Y` and `RS_WorldToRasterCoord`/`X`/`Y` treat pixel coordinates as **0-based**, while Sedona Spark (following PostGIS) is **1-based**. Every result is exactly one pixel off, including extrapolation outside the grid. On the standard parity grid (origin (100, 500), 2x3 pixels):

| query | SedonaDB | Sedona Spark |
|---|---|---|
| `RS_RasterToWorldCoordX(rast, 1, 1)` | 102 | 100 |
| `RS_RasterToWorldCoordY(rast, 1, 1)` | 497 | 500 |
| `RS_WorldToRasterCoordX(rast, 100, 500)` | 0 | 1 |
| `RS_WorldToRasterCoordY(rast, 100, 500)` | 0 | 1 |
| `RS_WorldToRasterCoordX(rast, 90, 505)` | -5 | -4 |

Meanwhile `RS_PixelAsPoint`/`Centroid`/`Polygon` are **1-based in both engines** (`RS_PixelAsPoint(rast, 1, 1)` → `POINT (100 500)` on both) — so SedonaDB is also internally inconsistent between its coordinate mappers and its pixel-geometry functions. Neither kernel documents a basis choice (`rs_rastercoordinate.rs`, `rs_worldcoordinate.rs`).

Surfaced by the `integration/spark-parity` suite; the coordinate-mapper modules xfail-catalog this divergence referencing this issue. Proposal: align the mappers to 1-based (Spark/PostGIS convention, and consistent with `RS_PixelAs*`), or document 0-basing loudly if it is deliberate.

paleolimbot:
> Proposal: align the mappers to 1-based (Spark/PostGIS convention, and consistent with RS_PixelAs*), or document 0-basing loudly if it is deliberate.

1-basing seems fine for consistency although confusing (do any other tools deal with pixel indices in this way?). I thought there was an x/y ordering thing...does this ordering match the other PostGIS/Sedona overloads?)

james-willis:
IIRC sedona does XY because Postgis does XY. as youve pointed out in the past, other systems like xarray use yx. I think 0 vs 1 indexing is orthogonal to XY YX order.

PR #1264, opened 2026-09-10
test(integration): geometry results in the parity harness; RS_WorldToRasterCoord + RS_RasterToWorldCoord catalogs

james-willis:
Unblocks geometry-returning raster functions in the spark-parity suite and uses that to catalog the `RS_WorldToRasterCoord` / `RS_RasterToWorldCoord` combined forms — including the missing point-geometry overload.

## Harness: geometry results (first commit)

Probing showed the two sides fail asymmetrically:

- **SedonaDB needed nothing**: its geometry results already arrive as `geoarrow.wkb` and `result_to_tuples` already renders them as WKT.
- **Sedona Spark was the broken side**: `DataFrame.toArrow()` ships a `GeometryUDT` column as the UDT's *internal* serialization bytes — not ISO WKB — which the tuple path then tried to cast to string ("Invalid UTF8 payload").

`SedonaSpark.result_to_table` now converts geometry columns to WKB in the JVM (`ST_AsBinary`, transport-only — the same pattern `decode_raster_result` uses with `RS_AsGeoTiff`; the shared test SQL stays RS-only) and re-tags the bytes as `geoarrow.wkb`, so both engines render WKT through the same geoarrow formatter. Columns are renamed positionally before the select because Spark's generated names (`rs_worldtorastercoord(rast, 104.0, 494.0)`) contain dots that every name-based lookup re-parses as a nested path; the original names are restored on the Arrow table.

Everything downstream (`result_to_tuples`, `assert_result`, WKT `expected=` anchors, `compare()`) works unchanged. Full suite on this branch: `174 passed, 84 xfailed` — main's counts plus exactly the 7 new xfails.

## RS_WorldToRasterCoord (second commit)

All probed, every case an xfail stating both observed behaviors:

- **0- vs 1-based** (#1235, now in geometry form): the origin answers `POINT (0 0)` from SedonaDB and `POINT (1 1)` from Sedona Spark; interior points likewise one pixel apart.
- **Outside the grid the offset is not even uniform** (new finding): (90, 505) sits −5 columns and −5/3 rows from the origin. SedonaDB truncates the fraction toward zero (`POINT (-5 -1)`); Spark floors before its 1-based shift (`POINT (-4 -1)`) — the rows agree by coincidence, the columns differ. A positive-fraction probe ((105, 493) → 2 vs 3) confirms the rules are truncation vs floor+1, not something else.
- **Missing point-geometry overload**: `RS_WorldToRasterCoord(raster, point)` raises "No kernel matching arguments" on SedonaDB; Spark accepts it. The xfail notes #1235 still applies once the overload exists.

## RS_RasterToWorldCoord

The input side of #1235: `(1, 1)` answers `POINT (102 497)` from SedonaDB (0-based input) and `POINT (100 500)` — the origin corner — from Spark (1-based input).

## Follow-ups this unblocks

The remaining geometry-returning functions can now get parity files the same way: RS_Envelope, RS_ConvexHull, RS_PixelAsPoint / RS_PixelAsCentroid / RS_PixelAsPolygon. (Item-level-CRS geometry outputs may need another look — these fixtures are CRS-less.) The `test_rs_worldtorastercoordx/y` docstrings still say the combined form is "deferred"; updating that line is left out here because #1262 touches the same sentence — whichever lands second gets the one-line fix.

### DIFF
--- a/python/sedonadb/python/sedonadb/testing_spark.py
+++ b/python/sedonadb/python/sedonadb/testing_spark.py
@@ -167,7 +167,41 @@ class SedonaSpark(DBEngine):
     def result_to_table(self, result) -> pa.Table:
         # toArrow() preserves nulls Arrow-natively; toPandas() would turn a
         # nodata None into a float NaN and mask the value under test.
-        return result.toArrow()
+        #
+        # toArrow() ships a GeometryUDT column as the UDT's internal
+        # serialization bytes, not ISO WKB, so convert in the JVM first
+        # (transport-only, like decode_raster_result's RS_AsGeoTiff) and
+        # re-tag the bytes as geoarrow.wkb — geometry results then render
+        # WKT through the same geoarrow path as every other engine.
+        import geoarrow.pyarrow as ga
+        from sedona.spark.sql.st_functions import ST_AsBinary
+        from sedona.spark.sql.types import GeometryType
+
+        fields = result.schema.fields
+        geometry_indices = {
+            i for i, f in enumerate(fields) if isinstance(f.dataType, GeometryType)
+        }
+        if geometry_indices:
+            # Rename positionally first: the generated column names contain
+            # dots ("rs_worldtorastercoord(rast, 104.0, 494.0)"), which every
+            # name-based column lookup would re-parse as a nested path.
+            names = [f.name for f in fields]
+            renamed = result.toDF(*[f"c{i}" for i in range(len(fields))])
+            result = renamed.select(
+                *[
+                    ST_AsBinary(column).alias(f"c{i}")
+                    if i in geometry_indices
+                    else column
+                    for i, column in enumerate(renamed[c] for c in renamed.columns)
+                ]
+            )
+        table = result.toArrow()
+        if geometry_indices:
+            table = table.rename_columns(names)
+        for i in geometry_indices:
+            wkb = ga.wkb().wrap_array(table.column(i).combine_chunks())
+            table = table.set_column(i, pa.field(table.schema.names[i], wkb.type), wkb)
+        return table
 
     def result_has_raster(self, sql) -> bool:
         from sedona.spark.sql.types import RasterType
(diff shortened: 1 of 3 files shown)


python/sedonadb/python/sedonadb/testing_spark.py
```diff
@@ -167,7 +167,41 @@ def execute_and_collect(self, query):
...
         # toArrow() preserves nulls Arrow-natively; toPandas() would turn a
         # nodata None into a float NaN and mask the value under test.
-        return result.toArrow()
+        #
+        # toArrow() ships a GeometryUDT column as the UDT's internal
+        # serialization bytes, not ISO WKB, so convert in the JVM first
+        # (transport-only, like decode_raster_result's RS_AsGeoTiff) and
+        # re-tag the bytes as geoarrow.wkb — geometry results then render
+        # WKT through the same geoarrow path as every other engine.
+        import geoarrow.pyarrow as ga
```
paleolimbot: This one can be a module level import (since pyarrow already is)

Also, isn't there already a "to geoarrow"? I PRed one into Sedona a few years ago but maybe it's gone now.

paleolimbot (approved):

james-willis:
Follow-up pushed (f7ebe5c): the hand-rolled ST_AsBinary + re-tag path is gone in favor of `sedona.spark.geoarrow.dataframe_to_arrow`, which was indeed already there and does the EWKB conversion + geoarrow tagging (with CRS inference as a bonus). The `geoarrow.pyarrow` import disappears with it.

One wrinkle: `dataframe_to_arrow` resolves columns by name (`list(df)` → `__getitem__`), so it fails on Spark's dotted generated names like `rs_worldtorastercoord(rast, 104.0, 494.0)` — filed as apache/sedona#3351 with a repro. Until that lands, the harness renames columns positionally around the call and restores the names on the Arrow table. Full suite unchanged: 174 passed, 84 xfailed.

PR #1268, opened 2026-09-11
chore(rust/sedona-raster): migrate to Rust 2024 edition

paleolimbot:
Migrates sedona-raster independently to Rust 2024 as part of #258, applies standard formatter output, and adopts a Rust 2024 let-chain required by Clippy. Validated with 190 package tests, all-target checks, and Clippy with warnings denied.


jiayuasu (approved):
LGTM

PR #1269, opened 2026-09-11
chore(c/sedona-geoarrow-c): migrate to Rust 2024 edition

paleolimbot:
Migrates sedona-geoarrow-c independently to Rust 2024 as part of #258 and applies standard formatter output. Also enables the sedona-testing criterion feature explicitly so standalone all-target checks include the benchmark helper without workspace feature unification. Validated with package tests, all-target checks, and Clippy with warnings denied.


jiayuasu (approved):
LGTM

PR #1270, opened 2026-09-11
chore(c/sedona-s2geography): migrate to Rust 2024 edition

paleolimbot:
Migrates sedona-s2geography independently to Rust 2024 as part of #258 and applies standard formatter output. Validated with 65 package tests, all-target checks, and Clippy with warnings denied.


jiayuasu (approved):
LGTM

PR #1271, opened 2026-09-11
chore(rust/sedona-geometry): migrate to Rust 2024 edition

paleolimbot:
Migrates sedona-geometry independently to Rust 2024 as part of #258 and applies standard formatter output. Validated with package tests, all-target checks, and Clippy with warnings denied.


jiayuasu (approved):
LGTM

PR #1273, opened 2026-09-11
chore(c/sedona-libgpuspatial): migrate to Rust 2024 edition

paleolimbot:
Migrates sedona-libgpuspatial independently to Rust 2024 as part of #258 and applies standard formatter output. Validated with package tests, all-target checks, and Clippy with warnings denied.

### DIFF
--- a/c/sedona-libgpuspatial/Cargo.toml
+++ b/c/sedona-libgpuspatial/Cargo.toml
@@ -23,7 +23,7 @@ homepage.workspace = true
 repository.workspace = true
 description = "GPU spatial operations wrapper for libgpuspatial"
 readme.workspace = true
-edition.workspace = true
+edition = "2024"
 rust-version.workspace = true
 
 [features]
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -56,6 +56,7 @@ repos:
       - id: fmt
         name: rustfmt
         args: ["--all", "--"]
+        pass_filenames: false
 
   - repo: https://github.com/cheshirekow/cmake-format-precommit
     rev: v0.6.13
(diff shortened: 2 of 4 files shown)


c/sedona-libgpuspatial/Cargo.toml
```diff
@@ -23,7 +23,7 @@ homepage.workspace = true
 repository.workspace = true
 description = "GPU spatial operations wrapper for libgpuspatial"
 readme.workspace = true
-edition.workspace = true
+edition = "2024"
```
jiayuasu, line 26: [P2] Could we also add explicit `unsafe` blocks in `check_ffi_call` and `probe_callback_wrapper` as part of this switch?

On a checkout with libgpuspatial and the CUDA build dependencies installed, including the driver library, run:

```sh
cargo clippy --locked -p sedona-libgpuspatial --lib --features gpu -- -Dwarnings
```

The edition change makes five `unsafe_op_in_unsafe_fn` warnings into errors with this command, at `libgpuspatial.rs:513,515,538,547,548`. For example, `let err_ptr = get_err(obj_ptr);` needs `let err_ptr = unsafe { get_err(obj_ptr) };`. The `CStr::from_ptr` call, raw pointer dereference, and two `from_raw_parts` calls need the same treatment.

The [GPU build log](https://github.com/apache/sedona-db/actions/runs/34639683474/job/103396046740) already shows all five warnings. That job stays green because it runs `cargo build`, despite being named `clippy`. Without CUDA, this module is skipped through `cfg(gpu_available)`.

paleolimbot: Updated, and I also updated the gpu workflow to do what it says it's doing (e.g., actually run clippy)

PR #1274, opened 2026-09-11
chore(c/sedona-tg): migrate to Rust 2024 edition

paleolimbot:
Migrates sedona-tg independently to Rust 2024 as part of #258, applies standard formatter output, and scopes unsafe operations explicitly inside unsafe functions. Validated with 14 package tests, all-target checks, and Clippy with warnings denied.


jiayuasu (approved):
LGTM

PR #1275, opened 2026-09-11
chore(rust/sedona-geo-traits-ext): migrate to Rust 2024 edition

paleolimbot:
Migrates sedona-geo-traits-ext independently to Rust 2024 as part of #258, applies standard formatter output, and adds explicit unsafe scopes required by the new edition. Validated with package tests, all-target checks, and Clippy with warnings denied.


jiayuasu (approved):
LGTM

PR #1276, opened 2026-09-11
chore(rust/sedona-geo-generic-alg): migrate to Rust 2024 edition

paleolimbot:
Migrates sedona-geo-generic-alg independently to Rust 2024 as part of #258, applies standard formatter output, and adopts Rust 2024 let-chains required by Clippy. Validated with package tests, all-target checks, and Clippy with warnings denied.


jiayuasu (approved):
LGTM

PR #1277, opened 2026-09-11
chore(rust/sedona-adbc): migrate to Rust 2024 edition

paleolimbot:
Migrates sedona-adbc to the Rust 2024 edition as part of #258. Applies standard Rust 2024 formatting without compatibility overrides. Validated with cargo check, Clippy with warnings denied, and package tests.


jiayuasu (approved):
LGTM

PR #1279, opened 2026-09-11
chore(rust/sedona-expr): migrate to Rust 2024 edition

paleolimbot:
Migrates sedona-expr independently to Rust 2024 as part of #258. Applies standard formatter output and Rust 2024 let chains. Validated with 101 package tests, all-target checks, and Clippy with warnings denied.


jiayuasu (approved):
LGTM

PR #1280, opened 2026-09-11
chore(rust/sedona-functions): migrate to Rust 2024 edition

paleolimbot:
Migrates sedona-functions independently to Rust 2024 as part of #258. Applies standard formatting, explicit unsafe scopes, and Rust 2024 let chains. Validated with 560 package tests, all-target checks, and Clippy with warnings denied.


jiayuasu (approved):
LGTM

PR #1282, opened 2026-09-11
chore(rust/sedona-geoparquet): migrate to Rust 2024 edition

paleolimbot:
Migrates sedona-geoparquet independently to Rust 2024 as part of #258. Declares the DataFusion SQL test feature locally, eliminating reliance on workspace feature unification. Applies standard formatting and let chains. Validated with 65 tests, all-target checks, and strict Clippy.


jiayuasu (approved):
LGTM

PR #1284, opened 2026-09-11
chore(rust/sedona-query-planner): migrate to Rust 2024 edition

paleolimbot:
Migrates sedona-query-planner independently to Rust 2024 as part of #258. Declares its DataFusion SQL test feature directly, fixes match ergonomics, and applies standard let-chain formatting. Validated with 82 tests, all-target checks, and strict Clippy.


jiayuasu (approved):
LGTM

PR #1285, opened 2026-09-11
chore(rust/sedona-raster-functions): migrate to Rust 2024 edition

paleolimbot:
Migrates sedona-raster-functions independently to Rust 2024 as part of #258. Applies standard formatter output and a Rust 2024 let chain. Validated with 269 package tests, all-target checks, and Clippy with warnings denied.


jiayuasu (approved):
LGTM

PR #1286, opened 2026-09-11
test(integration): RS_Envelope, RS_ConvexHull, RS_PixelAs* parity (item-level-CRS geometry in the harness)

james-willis:
Completes parity coverage of the geometry-returning raster functions, closing out the batch #1264 unblocked: RS_Envelope, RS_ConvexHull, RS_PixelAsPoint, RS_PixelAsCentroid, RS_PixelAsPolygon — one file per function, every case probed on both engines first.

## Harness: item-level-CRS geometry (first commit)

These five functions return geometry whose CRS can vary per row, which SedonaDB represents as `struct<item: geoarrow.wkb, crs>` — the tuple stringifier could not cast it ("Unsupported cast from struct"). `result_to_tuples` now unwraps the struct and renders the geometry child as WKT like any other geometry column (parent nulls propagated via `pc.struct_field`); the per-item crs field has no counterpart in the other engine's result and stays covered through the RS_CRS suite. RS_WorldToRasterCoord was unaffected because its pixel-space output carries no CRS.

## Harness: row-level CRS stays distinct from column-level CRS

Item-level CRS output is rendered as a `(crs, wkt)` tuple rather than being
folded into a bare WKT string, so a kernel that attaches a CRS at the wrong
level is caught rather than silently compared equal. Sedona Spark carries its
per-geometry SRID inside the EWKB bytes, so reading a row-level CRS means
reading that header.

That SRID is now read directly from the header bytes — EWKB tags the
geometry-type word with `0x20` and follows it with a four-byte SRID, so it is
nine bytes of arithmetic. Two earlier approaches are worth recording because
both failed in ways the replacement cannot:

- **shapely.** Reaching the SRID by constructing the geometry raises on
  geometry GEOS rejects but this harness compares happily as WKT (the unclosed
  LinearRing `ST_TessellateGeog` can emit), which broke `test_st_tessellategeog_no_split`.
  `on_invalid="fix"` avoids that but landed in shapely 2.1, whose Python floor
  of 3.10 is above this package's own 3.9. Declaring the floor would not have
  helped either: the wheel job installs an explicit `CIBW_TEST_REQUIRES` list
  rather than the `test` extra, so a cp39 wheel test would have resolved
  shapely 2.0 through geopandas and then failed on the unknown keyword.
- **pyarrow.compute.** Generalizing the existing column-level reader to
  per-row SRIDs meant vectorizing the endian flag, and `binary_slice` has no
  kernel for the `binary_view` storage SedonaDB returns.

Reading each blob in its own byte order also makes the old mixed-endian guard
unreachable — it existed only because the endian flag had been hoisted to a
column-level scalar — so it is gone rather than ported forward. The
column-level reader used by the PostGIS path now shares the same function.

## What the probes found

Mostly good news — four of the five functions agree exactly, anchored with exact WKT:

- **RS_ConvexHull**: identical ring on both engines (clockwise from the upper-left corner), with and without a raster CRS.
- **RS_PixelAsPoint / Centroid / Polygon (in grid)**: both engines read the pixel coordinate 1-based; corners, centroids, and footprints match string-for-string.
- **RS_PixelAsCentroid / Polygon (out of grid)**: both engines extrapolate along the geotransform, identically — anchored.

Two divergences, xfail-cataloged:

- **RS_Envelope ring order**: same rectangle, opposite winding — SedonaDB counter-clockwise from the lower-left, Sedona Spark clockwise. (Spark's envelope winding also differs from Spark's own convex-hull winding.)
- **RS_PixelAsPoint out of grid**: SedonaDB extrapolates (`(0, 0)` → `POINT (98 503)`); Sedona Spark raises `IndexOutOfBoundsException` ("Specified pixel coordinates (0, 0) do not lie in the raster") — even though Spark's own RS_PixelAsCentroid and RS_PixelAsPolygon extrapolate happily. An internal inconsistency on the Spark side.

Full suite: `198 passed, 103 xfailed`.

With this, every dual-engine RS_ function has a parity file except RS_Values (list-column results; its coordinate-array overload also has no SQL spelling both engines parse) and RS_AsGeoTiff (binary output, no shared round-trip SQL).

### DIFF
--- a/python/sedonadb/python/sedonadb/testing.py
+++ b/python/sedonadb/python/sedonadb/testing.py
@@ -290,15 +290,23 @@ class DBEngine:
         generally asserting a query result or verifying results between engines
         that have (e.g.) differing integer handling.
 
-        Geometry columns are rendered as WKT strings. List columns (e.g. the
-        `List<Double>` returned by `RS_Values`) can't be cast to string, so they
-        pass through as Python lists and are compared by value — assert them with
-        an expected cell that is itself a list, e.g. ``[([1.0, None],)]``.
+        Geometry columns are rendered as WKT strings — including SedonaDB's
+        item-level-CRS geometry (``struct<item: geoarrow.wkb, crs>``, returned
+        by RS_Envelope and friends whose output CRS can vary per row), which is
+        unwrapped to its geometry child first; the per-item crs field has no
+        counterpart in other engines' results and is asserted through RS_CRS
+        coverage instead. List columns (e.g. the `List<Double>` returned by
+        `RS_Values`) can't be cast to string, so they pass through as Python
+        lists and are compared by value — assert them with an expected cell
+        that is itself a list, e.g. ``[([1.0, None],)]``.
         """
         tab = self.result_to_table(result)
         columns = []
         for col in tab.columns:
             # isinstance() does not always work with pyarrow in pytest
+            unwrapped = _unwrap_item_crs_geometry(col)
+            if unwrapped is not None:
+                col = unwrapped
             if _type_is_geoarrow(col.type):
                 columns.append(ga.format_wkt(col, precision=wkt_precision).to_pylist())
             elif pa.types.is_list(col.type) or pa.types.is_large_list(col.type):
@@ -1107,6 +1115,31 @@ def _type_is_geoarrow(type):
     return hasattr(type, "extension_name") and type.extension_name.startswith(
         "geoarrow"
     )
+
+
+def _unwrap_item_crs_geometry(col):
+    """The geometry child of an item-level-CRS column, or None.
+
+    SedonaDB returns geometry whose CRS can vary per row (RS_Envelope,
+    RS_PixelAsPoint, ...) as ``struct<item: geoarrow.wkb, crs>``. Tuple
+    comparison wants the geometry values; parent-level nulls are propagated
+    into the child.
+    """
+    import pyarrow.compute as pc
+
+    if not pa.types.is_struct(col.type):
+        return None
+    fields = {
+        col.type.field(i).name: col.type.field(i).type
+        for i in range(col.type.num_fields)
+    }
+    if set(fields) != {"item", "crs"} or not _type_is_geoarrow(fields["item"]):
+        return None
+    child = pc.struct_field(col, "item")
+    if not _type_is_geoarrow(child.type):
+        # struct_field can strip the extension type; re-wrap its storage.
+        child = fields["item"].wrap_array(child)
+    return child
 
 
 def _crs_equal(actual, expected):
--- a/integration/spark-parity/test_rs_convexhull.py
+++ b/integration/spark-parity/test_rs_convexhull.py
@@ -1,0 +1,55 @@
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+"""SedonaDB vs Sedona Spark parity for RS_ConvexHull.
+
+For a north-up grid the hull is the footprint rectangle, and the engines
+emit an identical ring — clockwise from the upper-left corner — so the
+anchor is the exact WKT (contrast RS_Envelope, where only the ring
+order differs). SedonaDB returns the hull with an item-level CRS; the
+harness compares the geometry and leaves the crs field to the RS_CRS
+coverage.
+"""
+
+import pytest
+
+from sedonadb.raster_testing import write_random_geotiff
+from sedonadb.testing import SedonaDB, compare
+from sedonadb.testing_spark import SedonaSpark
+
+HULL = "POLYGON ((100 500, 114 500, 114 482, 100 482, 100 500))"
+
+
+@pytest.mark.parametrize(
+    "crs", [pytest.param(None, id="crsless"), pytest.param("EPSG:3857", id="epsg3857")]
+)
+def test_rs_convexhull(crs, tmp_path):
+    """The footprint hull reads identically from both engines, with or
+    without a raster CRS."""
+    path = tmp_path / "hull_src.tif"
+    write_random_geotiff(
+        path,
+        "uint8",
+        bands=1,
+        height=6,
+        width=7,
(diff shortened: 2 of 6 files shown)


python/sedonadb/python/sedonadb/testing.py
```diff
@@ -290,15 +290,23 @@ def result_to_tuples(
...
-        Geometry columns are rendered as WKT strings. List columns (e.g. the
-        `List<Double>` returned by `RS_Values`) can't be cast to string, so they
-        pass through as Python lists and are compared by value — assert them with
-        an expected cell that is itself a list, e.g. ``[([1.0, None],)]``.
+        Geometry columns are rendered as WKT strings — including SedonaDB's
+        item-level-CRS geometry (``struct<item: geoarrow.wkb, crs>``, returned
+        by RS_Envelope and friends whose output CRS can vary per row), which is
+        unwrapped to its geometry child first; the per-item crs field has no
+        counterpart in other engines' results and is asserted through RS_CRS
+        coverage instead. List columns (e.g. the `List<Double>` returned by
```
paleolimbot: In this particular context, item crs could come back as `{"item": <wkt>, "crs": <crs>}` or `(<wkt>, <crs>)`. I am not sure that ditching the CRS is a good idea for this specific function (you could always issue a separate query with just the geometry field if this is what you wanted to test?)

paleolimbot (commented):
Integration tests look great! I'm not sold on the handling of item crs in the tuples converter, but also our current handling is not great either.

python/sedonadb/python/sedonadb/testing.py
james-willis: I tried to redo this such that CRSes are measured. We dont care if the CRS is column level or row level, just that the geoms agree (wkt and crs). we rely on proj for equality. 

LKM what you think

python/sedonadb/tests/test_testing.py
```diff
@@ -268,3 +268,54 @@ def test_table_parquet(eng):
...
+
+
+def _ewkb_point(x, y, srid):
+    """A little-endian EWKB point carrying `srid` (0 for no SRID)."""
+    import struct
+
+    geometry_type = 0x00000001 | (0x20000000 if srid else 0)
+    header = struct.pack("<BI", 1, geometry_type)
+    srid_part = struct.pack("<I", srid) if srid else b""
+    return header + srid_part + struct.pack("<dd", x, y)
```
paleolimbot: I believe we have an existing way to create EWKB (probably using shapely). Can you defer to that?

paleolimbot (commented):
I'd like to keep item CRS output from the tester engine distinct from type CRS output because returning the wrong one is actually kind of easy to do in certain pieces of the kernel code. In forthcoming Sedona/Spark 4.2 this will also be distinguished, so we can implement the same output then.

For now, I'd rather avoid this change and have the parity test more verbose as a result (probably far less verbose than the implementation here, or the implementation I'm hoping for that just returns tuples of `(item, crs)`).

python/sedonadb/tests/test_testing.py
james-willis: yeah shapely had. thanks!

python/sedonadb/python/sedonadb/testing.py
```diff
@@ -1122,6 +1175,129 @@ def _crs_equal(actual, expected):
...
+def _normalize_crs(raw):
+    """A raw CRS as a `pyproj.CRS`, or None when the value means "no CRS".
+
+    The per-row CRS arrives in whatever form its source uses — a geoarrow
+    ``Crs`` object (from a column's ``type.crs``), a PROJJSON or authority
+    string (SedonaDB's item ``crs`` field), or an integer SRID (parsed from
+    EWKB) — and an absent CRS as ``None``, ``""``, or ``"0"`` (SedonaDB's
+    SRID-0 sentinel). Fold the absent forms to None and hand every present
+    form to pyproj, which decides semantic equality across the spellings.
+    """
```
james-willis, line 1190: flagging that we use pyproj's concept of equality, which is somewhat permissive.

james-willis:
I'll flag here the pre-existing limitation that we dont check column-level CRS equality now.

python/sedonadb/pyproject.toml
```diff
@@ -44,6 +44,9 @@ test = [
     "pyyaml",
     "rasterio",
     "sedonadb-expr",
+    # >=2.1 for from_wkb(on_invalid="fix"), used to read a row-level SRID
+    # without raising on geometry GEOS rejects (see testing._row_level_crs).
+    "shapely>=2.1",
```
paleolimbot: This may cause issues for the Python 3.9 wheel tests. (We arguably should drop 3.9 support soon anyway and add 3.15 support)

python/sedonadb/tests/test_testing.py
```diff
@@ -268,3 +269,88 @@ def test_table_parquet(eng):
...
+    """Reading the CRS must not blow up on geometry GEOS rejects.
+
+    Some geometry this harness compares happily as WKT is invalid to GEOS —
+    e.g. the unclosed LinearRing ST_TessellateGeog can produce. A plain
+    shapely.from_wkb raises there and failed the whole comparison, hence
+    on_invalid="fix", which still yields the SRID.
+    """
+    import struct
+
+    from sedonadb.testing import _row_level_crs
```
paleolimbot: Can these be module level imports? (Also others in this file)

python/sedonadb/python/sedonadb/testing.py
```diff
@@ -1122,6 +1178,85 @@ def _crs_equal(actual, expected):
...
+        srids = shapely.get_srid(shapely.from_wkb(blobs, on_invalid="fix"))
+        # 0 is "no SRID"; -1 is shapely's missing-geometry sentinel.
+        crs = [
+            None if blob is None or srid <= 0 else _normalize_crs(int(srid))
+            for blob, srid in zip(blobs, srids)
+        ]
+    else:
+        return None
+
+    return crs if any(c is not None for c in crs) else None
```
paleolimbot, line 1243: This is great...there's also a zero dependency version of EWKB srid extraction in sedona-not-db if useful here:

https://github.com/apache/sedona/blob/45617c138d8efc6afd279fef4ebd083d1585e71e/python/sedona/spark/geoarrow/geoarrow.py#L191-L226

python/sedonadb/python/sedonadb/testing.py
paleolimbot: You could use pyproj's concept of equality if the equals target is a `pyproj.CRS` or raw string equality if a string is provided as the expected target. (Either is great though)

paleolimbot (approved):
Thank you!

Probably worth handling shapely on Python 3.9 via a skip or by using the pyarrow ewkb extraction before these fail the wheel tests

python/sedonadb/python/sedonadb/testing.py
james-willis: I decided to go back to our own helper method for this.

shapely kind of needs the 2.1 version for deal with invalid geoms.

the sedona spark method has complexities of dealing with arrow that make integration kind of annoying.

the bit math method seems to me to be the simplest way to actually extract the srid for this exercise

python/sedonadb/pyproject.toml
james-willis: went back to our own ewkb srid parser. explained in other comment

paleolimbot (approved):

PR #1288, opened 2026-09-11
chore(rust/sedona-raster-zarr): migrate to Rust 2024 edition

paleolimbot:
Migrates sedona-raster-zarr independently to Rust 2024 as part of #258. Applies standard formatter output and a Rust 2024 let chain. Validated with 91 passing tests, all-target checks, strict Clippy, and expected ignored cloud tests.


jiayuasu (approved):
LGTM

PR #1289, opened 2026-09-11
chore(rust/sedona-spatial-join): migrate to Rust 2024 edition

paleolimbot:
Migrates sedona-spatial-join independently to Rust 2024 as part of #258. Declares the SQL test feature directly and applies standard formatter and let-chain output. Validated with 486 unit, integration, and doc tests plus strict all-target Clippy.


jiayuasu (approved):
LGTM

PR #1305, opened 2026-09-14
feat(raster): keep FilterExec from re-merging byte-bounded raster batches

james-willis:
**Follows #1252** (merged as a115fc3f0). Rebased onto main as a single commit. Step 2 of the plan on #1220.

## Problem

`EnsureLoadedExec` (#1252) bounds the bytes created at the materialization point, but a `FilterExec` above it coalesces its output by row count: its internal `LimitedBatchCoalescer` buffers filtered rows up to `datafusion.execution.batch_size` before emitting, so it rebuilds an oversized batch from the byte-bounded slices. That happens whether or not the predicate reads the raster — a filter on an id column carrying an in-db raster column still holds 8192 rows of pixels. Memory cost is a property of the bytes in a batch, not of the operator.

## What this does

- `RasterBatchBudgetRule` now also sets the batch size of every `FilterExec` whose output schema carries a raster column to `PASSTHROUGH_BATCH_SIZE = 1`. The coalescer's bypass threshold is `target / 2`, so at 1 every non-empty batch is forwarded untouched and slices keep their boundaries. The filter's `fetch` and projection are preserved (`FilterExec::with_batch_size`).
- Detection is by schema only. A stream whose raster column is still OutDb (cheap references) is included too, which costs nothing beyond skipping the merge of small filtered remainders. Rasters are recognised by the `sedona.raster` extension metadata **or** structurally by the raster storage type, because DataFusion drops field metadata on async UDF outputs (the `__async_fn_N` columns; that is what `sd_restore_metadata` exists for).
- Since DataFusion 54 the separate `CoalesceBatchesExec` is deprecated and no longer inserted by the planner, so on a linear pipeline `FilterExec` is the only merger. `RepartitionExec` reads the session batch size at execute time and is not per-node configurable; it remains the documented gap, and with `EnsureLoadedExec` keeping the default partitioning preference the planner places it below the exec, over OutDb references.
- Module docs point at [apache/datafusion#23385](https://github.com/apache/datafusion/issues/23385) as the long-term home for this: a byte target in arrow's `BatchCoalescer` would make every merger byte-aware without Sedona-owned rewrites.

## Testing

- A stock `FilterExec` over the budgeted exec merges three 2-row slices back into one 6-row batch (the baseline that proves the problem); with the rule the same plan yields `[2, 2, 2]` and the filter's batch size is 1 with an `EnsureLoadedExec` beneath it.
- A vector-only `FilterExec` is returned untouched (same `Arc`, batch size still 8192).
- `has_raster_column` recognises a scan column with the extension metadata and an async-UDF output column with the storage type but no metadata, and rejects a plain Int32 field.
- `sedona-query-planner`: 93 tests pass; clippy and fmt clean. (Locally the crate's tests need `--features datafusion/sql` since #1263 scoped that feature; CI runs `--all-features`.)


james-willis:
ready aside from being stacked

jiayuasu (approved):
LGTM

PR #1307, opened 2026-09-15
chore(ci): Widen timeout for child processes in pyogrio tests

paleolimbot:
We have a failing wheel job on macos-amd64, which has some very slow runners.

It's also possible the hang is real (and only realized on MacOS/amd64), but if this is the case extending the timeout should still catch any issue that occurs.


jiayuasu (approved):
LGTM

PR #1309, opened 2026-09-15
feat(rust/sedona-raster-functions): add RS_MakeEmptyRaster

james-willis:
## What

Adds `RS_MakeEmptyRaster`: build an in-database raster from nothing but a grid definition, with `num_bands` zero-filled bands (no nodata) of an optional `band_data_type` (default `float64`).

Three placements, each with and without the type argument:

| form | arguments |
|---|---|
| extent geometry | `(num_bands[, type], width, height, extent)` |
| upper-left + cell size | `(num_bands[, type], width, height, ulx, uly, cell_size)` |
| affine + SRID | `(num_bands[, type], width, height, ulx, uly, scale_x, scale_y, skew_x, skew_y, srid)` |

The last two are argument-for-argument the Sedona Spark signatures. The extent form is a SedonaDB extension: the grid covers the geometry's envelope with north-up pixels and takes the geometry's CRS, which is the direct spelling of "an abstract grid from ncol, nrow, bbox and crs" that a user asked for (their workaround was an inline sourceless VRT through `RS_FromPath`). It composes with `ST_MakeEnvelope`, `ST_Envelope` and `RS_Envelope` (item-CRS geometry is accepted), so grid-matching another raster is `RS_MakeEmptyRaster(1, w, h, RS_Envelope(other))`.

```sql
SELECT RS_AsRaster(
  geom,
  RS_MakeEmptyRaster(1, 'uint8', 400, 400, ST_MakeEnvelope(0, 0, 400, 400, 4326)),
  'uint8', false, 1.0, 0.0, false
) FROM t;
```

## Notes

- `num_bands = 0` is allowed and yields a bandless grid template.
- Band data is one zero `Buffer` per distinct band size, appended zero-copy through `append_band_data_buffer`, so a batch of N rows × B bands holds a single block of zeros. A band is capped at 4 GiB (the BinaryView length limit) with a clear error.
- Where Sedona Spark silently defaults an unknown `bandDataType` to double, this errors, matching `RS_AsRaster`.
- The pixel-type parser moved out of `rs_as_raster.rs` into `sedona_raster_functions::pixel_type` so both functions share it, and it now also accepts the names `RS_BandPixelType` returns (`UNSIGNED_8BITS`, ...), so a band's type can be fed straight back into a constructor. That is additive for `RS_AsRaster` (those spellings errored before).

## Tests

- Rust unit tests per form: transform/CRS/band layout, SRID mapping (0 / 4326 / other), envelope-vs-shape, item-CRS extent, arrays with nulls, null propagation, error cases, and the shared-zero-block property.
- `python/sedonadb/tests/functions/test_rs_makeemptyraster.py`: SQL surface (literal typing, extent from `RS_Envelope`, table-driven grids, rasterize-onto-grid flow).
- `integration/spark-parity/test_rs_makeemptyraster.py`: the two Spark-shaped forms anchored to literal zero grids, plus SRID/band-count/type scalars.


docs/reference/sql/rs_makeemptyraster.qmd
```diff
@@ -0,0 +1,214 @@
...
+      type: integer
+      description: Grid width in pixels.
+    - name: height
+      type: integer
+      description: Grid height in pixels.
+    - name: extent
+      type: geometry
+      description: >
+        The grid covers this geometry's envelope with north-up pixels, in the
+        geometry's CRS.
```
paleolimbot, line 41: None of these remotely match PostGIS: https://postgis.net/docs/RT_ST_MakeEmptyRaster.html . It looks like postgis favors a `st_makeemtpyraster(...).rst.add_band(...)..rst.add_band(...)` approach, although the extent as a geometry is probably much nicer.

Is this an intentional departure or an accidental one?

python/sedonadb/tests/functions/test_rs_makeemptyraster.py
```diff
@@ -0,0 +1,220 @@
...
+
+def _meta(con, expr):
+    return tuple(
+        con.sql(GRID_META.format(expr=expr)).to_arrow_table().to_pylist()[0].values()
+    )
+
+
+def _raster(con, expr):
+    table = con.sql(f"SELECT {expr} AS r").to_arrow_table()
+    return Raster(table["r"], 0)
```
paleolimbot, line 47: This is too much indirection for understanding what each test is doing without scrolling out of the test. You should be able to create a `Raster` with numpy arrays and assert them as equal, or write assertions for the pieces you care about.

rust/sedona-raster-functions/src/pixel_type.rs
```diff
@@ -0,0 +1,118 @@
...
+/// Parse a pixel type name into a [`BandDataType`].
+///
+/// Three spellings are accepted, all case-insensitively:
+///
+/// - Sedona Spark's letter codes: `D`, `F`, `I`, `UI`, `S`, `US`, `B`, `I8`,
+///   `U64`, `I64`.
+/// - Width-and-signedness names: `int8`, `uint8`, `int16`, `uint16`, `int32`,
+///   `uint32`, `int64`, `uint64`, `float32`, `float64`.
+/// - The names `RS_BandPixelType` returns (`UNSIGNED_8BITS`, `REAL_64BITS`,
+///   ...), so a band's type can be fed straight back into a constructor.
```
paleolimbot, line 33: This is madness that I would prefer not to propagate here (if this is new behaviour) or fix (if this was moved). At the very least RS_BandPixelType should match what PostGIS hands out, which doesn't match any of the values listed here.

rust/sedona-raster-functions/src/rs_make_empty_raster.rs
```diff
@@ -0,0 +1,838 @@
...
+) -> Result<(usize, usize)> {
+    if num_bands < 0 {
+        return exec_err!("RS_MakeEmptyRaster: num_bands must be >= 0, got {num_bands}");
+    }
+    if width <= 0 || height <= 0 {
+        return exec_err!(
+            "RS_MakeEmptyRaster: width and height must be positive, got {width} x {height}"
+        );
+    }
+    // Band data is a BinaryView value, whose length is a u32.
```
paleolimbot, line 362: Are you sure this isn't 2gb?

rust/sedona-raster-functions/src/rs_make_empty_raster.rs
```diff
@@ -0,0 +1,838 @@
...
+                }))
+            }
+        }
+    }
+}
+
+/// The `(xmin, ymin, xmax, ymax)` envelope of an extent geometry, which must
+/// span a positive width and height for the pixel size to be defined.
+fn extent_bounds(wkb: &[u8]) -> Result<(f64, f64, f64, f64)> {
+    let bbox = wkb_bounds_xy(wkb)
```
paleolimbot, line 330: If you want to be fancy you can accept geography here with the bounder in the config options

rust/sedona-raster-functions/src/rs_make_empty_raster.rs
```diff
@@ -0,0 +1,838 @@
...
+
+impl RsMakeEmptyRaster {
+    fn new(grid: Grid, typed: bool) -> Self {
+        Self { grid, typed }
+    }
+
+    /// Index of the first grid-placement argument (the one after `height`).
+    fn grid_arg_index(&self) -> usize {
+        if self.typed { 4 } else { 3 }
+    }
```
paleolimbot, line 100: This kind of matcher construction is an indicator of an API that nobody will be able to remember

docs/reference/sql/rs_makeemptyraster.qmd
james-willis: the non-geom ones are just from sedona. The geometry one is designed to service michael's request about empty raster from bbox.

rust/sedona-raster-functions/src/pixel_type.rs
james-willis: Mostly this was moved other than folding the RS_BandPixelType in. Agree its not great. 

If I was going to design this myself we would standardize on only the Width-and-signedness set.

paleolimbot: I shouldn't review these when tired...a more productive comment would be: let's align this behaviour in a future release...can you create an issue in apache/sedona to unify them? (I think the width-and-signedness names came from one of my review suggestions at some point)

paleolimbot: Also one for the 1 based indexing vs 0 based indexing if there isn't one (Claude flagged the PixelAs function in a different PR as inconsistent in a comment)

docs/reference/sql/rs_makeemptyraster.qmd
paleolimbot: From a geometry is definitely a useful overload...I do like the composable API for band definitions a bit better (although works better in Python or DuckDB style SQL where you can chain things).

james-willis: I dont think our parser(s) support the dot notation so I think thats kind of a non-starter. IMO it wouldnt feel idiomatic to sql users anyway.

what does the add_band syntax give us in an empty raster anyway? what fields do we want to set per-band? 

CC @jiayuasu for a third opinion on what the right way forward on this function is.

jiayuasu: I'd keep the Spark-compatible forms and the geometry overload for this PR. A separate add-band function can follow later.

rust/sedona-raster-functions/src/rs_make_empty_raster.rs
```diff
@@ -0,0 +1,838 @@
...
+                geom.upper_left_y,
+                geom.scale_x,
+                geom.scale_y,
+                geom.skew_x,
+                geom.skew_y,
+                geom.crs.as_deref(),
+            )?;
+            let buffer = zeros
+                .entry(band_len)
+                .or_insert_with(|| MutableBuffer::from_len_zeroed(band_len).into());
```
jiayuasu, line 216: Could we skip the pixel buffer and the per-band size check when `num_bands` is 0? A bandless template should only need the grid metadata, but this currently fails:

```sql
SELECT RS_MakeEmptyRaster(0, 100000, 100000, 0.0, 0.0, 1.0);
```

On `57ed532`, I get `a 100000 x 100000 band of REAL_64BITS pixels exceeds the 4 GiB per-band limit`, even though there are no bands. I also checked a smaller 1024 x 1024 zero-band grid with allocation tracking: it requests an 8,388,608-byte buffer before the empty band loop. That seems unnecessary for the lightweight template use case.

rust/sedona-raster-functions/src/rs_make_empty_raster.rs
```diff
@@ -0,0 +1,838 @@
...
+            Some(string_column(&args[1], n)?)
+        } else {
+            None
+        };
+        let g = self.grid_arg_index();
+        let width = int_column(&args[g - 2], n)?;
+        let height = int_column(&args[g - 1], n)?;
+
+        let mut placement = match self.grid {
+            Grid::Extent => Placement::Extent(executor.make_geom_wkb_crs_accessor(g)?),
```
jiayuasu, line 155: Could we handle an all-null extent column here too? This fails on `57ed532`:

```sql
SELECT RS_MakeEmptyRaster(1, 2, 2, extent)
FROM (VALUES (NULL), (NULL)) t(extent);
```

I get `SedonaDB internal error: Unsupported geometry type: Arrow(Null)`. The scalar version works and returns a null raster:

```sql
SELECT RS_MakeEmptyRaster(1, 2, 2, NULL);
```

I'd expect the first query to return two null rasters as well. It looks like the signature accepts the null type, but the geometry accessor doesn't handle its array form.

jiayuasu (commented):
Two P2 issues from local testing, with SQL reproductions.

rust/sedona-raster-functions/src/pixel_type.rs
james-willis: yes as sedona 2.0 gears up we can align indexing across the engines. I have a summary of indexing in raster functions saved somewhere. Found a misalignment in vector too!

rust/sedona-raster-functions/src/rs_make_empty_raster.rs
james-willis: I'll take a quick look but my guess is this is type system related. Probably best as a separate gh issue

rust/sedona-raster-functions/src/rs_make_empty_raster.rs
james-willis: done

rust/sedona-raster-functions/src/rs_make_empty_raster.rs
james-willis: good callout. fixed

python/sedonadb/tests/functions/test_rs_makeemptyraster.py
james-willis: fixed.

rust/sedona-raster-functions/src/rs_make_empty_raster.rs
james-willis: it was really easy fix so added to this PR

rust/sedona-raster-functions/src/rs_make_empty_raster.rs
james-willis: you are right. arrow-rs is deceiving by using the u32 type here since arrow enforces i32 for cross compat with various languages. fixed.

Issue #1265, opened 2026-09-11
WHERE on a struct field returned by RS_ZonalStatsAll fails: async functions should not be called directly


jiayuasu:
## Expected behavior

A `WHERE` predicate on a field of the struct returned by `RS_ZonalStatsAll` (or any other GDAL-backed, async raster function) should filter rows like any other predicate.

## Actual behavior

Any `WHERE` that references a field of the struct fails with:

```
sedonadb._lib.SedonaError: Internal error: async functions should not be called directly.
This issue was likely caused by a bug in DataFusion's code. Please help us to resolve this by filing a bug report in our issue tracker
```

Aggregating over the same struct fields (`SUM(s.count)`, `GROUP BY name`) works. Filtering on a plain column next to the struct works. The failure is specific to a predicate that touches the async function's output; the planner appears to push the filter into the projection and evaluate the async UDF on the synchronous path.

Matrix (each query wraps the same inner join):

| shape | result |
|---|---|
| `SELECT SUM(s.count) FROM (inner)` | OK |
| `... GROUP BY name` | OK |
| `... WHERE s.count > 0` | FAIL |
| `... WHERE s.count > 0 GROUP BY name` | FAIL |
| CTE + `SUM` / `GROUP BY` | OK |
| CTE + `WHERE s.count > 0` | FAIL |
| `SELECT name, s.count FROM (inner) WHERE s.count > 0` (no aggregate) | FAIL |
| `... WHERE name <> 'x'` (predicate on a plain column) | OK |
| view over inner + `WHERE s.count > 0` | FAIL |

Workaround: materialize first (`df.to_memtable()`), then filter.

## Steps to reproduce

```python
import sedonadb
sd = sedonadb.connect()
# any GeoParquet with a geometry column; here 26 polygons
sd.read_parquet("cantons.parquet").to_view("cantons")
p = ("s3://copernicus-dem-90m/Copernicus_DSM_COG_30_N46_00_E006_00_DEM/"
     "Copernicus_DSM_COG_30_N46_00_E006_00_DEM.tif")   # public bucket, AWS_NO_SIGN_REQUEST=YES
sd.sql(f"SELECT RS_FromPath('{p}') AS rast").to_view("dem")

inner = """SELECT c.name, RS_ZonalStatsAll(d.rast, c.geometry) AS s
           FROM dem d JOIN cantons c ON RS_Intersects(d.rast, c.geometry)"""
[...]

PR #1310, opened 2026-09-15
fix(rust/sedona-query-planner): treat columns produced by RS_EnsureLoaded as loaded

james-willis:
## Summary

Two `needs_pixels` raster calls that share a raster column over a scan fail at execution:

```python
sd.sql("SELECT RS_Value(r, p1, 1) a, RS_Value(r, p2, 1) b FROM ex").to_pandas()
# Internal error: async functions should not be called directly.
```

One call works. A read over a subquery that already called `RS_EnsureLoaded` fails the same way, even with a single call:

```sql
SELECT RS_Value(r, p, 1) FROM (SELECT RS_EnsureLoaded(r) AS r FROM ex)   -- same error
```

## Root cause

`EnsureLoadedOptimizerRule` wraps the raster argument of every `needs_pixels` call in the async `RS_EnsureLoaded`, and relies on recognising its own wrap by *shape* (`rs_ensureloaded(..)`, also behind aliases and the `sd_restore_metadata` wrapper) to stay idempotent across the optimizer's fixpoint passes. Two DataFusion rules destroy that shape between passes and leave a plain column reference behind:

- `CommonSubexprEliminate` hoists the loader shared by the two calls into a projection below, leaving each call with a bare `__common_expr_N` column.
- `OptimizeProjections` merges a user-written `RS_EnsureLoaded(r) AS r` subquery projection into its parent.

The column is raster-typed and, by shape, not loaded, so the next pass wraps it again. The extra loader is hoisted or merged the same way, and every pass (there are three) stacks one more:

```
Projection: sd_restore_metadata(rs_ensureloaded(__common_expr_2)) AS __common_expr_3
  Projection: sd_restore_metadata(rs_ensureloaded(sd_restore_metadata(rs_ensureloaded(ex.r)))) AS __common_expr_2
```

DataFusion's physical planner cannot hoist nested async calls (apache/datafusion#20031): `AsyncMapper` rewrites bottom-up, so only the innermost call lands in `AsyncFuncExec`; the outer one no longer matches, stays in the synchronous `ProjectionExec`, and `AsyncScalarUDF::invoke_with_args` raises the internal error.

## Fix

`already_loaded` now also accepts a column reference when the plan producing that column assigns it a loaded expression. `loaded_columns` resolves this through the input: a `Projection`'s output column is loaded when its expression is (with the projection's own columns resolved the same way), and schema-preserving nodes (`SubqueryAlias`, `Filter`, `Sort`, `Limit`) forward their input's set. Anything else contributes nothing, which is the conservative answer (wrap). The optimizer now reaches its fixpoint after one pass with a single shared loader:

```
Projection: rs_value(__common_expr_1 AS r, p1, 1) AS a, rs_value(__common_expr_1 AS r, p2, 1) AS b
  Projection: sd_restore_metadata(rs_ensureloaded(ex.r)) AS __common_expr_1
```

## Tests

- `rust/sedona-query-planner/src/ensure_loaded.rs`: rule-level tests for the CSE-hoisted column, the user-loaded subquery column (through a `SubqueryAlias`), the pass-through nodes, and a full-optimizer fixpoint test with a fake `AsyncScalarUDF` `rs_ensureloaded` that asserts one un-nested loader in the logical plan and one `rs_ensureloaded` in the physical plan (inside `AsyncFuncExec`). All three shape tests fail without the fix.
- `python/sedonadb/tests/functions/test_rs_value.py`: the reported query over a materialised table and over the user-loaded subquery, cross-checked against rasterio.

## Related

apache/sedona-db#1265 fails with the same error message but by a different mechanism (`PushDownFilter` moves a predicate containing the async call into a join filter, which the physical planner evaluates synchronously). That is fixed separately.


jiayuasu (approved):
LGTM

PR #1311, opened 2026-09-15
fix(rust/sedona-query-planner): hoist async UDF calls out of inner join filters

### LINKED ISSUE #1265: WHERE on a struct field returned by RS_ZonalStatsAll fails: async functions should not be called directly

james-willis:
## Summary

Closes #1265.

A `WHERE` on the output of an async raster function over a join fails at execution:

```python
inner = """SELECT c.name, RS_ZonalStatsAll(d.rast, c.geometry) AS s
           FROM dem d JOIN cantons c ON RS_Intersects(d.rast, c.geometry)"""
sd.sql(f"SELECT name, s.count FROM ({inner}) WHERE s.count > 0").to_pandas()
# Internal error: async functions should not be called directly.
```

Aggregating over the same struct field works, and so does a predicate on a plain column. A cross join (`FROM dem d, cantons c`) fails the same way.

## Root cause

DataFusion's `PushDownFilter` rewrites the predicate in terms of the join inputs, `get_field(rs_zonalstatsall(rs_ensureloaded(d.rast), c.geometry), 'count') > 0`, and because it references both sides of an inner join, moves it into the join's filter. DataFusion's physical planner only hoists async scalar UDFs into `AsyncFuncExec` for `Projection`, `Filter` and `Aggregate` expressions (`try_plan_async_exprs`); a join filter is evaluated synchronously inside the join operator (`NestedLoopJoinExec`, `HashJoinExec`, and our `SpatialJoinExec`), so the async `RS_EnsureLoaded` the `ensure_loaded` rule injects around the raster argument ends up called through the sync path:

```
SpatialJoinExec: join_type=Inner, on=ST_intersects(geometry@1, r@0),
    filter=get_field(rs_zonalstatsall(sd_restore_metadata(rs_ensureloaded(r@0)), geometry@1), count) > 0
```

This is a different mechanism from apache/sedona-db#1310 (nested loaders from re-wrapping a CSE-hoisted column), which produces the same error message; that fix does not cover this case and this fix does not cover that one.

## Fix

A new logical optimizer rule, `HoistAsyncJoinFilterRule`, moves the conjuncts of an inner join's filter that call an `AsyncScalarUDF` into a `Filter` above the join. For an inner join the filter is a plain post-join predicate, so this is semantics-preserving; the physical planner then plans the `Filter` as `FilterExec` over `AsyncFuncExec` as usual.

- It runs after `EnsureLoadedOptimizerRule` (the raster argument of a `needs_pixels` call only becomes an async call there) and before CSE, in `register_ensure_loaded_optimizer`.
- On the next optimizer pass `PushDownFilter` pushes the predicate back into the join and this rule hoists it again; the pass is a net no-op, so DataFusion's plan-signature check ends the fixpoint.
- Outer joins are left alone: their filter decides which rows get null-extended, and `PushDownFilter` never moves a `WHERE` predicate into an outer join's filter anyway. An async call written directly in an outer join's `ON` clause, or inside a KNN join's filter, remains unsupported.
- Only conjuncts whose async calls are evaluated **unconditionally** move (review finding below). `AsyncFuncExec` evaluates an async call for every row, whereas the join's synchronous evaluator skips some sub-expressions: every `CASE` branch after the first `WHEN` condition runs only on the rows that reach it, the right side of `AND`/`OR` is short-circuited, and a `ScalarUDF` with `short_circuits()` (`coalesce`, `nvl`, `nvl2`) declares lazy arguments through `conditional_arguments()`. Hoisting an async call from such a position would run it for rows the sync path never touches (`RS_EnsureLoaded` fetching a raster an unreached `ELSE` names). Those conjuncts stay in the join filter and behave exactly as on `main`: fine while the branch is never taken, the internal error once it is. Evaluating async calls under a selection is DataFusion's to add.

The general fix belongs upstream. Async UDFs in join expressions are an unchecked item on DataFusion's async-UDF epic (apache/datafusion#16520) with no dedicated issue yet; the planner-rewrite refactor (apache/datafusion#18150) is where it would land. This rule is the SedonaDB-side workaround until then.

## Tests

- `rust/sedona-query-planner/src/hoist_async_join_filter.rs`: rule-level tests (async conjunct hoisted, sync conjunct kept, whole filter hoisted, outer join untouched, schema preserved; a `CASE`-guarded conjunct kept while its unconditional sibling moves; an async call in the first `WHEN` condition hoisted; right side of `OR` and lazy `coalesce` arguments kept, eager ones hoisted), a full-optimizer test with a fake `AsyncScalarUDF` asserting no async call remains in any join filter and the physical plan has an `AsyncFuncExec` with no async call on the join line, and the reviewer's `case_keeps_unselected_async_branch` regression (the stub panics if ever invoked). All fail with the rule, or the conditional check, neutered.
- `python/sedonadb/tests/functions/test_rs_zonalstats.py`:
  - the issue's shape, `WHERE s.count > 0` over an `RS_Intersects` join, for `JOIN` (fails on the current release) and `LEFT JOIN` (always worked: DataFusion keeps a both-sides predicate above an outer join); the filtered result must match the unfiltered one, cross-checked against the numpy reference.
  - the same call written in the `ON` clause: `JOIN` passes (the conjunct is hoisted), `LEFT JOIN` is a strict `xfail` with the expected rows asserted, so it flips to a failure once DataFusion supports async calls in join expressions.
  - `RS_ZonalStatsAll` in a `CASE` branch of an `RS_Intersects` join's `ON` clause that no row reaches, over a VRT whose only source file is missing: the query returns its row and never loads (the reviewer's second reproduction; fails at the loader with the conjunct hoisted). The join is a spatial join on purpose: over a nested-loop join DataFusion's join planner pushes the one-sided `rs_ensureloaded(rast)` into a synchronous `ProjectionExec` below the join (`join_proj_push_down_N`), which fails with the same internal error on `main` and is a separate path this rule does not touch.


rust/sedona-query-planner/src/hoist_async_join_filter.rs
```diff
@@ -0,0 +1,358 @@
...
+        let (hoist, keep): (Vec<Expr>, Vec<Expr>) = split_conjunction_owned(filter.clone())
+            .into_iter()
+            .partition(calls_async_udf);
+        let Some(hoisted) = conjunction(hoist) else {
+            join.filter = Some(filter);
+            return Ok(Transformed::no(LogicalPlan::Join(join)));
+        };
+
+        join.filter = conjunction(keep);
+        let filter = Filter::try_new(hoisted, Arc::new(LogicalPlan::Join(join)))?;
```
jiayuasu, line 116: This runs an async call in a `CASE` branch that should be skipped. After the filter moves above the join, `AsyncFuncExec` evaluates the call before `CASE` checks the condition.

For example, with `a.x = [1, 2]` and `b.y = [10, 20]`:

```sql
SELECT a.x, b.y
FROM a JOIN b ON
  CASE WHEN a.x < b.y THEN CAST(1 AS BIGINT)
       ELSE fake_async(a.x + b.y)
  END > 0;
```

The condition is true for all four pairs, so `fake_async` should never run. On base `e20b11f`, this returns four rows. On `1ec8996`, it calls the existing test helper and panics with `internal error: entered unreachable code: stub; never executed`.

I also reproduced this with `RS_EnsureLoaded` and a local-file loader: the base makes no load requests, while this commit tries to load a missing file from the unused `ELSE` branch and fails with `No such file or directory`.

Can we preserve the branch condition when hoisting these calls?

<details>
<summary>Copy/paste regression test</summary>

Add this to the existing `tests` module in this file. It uses the existing `fake_async_udf()` helper, which panics if called.

```rust
#[tokio::test]
async fn case_keeps_unselected_async_branch() {
    use arrow::{array::Int64Array, record_batch::RecordBatch};
    use datafusion::{
        catalog::MemTable, execution::session_state::SessionStateBuilder,
        prelude::SessionContext,
    };

    let builder = crate::optimizer::register_ensure_loaded_optimizer(
        SessionStateBuilder::new().with_default_features(),
    )
    .unwrap();
    let ctx = SessionContext::new_with_state(builder.build());
    ctx.register_udf(fake_async_udf());
    for (name, column, values) in [("a", "x", vec![1_i64, 2]), ("b", "y", vec![10_i64, 20])] {
        let schema = Arc::new(Schema::new(vec![Field::new(column, DataType::Int64, false)]));
        let batch = RecordBatch::try_new(
            schema.clone(),
            vec![Arc::new(Int64Array::from(values))],
        )
        .unwrap();
        ctx.register_table(
            name,
            Arc::new(MemTable::try_new(schema, vec![vec![batch]]).unwrap()),
        )
        .unwrap();
    }
    let batches = ctx.sql(
        "SELECT a.x, b.y FROM a JOIN b ON \
         CASE WHEN a.x < b.y THEN CAST(1 AS BIGINT) \
         ELSE fake_async(a.x + b.y) END > 0",
    )
    .await.unwrap().collect().await.unwrap();
    assert_eq!(batches.iter().map(RecordBatch::num_rows).sum::<usize>(), 4);
}
```

Run:

```sh
cargo test -p sedona-query-planner --lib case_keeps_unselected_async_branch -- --nocapture
```

</details>

james-willis: This is really tricky. the unconditional evaluation of async UDFs in conditionally evaluated blocks is a function of how Datafusion itself plans async UDFs. So you basically have two options here:

1. Accept the risk that the conditional evaluation was load bearing to avoid evaluating expressions that will throw. Thus unconditional evaluation may blow up the query. This what the original version of this PR did.

2. Hoist only expressions that are evaluated unconditionally. This means we won't hoist some async UDFs out of the inner join. This results in the query failing if the conditional branch with the async UDF is ever invoked. 


The third option is to make ensure loaded not an async UDF. async UDFs are really poorly supported in datafusion and they're not making good progress on improving it: https://github.com/apache/datafusion/issues/16520

I'm really starting to lean toward this 3rd option after some of the bugs I've fixed with using async UDFs. cc @paleolimbot

paleolimbot: For now, I think it's fine to eat the extra evaluation for CASE WHEN. I also think it's reasonable to make RS_EnsureLoaded not async (must be constructed using a runtime, which is an extra step in each application layer like R, Python, ADBC, or the CLI, but there are possibly some other components that will need this). When we started talking about this a year ago the support had just been added; however, it seems like there was less enthusiasm for this than previously expected.

Separately from that, it would be great to report any bugs we come across if they haven't already been reported.

james-willis: The epic seems to cover all of the issues we've hit. I check each time I have a problem.

just no one working on it

PR #1312, opened 2026-09-15
feat(rust/sedona-raster-functions): RS_Value grid-coordinate overload

james-willis:
Sedona Spark's `RS_Value` has a grid-coordinate overload — `RS_Value(raster, colX, rowY, band)` — alongside the point forms; SedonaDB registered only `(raster, point[, band])`. This adds the missing overload.

## On XY ordering
We use XY ordering to match: existing sedonadb functions, sedona spark's implementation, and postgis.

Since theyre the same type, flipping to YX would be a silent breaking change that would give insidiously incorrect results for users who expect the existing XY ordering. I don't think it would be reasonable to more to a YX pattern, neither here nor in sedona spark.

## Behavior (probed against Sedona Spark)

The grid overload reads the pixel coordinate **0-based** — column 0 / row 0 is the origin pixel — with a **1-based** band. That's Spark's convention for this specific overload, and notably *different* from its own 1-based `RS_PixelAs*` / `RS_WorldToRasterCoord` functions (the 0-vs-1-based catalog, apache/sedona-db#1235). Verified on the standard grid: `(0,0,1)` → the origin pixel, `(2,3,1)` → `data[0][3][2]`, out-of-grid → NULL.

## Implementation

New `RsValueGrid` kernel matching `(raster, int, int, int)`. No CRS or affine transform is involved — the coordinates index the band grid directly — so it delegates to the existing `sample_pixel`, sharing its bounds-checking (out-of-grid → NULL) and nodata → NULL handling. Registered on the same UDF, so it inherits the `NEEDS_PIXELS_METADATA` tag that injects `RS_EnsureLoaded`.

## Tests

- Rust: `grid_coordinate_overload_is_zero_based` exercises the 4-arg dispatch and the 0-based origin + out-of-grid NULL (sample_pixel's own bounds/nodata behavior is already covered directly). `cargo test -p sedona-raster-functions` — 51 passed.
- Parity: `test_rs_value_grid_coordinate_overload` flips from xfail to an anchored parametrized test (origin, interior, last pixel, band 2, out-of-grid), green against Sedona Spark. Full `test_rs_value.py`: 17 passed, 2 xfailed (the remaining two — the band-default and CRS-mismatch cases — are separate divergences, untouched).


rust/sedona-raster-functions/src/rs_value.rs
```diff
@@ -381,6 +383,84 @@ fn resolve_point_xy(
 /// [`NdBuffer`](sedona_raster::traits::NdBuffer) strides — zero-copy and O(1),
 /// no whole-band materialisation. Errors if the band index is out of range or
 /// the band is not 2-D.
+/// Kernel for `RS_Value(raster, colX, rowY, band)` — sample by grid coordinate.
+///
+/// Mirrors Sedona Spark's grid overload, which reads the pixel coordinate
+/// **0-based** (column 0 / row 0 is the origin pixel) — unlike its 1-based
+/// `RS_PixelAs*` functions — with a 1-based band. No CRS or affine transform is
```
paleolimbot: This seems unintentional?

jiayuasu (approved):
LGTM

PR #1313, opened 2026-09-15
feat(rust/sedona-raster-functions): RS_SetGeoReference numeric overload

james-willis:
Sedona Spark's `RS_SetGeoReference` has a numeric 7-argument form — `RS_SetGeoReference(raster, upperLeftX, upperLeftY, scaleX, scaleY, skewX, skewY)` — alongside the world-file string forms; SedonaDB registered only the string forms. This adds the missing overload.

## Implementation

New `RsSetGeoReferenceNumeric` kernel matching `(raster, numeric×6)`. The six components map straight onto the GDAL geotransform `[upperLeftX, scaleX, skewX, upperLeftY, skewY, scaleY]` — always corner coordinates (there is no ESRI/center convention on the numeric form). Verified against Sedona Spark: `RS_SetGeoReference(rast, 10, 20, 2, -2, 0.1, 0.2)` yields transform `(10, 2, 0.1, 20, 0.2, -2)` on both engines.

The array-build tail (transform ListArray + `with_column_overrides`) is factored into a shared `finish_transform` so the string and numeric kernels don't duplicate it. A null in any numeric component makes that row a null raster, matching the string form's null-georef behavior; the raster's CRS, band names and pixels are carried over untouched.

## Tests

- Rust: `numeric_form_sets_gdal_transform` (asserts the transform mapping with CRS/band/pixels preserved) and `numeric_form_null_component_nulls_raster`. `cargo test -p sedona-raster-functions` — 14 passed.
- Parity: `test_rs_setgeoreference_numeric_overload` flips from xfail to a truth-anchored parity test (exact `DecodedRaster`), green against Sedona Spark. Full `test_rs_setgeoreference.py`: 7 passed, 2 xfailed (the remaining two — ESRI-skew offset and malformed-string handling — are separate divergences, untouched).


paleolimbot (commented):
Thank you!

Does this need to be documented in the SQL docs?

Are there any Python tests of the existing overload that can be modified to also include the new overload?

paleolimbot (approved):

PR #1314, opened 2026-09-15
feat(rust/sedona-raster-functions): RS_WorldToRasterCoord(X/Y) point-geometry overload

james-willis:
Sedona Spark's `RS_WorldToRasterCoord`, `RS_WorldToRasterCoordX`, and `RS_WorldToRasterCoordY` each accept a point geometry in place of the `(worldX, worldY)` numeric pair; SedonaDB registered only the numeric forms. This adds the `(raster, point)` overload to all three.

## Implementation

Each of the three UDFs gains a second, `from_point` kernel. It reads the world coordinate from the point (`read_point_xy`) and feeds the existing `to_raster_coordinate` — raw coordinates, no reprojection, matching the numeric form's no-CRS behavior. Shared input matchers (`coord_input_matchers`) and numeric-argument extraction (`world_xy_arrays`) keep the two input modes from duplicating.

## Parity status

The overload now exists on both engines, but the parity tests **stay xfail** — on the *separate* 0- vs 1-based coordinate-basis divergence (apache/sedona-db#1235), the same one the numeric forms already catalog: SedonaDB reads pixel coordinates 0-based where Sedona Spark is 1-based. So this closes the overload gap without a green flip; the xfail reasons are updated from "no kernel matching arguments" to the coordinate-basis mismatch, and the stale "SedonaDB lacks this overload" module docstrings are corrected.

## Tests

- Rust: `udf_invoke_xy_from_point` and `udf_invoke_pt_from_point` exercise the point overload for X, Y, and the combined form, asserting it matches the numeric form (world (2, 3) → raster (0, 0) on the test raster). `cargo test -p sedona-raster-functions` — 12 passed in the module.
- Parity: the three `*_point_overload` tests now reach the kernel (no longer error) and xfail on apache/sedona-db#1235; `test_rs_worldtorastercoord*.py` together — 10 xfailed, 0 xpassed.


rust/sedona-raster-functions/src/rs_rastercoordinate.rs
```diff
@@ -184,6 +188,92 @@ impl SedonaScalarKernel for RsCoordinatePoint {
...
 
+/// Map each row's world coordinate to a raster coordinate, from whichever
+/// input form the kernel takes, and hand it to `emit`.
+///
+/// This is the whole difference between the two input forms — the numeric
+/// `(raster, worldX, worldY)` and the `(raster, point)` overload — so the
+/// kernels differ only in what they build from the result. `emit` receives
+/// `None` for a row with no coordinate to map: a null raster, a null or empty
+/// point, or a null ordinate.
+fn for_each_raster_coordinate<F>(
```
paleolimbot: There's some existing functions that do vaguely this named `visit_...()` if that feels right here

paleolimbot (approved):
Thank you!

Does this overload need to be added to the .qmd docs?

Are there any (non-parity) Python tests for these functions that currently exist and could be copied or augmented with the new overload?

james-willis:
prior overload did not have tests so I added a simple one for each

PR #1318, opened 2026-09-18
fix(rust/sedona-schema): accept LargeBinary WKB storage

### LINKED ISSUE #1215: WKB import rejects LargeBinary storage
## Summary

Importing WKB whose Arrow storage is `LargeBinary` fails:

```
SedonaDB internal error: Expected Wkb type with Binary storage but got LargeBinary.
```

Reproducer through a literal (the same applies to any LargeBinary-backed `geoarrow.wkb` input):

```python
import geoarrow.pyarrow as ga, pyarrow as pa
from shapely.geometry import Point
from sedonadb.expr import lit

df.select(lit(pa.array([Point(1, 1).wkb], ga.large_wkb())))
```

`ga.large_wkb()` is a public GeoArrow type, and sources such as Iceberg return BLOB columns as LargeBinary, so this shows up with real data, not only constructed literals.

## Expected

WKB with `LargeBinary` storage is accepted wherever `Binary`-backed WKB is, keeping the CRS and edge-type metadata of the extension type.

## Related

- #1195 works around this on the Python side by rebuilding the type on `Binary` storage (with the same CRS and edge type) before resolving the literal; that can go once the importer accepts LargeBinary.

paleolimbot:
Accept GeoArrow WKB backed by LargeBinary while preserving CRS and edge metadata. Add LargeBinary executor paths for scalar functions and raster functions, binary matcher support, GeoArrow C conversion, and GDAL raster input. Extend focused coverage across schema round-trips, unary and binary geometry kernels, raster geometry access, and binary-consuming functions.

Closes #1215


rust/sedona-schema/src/datatypes.rs
```diff
@@ -388,9 +425,10 @@ fn display_geometry(
 fn sedona_type_wkb(edges: Edges, crs: Crs, storage_type: DataType) -> Result<SedonaType> {
     match storage_type {
         DataType::Binary => Ok(SedonaType::Wkb(edges, crs)),
+        DataType::LargeBinary => Ok(SedonaType::WkbLarge(edges, crs)),
```
jiayuasu, line 428: Could you also add `WkbLarge` to the `edge_type` getter in `python/sedonadb/src/schema.rs`? It only matches `Wkb` and `WkbView`, so LargeBinary geography returns `None`.

Example with the Python package built from this branch:

```python
import geoarrow.pyarrow as ga
import pyarrow as pa
import sedonadb
from shapely.geometry import Point

ty = ga.large_wkb().with_edge_type(ga.EdgeType.SPHERICAL).with_crs("OGC:CRS84")
geom = pa.array([Point(1, 2).wkb], type=ty)
df = sedonadb.connect().create_data_frame(pa.table({"geom": geom}))
print(df.schema.field("geom").type.edge_type)  # None
```

This should be `EdgeType.SPHERICAL`, as it is when using `ga.wkb()`. I confirmed the getter result by compiling this branch's Python schema binding.

GeoPandas also reads this property in `_scalar_expr()` to preserve geography on assignment. With `None` here, replacing the column with a Shapely geometry or a null takes the planar branch instead.

PR #1319, opened 2026-09-18
fix(rust/sedona-spatial-join): account for sliced list children

### LINKED ISSUE #519: `get_array_data_memory_size` does not return correct results for listarray slices
arrow-rs propagates the slice for a StructArray on slice(): https://docs.rs/arrow-array/57.2.0/src/arrow_array/array/struct_array.rs.html#337

It doesn't do this for list arrays: https://docs.rs/arrow-array/57.2.0/src/arrow_array/array/list_array.rs.html#396 ...and possibly not for some other types like dictionaries, list views, or run-end-encoded

See this comment https://github.com/apache/sedona-db/pull/515#discussion_r2694835218 and the failing test case: https://github.com/apache/sedona-db/pull/515/files#diff-5fdc254a5ea41059acecb8588e449aae6b354d4886deecbe3b2ffec3628c4b06R257-R290

paleolimbot:
Fixes memory accounting for sliced List, LargeList, and Map arrays by propagating the parent offset range to child data. The estimator remains a single tree traversal and reads only two offsets per list-like node.

Restores the issue regression test and adds LargeList and Map coverage.

closes #519


james-willis (approved):
Thank you! every byte on the war on OOMs matters!

jiayuasu (approved):
LGTM

Ran 16 `arrow_utils` tests and 8 extra probes for nested lists/maps, nulls, empty slices, and nonzero offsets. All passed. Also confirmed the slice regression fails on the base commit and passes here.

PR #1321, opened 2026-09-18
feat(rust/sedona-raster-zarr): keep store clients and opened arrays across load calls

james-willis:
First of the loader-cache PRs from #1224: the two costs that live *inside* the Zarr loader and are paid on every `load` call. Chunk bytes are not cached here; that is the next PR, on the host side.

## What changed

**`sedona-raster-zarr`** (first commit)

`ZarrLoader::load` used to build a fresh `object_store` client for every distinct store and re-open every distinct array, a metadata round trip each, on every call. Since #1252 slices `RS_EnsureLoaded` input by bytes, a query makes more `load` calls than before, and a scan pays these costs on every slice even though it touches each chunk exactly once.

The loader now keeps two handle caches for the life of the registered instance, and never chunk bytes:

- the `ObjectStore` client per scheme and authority (`store_client_key`), so every group in one bucket or on one host shares a client and its connection pool;
- the opened `Array` per (store URI, array path), LRU-bounded at 256 entries with a 5-minute TTL as the only guard against an array being rewritten under a running session.

`ZarrLoader` stops being `Copy`; clones share the handles. `with_array_handle_capacity` and `with_array_handle_ttl` configure a fresh cache before the loader is shared, and `handle_stats()` exposes the reuse counters the tests assert on. Locks are held only around map access, never across an await; two calls racing to open the same array both open it and the later insert wins.

The store map is deliberately a stopgap: once host object stores can be handed across the loader boundary (the direction of #890), the host's registry supplies clients with credentials and pools and this map goes away.

**`sedonadb-zarr`** (second commit)

The extension's private tokio runtime was a current-thread runtime. `ZarrLoader::load` is driven through it by one blocking thread per DataFusion partition at once. On a current-thread runtime every caller's I/O readiness is delivered by whichever caller currently owns the driver, so a CPU-bound chunk decode on that thread stalled the other partitions' fetch completions. It is now a two-worker multi-thread runtime; each `block_on` drives only its own future.

## Measurements

`cargo bench -p sedona-raster-zarr --bench zarr_loader`, one `load` of 64 chunks across 4 arrays of one group on local disk, release build:

| loader | per call |
|---|---|
| handle cache (default) | 798 µs |
| re-open every call (`with_array_handle_ttl(Duration::ZERO)`) | 949 µs |

On local disk an array open is about 38 µs, so the cache saves ~150 µs per call here. On object storage each open is a network round trip, which is the case this is for; the bench pins the mechanism rather than the S3 gain.

## Tests

- `zarr_loader_reuses_array_handles_across_calls`: second call reuses both arrays, one store client built, bytes identical, a clone shares the handles.
- `zarr_loader_reopens_arrays_once_the_handle_ttl_expires`: zero TTL re-opens arrays but keeps the client.
- `zarr_loader_bounds_the_number_of_kept_array_handles`: capacity one with two arrays per call never reuses; zero clamps to one.
- `store_client_key_is_scheme_and_authority`: path-insensitive, host-lowercased keys.

The pre-existing rustdoc error for the `ObjectStoreRegistry::get_store` link in `source_uri.rs` is untouched.


rust/sedona-raster-zarr/src/raster_loader.rs
```diff
@@ -67,16 +78,34 @@ pub const ZARR_FORMAT: &str = "zarr";
...
+///
+/// Building the store client here is the in-process bridge until a
+/// credentialed store can be passed in from DataFusion's
+/// `ObjectStoreRegistry`; see the loader-FFI follow-up. Because the client
+/// is kept, the environment (credentials, endpoint, region) is read once
+/// per scheme and authority for the life of the loader, not on every call.
+#[derive(Debug, Clone)]
 pub struct ZarrLoader {
     concurrency: usize,
+    handles: Arc<HandleCache>,
```
paleolimbot, line 108: If I'm reading this correctly, it's session scoped and not statically scoped, which is great!

python/sedonadb-zarr/src/lib.rs
```diff
@@ -38,14 +38,27 @@ use tokio::runtime::{Builder, Runtime};
...
+/// Multi-thread rather than current-thread: `load` is called from one
+/// blocking thread per DataFusion partition at once, and on a
+/// current-thread runtime every caller's I/O readiness is delivered by
+/// whichever caller happens to own the driver, so a CPU-bound chunk decode
+/// on that thread stalls the others' fetches. With workers owning the
+/// driver, each `block_on` only drives its own future.
+///
+/// Like any tokio runtime held in a Python process, the workers do not
+/// survive `fork()`: a forked child that inherits an initialized runtime
+/// must not use this module. Spawn-start multiprocessing is unaffected.
```
paleolimbot, line 55: I believe there's something in either sedonadb python or datafusion python (from whence that came) that detects the forked case and ensures an error occurs (and not a crash or panic).

paleolimbot (approved):
Thank you!

Would it make sense to make the cache carry stats (hits/misses?) that can be queried in some way? Maybe as part of the display or debug string for the loader and allow that debug string to get surface in Python somehow?

python/sedonadb-zarr/src/lib.rs
james-willis: The AI agent wasn't able to find such a feature in a brief search

paleolimbot: Ah, it's delta-rs: https://github.com/delta-io/delta-rs/blob/87010461cfe01563d91a4b9cd6fa468e2ad5f283/python/src/utils.rs#L10-L31

datafusion-python just links there and we don't handle that either.

PR #1322, opened 2026-09-18
feat(r/sedonadb): expose parquet read options

### LINKED ISSUE #930: feat(r/sedonadb): sd_read_parquet() doesn't expose all Parquet reading options available to Python sd.read_parquet()
Most of the time the defaults are fine, but we do have some options and they should be exposed here, too.

paleolimbot:
Exposes options, geometry_columns, validate, and partitioning through sd_read_parquet() and sd_ctx_read_parquet(). Adds R input validation, GeoParquet option construction in the Rust binding, generated binding updates, documentation, and coverage for reader options and hive partitioning.

closes #930

```r
library(sedonadb)
path <- system.file("files/natural-earth_cities_geo.parquet", package = "sedonadb")
sd_read_parquet(
  path,
  geometry_columns = list(geometry = list(encoding = "WKB")),
  validate = TRUE
) |> sd_count()
#> [1] 243
```


r/sedonadb/R/context.R
```diff
@@ -119,18 +129,114 @@ sd_connect <- function(
...
+      stop(
+        "The `jsonlite` package is required when `geometry_columns` is a list. ",
+        "Install it with install.packages(\"jsonlite\"), or supply a JSON string."
+      )
+    }
+
+    geometry_columns <- as.character(jsonlite::toJSON(
+      geometry_columns,
+      auto_unbox = TRUE,
+      null = "null"
```
jiayuasu: Could we preserve numeric precision here, perhaps with `digits = NA`? I'm seeing the default serialization round bounding-box coordinates to four decimal places:

```r
x <- 1.123456789
y <- 2.234567891
bbox <- c(x, y, x, y)
metadata <- list(geometry = list(encoding = "WKB", bbox = bbox))

encoded <- jsonlite::toJSON(metadata, auto_unbox = TRUE, null = "null")
cat(encoded)
# {"geometry":{"encoding":"WKB","bbox":[1.1235,2.2346,1.1235,2.2346]}}

rounded <- jsonlite::fromJSON(encoded)$geometry$bbox
c(
  original_contains_x = bbox[1] <= x && x <= bbox[3],
  serialized_contains_x = rounded[1] <= x && x <= rounded[3]
)
# original_contains_x   serialized_contains_x
#                TRUE                  FALSE
```

Since the overridden bbox feeds file-level spatial pruning, this looks like it could cause a matching point to be skipped. Numeric parameters inside a supplied CRS would also be rounded by this call.

I verified the serialization and the corresponding native bounding-box predicate locally, but haven't run the complete R/native integration path. Would a full-precision setting plus a fractional-bbox regression test make sense?

r/sedonadb/R/context.R
```diff
@@ -119,18 +129,114 @@ sd_connect <- function(
...
+    if (!jsonlite_available()) {
+      stop(
+        "The `jsonlite` package is required when `geometry_columns` is a list. ",
+        "Install it with install.packages(\"jsonlite\"), or supply a JSON string."
+      )
+    }
+
+    geometry_columns <- as.character(jsonlite::toJSON(
+      geometry_columns,
+      auto_unbox = TRUE,
```
jiayuasu: Could we preserve the array shape for `geometry_types` here? I may be missing an intended list convention, but a one-element character vector gets unboxed into a string:

```r
metadata <- list(
  geometry = list(encoding = "WKB", geometry_types = c("Point"))
)
jsonlite::toJSON(metadata, auto_unbox = TRUE, null = "null")
# {"geometry":{"encoding":"WKB","geometry_types":"Point"}}
```

The PR's native geometry-type deserializer rejected that output locally with `invalid type: string "Point", expected a sequence`. The two-type vector in the new test remains an array, so it doesn't exercise this case.

With the R package built, a small integration case to try is:

```r
path <- system.file(
  "files/natural-earth_cities_geo.parquet", package = "sedonadb"
)
sedonadb::sd_read_parquet(path, geometry_columns = metadata)
```

I haven't run that full R call locally because package build dependencies are missing; the serialization and native-parser checks above were run. Would normalizing this field to retain its array shape, while still unboxing scalar fields such as `encoding`, be reasonable?

PR #1323, opened 2026-09-18
feat(rust/sedona-raster): session cache of loaded OutDb band bytes with refcount-aware eviction

james-willis:
**Rebased onto main after #1252 merged.** `RasterOptions` now carries both `max_batch_bytes` and `cache_max_bytes`, and `SedonaContext` derives both under a memory limit. The bounded-batch end-to-end test from #1252 now gives each raster its own URI: with one shared URI its six rows collapse to a single loader call (identical requests are deduplicated within a slice and served from the cache across slices), which is not what that test measures. The planner's tests and bench from #1252 built `RasterOptions` as a struct literal; they now spread `Default` so the new field does not break them.

Second of the loader-cache PRs from #1224, and the one that caches bytes. #1321 covers the per-call handle costs inside the Zarr loader; this PR keeps loaded OutDb band bytes across calls, batches, partitions and queries within a session, on the host side, for every loader.

One commit per crate:

1. **`sedona-raster`** — `chunk_cache::RasterChunkCache`, and a slot for it on the `RasterLoaderConfig` config extension next to the loader registry.
2. **`sedona-common`** — `sedona.raster.cache_max_bytes` (default 512 MiB).
3. **`sedona-raster-functions`** — `ensure_loaded` probes the cache in pass 1, dedupes identical requests within a batch, dispatches only misses, and stores whole-source results after pass 2.
4. **`sedona`** — the context builds the cache with the session's memory pool, derives the budget under a memory limit, exposes `SedonaContext::raster_chunk_cache()`, and enables arrow's `pool` feature.
5. **`sedonadb` (Python bridge)** — result bytes from Python loaders are copied into Rust-owned buffers instead of wrapping the Python object.

## The cache

**Key and value.** `(outdb_uri, data_type)` → bytes plus the source shape and source-order dimension names they are laid out over. The Zarr anchor URI names one chunk and the GDAL URI one band, so the unit cached is the unit loaded. The value never carries a view: a loader result echoes the *request's* view, so caching a whole result would hand the first requester's view to a later request with a different window over the same chunk. On a hit the band's own view is applied over the cached full-source bytes. A hit only counts when the entry's shape and dimension order match the requesting band; a stale entry is a miss, and an insert with a different layout replaces it (a transposed shape with the same byte count would otherwise pass the byte-count check and carry the wrong layout). Only whole-source results are inserted (source shape and view equal to the request's, byte count as expected); a loader that resolves a view returns something that is not the full source and is left alone. The cache assumes the bytes behind a URI do not change during a session; that is now stated on the option and in the module docs.

**Zero-copy on hit.** The band builder attaches the cached `Buffer` as a shared data block of the output view column, so a hit is a refcount increment. `insert` returns the cached buffer and `ensure_loaded` builds the output from that, so the first batch and the cache share one allocation too.

**Refcount-aware eviction.** Because hits share the allocation, evicting an entry only frees memory when the cache is its last holder. `Buffer::strong_count` tells the two apart. Entries with count one are *idle*: they are what the budget counts and what eviction removes, least recently used first. Entries with a higher count are *shared* with some in-flight batch: they cost the cache nothing extra, are by definition hot, and are left alone until their last holder drops. Idle bytes are recomputed by scanning on insert (there is no callback on the last drop). This is why the structure is a map plus an LRU list from the `lru` crate already in the workspace rather than an off-the-shelf cache: a weigher fixed at insert cannot see refcounts, so evicting a shared entry would look like progress and free nothing.

**Accounting.** Behind arrow-buffer's `pool` feature, `Buffer::claim` attaches one reservation to the allocation for its capacity and releases it when the allocation is freed — when the last holder drops, not when the cache evicts. That is the semantics we want and it comes from the allocation, so a view array's variadic blocks need no special handling. DataFusion 54.1's `ArrowMemoryPool` adapter (feature `arrow_buffer_pool`) bridges it to the session pool. Arrow's pool interface accounts but cannot refuse, so `insert` checks the pool's room first, evicting idle entries to make some, and skips the insert when there is none; a load never fails because of the cache. Two caveats: DataFusion's own operators still estimate build sides from block capacities, so a claimed block sitting in a join build side is counted twice until upstream migrates to claims (conservative); and the adapter's `available()` saturates to `isize::MIN` on an unbounded pool, so room is computed from `capacity` and `used`.

**Budget.** `sedona.raster.cache_max_bytes`, 512 MiB by default. Under a memory limit the context derives `min(512 MiB, limit / 8)`, provisional and to be tuned with the rest of bounded execution. The UDF re-reads the option on each call, so `SET` applies live; zero clears and disables. Entries only go idle when a batch drops, which the cache cannot observe, so every `RS_EnsureLoaded` call also trims idle bytes down to the budget first; without that, a session that only hits after one large query would keep everything that query loaded. Pressure handling is self-pressure only in this version: an insert that would push idle bytes over the budget evicts idle entries first, and all of a load's results go in as one batch under one idle scan. DataFusion's pool has no reclaim hook, so a pool wrapper that asks the cache to give back idle bytes on another consumer's failed `try_grow` is a follow-up, shared with the halve-and-retry step of #1220.

**Within-batch dedupe.** Identical requests in one batch (same key, source shape, dimension order and view) reach the loader once and share one buffer, cache or not. This generalizes what the Zarr loader's `LoadPlan` already did to every loader; GDAL stops reading the same band twice inside one bundle. Requests that differ only in view are *not* deduped, since a loader may resolve a view.

**Concurrency.** Misses are still bundled into one `load` per loader group, so per-key single-flight is not attempted: two partitions missing the same chunk in the same window fetch it twice. A scan never produces that shape; a join can, bounded by partitions minus one extra fetches per chunk during the first fetch's latency. An in-flight map is a local addition later.

## Python loaders copy on insert

The bridge used to wrap a buffer-protocol view of the Python object the loader returned as the Arrow buffer, zero-copy, owned by that object. Cached, that object would live until eviction, its release would attach to the interpreter from whichever thread evicts, and for user loaders returning numpy the bytes could be mutated after return (the bridge did not enforce read-only despite its comment) or could pin a much larger base array than the entry's weight says. The bridge now copies into a Rust-owned buffer (`PyBuffer::to_vec`, so a non-contiguous buffer lands in C order too) while it holds the GIL and releases the Python object right there; the raw-pointer wrapper and its `unsafe impl Send/Sync` are gone. One memcpy per miss, small next to the round trip that produced it. The Zarr loader reaches the host through this bridge today, so this applies to it until the loader ABI lands and results cross with a release callback instead.

## What this does not do

- Sub-request sharing: two windows over one chunk are one entry today because loaders return whole chunks; once view pushdown lands the view must join the key.
- Invalidation: keys assume the object behind a URI does not change during a session. `SET sedona.raster.cache_max_bytes = 0` or `raster_chunk_cache().clear()` are the escape hatches.
- The estimator interaction: `estimated_row_bytes` on #1252 sums raw source shape per band row, so a join that duplicates one chunk across a batch estimates rows × chunk while the resident cost is one chunk. Counting distinct keys per slice is a follow-up there.

## Tests

- `sedona-raster`: hit shares the allocation; disabled cache; oversized entry skipped; idle-only LRU eviction with a held entry surviving; shared entries not counted; trim without an insert; budget lowering and zero; same-layout duplicate keeps the existing entry, different-layout duplicate replaces it; spare capacity shrunk on insert; batch insert under one scan; pool charged until the last holder drops; full pool evicts idle then skips.
- `sedona-raster-functions`: repeated calls served from the cache with pointer-identical output; within-batch dedupe; two views over one cached chunk each keep their own view; a view-resolving loader's result is not cached; one entry per data type for a URI; InDb bands and null rows leave the cache untouched; the budget option is applied from config.
- `sedona`: end to end, the same query twice against a counting loader loads once, the pool shows the cached bytes, and `SET sedona.raster.cache_max_bytes = 0` releases them; the derived budget is `128 MiB` at a 1 GiB limit and the default at 16 GiB or no limit.

Local build note: `sedona-testing` carries an `#[allow(clippy::chunks_exact_to_as_chunks)]` that the current toolchain reports as an unknown lint; clippy on test targets needs `-A unknown_lints` locally. Untouched here.


rust/sedona/src/context.rs
```diff
@@ -159,6 +167,13 @@ impl SedonaContext {
         // `SET sedona.raster.max_batch_bytes` overrides this.
         const RASTER_BATCH_BUDGET_DIVISOR: usize = 8;
         const MIN_RASTER_MAX_BATCH_BYTES: usize = 16 * 1024 * 1024; // 16MB
+
+        // The raster chunk cache keeps loaded OutDb bytes across queries. Its
+        // budget only counts entries no batch references, and every entry is
```
paleolimbot:
```suggestion
        // The raster chunk cache keeps loaded OutDb bytes across queries. Its
        // budget only counts entries not batch references, and every entry is
```

rust/sedona-raster/src/chunk_cache.rs
```diff
@@ -0,0 +1,730 @@
...
+pub struct RasterChunkCache {
+    inner: Mutex<Inner>,
+    pool: Option<Arc<dyn MemoryPool>>,
+    hits: AtomicU64,
+    misses: AtomicU64,
+    inserts: AtomicU64,
+    skipped: AtomicU64,
+}
+
+impl RasterChunkCache {
```
paleolimbot: This makes more sense at the top of the file near the docs

This should also probably be a `trait`...this is a great implementation of a raster chunk cache but putting the trait at the top of the file would nicely communicate what this does and how it is interacted with. Also reasonable that a client will want to provide an implementation (e.g., to spill to disk or S3)

rust/sedona-raster/src/chunk_cache.rs
```diff
@@ -0,0 +1,730 @@
...
+
+            out.push(chunk.bytes.clone());
+            inner.entries.put(key, chunk);
+            self.inserts.fetch_add(1, Ordering::Relaxed);
+        }
+        out
+    }
+
+    /// Counters plus a scan of what is idle and what is shared right now.
+    pub fn stats(&self) -> ChunkCacheStats {
```
paleolimbot: Not sure if it makes sense to reuse the Metrics interface here but I think it is solving a similar problem

rust/sedona-raster/src/chunk_cache.rs
```diff
@@ -0,0 +1,730 @@
...
+use sedona_schema::raster::BandDataType;
+
+/// Identity of a cached load: the band's `outdb_uri` and data type. The
+/// data type is defensive — a URI identifies its bytes, but a mismatched
+/// dtype request must never alias an entry.
+#[derive(Debug, Clone, PartialEq, Eq, Hash)]
+pub struct ChunkKey {
+    pub uri: String,
+    pub data_type: BandDataType,
+}
```
paleolimbot, line 153: Is there a reason a mismatched dtype is counted here? It's possible, although unlikely, that a cast between int32/uint32 is a noop bytewise.

rust/sedona-raster/src/chunk_cache.rs
```diff
@@ -0,0 +1,730 @@
...
+//! when the cache is its last holder. [`Buffer::strong_count`] tells the
+//! two apart: an entry with count one is *idle* (only the cache holds it)
+//! and is what the budget counts and what eviction removes; an entry with
+//! a higher count is *shared* with some in-flight batch, costs the cache
+//! nothing extra, is by definition hot, and is left alone until its last
+//! holder drops. There is no callback on that drop, so idle bytes are
+//! recomputed by scanning on insert and on request. The count is a
+//! snapshot: evicting an entry someone just cloned frees nothing and
+//! skipping one whose batch just dropped costs one more pass, neither of
+//! which is unsafe.
```
paleolimbot, line 59: I believe sedona also has an on disk cache layer and it's worth considering what that looks like here even if you don't implement it yet, since the logic you have here is entirely concerned with in-memory.

rust/sedona-raster-functions/src/rs_ensure_loaded.rs
```diff
@@ -144,6 +147,23 @@ fn lookup_loader(
     guard.get_or_error(format)
 }
 
+/// The session's chunk cache, with `sedona.raster.cache_max_bytes` applied
+/// so a `SET` takes effect on the next call. `None` when the session has
+/// no cache (a bare DataFusion context, or one built without it).
+fn chunk_cache_from_config(config: &ConfigOptions) -> Option<Arc<RasterChunkCache>> {
```
paleolimbot: If you have the raster chunk cache as a trait, you can make a void cache to handle the `None` case here.

python/sedonadb/src/raster_loader.rs
```diff
@@ -133,10 +132,15 @@ impl AsyncRasterLoader for PyRasterLoader {
...
-                    let bytes = Bytes::from_owner(py_buffer);
+                    // Accepts any object implementing the buffer protocol
+                    // (bytes, memoryview, numpy array, ...) and copies it into
+                    // a Rust-owned buffer, in C order, while the GIL is held.
+                    let py_buffer: PyBuffer<u8> = PyBuffer::get(&bytes_obj).map_err(|e| {
+                        ArrowError::InvalidArgumentError(format!(
+                            "Failed to get buffer from Python object: {e}"
+                        ))
+                    })?;
+                    let bytes = Buffer::from_vec(py_buffer.to_vec(py).map_err(py_err)?);
```
paleolimbot: I think this `to_vec()` introduces a copy that we had been trying hard to avoid in this section of code

rust/sedona-raster-functions/src/rs_ensure_loaded.rs
```diff
@@ -487,8 +560,37 @@ where
...
+                            data: Override::Set(&chunk.bytes),
+                            ..Default::default()
+                        },
+                    )
+                    .map_err(|e| {
+                        sedona_internal_datafusion_err!(
+                            "RS_EnsureLoaded: cached derive failed at \
+                             ({raster_idx},{band_idx}): {e}"
+                        )
+                    })?;
```
paleolimbot, line 591: Does this copy_into copy bytes? If so, can it be avoided for a slice of the cached chunk? Also OK if it is not possible to calculate that here (but possibly put a comment). Or error since it would mean two outdb rasters have conflicting metadata?

paleolimbot (commented):
A few questions but in general this looks very cool...an outdb raster buffer pool.

The one thing I couldn't parse from this is how it interacts with view pushdown. I know we don't *do* view pushdown right now, but in some future if a loader decides it knows how to read a few bytes off of a remote buffer effectively, how would that be cached and/or restored? (Fine if the answer is error...a comment is probably a good outlet for this)

james-willis:
> one thing I couldn't parse from this is how it interacts with view pushdown

Correction to my reply: it's handled. Only whole-source results are inserted (shape and view must equal the request's), so a loader that pushes the view down is simply bypassed on insert, and a hit re-applies the band's view over the whole chunk either way. Test: ensure_loaded_does_not_cache_a_loader_resolved_result. Once pushdown exists the view joins the key, which the description notes.

Edited: Original reply was wrong.

rust/sedona-raster-functions/src/rs_ensure_loaded.rs
james-willis: no it creates a new View in the builder that uses the reference of the bytes in `band`

python/sedonadb/src/raster_loader.rs
james-willis: yes this is intentional but I am ambivalent. The options (as long as we are going through python here) are:

1) no copy. python owns the object which means to free the memory requires the application to acquire the GIL again.
2) we copy to a rust-owned vector eagerly and never have to acquire the GIL again.

I don't really know how to weight these two things. The fact these data tend to come from the disk or network and have to be decoded makes me think the copy between two memory locations is not a big deal but I have no benchmark. I have some evidence that GIL contention can be a real bottleneck in datafusion queries but it depends on the workload pattern (e.g. UDFs that are GIL hungry).

The agent was quite convinced that reacquiring the GIL was much worse than the eager copy.

One thing to note is that is only affects the zarr loader and python data sources, but NOT the from numpy constructors. Zarr wont be in this code path once we move the ABI to C

rust/sedona/src/context.rs
james-willis: I think the wording is confusing here. maybe "Its budget only count entries which are not referenced by any batch..."

rust/sedona-raster/src/chunk_cache.rs
james-willis: good idea. probably some time in the future we ourselves will want other impls (spillable...)

python/sedonadb/src/raster_loader.rs
paleolimbot: Right, and we're going to move that to C soon.

I am inclined to leave the zero-copy-ness in for now (although you can leave a note with your leanings) and solve the gil contention if or when we see it. There are also other things we can do (pybuffer trash can that has to fill up before it aquires a gil to release; copy only small buffers but leave big ones zero copy).

rust/sedona-raster/src/chunk_cache.rs
james-willis: I assume you mean operator metrics?

This is different because it is metrics for the cache lifetime.

Do you want to file a followup issue to add operator metrics to the RS_EnsureLoaded operator that show in `EXPLAIN ANALYZE`?

rust/sedona-raster/src/chunk_cache.rs
james-willis: It's defensive so that if for some reason we have a mismatch between the dtype of the byte and the dtype of the request we dont silently get nonsensical values. IIRC there is a similar guard in the zarr loader path to make sure  we dont load garbage bytes for the request.

Imagine the case where between df construction and byte loading (loader case) or vice versa (cache case) the array is rewritten with a new dtype.

python/sedonadb/src/raster_loader.rs
james-willis: ok reverted. These still some issues about writable buffers might behave and how pool reservations might be inaccurate but maybe we dont deal with those for now.

PR #1324, opened 2026-09-18
feat(rust/sedona): unify registered-format reads

### LINKED ISSUE #48: Ensure that signed and unsigned S3 requests can coexist
From @zhangfengcdt in https://github.com/apache/sedona-db/pull/44 :

```
            // TODO: Consider registering object stores per-bucket instead of per-scheme to avoid
            // authentication conflicts. Currently, if a user first accesses a public S3 bucket with
            // aws.skip_signature=true and then tries to access a private bucket, the cached object
            // store will still have skip_signature enabled, preventing authentication to the private
            // bucket. A per-bucket registration approach would solve this by using bucket-specific
            // cache keys like "s3://bucket-name" instead of just "s3://".
```

(Opening an issue so that we can remember/discover this and circle back!)

### LINKED ISSUE #1245: read_parquet rejects URLs without a .parquet extension and offers no way to bypass the check
## Problem

`read_parquet` refuses any path that does not end in `.parquet`, and there is no way to relax or bypass that check. This breaks reading GeoParquet from APIs that serve it from a URL without a file extension, for example OGC API - Features style endpoints:

```python
import sedona.db

sd = sedona.db.connect()
df = sd.read_parquet("https://example.com/collections/collection_id/items?f=parquet")
```

```
sedonadb._lib.SedonaError: File path 'https://example.com/collections/collection_id/items?f=parquet' does not match the expected extension '.parquet'
```

Reported on Discord: https://discord.com/channels/1034906732264181760/1545393838678478888

## Where the check lives

`geoparquet_listing_table` in `rust/sedona-geoparquet/src/provider.rs` strips the query string and then requires the remaining path to end in `listing_options.file_extension`, which is always the `ParquetReadOptions` default of `.parquet`. `GeoParquetReadOptions` never exposes `file_extension`, and the `options` dict from Python only feeds DataFusion `TableOptions` (`aws.*`, `azure.*`, `format.*`), so neither `{"file_extension": ""}` nor `{"format.file_extension": ""}` has any effect.

`sd.read()` has the same limitation on the Python side because it guesses the format from the extension and raises `Can't guess format from paths where no item has an extension`.

## Workaround that works today

The SQL path does not apply the check, and the resulting table still goes through the GeoParquet format (geometry column typed with CRS, `ST_*` functions work):

```python
sd.sql("""
    CREATE EXTERNAL TABLE items
    STORED AS PARQUET
    LOCATION 'https://example.com/collections/collection_id/items?f=parquet'
""")
sd.sql("SELECT * FROM items").show()
```

Verified against a GeoParquet file served over HTTP from a URL with no extension and a `?f=parquet` query string (the server must support Range requests, as with any remote Parquet read).

## Proposal

Expose a way to skip or override the extension check in `read_parquet`:

- Rust: add `file_extension` (or a `check_extension: bool`) to `GeoParquetReadOptions` and thread it into `ParquetReadOptions` / the check in `geoparquet_listing_table`. `sedona-datasource` already has a `check_extension` flag on `RecordBatchReaderTableOptions`, so GeoParquet is currently the odd one out.
- Python: accept it in `SedonaContext.read_parquet(...)`, either as a keyword argument or as a recognized key in `options`.
- Optionally let `sd.read()` accept an explicit `format=` so it does not have to guess from the path.

paleolimbot:
Adds one Rust provider resolver behind SedonaContext::read, Python sd.read(), R sd_read(), and SQL URL tables. File options flow through the same FileFormatFactory::create key/value path as CREATE EXTERNAL TABLE, while external single-object formats such as Zarr avoid listing. Object stores are registered per URL authority so signed and unsigned S3 buckets coexist. Adds R partitioning and cross-language regression coverage.

Closes #48.

```python
from pathlib import Path
import tempfile
import sedona.db

with tempfile.TemporaryDirectory() as directory:
    path = Path(directory) / "values.csv"
    path.write_text("id;value\n1;one\n")
    db = sedona.db.connect()
    print(db.read(path, {"delimiter": ";"}, "csv").to_pandas())
#    id value
# 0   1   one
```

Closes #1245 .


rust/sedona/src/read.rs
```diff
@@ -0,0 +1,222 @@
...
+    GeoParquetReadOptions::from_table_options(options.clone()).map_err(DataFusionError::Plan)?;
+
+    for path in &table_paths {
+        ensure_object_store_registered_with_options(
+            &mut context.state(),
+            path.as_str(),
+            Some(options),
+        )
+        .await?;
+    }
```
paleolimbot: I have to double check on how this is done...we also have an issue where we may want to allow the store to be attached to a source (e.g., allow anonymous s3 for a single provider).

I'm also not sure we should be parsing GeoParquetReadOptions out of the gate here.

rust/sedona/src/read.rs
```diff
@@ -0,0 +1,222 @@
...
+/// This is the common implementation behind [`SedonaContext::read`](crate::context::SedonaContext::read)
+/// and SQL URL tables. External formats that opt into
+/// [`list_single_object`](sedona_datasource::spec::ExternalFormatSpec::list_single_object)
+/// are routed through the single-object provider; all other formats use a
+/// [`ListingTable`].
+pub(crate) async fn read_provider(
+    context: &SessionContext,
+    table_paths: Vec<ListingTableUrl>,
+    options: &HashMap<String, String>,
+    format: Option<&str>,
```
paleolimbot: FileFormatFactory I think should have been resolved before this point...even in the bindings themselves. (e.g., so a FileFormat doesn't have to be registered to make it to this point).

rust/sedona/src/read.rs
```diff
@@ -0,0 +1,222 @@
...
+    if format == "csv" {
+        if let Some(delimiter) = options
+            .get("delimiter")
+            .or_else(|| options.get("format.delimiter"))
+        {
+            if delimiter.len() != 1 {
+                return plan_err!("CSV delimiter must be a single byte, got {delimiter:?}");
+            }
+        }
+    }
```
paleolimbot: There's some way for SQL to create an external table and pass a hashmap of string/string options and this is probably not it

r/sedonadb/R/context.R
```diff
@@ -131,6 +131,58 @@ sd_ctx_read_parquet <- function(ctx, path) {
...
+#'   or `"json"`. By default the format is inferred from the path extension.
+#' @param ctx A SedonaDB context.
+#'
+#' @returns A sedonadb_dataframe
+#' @export
+#'
+#' @examples
+#' path <- system.file("files/natural-earth_cities_geo.parquet", package = "sedonadb")
+#' sd_read(path) |> head(5) |> sd_preview()
+sd_read <- function(file_or_files, options = list(), format = NULL) {
```
paleolimbot: This should also support partitioning like the Python API

rust/sedona/src/read.rs
paleolimbot: Changed this to mirror ListingTableFactory: unqualified reader keys are normalized into the format.* namespace and passed to FileFormatFactory::create(state, &options). ExternalFormatFactory now deserializes its own namespaced options and ignores unrelated table-extension namespaces.

rust/sedona/src/read.rs
paleolimbot: Removed the GeoParquetReadOptions parse from the generic path. Object-store settings are now parsed independently by scheme through DataFusion TableOptions, and each FileFormatFactory owns its format-specific option deserialization. I also added per-authority signed/unsigned S3 coverage for #48.

rust/sedona/src/read.rs
paleolimbot: Resolved: read_provider now receives an already-selected Arc<dyn FileFormatFactory>. Explicit factories can be passed directly without registration; the Python and R bindings resolve explicit format strings before entering the shared provider path. The Rust regression also reads CSV through an explicit factory from a differently suffixed path.

r/sedonadb/R/context.R
paleolimbot: Added partitioning to sd_read() and sd_ctx_read(), with the same three states as Python: NULL auto-discovers, character() disables discovery, and a character vector specifies columns. The R regression covers explicit and disabled partitioning.

rust/sedona/src/read.rs
```diff
@@ -0,0 +1,282 @@
...
+                    factory,
+                    compression,
+                    listing_extension: Some(extension),
+                });
+            }
+        }
+        return Ok(ResolvedReadFormat {
+            factory,
+            compression: None,
+            listing_extension: None,
```
jiayuasu, line 202: I'm less familiar with this reader setup, so I may be missing some context. Could we keep the format's suffix filter when the path is a directory?

I tried a directory with `part.csv` containing `id\n1\n` and `notes.txt` containing `id\n999\n`. The old CSV reader returns one row; the shared reader returns two, including `999` from the text file. A Parquet directory with a text sidecar similarly goes from returning 243 rows to `Invalid Parquet file. Corrupt footer`.

It looks like the exception for explicitly named files also removes filtering when listing a directory. This seems to affect Python's `read.csv()` / `read.parquet()` too; I tested the Rust path below and checked the Python routing in source.

<details>
<summary>Rust repro for the extra CSV row</summary>

Save as `rust/sedona/tests/pr1324_directory.rs`:

```rust
use datafusion::{
    arrow::util::pretty::pretty_format_batches, datasource::file_format::csv::CsvFormatFactory,
};
use sedona::context::SedonaContext;
use std::{collections::HashMap, sync::Arc};

#[tokio::test]
async fn explicit_csv_directory_keeps_extension_filter() {
    let dir = tempfile::tempdir().unwrap();
    std::fs::write(dir.path().join("part.csv"), "id\n1\n").unwrap();
    std::fs::write(dir.path().join("notes.txt"), "id\n999\n").unwrap();
    let path = dir.path().to_string_lossy().to_string();
    let ctx = SedonaContext::new_local_interactive().await.unwrap();
    let old = ctx
        .read_csv(vec![path.clone()], &HashMap::new(), true, ",")
        .await
        .unwrap()
        .collect()
        .await
        .unwrap();
    assert_eq!(old.iter().map(|b| b.num_rows()).sum::<usize>(), 1);
    let new = ctx
        .read(
            path,
            &HashMap::new(),
            Some(Arc::new(CsvFormatFactory::new())),
        )
        .await
        .unwrap()
        .collect()
        .await
        .unwrap();
    println!(
        "new explicit directory read:\n{}",
        pretty_format_batches(&new).unwrap()
    );
    assert_eq!(
        new.iter().map(|b| b.num_rows()).sum::<usize>(),
        1,
        "explicit CSV read included a .txt sidecar"
    );
}
```

Run from the repository root:

```sh
cargo test -p sedona --test pr1324_directory -- --nocapture
```

The old-reader assertion passes. The final assertion fails with `left: 2, right: 1`; the printed table contains both `1` and `999`.

</details>

rust/sedona/src/read.rs
```diff
@@ -0,0 +1,282 @@
...
+    // This is the same key/value entry point used by DataFusion's
+    // ListingTableFactory for CREATE EXTERNAL TABLE. Any format-specific
+    // deserialization belongs to the selected factory, not this generic path.
+    let file_format = format.factory.create(&state, &factory_options)?;
+    if let Some(external) = file_format.downcast_ref::<ExternalFileFormat>() {
+        return external_table(
+            external.spec().clone(),
+            context,
+            table_paths,
+            false,
```
jiayuasu, line 92: Could we pass the caller's `check_extension` value through here? I may be missing a check elsewhere, but `read_external_format(..., true, ...)` seems to lose it at this `false`.

With a format whose extension is `example`, I passed `one.example` and `two.wrong`, with one row per file. The previous `external_table(..., true, ...)` call rejects the second path. The new context method succeeds and returns only one row, even with `check_extension=true`, so the caller gets no indication that one input was skipped.

<details>
<summary>Rust repro with a one-row external format</summary>

Save as `rust/sedona/tests/pr1324_extension.rs`:

```rust
use sedona::context::SedonaContext;
use std::{collections::HashMap, sync::Arc};

#[derive(Debug)]
struct ExampleFormat;
#[async_trait::async_trait]
impl sedona_datasource::spec::ExternalFormatSpec for ExampleFormat {
    async fn infer_schema(
        &self,
        _location: &sedona_datasource::spec::Object,
    ) -> datafusion_common::Result<arrow_schema::Schema> {
        Ok(arrow_schema::Schema::new(vec![arrow_schema::Field::new(
            "value",
            arrow_schema::DataType::Int32,
            false,
        )]))
    }
    async fn open_reader(
        &self,
        _args: &sedona_datasource::spec::OpenReaderArgs,
    ) -> datafusion_common::Result<Box<dyn arrow_array::RecordBatchReader + Send>> {
        let batch = arrow_array::RecordBatch::try_from_iter([(
            "value",
            Arc::new(arrow_array::Int32Array::from(vec![1])) as arrow_array::ArrayRef,
        )])
        .unwrap();
        let schema = batch.schema();
        Ok(Box::new(arrow_array::RecordBatchIterator::new(
            vec![Ok(batch)],
            schema,
        )))
    }
    fn with_options(
        &self,
        _options: &HashMap<String, String>,
    ) -> datafusion_common::Result<Arc<dyn sedona_datasource::spec::ExternalFormatSpec>> {
        Ok(Arc::new(Self))
    }
    fn extension(&self) -> &str {
        "example"
    }
}

#[tokio::test]
async fn external_check_extension_rejects_mismatched_input() {
    let dir = tempfile::tempdir().unwrap();
    let one = dir.path().join("one.example");
    let two = dir.path().join("two.wrong");
    std::fs::write(&one, "unused").unwrap();
    std::fs::write(&two, "unused").unwrap();
    let ctx = SedonaContext::new_local_interactive().await.unwrap();
    let paths = vec![
        one.to_string_lossy().to_string(),
        two.to_string_lossy().to_string(),
    ];
    let direct_old = sedona_datasource::provider::external_table(
        Arc::new(ExampleFormat),
        &ctx.ctx,
        paths
            .iter()
            .map(|p| datafusion::datasource::listing::ListingTableUrl::parse(p).unwrap())
            .collect(),
        true,
        None,
    )
    .await;
    assert!(direct_old
        .unwrap_err()
        .to_string()
        .contains("does not match the expected extension"));
    match ctx
        .read_external_format(Arc::new(ExampleFormat), paths, None, true, None)
        .await
    {
        Err(error) => assert!(error
            .to_string()
            .contains("does not match the expected extension")),
        Ok(df) => {
            let batches = df.collect().await.unwrap();
            panic!(
                "check_extension=true silently accepted two paths and returned {} rows",
                batches.iter().map(|b| b.num_rows()).sum::<usize>()
            );
        }
    }
}
```

Run from the repository root:

```sh
cargo test -p sedona --test pr1324_extension -- --nocapture
```

The old helper's error assertion passes. The new call reaches the final panic with `check_extension=true silently accepted two paths and returned 1 rows`.

</details>

jiayuasu (commented):
Two questions about file selection in the shared reader, with local reproductions.

PR #1325, opened 2026-09-18
feat(python/sedonadb-geopandas): add Series arithmetic

jiayuasu:
## What changes

Third slice of the series (#1195 column assignment, #1246 dissolve). This one adds arithmetic on `Series` in the experimental `sedonadb-geopandas` package: `+`, `-`, `*`, `/`, their reflected forms, and negation, on numeric columns.

- **Division follows pandas rather than SQL.** `/` is true division: an integer/integer division casts the numerator to float first (dictionary encoding unwrapped, wrapper types resolved) so it does not silently truncate, while `Decimal` division stays `Decimal`. `//` is deliberately not implemented, since SQL division truncates toward zero where Python floors — better an `AttributeError` than a silently different answer.
- **NumPy defers to the column.** `__array_ufunc__ = None` on `Series`, so `np.array([...]) + series` goes to the reflected operator (and is rejected as an array-like) instead of NumPy broadcasting an array against a lazy column element by element.
- **The operand matrix.** Python and NumPy scalars, Arrow scalars, `Literal`s, and one-element containers (an Arrow array, a pandas `Series`) all resolve to their numeric payload; a `Literal` that resolves to several values propagates the resolver's own error rather than a generic type error; an integer past int64 raises `OverflowError` the way pandas does. Bare expressions and `Series` from another frame are rejected for the same provenance reasons as assignment.
- **Duration `*` / `/` are rejected for now** with a clear `NotImplementedError`: the tick-level overflow, precision, and missing-value handling they need arrives with the temporal follow-up.

## Testing

17 new tests (153 in the package) covering the operator surface, true and decimal division, the full operand matrix, and the rejections; run with warnings promoted to errors against both a source-built current main and the released sedonadb 0.4.1 in a clean environment (pandas 2 and 3). Ruff lint/format clean.


james-willis (approved):

python/sedonadb-geopandas/python/sedonadb_geopandas/_series.py
```diff
@@ -257,6 +319,102 @@ def __or__(self, other):
...
+        return Series(self._df, _operand(self._df, other) * self._expr, self._name)
+
+    # `/` is true division here, as in pandas. The engine follows SQL, where
+    # dividing two integers truncates (`1 / 2` is 0), so an integer expression is
+    # cast to double first — but only when the *other* operand is also
+    # integer-like. The cast is scoped that tightly because it is lossy
+    # elsewhere: an integer divided by a Decimal must stay in decimal arithmetic
+    # (forcing double gives 1/Decimal("0.1") -> binary rounding), and durations
+    # are handled separately below. Dictionary encoding is unwrapped before
+    # deciding, so a dictionary<int64> column does not silently truncate.
```
paleolimbot: Probably no REEs in common usage, but a run-end-encoded int is in this category too

python/sedonadb-geopandas/python/sedonadb_geopandas/_series.py
```diff
@@ -257,6 +319,102 @@ def __or__(self, other):
...
+            return self._expr.cast(pa.float64())
+        return self._expr
+
+    def _duration_arith(self, op, other):
+        # Tick-level arithmetic needs overflow, precision, and missing-value
+        # handling — the pandas NaT sentinel arrives from Arrow data as a
+        # representable tick — that lands with the dedicated temporal support.
+        raise NotImplementedError(
+            "duration arithmetic is not supported yet; temporal support "
+            "arrives in a follow-up change"
```
paleolimbot: probably leave the follow-up out of the error (users won't know what that means)

paleolimbot (approved):

python/sedonadb-geopandas/python/sedonadb_geopandas/_series.py
jiayuasu: Good catch. It was worse than a truncation: the engine refused arithmetic on a run-end-encoded column outright. `_dtype()` now unwraps it like dictionary encoding, with a test.

python/sedonadb-geopandas/python/sedonadb_geopandas/_series.py
jiayuasu: Done, it just says duration arithmetic isn't supported yet.

PR #1329, opened 2026-09-21
feat(rust/sedona-extension): Add Metrics support to execution plans over FFI

paleolimbot:
This PR implements metrics across the FFI boundary, so that EXPLAIN ANALYZE works for scans dervied from an FFI table provider or execution plan. Most scans can and should implement metrics for debugging performance...this makes it easier to do so.

As a side effect, the JSON produced here can also be used to programmatically inspect metrics in (e.g.) Python in a future PR.


c/sedona-extension/src/metrics.rs
```diff
@@ -0,0 +1,836 @@
...
+        let Some(other) = other.as_any().downcast_ref::<Self>() else {
+            return;
+        };
+        let other = other.snapshot();
+        let mut state = self
+            .state
+            .lock()
+            .unwrap_or_else(|poisoned| poisoned.into_inner());
+        if state.display.is_empty() {
+            state.display = other.display;
```
jiayuasu: Could we preserve the aggregate display across partitions here? For a custom counter with values 10 and 20, this adds the numeric values to 30 but keeps the display as `10`. `EXPLAIN ANALYZE` uses that display, so it shows only the first partition's count.

Repro: add this to the existing `metrics.rs` tests module and run `cargo test -p sedona-extension custom_metric_display_repro`:

```rust
#[test]
fn custom_metric_display_repro() {
    let snapshot: SerializableExecutionPlanMetricsSet = serde_json::from_value(
        serde_json::json!({"metrics": [
            {"partition": 0, "metric_type": "summary",
             "value": {"type": "custom", "name": "custom_count", "display": "10", "value": 10}},
            {"partition": 1, "metric_type": "summary",
             "value": {"type": "custom", "name": "custom_count", "display": "20", "value": 20}}
        ]}),
    ).unwrap();
    let metrics = snapshot.into_metrics_set().aggregate_by_name();
    assert_eq!(metrics.iter().next().unwrap().value().as_usize(), 30);
    assert_eq!(metrics.to_string(), "custom_count=30");
}
```

The numeric assertion passes; the last assertion fails with `custom_count=10` instead of `custom_count=30`. I also checked this with a native custom counter: it displays 30 before the JSON round trip. Could we retain the producer-side aggregate while the original custom implementation is still available?

c/sedona-extension/src/execution_plan.rs
```diff
@@ -431,10 +432,14 @@ impl ExecutionPlan for ImportedSedonaCExec {
     }
 
     fn metrics(&self) -> Option<MetricsSet> {
-        get_plan_property::<Option<SerializedMetrics>, ()>(&self.inner, "metrics", None)
-            .ok()
-            .flatten()
-            .map(|sm| sm.into_metrics_set())
+        get_plan_property::<Option<SerializableExecutionPlanMetricsSet>, ()>(
```
jiayuasu, line 435: Could we keep the old `metrics` format working here? Extensions built with SedonaDB 0.4.1 return `{"display":"..."}`. This decoder rejects it, and `.ok()` below turns the error into `None`, so the full metrics disappear when the extension and host use different versions.

Repro: add this to the existing `metrics.rs` tests module and run `cargo test -p sedona-extension legacy_metrics_repro`:

```rust
#[test]
fn legacy_metrics_repro() {
    let json = r#"{"display":"output_rows=42"}"#;
    let result = serde_json::from_str::<Option<SerializableExecutionPlanMetricsSet>>(json);
    assert!(result.is_ok(), "{result:?}");
}
```

This fails with `missing field metrics`. The other direction also fails: a 0.4.1 reader expects `display`, which the new response omits. Could we accept either format and keep `display` in the response too, or put the structured format under a new property?

PR #1332, opened 2026-09-21
docs(r): publish pkgdown site with documentation

### LINKED ISSUE #1331: Upload pkgdown derived docs for R bindings in documentation build workflow
Currently our R API has no docs!

For nanoarrow we build with pkgdown and link directly to the site:

https://github.com/apache/arrow-nanoarrow/blob/ec8a58cae18beaa241c7fea7cb26816ac27c280f/ci/scripts/build-docs.sh#L105-L111

The guts are:

```shell
  show_header "Build R documentation"

   # Install the R package from source
   R CMD INSTALL ../r --preclean

   # Build R documentation
   Rscript -e 'pkgdown::build_site("../r", override = list(destination = "../docs/_build/html/r"), new_process = FALSE, install = FALSE)'
```

paleolimbot:
Adds the R toolchain and pkgdown dependencies to the documentation job. The build now installs the sedonadb R package, generates its pkgdown site inside the MkDocs source tree, and links it from the API navigation. This ensures both documentation artifacts and versioned mike deployments include the R reference under r/.

Validated with shell syntax checking, YAML parsing, and git diff checks.

Closes #1331


ci/scripts/build-docs.sh
```diff
@@ -55,10 +55,19 @@ fi
...
-else
+
+# Install the R package from source
+R CMD INSTALL r/sedonadb --preclean
+
+# Build R documentation inside the MkDocs source directory so both the MkDocs
+# build and the mike deployment include it. pkgdown paths are relative to the
+# package root (r/sedonadb).
+rm -rf docs/r
+Rscript -e 'pkgdown::build_site("r/sedonadb", override = list(destination = "../../docs/r"), new_process = FALSE, install = FALSE)'
```
jiayuasu: Could we check the exit status here and for `R CMD INSTALL` above? The script doesn't use `set -e`, so a failed pkgdown run can still reach `Success!` and publish a partial R site.

Here's a repro from the repo root with MkDocs installed. It stubs R/Quarto and makes Rscript write a partial page before failing; MkDocs is real:

```bash
t=$(mktemp -d)
mkdir -p "$t"/{ci/scripts,bin,docs/reference/sql}
cp ci/scripts/build-docs.sh "$t/ci/scripts/"
printf 'site_name: Repro\nnav:\n  - R: r/index.html\n' > "$t/mkdocs.yml"
printf '#!/bin/sh\nexit 0\n' > "$t/bin/R"
printf '#!/bin/sh\necho "# SQL" > generated.md\n' > "$t/bin/quarto"
cat > "$t/bin/Rscript" <<'SH'
#!/bin/sh
mkdir -p docs/r
printf '<a href="missing.html">Missing page</a>' > docs/r/index.html
echo 'Simulated pkgdown failure' >&2
exit 24
SH
chmod +x "$t"/bin/*
PATH="$t/bin:$PATH" bash "$t/ci/scripts/build-docs.sh"
echo "build exit: $?"
test -f "$t/site/r/index.html" && echo 'Partial site was copied'
```

I get `Simulated pkgdown failure`, then `Success!`, `build exit: 0`, and `Partial site was copied`. MkDocs doesn't flag the link to the missing page. Could we guard both R commands the same way as the MkDocs call below?

